PR #34, opened 2018-02-22
PR: Create an initial package of InputSystem

joelpryde:
Tasks:
- Move all code into Packages/com.unity.inputsystem
- Generate package.json and move over core package template files
- Export docs from GitHub wiki as markdown and add to Packages/com.unity.inputsystem/Documentation
- Fix asmdef files to conform to new package naming
- Add missing individual asmdef files for Extras
- Fix .gitignore/.npmignore files
- Give access to core assembly for tests with InternalsVisibleTo attribute


Packages/com.unity.inputsystem/Extras/DualShock/Unity.InputSystem.DualShock.asmdef
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "Unity.InputSystem.DualShock",
+    "references": [
+        "Unity.InputSystem"
```
Rene-Damm, line 4: Was wondering... should the DLLs be named UnityEngine.XXX or Unity.XXX? The thing is that nothing in this package goes in the new Unity namespace and likely nothing ever will. At this point it seems there will be a new package which adds an ECS-compatible system in the Unity namespace. Seems more fitting to name that one Unity.Input or something and use UnityEngine.InputSystem for this one. Not sure what the convention is, though.

Packages/com.unity.inputsystem/InputSystem/AssemblyInfo.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("UnityEngine.Experimental.XR.Interaction.Tests")]
```
Rene-Damm: Please update the attribute in InputSystem.cs instead.

Rene-Damm (changes requested):
Minor thing about InternalsVisibleTo. Other than that, looks great to me.

Packages/com.unity.inputsystem/Extras/DualShock/Unity.InputSystem.DualShock.asmdef
joelpryde: According to the package template it is Unity.XXX (I just changed my package).  I think this is a new change going forward from the scripting team to move everything into Unity.XXX namespace eventually.
https://gitlab.internal.unity3d.com/upm-packages/upm-package-template/blob/master/Tests/Runtime/Unity.YourPackageName.Tests.asmdef

Rene-Damm (approved):

PR #37, opened 2018-02-28
Fix build errors with UWP/IL2CPP

joelpryde:
Looks like UWP/IL2CPP uses the NET_4_6 instead of NET_4_0 for 4.6 .net compat level.  This change fixes the build errors on UWP with IL2CPP as the scripting backend.


Rene-Damm (approved):

Rene-Damm:
Great, thanks Joel!

PR #43, opened 2018-03-13
Ios

todi1856:
* Add TouchType to TouchControl
* Add iOS.GameController for iOS specific controllers


Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/GameController.cs
```diff
@@ -0,0 +1,83 @@
...
+        [InputControl(name = "buttonSouth", template = "Button", bit = (uint)Button.A)]
+        [InputControl(name = "buttonWest", template = "Button", bit = (uint)Button.X)]
+        [InputControl(name = "buttonNorth", template = "Button", bit = (uint)Button.Y)]
+        [InputControl(name = "buttonEast", template = "Button", bit = (uint)Button.B)]
+        [InputControl(name = "leftStickPress", template = "Button", bit = (uint)Button.LeftStick)]
+        [InputControl(name = "rightStickPress", template = "Button", bit = (uint)Button.RightStick)]
+        [InputControl(name = "leftShoulder", template = "Button", bit = (uint)Button.LeftShoulder)]
+        [InputControl(name = "rightShoulder", template = "Button", bit = (uint)Button.RightShoulder)]
+        [InputControl(name = "leftTrigger", template = "Button", bit = (uint)Button.LeftTrigger)]
+        [InputControl(name = "rightTrigger", template = "Button", bit = (uint)Button.RightTrigger)]
```
Rene-Damm, line 56: Triggers don't come through as axes on iOS?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/GameController.cs
```diff
@@ -0,0 +1,83 @@
...
+        public uint buttons;
+        public fixed float buttonValues[kMaxButtons];
```
Rene-Damm, line 58: What's this for?

Packages/com.unity.inputsystem/InputSystem/Utilities/Touch.cs
```diff
@@ -3,6 +3,12 @@
 
 namespace ISX
 {
+    public enum TouchType
+    {
+        Direct,
+        Indirect,
+        Stylus
+    }
```
Rene-Damm, line 11: This one... IDK :) I have no idea what a "direct" touch and an "indirect" touch is. Seems very iOS-specific. Maybe it's just a naming thing. "Stylus touch" intuitively seems like it should go under pen support, but there's probably a good point to be made for discerning those kind of contacts on mobiles where pen support isn't as explicit as on, say, a Surface laptop.

Rene-Damm (commented):
Overall LGTM. Just a couple clarifications I'd like to have.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/GameController.cs
todi1856: They don't, OS returns them as buttons. And buttons on iOS both hold bool and float value.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/GameController.cs
todi1856: Like mentioned buttons on iOS are returned as bool and float float, so the plan is to add additional InputControls later, for ex, leftShoulderRaw, XRaw returning float values.

Packages/com.unity.inputsystem/InputSystem/Utilities/Touch.cs
todi1856: Direct touches are those usual touches performed via touchscreen.

An example of indirect touch is for ex from Siri remote, you click on touchpad and it's reported as indirect touch - https://developer.apple.com/documentation/uikit/uitouchtype/1618107-indirect

Mantas said that indirect touch also exist on Android, but haven't seen that yet.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/GameController.cs
Rene-Damm: I see. Why aren't we using the float value the triggers then?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/GameController.cs
Rene-Damm: Wondering if it doesn't make more sense to directly wire all the buttons to the float values then and ignore the bit values. ButtonControls are AxisControls and are fine having float values.

todi1856: I talked to Mantas regarding this... and he said, that OS might have a different logic for defining when button bool value is true or false... So I slowly backed away from this solution

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/GameController.cs
todi1856: So on iOS triggers fall in same category as other buttons  - dpad, shoulder buttons, start, select. So it didn't make sense to take float for triggers and not take floats for other buttons.

I know that Gamepad class treat triggers as float input and those other buttons as bit values... that's why it's a bit confusing.

Maybe, let's make it like this - wire all those buttons (shoulder, dpad, select, start, triggers) to buttonValues variable as FLT format, as far as I understand InputSystem converts float to bool value if needed.

If user needs a direct OS bool value, he can call inputcontrol which will be wired to uint buttons? Not sure how to call such control, for leftShoulder wired to buttons, maybe should be called leftShoulderRaw or whatever.

What do you think?

PR #45, opened 2018-03-15
Android

todi1856:


Packages/com.unity.inputsystem/InputSystem/Plugins/Android/GameController.cs
```diff
@@ -0,0 +1,337 @@
...
+        }
+    }
+
+    [InputTemplate(stateType = typeof(GameControllerState))]
+    public class GameController : Gamepad
+    {
+        protected override void FinishSetup(InputControlSetup setup)
+        {
+            // TODO: pick dpadAxes or dpadButtons? We don't upfront if dpad is bound to buttons or hat switch
+            //       only when we first receive event for ex., HAT_X or DPAD_UP, we can then assume that one or another should be used
```
todi1856: @Rene-Damm There is this little problem, when game controller is connected I don't know if DPad will be coming via buttons or axes

todi1856: bah. * clicked enter accidentally*, anyways only when i receive first event from either MOTION_HAT_X  or KEYCODE_DPAD_LEFT, I can then identify where the events will be coming from (buttons or axis), and then I know that I should take dpadAxes or dpadButtons control.

The problem is dpad  property is set in Gamepad.FinishSetup, at that point it's not clear which control to take.

How should I solve this, should I create Android.DPadControl which would do this resolve logic inside?

Rene-Damm: That's an interesting case. It seems like this would need a custom Dpad control which is wired to both the button and the axis state and dynamically decides which one to use.

I do wonder, is this the case on Android even when using an actual gamepad with a normal button layout (i.e. on one gamepad you get buttons and on another gamepad with the same layout you get axes; that'd be super weird) or is it just something that varies between joysticks and gamepads? Can we detect it by device type?

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/GameController.cs
```diff
@@ -0,0 +1,337 @@
+using System.Runtime.InteropServices;
+using ISX.Controls;
+using ISX.LowLevel;
+using ISX.Utilities;
+using UnityEngine;
+
+namespace ISX.Android
+{
+    [StructLayout(LayoutKind.Sequential)]
```
Rene-Damm: `LayoutKind.Sequential` doesn't really do much. It's kinda what the code does anyway (at least on Mono). It's not enough to enforce layout to match between native and managed. From my experience so far, the only thing that really invariably guarantees matching layouts regardless of platform is `LayoutKind.Explicit`.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/GameController.cs
```diff
@@ -0,0 +1,337 @@
...
+           GENERIC_8 = 39,
+           GENERIC_9 = 40,
+           GENERIC_10 = 41,
+           GENERIC_11 = 42,
+           GENERIC_12 = 43,
+           GENERIC_13 = 44,
+           GENERIC_14 = 45,
+           GENERIC_15 = 46,
+           GENERIC_16 = 47,
+        };
```
Rene-Damm: Please rename these enum values to conform to convention.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/GameController.cs
Rene-Damm: For now, let's just leave it as is.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
```diff
@@ -1,11 +1,11 @@
 using System;
-using ISX.Editor;
 using ISX.LowLevel;
 using ISX.Utilities;
 using Unity.Collections.LowLevel.Unsafe;
 using UnityEngine;
 
 #if UNITY_EDITOR
+using ISX.Editor;
```
Rene-Damm: Please resolve the merge conflicts with head.

Rene-Damm (changes requested):

Rene-Damm:
Ah, also noticed that whitespace in the iOS PR had some issues. Please run the formatter (format.ps1 script on Windows and format.sh script on OSX). It'll run the formatter from the Unity meta-tools.

todi1856 (commented):
That's exactly the case - 2 gamepads, one returns DPad as bool values, other one as float values... Even in old implementation, there was a check whatever first event comes - motion event or key event, appropriate code branch for resolving dpad values was picked.

I'll have more info tomorrow.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/GameController.cs
Rene-Damm: Ok, judging from your other comment, that's indeed a pretty messed up setup. Seems like it will indeed require an Android-specific DpadControl which taps both states and dynamically returns one or the other.

One problem we'll likely run into is that the two states don't form a contiguous chunk of memory. It's a bit over here and a float over there. Likely to cause some issues in combination with input actions. But let's see.

todi1856: Actually, I am stupid... there is a way, when device is connected, you can get a list of available motion event types from device, so in this case  if MOTION_HAT_X , MOTION_HAT_Y are available, we know that DPAD will be coming from there, so I'll guess I'll put that info into capabilities of device

todi1856:
I'll update the PR, found an easy way for distinguishing Hat and Dpad

Packages/com.unity.inputsystem/InputSystem/Devices/Gamepad.cs
```diff
@@ -146,7 +146,7 @@ public class Gamepad : InputDevice, IDualMotorRumble
         public ButtonControl startButton { get; private set; }
         public ButtonControl selectButton { get; private set; }
 
-        public DpadControl dpad { get; private set; }
+        public DpadControl dpad { get; protected set; }
```
Rene-Damm, line 149: This is not a good change. The accessors sit on top of a contract established by a specific template. For Gamepad, the guarantee is that there is a DpadControl that can be found by looking for the name "dpad". AndroidGamepad breaks that contract.

I'll fix it up and bring it over to using a template constructor after the merge.

Rene-Damm (approved):

PR #50, opened 2018-03-20
Few fixes & improvements

todi1856:
* Expose InputControl.previousValueAsObject, that why it's possible to make use of back & front state buffers, while comparing values
* Few fixes for Android


todi1856:
I am probably a noob, but why does it show this huge list of commits when making PR, if you look at the Files Changed, it only shows the actual change.

Did I need to close a branch or something?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -134,6 +134,11 @@ public virtual object valueAsObject
             get { return null; }
         }
 
+        public virtual object previousValueAsObject
```
Rene-Damm: If we add this, `previousValueAsObject` will have to be implemented at the `InputDevice` level and these things will need tests.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidGameController.cs
```diff
@@ -36,7 +36,9 @@ public unsafe struct AndroidGameControllerState : IInputStateTypeInfo
         [InputControl(name = "leftTrigger", offset = (uint)AndroidAxis.Ltrigger * sizeof(float) + kAxisOffset)]
         [InputControl(name = "rightTrigger", offset = (uint)AndroidAxis.Rtrigger * sizeof(float) + kAxisOffset)]
         [InputControl(name = "leftStick")]
-        [InputControl(name = "rightStick", offset = (uint)AndroidAxis.Z * sizeof(float) + kAxisOffset)]
+        [InputControl(name = "rightStick")]
+        [InputControl(name = "rightStick/x", offset = (uint)AndroidAxis.Z * sizeof(float))]
+        [InputControl(name = "rightStick/y", offset = (uint)AndroidAxis.Rz * sizeof(float))]
```
Rene-Damm, line 41: Please add ////FIXME or ////REVIEW making a note of the non-contiguous state problem here.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
todi1856: because it's not actually a previousValue, I'll revert this change

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidGameController.cs
Rene-Damm: Ah also, please update the tests. They use the wrong data and this change should actually break the tests.

PR #56, opened 2018-03-22
NEW: Initial Switch support

yebata:


Packages/com.unity.inputsystem/Extras/NPad/NPad.cs
```diff
@@ -0,0 +1,255 @@
...
+        [InputControl(name = "buttonSouth", displayName = "B", bit = (uint)Button.South, usage = "Back")]
+        [InputControl(name = "buttonWest", displayName = "Y", bit = (uint)Button.West, usage = "SecondaryAction")]
+        [InputControl(name = "buttonEast", displayName = "A", bit = (uint)Button.East, usage = "PrimaryAction")]
+        [InputControl(name = "leftStickPress", displayName = "StickL", bit = (uint)Button.StickL)]
+        [InputControl(name = "rightStickPress", displayName = "StickR", bit = (uint)Button.StickR)]
+        [InputControl(name = "leftShoulder", displayName = "L", bit = (uint)Button.L)]
+        [InputControl(name = "rightShoulder", displayName = "R", bit = (uint)Button.R)]
+        [InputControl(name = "leftTriggerButton", template = "Button", displayName = "ZL", bit = (uint)Button.ZL)]
+        [InputControl(name = "rightTriggerButton", template = "Button", displayName = "ZR", bit = (uint)Button.ZR)]
+		[InputControl(name = "start", displayName = "Plus", bit = (uint)Button.Plus, usage = "Menu")]
```
Rene-Damm: Please run format.sh/format.ps1 found in the root directory of the repo.

Packages/com.unity.inputsystem/Extras/NPad/NPadSupport.cs
```diff
@@ -0,0 +1,71 @@
+namespace ISX.Plugins.Switch
+{
+    /// <summary>
+    /// Adds support for Switch NPad controllers.
+    /// </summary>
+    [InputPlugin]
```
Rene-Damm: This won't compile anymore. Please remove this, move the NPad folder to Packages/com.unity.inputsystem/InputSystem/Plugins, remove the .asmdef file, and add the initialization to `InputSystem.PerformDefaultPluginInitialization`.

Packages/com.unity.inputsystem/Extras/NPad/NPadSupport.cs
```diff
@@ -0,0 +1,71 @@
...
+				{ ""name"" : ""leftStick/up"", ""offset"" : 2, ""format"" : ""SHRT"", ""parameters"" : ""invert=false,normalize=false"" },
+				{ ""name"" : ""leftStick/down"", ""offset"" : 2, ""format"" : ""SHRT"", ""parameters"" : ""invert=false,normalize=false"" },
+				{ ""name"" : ""rightStick"", ""offset"" : 8, ""format"" : ""VC2S"" },
+				{ ""name"" : ""rightStick/x"", ""offset"" : 0, ""format"" : ""SHRT"", ""parameters"" : ""clamp=false,invert=false,normalize=false"" },
+				{ ""name"" : ""rightStick/left"", ""offset"" : 0, ""format"" : ""SHRT"", ""parameters"" : ""invert=false,normalize=false"" },
+				{ ""name"" : ""rightStick/right"", ""offset"" : 0, ""format"" : ""SHRT"", ""parameters"" : ""invert=false,normalize=false"" },
+				{ ""name"" : ""rightStick/y"", ""offset"" : 2, ""format"" : ""SHRT"", ""parameters"" : ""clamp=false,invert=false,normalize=false"" },
+				{ ""name"" : ""rightStick/up"", ""offset"" : 2, ""format"" : ""SHRT"", ""parameters"" : ""invert=false,normalize=false"" },
+				{ ""name"" : ""rightStick/down"", ""offset"" : 2, ""format"" : ""SHRT"", ""parameters"" : ""invert=false,normalize=false"" }
```
Rene-Damm: This seems to have the same state layout as the base template it is extending. What's the reason for having this template?

Rene-Damm (changes requested):
Looking great. Few changes necessary as the plugin mechanism has changed in the repo.

Rene-Damm:
Also, heads-up... everything has moved to the UnityEngine.Experimental.Input namespace now.

yebata:
Fixed namespace, file locations and ran reformatter script.

Packages/com.unity.inputsystem/Extras/NPad/NPadSupport.cs
yebata: I expect the template isn't derived any templates so I have to define all state layout. Isn't it?

yebata:
There is a template layout issue remaining.

Packages/com.unity.inputsystem/Extras/NPad/NPadSupport.cs
yebata: And, there are some Switch specific buttons so I'd like them to expose to developer.

Packages/com.unity.inputsystem/InputSystem/Plugins/NPad/NPadSupport.cs
```diff
@@ -0,0 +1,70 @@
+namespace UnityEngine.Experimental.Input.Plugins.Switch
```
Rene-Damm, line 1: Breaks build outside of editor on players other than Switch. I'll fix it after the merge.

Packages/com.unity.inputsystem/Extras/NPad/NPadSupport.cs
Rene-Damm: From what I see, it's extending NPad which takes its state layout from the state struct. That layout and the one in the JSON template here appear to be the same. Is the idea with NPad that we might end up supporting the Switch controller on platforms other than the Switch? I think I'm a little confused why we have two templates for what seems to be a single, platform-specific piece of hardware in my mind (a picture that may well be wrong).

Rene-Damm (approved):
Thanks Yusuke.

Packages/com.unity.inputsystem/Extras/NPad/NPadSupport.cs
yebata: Ah, I didn't understood that a JSON template and a state layout are for a same use. Will fix at the next phase.

Rene-Damm: Yeah, documentation is "not great" ATM. I've started putting something together at https://github.com/Unity-Technologies/InputSystem/wiki/Input-Templates.

And no worries, I'm writing some tests for Npad and I'll bring it back together into a single template.

PR #60, opened 2018-03-28
IOS fixes

todi1856:
* Fixed iOS left and right triggers
* Added some preliminary tests
* Small refactors


Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/iOSTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        Assert.That(controller.rightTrigger.value, Is.EqualTo(0.456).Within(0.000001));
+        Assert.That(controller.leftStick.x.value, Is.EqualTo(0.789).Within(0.000001));
+        Assert.That(controller.leftStick.y.value, Is.EqualTo(0.987).Within(0.000001));
+        Assert.That(controller.rightStick.x.value, Is.EqualTo(0.654).Within(0.000001));
+        Assert.That(controller.rightStick.y.value, Is.EqualTo(0.321).Within(0.000001));
+
+        /// FIXME, these will fail saying start/or select buttons were not expected to be pressed.
+        /// The problem is iOSGameController doesn't have those, so mappings for those are not set.
+        /// Because iOSGameController is derived from Gamepad, derived mappings are picked which are basically wrong
+        /// Question: how to disable those controls?
```
Rene-Damm, line 44: ATM there's no support for entirely removing controls from state (eventually I'd like to have that) but what you can do ATM is "park" the controls you're not interested in. An example is `MouseState` which gets a number of controls from Pointer that it is not interested in (such as "pressure"). So what it does is sets `offset` to `InputStateBlock.kInvalidOffset`. All controls that have an invalid offset when state layouting is performed will receive an automatically computed state offset -- which, in effect, appends the control to the end of the state and thus outside the state you are actively sending in your events.

Not as good as really removing the state entirely but for most intents and purposes, it has the same effect.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/iOSTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using ISX;
+using ISX.Plugins.iOS;
+using ISX.Plugins.iOS.LowLevel;
+using NUnit.Framework;
+
+class iOSTests : InputTestFixture
+{
+    [Test]
+    [Category("Devices")]
+    public void Devices_SupportsiOSGamePad()
```
Rene-Damm, line 11: Nice! Thanks Tomas.

Rene-Damm (approved):

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/iOSTests.cs
todi1856: Cool, i'll fix that way then

PR #63, opened 2018-03-29
XR Devices and basic XR device detection for the new Input System

StayTalm:
This has basic rumble, a large chunk of devices, and the core structure that detects and determines which device to use for XR.

I also had to make a few small changes to core to access descriptors in FinishSetup, and removed the old VR placeholders.


StayTalm:
I need to make a few updates before doing the true PR.

StayTalm:
Fixed issues and reopened

Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceDescription.cs
```diff
@@ -135,10 +135,9 @@ public bool Matches(InputDeviceDescription other)
             return MatchPair(interfaceName, other.interfaceName)
                 && MatchPair(deviceClass, other.deviceClass)
                 && MatchPair(manufacturer, other.manufacturer)
-                && MatchPair(product, other.product)
-                // We don't match serials; seems nonsense to do that.
-                && MatchPair(version, other.version)
-                && MatchPair(capabilities, other.capabilities);
+                && MatchPair(product, other.product)
+                // We don't match serials or capabilities; seems nonsense to do that.
+                && MatchPair(version, other.version);
```
Rene-Damm, line 140: Not a fan of this change. Matching serials indeed doesn't make sense. Matching capabilities does make sense. We actually already have a case for that on Android. Anyway, going to accept this change for now.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Daydream.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+
+            return null;
+        }
+    }
+
+    [StructLayout(LayoutKind.Explicit, Size = 120)]
+    public struct DaydreamHMDState : IInputStateTypeInfo
+    {
+        [InputControl(template = "Integer")]
```
Rene-Damm, line 34: This one might warrant its own control type to expose it as an enum (like PointerPhaseControl).

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -0,0 +1,94 @@
...
+using UnityEngine.Experimental.Input.LowLevel;
+using UnityEngine.Experimental.Input.Utilities;
+using UnityEngine.Experimental.Input.Plugins.XR.Haptics;
+using System.Text;
+
+namespace UnityEngine.Experimental.Input.Plugins.XR
+{
+    public class XRHMD : InputDevice
+    {
+        public static XRHMD active { get; private set; }
```
Rene-Damm, line 13: What's the intention here? This one seems to overlap with `current` and is confusing.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        protected override void FinishSetup(InputControlSetup setup)
+        {
+            base.FinishSetup(setup);
+            active = this;
```
Rene-Damm, line 18: Device setups should not alter global state. A device can be set up and never actually added to the system. Creating a device through `InputControlSetup` should not make it current/active.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
```diff
@@ -0,0 +1,100 @@
...
+    class XRFeatureDescriptor
+    {
+        public string name;
+        public List<UsageHint> usageHints;
+        public EFeatureType featureType;
+        public System.UInt32 customSize;
+    }
+
+    // Sync to UnityXRInputDeviceRole in IUnityXRInput.h
+    enum EDeviceRole
```
Rene-Damm, line 43: No 'E' prefixes for enums in C#.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
```diff
@@ -0,0 +1,100 @@
...
+#pragma warning disable 0649
+    [Serializable]
+    class UsageHint
+    {
+        public string content;
+    }
+
+    //Sync to XRInputFeatureDefinition in XRInputDeviceDefinition.h
+    [Serializable]
+    class XRFeatureDescriptor
```
Rene-Damm, line 34: Seems like UsageHint and XRFeatureDescriptor can just be structs.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            return JsonUtility.FromJson<XRDeviceDescriptor>(json);
+        }
+    }
+#pragma warning restore 0649
+
+	public static class XRSupport
+	{
+		public static void Initialize()
+		{
```
Rene-Damm, line 80: Please run format.ps1 or format.sh before committing.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Daydream.cs
```diff
@@ -0,0 +1,241 @@
...
+                            InputSystem.SetUsage(this, CommonUsages.RightHand);
+                            rightHand = this;
+                            break;
+                        }
+                    default:
+                        break;
+                }
+            }
+            catch (Exception)
+            { }
```
Rene-Damm, line 221: This will need a better solution. I'll take a look at this and SetupWithDescription after merging.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/SimpleXRRumble.cs
```diff
@@ -0,0 +1,72 @@
+using System;
+using UnityEngine.Experimental.Input.Plugins.XR;
+using UnityEngine;
+
+namespace UnityEngine.Experimental.Input.Plugins.XR.Haptics
+{
+    public class SimpleXRRumble
```
Rene-Damm, line 7: Why not just a struct?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Daydream.cs
```diff
@@ -0,0 +1,241 @@
...
+        public Vector3Control deviceVelocity { get; private set; }
+        public Vector3Control deviceAcceleration { get; private set; }
+
+        protected override void FinishSetup(InputControlSetup setup)
+        {
+            base.FinishSetup(setup);
+
+            try
+            {
+                XRDeviceDescriptor deviceDescriptor = XRDeviceDescriptor.FromJson(description.capabilities);
```
Rene-Damm, line 200: Please use `var` for local variables.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+                case EDeviceRole.RightHanded:
+                    {
+                        deviceType = typeof(XRController);
+                    }
+                    break;
+                default:
+                    {
+                        deviceType = typeof(XRHMD);
+                    }
+                    break;
```
Rene-Damm, line 127: What you probably want to do instead is to register `XRController` and `XRHMD` as templates and choose them as a base template.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+                updateBeforeRender = true
+            };
+
+            List<string> currentUsages = new List<string>();
+
+            uint currentOffset = 0;
+            foreach (var feature in descriptor.inputFeatures)
+            {
+                currentUsages.Clear();
+                foreach (var usageHint in feature.usageHints)
```
Rene-Damm, line 143: Should null check list here.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+                {
+                    if (usageHint.content != null && usageHint.content.Length > 0)
```
Rene-Damm, line 145: Use `string.IsNullOrEmpty` instead.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        static string SanitizeTemplateName(string templateName)
+        {
+            int stringLength = templateName.Length;
+            StringBuilder sanitizedTemplateName = new StringBuilder(stringLength);
+            for (int i = 0; i < stringLength; i++)
+            {
+                char letter = templateName[i];
+                if(Char.IsUpper(letter) || Char.IsLower(letter) || Char.IsDigit(letter) || letter == ':')
```
Rene-Damm, line 56: Reference primitive types with their primitive type name instead of their value struct name. I.e. `char` instead of `Char`.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -0,0 +1,213 @@
...
+namespace UnityEngine.Experimental.Input.Plugins.XR
+{
+    [Serializable]
+    class XRTemplateBuilder
+    {
+        static List<Func<XRDeviceDescriptor, string>> availableTemplates = new List<Func<XRDeviceDescriptor, string>>();
+
+        public XRDeviceDescriptor descriptor;
+
+        static uint GetOffsetForFeatureType(XRFeatureDescriptor featureDescriptor)
```
Rene-Damm, line 18: Seems slightly misnamed. GetSizeOfFeatureType seems more accurate.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Daydream.cs
Rene-Damm: Please remove the try/catch block. Exceptions from InputSystem should not be swallowed here.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/SimpleXRRumble.cs
StayTalm: I'm not really passing this thing around, and I wanted to have a guaranteed InputDevice reference via the Constructor.  I wasn't a big fan of DualMotorRumble always requiring the device again and again when setting or updating.

Rene-Damm: You could just pass the InputDevice to the struct constructor. IMO adding a separate heap object for what should just be part of a class isn't a good thing.

PR #66, opened 2018-04-05
Previous PR Feedback, and general optimization and cleanup of layout methods

StayTalm:
I handled most of the PR feedback from previously, removing a few of the temporary workarounds, getting global state setting out of FinishSetup, and removing state structures that were unneeded.  This should cut down the bloat and make it easier to extend XR devices for now.


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Daydream.cs
```diff
@@ -1,79 +1,10 @@
...
-        public Quaternion centerEyeRotation;
-
-        public FourCC GetFormat()
-        {
-            return new FourCC('X', 'R', 'S', '0');
-        }
-    }
-
-    [InputControlLayout(stateType = typeof(DaydreamHMDState))]
+    [InputControlLayout()]
```
Rene-Damm, line 5: Minor: parenthesis are unnecessary when empty.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -338,7 +338,7 @@ public struct Builder
             public Type type;
             public FourCC stateFormat;
             public string extendsLayout;
-            public bool updateBeforeRender;
+            public bool? updateBeforeRender;
```
Rene-Damm, line 341: Good catch.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -77,14 +73,35 @@ public override void MakeCurrent()
         }
     }
 
-    public class XRControllerWithRumble : XRController
+    public class XRControllerWithRumble : XRController, IHaptics
     {
-        public SimpleXRRumble rumble { get; private set; }
+        SimpleXRRumble m_Rumble;
+        public SimpleXRRumble rumble { get { return m_Rumble; } }
```
Rene-Damm, line 79: Ah doh, I didn't catch before that this is publicly exposed and not just an implementation internal to the controller (like `DualMotorRumble`). Kinda obsoletes my arguments about struct being adequate.

Thinking about it, though, I think it's better to not expose this but have the interface directly on the device. For haptics APIs that can be shared between devices, there should be interfaces derived from `IHaptics` (like `IDualMotorRumble`).

"Is it paused?" is missing from the `IHaptics` interface itself, I think. Overall, the play/pause/resume API can probably be improved.

Rene-Damm (approved):
Nice!

PR #68, opened 2018-04-10
[PS4] Added additional id's to the PS4 gamepad

martinh-unity:
Match connected gamepads to a logged-in user on the console.

Had to change the size of QueryUserIdCommand, because the length of a string is only 2 bytes in the native runtime. 

Changed 'orientation' to use a QuaternionControl instead as the PS4 native gamepad code passes a quaternion rather than Euler angles. I'm not sure how this affects the PC. Also I'm not sure if the HID API on the PC is  Euler or Quaternion. Either gamepad is providing Euler and the library is converting it to Quaternion or the PC code is incorrect. This will need to be investigated. It would be nice to provide a Quaternion on the PS4 as this is what a developer would normally expect but it might mean moving the 'orientation' property outside of the DualShockGamepad class and into the DualShockGamepadPS4 and DualShockGamepadHID separately. 

I have implemented the 'all' property on the DualShockGamepadPS4 to keep a list of gamepads. I did this by implementing the OnDeviceChange method. I can then read the gamepads 'slotid' and use that to add the gamepad to the matching index. That way the indexes of the gamepads in the array is the same as their slotids. This kind of feels correct from an API point of view as this is how developers would access a gamepad by using it's index from 0-3.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
@@ -273,39 +273,40 @@ namespace UnityEngine.Experimental.Input.Plugins.DualShock
         ////TODO: move up into base
         public ReadOnlyArray<PS4TouchControl> touches { get; private set; }
 
-        static DualShockGamepadPS4[] devices = new DualShockGamepadPS4[4];
+        private static DualShockGamepadPS4[] s_Devices = new DualShockGamepadPS4[4];
 
         // Slot id for the gamepad. Once set will never change.
-        private int m_slotId = -1;
-        private int m_defaultColorId = -1;
-        private int m_userId = -1;
+        private int m_SlotId = -1;
+        private int m_DefaultColorId = -1;
+        private int m_SceUserId = -1;
+
         private void UpdatePadSettingsIfNeeded()
         {
-            if (m_slotId == -1)
+            if (m_SlotId == -1)
             {
                 var command = QuerySlotIdCommand.Create();
 
                 if (ExecuteCommand(ref command) > 0)
                 {
-                    m_slotId = command.slotId;
-                    m_defaultColorId = command.defaultColorId;
-                    m_userId = command.userId;
+                    m_SlotId = command.slotId;
+                    m_DefaultColorId = command.defaultColorId;
+                    m_SceUserId = command.userId;
 
                     if (m_LightBarColor.HasValue == false)
                     {
-                        m_LightBarColor = GetPlayStationColor(m_defaultColorId);
+                        m_LightBarColor = GetPlayStationColor(m_DefaultColorId);
                     }
                 }
             }
         }
 
-        public Color LightBarColor
+        public Color lightBarColor
         {
             get
             {
                 if (m_LightBarColor.HasValue == false)
                 {
-                    return GetPlayStationColor(m_defaultColorId);
+                    return GetPlayStationColor(m_DefaultColorId);
                 }
 
                 return m_LightBarColor.Value;
@@ -327,12 +328,13 @@ namespace UnityEngine.Experimental.Input.Plugins.DualShock
                     return Color.black;
             }
         }
+
         public int slotId
         {
             get
             {
                 UpdatePadSettingsIfNeeded();
-                return m_slotId;
+                return m_SlotId;
             }
         }
         public int sceUserId
@@ -340,46 +342,47 @@ namespace UnityEngine.Experimental.Input.Plugins.DualShock
             get
             {
                 UpdatePadSettingsIfNeeded();
-                return m_userId;
+                return m_SceUserId;
             }
         }
 
-        public static void OnDeviceChange(InputDevice device, InputDeviceChange change)
+        internal static void OnDeviceChange(InputDevice device, InputDeviceChange change)
         {
-            DualShockGamepadPS4 ps4Gamepad = device as DualShockGamepadPS4;
+            var ps4Gamepad = device as DualShockGamepadPS4;
 
             if (ps4Gamepad == null || ps4Gamepad.slotId == -1) return;
 
             if (change == InputDeviceChange.Added)
             {
                 // Check there is no other device already in that slot
-                if (devices[ps4Gamepad.slotId] == null)
+                if (s_Devices[ps4Gamepad.slotId] == null)
                 {
-                    devices[ps4Gamepad.slotId] = ps4Gamepad;
+                    s_Devices[ps4Gamepad.slotId] = ps4Gamepad;
                 }
             }
             else if (change == InputDeviceChange.Removed)
             {
                 // check to make sure the device in the expected array index matches the actual device in that slot.
-                if (devices[ps4Gamepad.slotId] == device)
+                if (s_Devices[ps4Gamepad.slotId] == device)
                 {
-                    devices[ps4Gamepad.slotId] = null;
+                    s_Devices[ps4Gamepad.slotId] = null;
                 }
             }
         }
 
         public static DualShockGamepadPS4 FindBySlotId(int slotId)
         {
-            if (devices[slotId] != null && devices[slotId].slotId == slotId)
+            if (s_Devices[slotId] != null && s_Devices[slotId].slotId == slotId)
             {
-                return devices[slotId];
+                return s_Devices[slotId];
             }
 
             return null;
         }
+
         public static new ReadOnlyArray<DualShockGamepadPS4> all
         {
-            get { return new ReadOnlyArray<DualShockGamepadPS4>(devices); }
+            get { return new ReadOnlyArray<DualShockGamepadPS4>(s_Devices); }
         }
 
         protected override void FinishSetup(InputDeviceBuilder builder)
@@ -470,6 +473,7 @@ namespace UnityEngine.Experimental.Input.Plugins.DualShock
             m_LargeMotor = largeMotor;
             m_SmallMotor = smallMotor;
         }
+
         public void ResetOrientation()
         {
             var command = DualShockPS4OuputCommand.Create();
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
@@ -42,7 +42,7 @@ namespace UnityEngine.Experimental.Input.Plugins.DualShock.LowLevel
 
         [InputControl(name = "leftStickPress", bit = (uint)Button.L3, displayName = "L3")]
         [InputControl(name = "rightStickPress", bit = (uint)Button.R3, displayName = "R3")]
-        [InputControl(name = "options", layout = "Button", bit = (uint)Button.Options)]
+        [InputControl(name = "start", layout = "Button", bit = (uint)Button.Options)]
         [InputControl(name = "dpad", layout = "Dpad", sizeInBits = 4)]
         [InputControl(name = "dpad/up", bit = (uint)Button.DpadUp)]
         [InputControl(name = "dpad/right", bit = (uint)Button.DpadRight)]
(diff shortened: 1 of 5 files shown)


Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/DualShockTests.cs
```diff
@@ -175,9 +175,16 @@ public void Devices_SupportsDualShockOnPS4()
         Assert.That(gamepad.acceleration.y.ReadValue(), Is.EqualTo(0.654).Within(0.00001));
         Assert.That(gamepad.acceleration.z.ReadValue(), Is.EqualTo(0.321).Within(0.00001));
 
-        Assert.That(gamepad.orientation.x.ReadValue(), Is.EqualTo(0.111).Within(0.00001));
-        Assert.That(gamepad.orientation.y.ReadValue(), Is.EqualTo(0.222).Within(0.00001));
-        Assert.That(gamepad.orientation.z.ReadValue(), Is.EqualTo(0.333).Within(0.00001));
+        Quaternion orientation = gamepad.orientation.ReadValue();
```
Rene-Damm, line 260: Sidenote: reading a quaternion as a whole is not necessarily equivalent to reading its individual x, y, z, and w controls (which is one reason I prefer reading individual controls in tests; you don't want to test the combination into a quaternion here but rather just ensure the controls have the expected value). Combined controls often end up with processors on them which then alter the values. E.g. the stick controls have deadzone processor on them so reading `leftStick` directly yields a different value than doing `new Vector2(leftStick.x.ReadValue(), leftStick.y.ReadValue())`. Anyway, don't think it's overly relevant here.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/DualShockTests.cs
```diff
@@ -175,9 +175,16 @@ public void Devices_SupportsDualShockOnPS4()
...
+
+        Assert.That(orientation.x, Is.EqualTo(0.111).Within(0.00001));
+        Assert.That(orientation.y, Is.EqualTo(0.222).Within(0.00001));
+        Assert.That(orientation.z, Is.EqualTo(0.333).Within(0.00001));
+        Assert.That(orientation.w, Is.EqualTo(0.444).Within(0.00001));
+
+        //Assert.That(gamepad.orientation.x.ReadValue(), Is.EqualTo(0.111).Within(0.00001));
+        //Assert.That(gamepad.orientation.y.ReadValue(), Is.EqualTo(0.222).Within(0.00001));
+        //Assert.That(gamepad.orientation.z.ReadValue(), Is.EqualTo(0.333).Within(0.00001));
```
Rene-Damm: Please remove dead code.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -231,6 +273,115 @@ public class DualShockGamepadPS4 : DualShockGamepad
         ////TODO: move up into base
         public ReadOnlyArray<PS4TouchControl> touches { get; private set; }
 
+        static DualShockGamepadPS4[] devices = new DualShockGamepadPS4[4];
+
+        // Slot id for the gamepad. Once set will never change.
+        private int m_slotId = -1;
+        private int m_defaultColorId = -1;
+        private int m_userId = -1;
```
Rene-Damm: All of them won't change?

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -231,6 +273,115 @@ public class DualShockGamepadPS4 : DualShockGamepad
         ////TODO: move up into base
         public ReadOnlyArray<PS4TouchControl> touches { get; private set; }
 
+        static DualShockGamepadPS4[] devices = new DualShockGamepadPS4[4];
```
Rene-Damm: Should be private s_Devices.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockSupport.cs
```diff
@@ -23,13 +23,17 @@ public static void Initialize()
...
+
             InputSystem.RegisterControlLayout<PS4TouchControl>("PS4Touch");
             InputSystem.RegisterControlLayout<DualShockGamepadPS4>("PS4DualShockGamepad",
                 deviceDescription: new InputDeviceDescription
             {
                 deviceClass = "PS4DualShockGamepad",
                 interfaceName = "PS4"
             });
+
+            InputSystem.onDeviceChange += DualShockGamepadPS4.OnDeviceChange;
```
Rene-Damm, line 35: This should not be necessary but the current API is indeed lacking a good way to do this. `MakeCurrent` only covers a subset of the functionality necessary to support this. I'll take a look after the PR is in.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -231,6 +273,115 @@ public class DualShockGamepadPS4 : DualShockGamepad
...
+            get
+            {
+                UpdatePadSettingsIfNeeded();
+                return m_userId;
+            }
+        }
+
+        public static void OnDeviceChange(InputDevice device, InputDeviceChange change)
+        {
+            DualShockGamepadPS4 ps4Gamepad = device as DualShockGamepadPS4;
```
Rene-Damm: Use var for local variables.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
Rene-Damm: Also, uppercase after m_.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -231,6 +273,115 @@ public class DualShockGamepadPS4 : DualShockGamepad
...
+
+                    if (m_LightBarColor.HasValue == false)
+                    {
+                        m_LightBarColor = GetPlayStationColor(m_defaultColorId);
+                    }
+                }
+            }
+        }
+
+        public Color LightBarColor
```
Rene-Damm: Lowercase first char. `lightBarColor`

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -319,6 +470,13 @@ public override void SetMotorSpeeds(float largeMotor, float smallMotor)
             m_LargeMotor = largeMotor;
             m_SmallMotor = smallMotor;
         }
+        public void ResetOrientation()
```
Rene-Damm: Will this reset just the orientation reading or the entire gyro data?

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -231,6 +273,115 @@ public class DualShockGamepadPS4 : DualShockGamepad
...
+        public int sceUserId
+        {
+            get
+            {
+                UpdatePadSettingsIfNeeded();
+                return m_userId;
+            }
+        }
+
+        public static void OnDeviceChange(InputDevice device, InputDeviceChange change)
```
Rene-Damm: Make internal.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -231,6 +273,115 @@ public class DualShockGamepadPS4 : DualShockGamepad
...
+                    return Color.red;
+                case 2:
+                    return Color.green;
+                case 3:
+                    return Color.magenta;
+                default:
+                    return Color.black;
+            }
+        }
+        public int slotId
```
Rene-Damm: Please run format.ps1 or format.sh.

Rene-Damm (changes requested):
Please add tests for new functionality.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
martinh-unity: All of those settings, once read from the IOCTL won't change. The slotId, userId and colourId won't alter once the controller is connected. They only change if the controller is disconnected and reconnected again. In this case the gamepad object is deleted anyway and a new one created. 

I'll fix the naming issue.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
martinh-unity: Done. 

As a side note I was under the impression that coding standards for C# were changing to match the Microsoft ones, but older code won't change. I'm sure I read it somewhere. Isn't this happening for the InputSystem as its going to be one of the first packages?

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
martinh-unity: Only resets the orientation. The other two, velocity and angular velocity, are accelerometers. Although the documentation says angular velocity gets recalculated from the new orientation it won't get reset to 0's. The native method is called scePadResetOrientation which only sets the orientation to an identity quaternion (0,0,0,1).

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/DualShockTests.cs
martinh-unity: Not sure what you are getting at here. I know I could't find any individual components on the QuaternionControl. The comment in that class confirmed that adding in individual quaternion fields wasn't correct for a Quaternion, which I totally understand. 

However I'm not sure the best way to test this. Should the test create a quaternion from Euler angles and then use ReadValue() to get the quaternion and convert that back into  Euler angles and then test those?

Or should it just be removed if its not a relevant test anyway.

Rene-Damm: Ah doh, you're right, that didn't connect properly in my brain. I didn't realize that QuaternionControl indeed works different that way from Vector3Control. So, all good.

martinh-unity:
Now updated runtime tests for the PS4 DualShock haptics.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/DualShockTests.cs
```diff
@@ -109,11 +109,93 @@ public void Devices_CanSetLightBarColorAndMotorSpeedsOnDualShockHID()
...
+
 #if UNITY_EDITOR || UNITY_PS4
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_CanSetLightBarColorAndMotorSpeedsOnDualShockPS4()
+    {
+        var device = InputSystem.AddDevice(new InputDeviceDescription
+        {
+            deviceClass = "PS4DualShockGamepad", ////REVIEW: this should be the product name instead
```
Rene-Damm, line 124: Feel free to cull those comments when copypasting :) This comment isn't even quite correct :)

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/DualShockTests.cs
```diff
@@ -109,11 +109,93 @@ public void Devices_CanSetLightBarColorAndMotorSpeedsOnDualShockHID()
...
+        testRuntime.SetDeviceCommandCallback(gamepad.id,
+            (id, commandPtr) =>
+            {
+                unsafe
+                {
+                    if (commandPtr->type == QuerySlotIdCommand.Type)
+                    {
+                        Assert.That(receivedCommand.HasValue, Is.False);
+                        receivedCommand = *((QuerySlotIdCommand*)commandPtr);
+                        return 1;
```
Rene-Damm, line 190: This should populate the fields in the command and the test should then check whether the device received the right values.

Rene-Damm (approved):

PR #72, opened 2018-04-11
Various XRInput backend updates (+ initial Tests)

StayTalm:
*Previous PR Feedback
*Added initial suite of XRTests
*Exclude AraxisMerge merge files
*Added ViveLighthouse and Oculus Tracking Reference
*XR devices are now available in more XR players
*Sanitize XR feature names to be property names


Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1035,10 +1035,10 @@ internal static void PerformDefaultPluginInitialization()
...
 
             #if UNITY_EDITOR || UNITY_SWITCH
-            Plugins.Switch.SwitchSupport.Initialize();
-            #endif
-
-            #if UNITY_EDITOR
+            Plugins.Switch.SwitchSupport.Initialize();
+#endif
+
+#if UNITY_EDITOR || UNITY_STANDALONE || UNITY_EDITOR || UNITY_IOS
```
Rene-Damm: Formatting

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -89,6 +88,14 @@ public void SetIntensity(float intensity)
             m_Rumble.intensity = intensity;
         }
 
+        public bool IsHapticsPaused
```
Rene-Damm: Lowercase

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Vive.cs
```diff
@@ -89,4 +89,23 @@ protected override void FinishSetup(InputDeviceBuilder builder)
             deviceAngularVelocity = builder.GetControl<Vector3Control>("deviceAngularVelocity");
         }
     }
+
+    [InputControlLayout]
```
Rene-Damm, line 93: Technically you don't even need these. When registering a type as a layout, it takes the type whether the attribute is there or not. The attribute just allows controlling some settings on the layout.

Rene-Damm: I don't mind them being there, though. Kinda makes it a little more explicit.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+                        featureType = FeatureType.Binary
+                    }
+                }
+            }.ToJson()
+        }.ToJson());
+
+        InputSystem.Update();
+
+        var generatedLayout = InputSystem.TryLoadLayout("XRInput::XRManufacturer::XRGenericDevice");
+        Assert.That(generatedLayout, Is.Not.EqualTo(null));
```
Rene-Damm: Alternatively there's `Is.Not.Null`.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+                inputFeatures = new List<XRFeatureDescriptor>()
+                {
+                    new XRFeatureDescriptor()
+                    {
+                        name = "SimpleFeature",
+                        featureType = FeatureType.Binary
+                    }
+                }
+            }.ToJson()
+        }.ToJson());
```
Rene-Damm: Might be worth building something equivalent to `HID.HIDDeviceDescriptorBuilder` for XR.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_HardwareTrackerCreatesDefaultInputDevice()
+    {
+        testRuntime.ReportNewInputDevice(
+            new InputDeviceDescription
+        {
+            interfaceName = XRUtilities.kXRInterface,
+            product = "HardwareTracker",
```
Rene-Damm: What's a "HardwareTracker"?

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+        });
+        InputSystem.Update();
+
+        Assert.That((device["Button"] as ButtonControl).isPressed, Is.EqualTo(false));
+        Assert.That(device["DiscreteState"].ReadValueAsObject(), Is.EqualTo(0));
+        Assert.That(device["Axis"].ReadValueAsObject(), Is.EqualTo(0f).Within(0.0001f));
+        Assert.That(device["Vector2"].ReadValueAsObject(), Is.EqualTo(Vector2.zero));
+        Assert.That(device["Vector3"].ReadValueAsObject(), Is.EqualTo(Vector3.zero));
+        Assert.That(device["Rotation"].ReadValueAsObject(), Is.EqualTo(Quaternion.identity));
+        Assert.That(device["Custom"], Is.EqualTo(null));
```
Rene-Damm: `Is.Null`

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+        Assert.That((device["Last"] as ButtonControl).isPressed, Is.EqualTo(false));
```
Rene-Damm: `Is.False`

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+            discreteState = 17,
+            axis = 1.24f,
+            axis2D = new Vector2(0.1f, 0.2f),
+            axis3D = new Vector3(0.3f, 0.4f, 0.5f),
+            rotation = new Quaternion(0.6f, 0.7f, 0.8f, 0.9f),
+            lastElement = byte.MaxValue,
+        });
+        InputSystem.Update();
+
+        Assert.That((device["Button"] as ButtonControl).isPressed, Is.EqualTo(true));
```
Rene-Damm: `Is.True`

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+        var generatedLayout = InputSystem.TryLoadLayout("XRInput::XRManufacturer::XRDevice1");
+        Assert.That(generatedLayout, Is.EqualTo(null));
+
+        generatedLayout = InputSystem.TryLoadLayout("XRInput::XRManufacturer::XRDevice2");
+        Assert.That(generatedLayout, Is.EqualTo(null));
+    }
+
+    [Test]
+    [Category("Layouts")]
+    public void Layouts_WhenMatchingAKnownDeviceTheGeneratedLayoutInheritsProperly()
```
Rene-Damm: Avoid "proper" and "correct" in test names and try to be specific instead. Also, this test seems to single out one case with no guarantee that the other cases work the same. Recommend converting this test into one that takes the product+manufacturer combination as well as the base layout name as a parameter and then uses `[TestCase(...)]`.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
```diff
@@ -1,29 +1,648 @@
...
+            {
+                deviceRole = DeviceRole.LeftHanded,
+                inputFeatures = new List<XRFeatureDescriptor>()
+                {
+                    new XRFeatureDescriptor()
+                    {
+                        name = "SimpleFeature",
+                        featureType = FeatureType.Binary
+                    }
+                }
```
Rene-Damm: This feature doesn't seem relevant to the test. Same for some other tests here.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
Rene-Damm: Also recommend breaking up test names when there's a condition+outcome kinda setup. I.e. Devices_WhenThisIsTheCase_ThenThisIsTheResult. Just helps navigating test names a bit.

Packages/com.unity.inputsystem/Tests/InputSystem/UnitTests.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
Rene-Damm: Please remove this file. The UnitTests folder has been removed.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
StayTalm: A hardware tracker is the hockey puck device we have that is attached to objects to render in game.  https://www.vive.com/us/vive-tracker/

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
StayTalm: I need a device state of at least 1 element, since it needs to have a size > 0 in the device state.
Although, that reminds me... I should have a test and feature that excludes devices of 0 features.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/XRTests.cs
StayTalm: Playing around with the XRDeviceDescriptorBuilder, and I'm finding it's about the same amount of lines to build, but now needs some whole other class.  I'm thinking of holding off on that one for now, since testing is the only time I'd really create a XRDeviceDescriptor in C#.

StayTalm:
@Rene-Damm  Updated tests and took feedback into account.  Hoping to leave XML Docs on this round, and tackle them on the next pass.

Rene-Damm (approved):

PR #75, opened 2018-04-20
Merge iOS back to Develop

todi1856:
* Make lower case for iOS based classes as discussed for consistency
* Move out non gyro values out of gyroscope


Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSensors.cs
```diff
@@ -0,0 +1,25 @@
...
+        [InputControl][FieldOffset(0)] public Vector3 gravity;
+        [InputControl][FieldOffset(12)] public Quaternion attitude;
+        [InputControl][FieldOffset(28)] public Vector3 acceleration;
+        [InputControl][FieldOffset(40)] public Vector3 rotationRateUnbiased;
+
+        public FourCC GetFormat()
+        {
+            return kFormat;
+        }
+    }
```
todi1856, line 23: So just to confirm, how should I approach this, you seems to agreed that we should create artifical sensors for each of those properties, right?

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
```diff
@@ -3926,33 +3926,34 @@ public void Devices_CanGetAccelerometerReading()
     [Category("Devices")]
     public void Devices_CanGetGyroReading()
     {
+        ////FIXME: Move non gyro values to other tests
```
Rene-Damm, line 4111: Recommend just nuking the code here in the test. It's not that precious :)

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSensors.cs
Rene-Damm: I didn't come up with a good solution on the managed side. ATM my thinking is that the iOS side should split up its motion sensor code path such that it registers and feeds several sensors from the motion data.

There probably is a more elegant solution somewhere in here but ATM, having the backend and the C# system develop different views on the device setup and on where to route state would require some more invasive changes.

PR #88, opened 2018-04-26
WIP sensors

todi1856:
* Added generic sensors gravity/attitude/linearacceleration + tests
* Split Android tests into seperate files... don't lump everything into one
* Add InputConfiguration.compensateSensorsForScreenOrientation


Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
```diff
@@ -118,14 +156,95 @@ protected override void OnRemoved()
...
+        }
+
+        protected override void FinishSetup(InputDeviceBuilder builder)
+        {
             gravity = builder.GetControl<Vector3Control>("gravity");
             base.FinishSetup(builder);
         }
     }
 
+    //// FIXME: Is this name good enough, possible other name RotationVector, here's how Android docs describe it. "A rotation vector sensor reports the orientation of the device relative to the East-North-Up coordinates frame."
```
Rene-Damm: I'd prefer RotationVector. IMO technical/fancy names are usually BS. 'attitude' and 'azimuth' and whatnot are all bad names IMO. An API should not require people to look up the technical definition of a term to be able to understand what it's about.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/AndroidSensorTests.cs
```diff
@@ -0,0 +1,69 @@
+#if UNITY_EDITOR || UNITY_ANDROID
+using System;
+using UnityEngine.Experimental.Input;
+using UnityEngine.Experimental.Input.Plugins.Android;
+using UnityEngine.Experimental.Input.Plugins.Android.LowLevel;
+using NUnit.Framework;
+
+class AndroidSensorTests : InputTestFixture
```
Rene-Damm, line 8: Please keep the tests together in AndroidTests. I'll move it over.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/AndroidTests.cs
```diff
@@ -41,16 +49,7 @@ public void Devices_CanDifferentiateAndroidGamepadFromJoystick()
...
-                new InputDeviceDescription
-        {
-            interfaceName = "Android",
-            deviceClass = "AndroidGameController",
-            capabilities = new AndroidDeviceCapabilities
-            {
-                inputSources = AndroidInputSource.Gamepad | AndroidInputSource.Joystick,
-            }.ToJson()
-        });
+        var device = InputSystem.AddDevice(GetGamepadDeviceDescriptor());
```
Rene-Damm, line 52: Sorry but I'l probably do the same as last time with that :)

Same rationale. Reading a test should not make you have to learn anything but the API of the system being tested and the API of your test framework. All these little helpers do is decrease readability in order to save a few lines of code. Code duplication in test code is not an issue. Readability and understandability of a test just by itself is. And these little helpers just end up proliferating and turn into test frameworks of their own.

If an API requires lots of code duplication in tests, opt for addressing that in the API, not in the tests.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidSensors.cs
```diff
@@ -92,12 +124,49 @@ public class AndroidAccelerationProcessor : IInputControlProcessor<Vector3>
...
+            if (InputConfiguration.CompensateSensorsForScreenOrientation)
+            {
+                const float kSqrtOfTwo = 1.4142135623731f;
+                Quaternion q = Quaternion.identity;
+
+                switch (Screen.orientation)
+                {
+                    case ScreenOrientation.PortraitUpsideDown: q = new Quaternion(0.0f, 0.0f, 1.0f /*sin(pi/2)*/, 0.0f /*cos(pi/2)*/); break;
+                    case ScreenOrientation.LandscapeLeft:      q = new Quaternion(0.0f, 0.0f, kSqrtOfTwo * 0.5f /*sin(pi/4)*/, -kSqrtOfTwo * 0.5f /*cos(pi/4)*/); break;
+                    case ScreenOrientation.LandscapeRight:     q = new Quaternion(0.0f, 0.0f, -kSqrtOfTwo * 0.5f /*sin(3pi/4)*/, -kSqrtOfTwo * 0.5f /*cos(3pi/4)*/); break;
```
Rene-Damm, line 160: What's the commented out code about?

todi1856:
"Please keep the tests together in AndroidTests. I'll move it over."

Why? This makes the file huge for no good reason, I just looked at CoreTests.cs and it's over 7000 lines?! Why would you want such huge files?

todi1856:
And boo regarding code duplication, it's obvious there will be more tests required for gamepad, and that code duplication problem will be even bigger. I don't think that should be fixed in API, because in this the code duplication happens due tests itself.

Rene-Damm:
>Why? This makes the file huge for no good reason, I just looked at CoreTests.cs and it's over 7000 lines?! Why would you want such huge files?

The alternative is tests that are sliced up in arbitrary fashion for no reason that is apparent to the reader. `AndroidTests` tests a piece of functionality that stands on its own (represented by `AndroidSupport`). `AndroidSensorTests` slices out sensors for no reason other than that the programmer thought it feels better :)

I don't mind large C# files as long as what's in them belongs together.

todi1856:
That programmer was me, I put sensors to AndroidSensorsTests, because it's a big group of stuff, sure it's Android. If I only care about testing sensors and not other stuff, I want the only run them, when all tests are in AndroidTests, I am forced additionally to run other non related tests (like gamepad/mouse/touches.. etc).

The same is with CoreTests... there's so many tests there, which could have been categorized in smaller groups making it easier to navigate and testing individual group of features.

PR #90, opened 2018-05-02
CHANGE: Apply override bindings for sets

AlexanderVott:
Added an implementation override of bindings for sets.
Added the implementation of removing the override of bindings for sets.


Rene-Damm:
Looking at this. We're currently in the midst of a larger refactor of the action and binding stuff but I think the changes here still fit in.

Rene-Damm:
Awesome PR BTW.

AlexanderVott:
Thanks)

PR #91, opened 2018-05-02
Fix apply binding override

AlexanderVott:
Fixed checking the index on the range of the array of bindings.
If the index is 1, and the number of bindings is 1, then the check was successful, but when the value was filled from the array, an exception was thrown.

### DIFF
--- a/Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
+++ b/Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
@@ -6269,6 +6269,24 @@ class CoreTests : InputTestFixture
 
     [Test]
     [Category("Actions")]
+    public void Actions_CanOverrideBindings()
+    {
+        var gamepad = (Gamepad)InputSystem.AddDevice("Gamepad");
+        var action = new InputAction(binding: "/gamepad/leftTrigger");
+        action.ApplyBindingOverride(new InputBindingOverride {binding = "/gamepad/rightTrigger"});
+        action.Enable();
+
+        var wasPerformed = false;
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
```diff
@@ -6447,7 +6447,34 @@ public void Actions_CanSerializeAndDeserializeActionsWithCompositeBindings()
...
+	[Test]
+	[Category("Sets")]
+	public void Sets_OnSetWithMultipleBindings_ApplyOverrides()
+	{
+		var set = new InputActionSet();
+		var action1 = set.AddAction("action1", "/<keyboard>/enter");
+		var action2 = set.AddAction("action2", "/<gamepad>/buttonSouth");
+
+		var listOverrides = new List<InputBindingOverride>(3);
+		listOverrides.Add(new InputBindingOverride {action = "action3", binding = "/gamepad/buttonSouth"});
```
StephenHodgson, line 6459: Def not a fan of these string bindings. @Rene-Damm any ideas on alternatives?

AlexanderVott: I closed this pullrequest. There was garbage.
This implementation is in another pullrequest. Tell me how it will be correct, I will change.

AlexanderVott: @StephenHodgson You about?
```
<gamepad>/buttonSouth
```
Maybe then create a device and bind it through its properties?

PR #92, opened 2018-05-02
FIX: Check index of range array.

AlexanderVott:
Fixed checking the index on the range of the array of bindings.
If the index is 1, and the number of bindings is 1, then the check was successful, but when the value was filled from the array, an exception was thrown.


AlexanderVott:
screenshot_4

Rene-Damm (approved):

Rene-Damm:
Thanks :)

PR #98, opened 2018-05-06
FIX: When cloning an InputActionSet, ensure the cloned actions refer to the newly created set

periodyctom:
…, as they'll default to the SingletonActionSet (null) when initially cloned

CHANGED: Updated tests to verify changes

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSet.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSet.cs
@@ -185,6 +185,17 @@ namespace UnityEngine.Experimental.Input
                 m_Actions = ArrayHelpers.Clone(m_Actions)
             };
 
+            if(clone.m_Actions != null)
+            {
+                // ensure cloned actions refer to the new set and are not singletons
+                var cloneActions = clone.m_Actions;
+                var length = cloneActions.Length;
+                for(int i = 0; i < length; ++i)
+                {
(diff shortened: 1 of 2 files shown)


periodyctom:
Discovered the bug while implementing handling for multiple players, using a variation on this post:

https://forum.unity.com/threads/input-system-update.508660/page-5#post-3476667

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSet.cs
```diff
@@ -185,6 +185,17 @@ public InputActionSet Clone()
...
+            if(clone.m_Actions != null)
+            {
+                // ensure cloned actions refer to the new set and are not singletons
+                var cloneActions = clone.m_Actions;
+                var length = cloneActions.Length;
+                for(int i = 0; i < length; ++i)
+                {
+                    cloneActions[i].m_ActionSet = clone;
+                }
+            }
```
Rene-Damm, line 197: Took me a moment to realize but this unfortunately is a symptom fix. `ArrayHelpers.Clone` is calling `Clone` on each action which results in the action set being populated with actions that think they're singleton actions instead of being part of a set. I.e. the root problem is that the set isn't cloning actions correctly.

Good catch, though. The action stuff is currently undergoing a bigger refactor and I'll make sure to address this problem as part of that.

periodyctom:
Got it, I'll close this and just hold onto my patch locally until the refactor is done.

PR #99, opened 2018-05-07
Improvements to InputDebugger

todi1856:
* Properly show enum values for specific events
* Add missing types for input debugger


Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -195,6 +195,11 @@ public bool noisy
             get { return InputControlPath.TryFindChild(this, path); }
         }
 
+        public Type valueType
```
Rene-Damm: Please make virtual and remove from constructor. IMO no need to store this on InputControl.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputControlTreeView.cs
```diff
@@ -321,39 +321,50 @@ private unsafe string ReadRawValueAsString(InputControl control, byte[] state)
...
+                // Stringify enum values, for. ex., PointerPhase
+                if (value != null && control.valueType.IsEnum)
                 {
-                    return (*((double*)ptr)).ToString();
+                    var intValue = Convert.ToInt32(value);
+                    value = Enum.ToObject(control.valueType, intValue);
                 }
 
-                return null;
+                return value != null ? value.ToString() : null;
```
Rene-Damm, line 367: Neither the old nor the new version are quite that ideal; both generate more garbage than they need to. Fine for now. Ultimately, we want to get rid of this entire method.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
Rene-Damm: Actually, not virtual but abstract.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -195,10 +195,10 @@ public bool noisy
             get { return InputControlPath.TryFindChild(this, path); }
         }
 
-        public Type valueType
-        {
-            get { return m_ValueType; }
-        }
+        /// <summary>
+        /// Returns the underlying value type of this control.
```
Rene-Damm: For doc comments in general, please go beyond the obvious. For this one, the fact it can be null is undocumented, for example. Also noteworthy that it's not necessarily the same as `ReadValueAsObject().GetType()`. Etc.

PR #100, opened 2018-05-08
Implement InputConfiguration.CompensateSensorsForScreenOrientation for all sensors

todi1856:


Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
```diff
@@ -77,13 +81,56 @@ public static FourCC kFormat
...
+        [InputControl(processors = "Sensor")] 
+        public Vector3 acceleration;
 
         public FourCC GetFormat()
         {
             return kFormat;
         }
     }
+    
+    public class SensorProcessor : IInputControlProcessor<Vector3>
```
Rene-Damm: This needs a more specific name.

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
```diff
@@ -77,13 +81,56 @@ public static FourCC kFormat
...
+    {
+        public Vector3 Process(Vector3 value, InputControl control)
+        {
+            if (!InputConfiguration.CompensateSensorsForScreenOrientation)
+                return value;
+            
+            Quaternion rotation = Quaternion.identity;
```
Rene-Damm: Use var for local variables.

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
```diff
@@ -77,13 +81,56 @@ public static FourCC kFormat
...
+            {
+                case ScreenOrientation.PortraitUpsideDown: rotation = Quaternion.Euler(0, 0, 180); break;
+                case ScreenOrientation.LandscapeLeft: rotation = Quaternion.Euler(0, 0, 90); break;
+                case ScreenOrientation.LandscapeRight: rotation = Quaternion.Euler(0, 0, 270); break;
+            }
+            return rotation * value;
+        }
+    }
+    
+    public class SensorRotationProcessor : IInputControlProcessor<Quaternion>
```
Rene-Damm: Should have better name. SensorRotationProcessor gives no indication of how it processes rotation. Also no indication how this is really different from the processor above. Recommend weaving part of the name of the configuration property into the names. Stuff that makes it clear one processor is compensating a direction vector for screen orientation and the other processor is compensating a rotation for screen orientation.

IMO this doesn't even need to be specific to sensors and wouldn't have to have 'sensor' in the name. Any direction or rotation vector could be compensated for device orientation.

Also makes me wonder whether 'device orientation' is a better term than 'screen orientation'?

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
Rene-Damm: (Goes for elsewhere, too)

Rene-Damm (changes requested):

Rene-Damm:
Also, tests.

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
todi1856: I am really bad with names, but I'll try. Also device orientation is a different thing compared to screen orientation, so better not to mix them

todi1856:
@Rene-Damm regarding tests, I can't trigger Screen.orientation changes via tests, if I would set Screen.orientation = Landscape, Screen.orientation would still return Portrait in Editor at least, in other words I somehow need to mock that value, any ideas?

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidSensors.cs
```diff
@@ -115,7 +116,7 @@ public FourCC GetFormat()
         }
     }
 
-    public class AndroidSensorProcessor : IInputControlProcessor<Vector3>
+    public class AndroidCompensateDirectionProcessor : CompensateDirectionProcessor
```
Rene-Damm, line 119: Ideally, should indicate functional difference to base processor by its name (or at least comment).

Rene-Damm (approved):

Rene-Damm:
@todi1856 Not sure. Think ideally `IInputRuntime` should have native functionality needed by the input system. So, one solution would be to put screen orientation on there.

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
```diff
@@ -77,13 +81,56 @@ public static FourCC kFormat
...
+        {
+            if (!InputConfiguration.CompensateSensorsForScreenOrientation)
+                return value;
+            
+            var rotation = Quaternion.identity;
+            switch (Screen.orientation)
+            {
+                case ScreenOrientation.PortraitUpsideDown: rotation = Quaternion.Euler(0, 0, 180); break;
+                case ScreenOrientation.LandscapeLeft: rotation = Quaternion.Euler(0, 0, 90); break;
+                case ScreenOrientation.LandscapeRight: rotation = Quaternion.Euler(0, 0, 270); break;
```
Rene-Damm, line 105: Stuff like this really needs comments.

todi1856: You should have commented before merging the PR :D

PR #101, opened 2018-05-08
FIX: Expanded the unsafe block in some tests

douglas-piconi:
mono runtime 4.x appears to be more strict with what needs to go into unsafe block.   Expanded
the blocks to get tests to compile.  Ran tests in 4.x mode mono backed in Editor and with Windows64 player.


Rene-Damm (approved):
Awesome, thanks.

PR #102, opened 2018-05-10
Tests for sensors

todi1856:
* Move Screen.orientation setup to InputRuntime
* And added test thanks to that for sensors


Packages/com.unity.inputsystem/InputSystem/IInputRuntime.cs
```diff
@@ -93,6 +93,8 @@ public unsafe interface IInputRuntime
         float pollingFrequency { set; }
 
         InputUpdateType updateMask { set; }
+
+        ScreenOrientation screenOrientation { set; get; }
```
Rene-Damm, line 97: Do we actually want this to be settable? I don't think the input system should get an API for modifying this, so I'd prefer for this to be readonly. `InputTestRuntime` can still have it read/write for tests.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
```diff
@@ -4247,6 +4258,88 @@ public void Devices_CanGetLinearAccelerationReading()
         Assert.That(sensor.acceleration.ReadValue(), Is.EqualTo(value).Within(0.00001));
     }
 
+    private void ValidateSensorControl(Vector3Control control, Vector3 targetValue)
```
Rene-Damm, line 4261: Umpff...

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
```diff
@@ -4247,6 +4258,88 @@ public void Devices_CanGetLinearAccelerationReading()
...
+        InputRuntime.s_Instance.screenOrientation = ScreenOrientation.LandscapeLeft;
+        Assert.That(control.ReadValue(), Is.EqualTo(new Vector3(-targetValue.y, targetValue.x, targetValue.z)).Using(new Vector3Comparer(0.0001f)));
+
+        InputRuntime.s_Instance.screenOrientation = ScreenOrientation.PortraitUpsideDown;
+        Assert.That(control.ReadValue(), Is.EqualTo(new Vector3(-targetValue.x, -targetValue.y, targetValue.z)).Using(new Vector3Comparer(0.0001f)));
+
+        InputRuntime.s_Instance.screenOrientation = ScreenOrientation.LandscapeRight;
+        Assert.That(control.ReadValue(), Is.EqualTo(new Vector3(targetValue.y, -targetValue.x, targetValue.z)).Using(new Vector3Comparer(0.0001f)));
+
+        InputRuntime.s_Instance.screenOrientation = ScreenOrientation.Portrait;
```
Rene-Damm, line 4274: This should be `testRuntime`.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
todi1856: But so much less code, come on... :/

Packages/com.unity.inputsystem/InputSystem/IInputRuntime.cs
todi1856: Screen.screenOrientation is settable, so I thought that should be the same? not sure though

Rene-Damm: I've removed the setting from the interface. While overall it seems a slightly odd thing to be settable, allowing the user-facing API in Unity to set it seems fine. However, the input system itself shouldn't have need for setting screen orientation.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests.cs
Rene-Damm: See, I've yet to come across a case where a helper function added next to tests doesn't indicate one of three things:

1. The tests need improvement.
2. The system API needs improvement.
3. The test framework/rig API needs improvement.

Let's take the tests here. All tests using ValidateSensorControl (that's another gripe I have with these helper functions -- the names don't tell you anything about what they actually do) run the exact same verification. All that is different is the input data. That makes it a prime target for writing a test using `TestCase`. So now we have three tests that become a single method annotated with three TestCases. *That* is much less code :) And no helper function.

todi1856: I am curious, how would I use TestCase attribute in this case, I would somehow need to pass a property of a sensor  to TestCase? For ex., sensor.acceleration, sensor.angularVelocity

Should I pass a name and then resolve it via reflection, or there's some other way?

Rene-Damm: Yup, one way would be to just pass a `string` giving the test a control name (or even path) and then do `device[controlName]` to fetch the control in the tests.

todi1856: Oh I didn't know you can access controls like this, I'll give it a shot. Thx

Rene-Damm: It's a pretty flexible way as it supports the same path language that the binding system uses. So you can do stuff like `myDevice["leftStick/x"]` or `myDevice["{PrimaryAction}"]` and stuff like that.

todi1856: So I am stuck at 
 InputSystem.QueueStateEvent(sensor, new AccelerometerState { acceleration = new Vector3(0,1, 2)});

I can't use new AccelerometerState { acceleration =  new Vector3(0,1, 2)} in TestCaseAttribute because it's not constant, even if I could , InputSystem.QueueStateEvent expects type to be known upfront, it wouldn't accept an object type.

Any ideas?

`
    [Test]
    [Category("Devices")]
    [TestCase("Accelerometer", "acceleration", new AccelerometerState { acceleration = new Vector3(0.123f, 0.456f, 0.789f)})]
    public void Devices_CanCompensateSensorDirection(string layout, string controlName, object state)
    {
        var sensor = InputSystem.AddDevice(layout);
        var value = new Vector3(0.123f, 0.456f, 0.789f);
        InputSystem.QueueStateEvent(sensor, state);
        InputSystem.Update();

        var control = (Vector3Control) sensor[controlName];
        InputConfiguration.CompensateSensorsForScreenOrientation = true;

        testRuntime.screenOrientation = ScreenOrientation.LandscapeLeft;
        Assert.That(control.ReadValue(), Is.EqualTo(new Vector3(-value.y, value.x, value.z)).Using(new Vector3Comparer(0.0001f)));

        testRuntime.screenOrientation = ScreenOrientation.PortraitUpsideDown;
        Assert.That(control.ReadValue(), Is.EqualTo(new Vector3(-value.x, -value.y, value.z)).Using(new Vector3Comparer(0.0001f)));

        testRuntime.screenOrientation = ScreenOrientation.LandscapeRight;
        Assert.That(control.ReadValue(), Is.EqualTo(new Vector3(value.y, -value.x, value.z)).Using(new Vector3Comparer(0.0001f)));

        testRuntime.screenOrientation = ScreenOrientation.Portrait;
        Assert.That(control.ReadValue(), Is.EqualTo(value).Using(new Vector3Comparer(0.0001f)));

        InputConfiguration.CompensateSensorsForScreenOrientation = false;
        Assert.That(control.ReadValue(), Is.EqualTo(value).Using(new Vector3Comparer(0.0001f)));
    }
`

Rene-Damm: One solution could be to add an overload to `InputSystem.QueueStateEvent` that takes an arbitrary struct plus a `FourCC` state type code. This way you could simply pass a `Vector3` and take the `FourCC` format code from the device.

A possibly better solution would be to add an API that allows grabbing the current state of an `InputDevice` as a state event and then use `WriteValueInto` to write the `Vector3` into the state event.

Rene-Damm: I've added the ability to do the latter thing in https://github.com/Unity-Technologies/InputSystem/commit/ea4856579d3dca62de22b5f32fbb722b2199be3e.

Rene-Damm: I've also added the missing implementation for `WriteValueInto` and added tests that do the thing I was suggesting (`Devices_DeltaControlsAccumulateBetweenUpdates` and `Devices_DeltaControlsResetBetweenUpdates`) in https://github.com/Unity-Technologies/InputSystem/commit/dd71b04aa2e8b7396748f8c78f74dc2f7b5f03f2.

PR #110, opened 2018-05-17
[android] Various fixes for android

todi1856:
- [android] Apply process for touch positions, on Android y value is inverted
- Added InputRuntime.screenSize
- [android] Added input controls, comments, tests for sensors
- Rewrote sensor tests in CoreTest.cs


Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs
```diff
@@ -62,6 +62,21 @@ public FourCC GetFormat()
...
+        {
+#if UNITY_EDITOR
+            return value;
+#elif PLATFORM_ANDROID
+            return new Vector2(value.x, InputRuntime.s_Instance.screenSize.y - value.y);
+#else
+            return value;
+#endif
+        }
+    }
```
Rene-Damm, line 77: This one I don't like so much. If there's no need for a processor, there simply shouldn't be a processor on the control. For now, let's take it and then refine it.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/IInputRuntime.cs
```diff
@@ -112,6 +112,7 @@ public unsafe interface IInputRuntime
         InputUpdateType updateMask { set; }
 
         ScreenOrientation screenOrientation { get; }
+        Vector2 screenSize { get; }
```
Rene-Damm, line 115: Looking at these two properties I find myself hesitant. They proliferate what I think is simply a stupid assumption in the old Unity APIs -- that there is a single screen on the system. We've hardcoded and spread that into several places and then awkwardly worked around it with the "multi display" APIs.

Not sure how to best handle this but would prefer a solution that doesn't involve adding more APIs cementing what I think is a design fallacy.

todi1856:
@Rene-Damm you're right, at least for screenSize which is per screen, and we can have multiple screens...

as for screenOrientation, there cannot be multiple instances of it.

todi1856:
So looking ahead, we'll probably need screen size and window size, and probably screen offset in some case, as on some platforms (for ex., UWP), you can have unity screen occupy 1/3 of screen with specific offset.

todi1856:
Maybe it's worth to add ////FIXME: at those lines with comments

PR #114, opened 2018-05-18
Refactor of input bindings

Rene-Damm:


Rene-Damm:
Crap, the history on this thing should not have been squashed. Well, learning every day.

PR #123, opened 2018-05-23
Clean up Input Test

xjdonne:
Clean up scripts: Less repetitive functions.
Remove useless Prefabs


ProjectSettings/EditorBuildSettings.asset
```diff
@@ -8,4 +8,16 @@ EditorBuildSettings:
...
+    guid: 61c2d094f0992ee4799446a1b5fd10a0
+  - enabled: 1
+    path: Assets/QA/Input_Test/Scenes/Controller Xbox.unity
+    guid: 8f5a18c7b34093e4b8e68821516bb499
+  - enabled: 1
+    path: Assets/QA/Input_Test/Scenes/Mouse Keyboard.unity
+    guid: 401d9a8859c068f4db3331d02e2854a1
+  - enabled: 1
+    path: Assets/QA/Input_Test/Scenes/Pen.unity
+    guid: bfca4f66fa1693b4fa56be497d438d52
```
Rene-Damm: Without a mechanism to choose these scenes in the player, this does not seem useful.

Would recommend adding a MenuItem that builds a QA player build with the QA test scenes.

JiaUnity: If the UI is supported, I'll put the drop down menu back.
I separated them apart because UI didn't work.

Rene-Damm: Which drop down menu?

What I meant is keeping the QA scenes out of player builds by default. They just increase build time and size and shouldn't really be in the demo scene build. Instead, let's have a button in the editor main menu that performs a QA build programmatically using just the QA scenes.

JiaUnity: Ahhh, I see what you mean. I misread. 
I'll remove them from the build settings.

PR #126, opened 2018-05-30
XInput Deadzone values

douglas-piconi:
I don't know if you want to merge this now to close those two specific Xbone controller bugs or do a PR with all platforms.  I only have decent xinput numbers.  The PS4 just doesn't drift at all.   I think we can just go with the default values for PS4 for now.


Rene-Damm:
>The PS4 just doesn't drift at all. I think we can just go with the default values for PS4 for now.

Probably means we can make the values tighter there instead of going with the defaults.

PR #129, opened 2018-05-31
NEW: Warn the user if the old input system is the active.

douglas-piconi:
If the new input system package is installed, but the old input system is the active player setting, it can cause confusion with the user.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -88,6 +88,14 @@ namespace UnityEngine.Experimental.Input
             }
         }
 
+        #if UNITY_EDITOR
+        public bool oldInputWarnUserFlag
+        {
+            get { return m_OldInputWarnUserFlag; }
+            set { m_OldInputWarnUserFlag = value; }
+        }
+        #endif
+
         public event DeviceChangeListener onDeviceChange
         {
             add { m_DeviceChangeListeners.Append(value); }
@@ -1227,6 +1235,7 @@ namespace UnityEngine.Experimental.Input
         [NonSerialized] internal IInputRuntime m_Runtime;
 
         #if UNITY_EDITOR
+        [NonSerialized] private bool m_OldInputWarnUserFlag;  // Let the user know if they have the old input manager set, but the new input system installed.
         [NonSerialized] internal IInputDiagnostics m_Diagnostics;
         #endif
 
@@ -2272,6 +2281,10 @@ namespace UnityEngine.Experimental.Input
             public InputUpdate.SerializedState updateState;
             public InputUpdateType updateMask;
 
+            #if UNITY_EDITOR
+            public bool oldInputWarnUserFlag;
+            #endif
+
             // The rest is state that we want to preserve across Save() and Restore() but
             // not across domain reloads.
 
@@ -2371,6 +2384,7 @@ namespace UnityEngine.Experimental.Input
                 runtime = m_Runtime,
 
                 #if UNITY_EDITOR
+                oldInputWarnUserFlag = m_OldInputWarnUserFlag,
                 diagnostics = m_Diagnostics
                 #endif
             };
@@ -2399,8 +2413,9 @@ namespace UnityEngine.Experimental.Input
             m_LayoutChangeListeners = state.layoutChangeListeners;
             m_EventListeners = state.eventListeners;
             m_UpdateMask = state.updateMask;
-
+            
             #if UNITY_EDITOR
+            m_OldInputWarnUserFlag = state.oldInputWarnUserFlag;
             m_Diagnostics = state.diagnostics;
             #endif
 
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -88,14 +88,6 @@ namespace UnityEngine.Experimental.Input
             }
         }
 
-        #if UNITY_EDITOR
-        public bool oldInputWarnUserFlag
-        {
-            get { return m_OldInputWarnUserFlag; }
-            set { m_OldInputWarnUserFlag = value; }
-        }
-        #endif
-
         public event DeviceChangeListener onDeviceChange
         {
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -88,6 +88,14 @@ public InputUpdateType updateMask
             }
         }
 
+        #if UNITY_EDITOR
+        public bool oldInputWarnUserFlag
```
Rene-Damm: This is better kept directly on InputSystemObject and should have a better name.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1096,6 +1096,23 @@ private static void RunInitializeInPlayer()
 #if UNITY_EDITOR
         private static InputSystemObject s_SystemObject;
 
+        private static void CheckCurrentPlayerSettingForInputSystem()
```
Rene-Damm: Method names like this are bad as it doesn't really tell what the method does. Check how? And do what with the check? (given there is no return value to indicate the result of the check)

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1096,6 +1096,23 @@ private static void RunInitializeInPlayer()
...
+            PlayerSettings[] array = Resources.FindObjectsOfTypeAll<PlayerSettings>();
+            var playerSettings = array.Length > 0 ? array[0] : null;
+            if (playerSettings)
+            {
+                var s_PlayerSettings = new SerializedObject(playerSettings);
+
+                // Only warn the user once.  Save the warning flag across domain reloads.
+                if (s_PlayerSettings.FindProperty("enableNativePlatformBackendsForNewInputSystem").boolValue == false && !s_Manager.oldInputWarnUserFlag)
+                {
+                     EditorUtility.DisplayDialog("Input Manager Active", "You currently have the input system package installed, however it is not active due to the Active Input Handling player setting set to Input Manager.", "OK");
```
Rene-Damm: This should both explain the situation to the user ("This project is using the new input system package but the native platform backends for the new input system are not enabled in the player settings. This means that no input from native devices will come through.") and also give the user a choice for what to do ("Do you want to enable the backends. Doing so requires a restart of the editor."; choice "Enable New&Old Backends", "Enable New Backends", "Do Nothing").

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: Also recommend moving the method into `InputSystemObject`.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/InputSystemObject.cs
```diff
@@ -16,9 +16,11 @@ internal class InputSystemObject : ScriptableObject, ISerializationCallbackRecei
         [SerializeField] public InputManager manager;
         [NonSerialized] public InputRemoting remote;
         [SerializeField] public RemoteInputPlayerConnection playerConnection;
+        [SerializeField] private bool oldInputSystemWarningTriggered;
```
Rene-Damm: Non-public fields use m_ convention.

Packages/com.unity.inputsystem/InputSystem/InputSystemObject.cs
```diff
@@ -40,6 +42,50 @@ public void ReviveAfterDomainReload()
...
+
+            if (serializedPlayerSettings != null)
+                return serializedPlayerSettings.FindProperty("enableNativePlatformBackendsForNewInputSystem").boolValue;
+
+            return false;
+        }
+
+        public void DisplayOldInputSystemWarningDialog()
+        {
+            var dialogText = "This project is using the new input system package but the native platform backends for the new input system are not enabled in the player settings." +
```
Rene-Damm: const string

Packages/com.unity.inputsystem/InputSystem/InputSystemObject.cs
```diff
@@ -40,6 +42,50 @@ public void ReviveAfterDomainReload()
...
+
+            return null;
+        }
+
+        public bool IsNewInputSystemActiveInPlayerSettings()
+        {
+            var serializedPlayerSettings = GetPlayerSerializedPlayerSettings();
+
+            if (serializedPlayerSettings != null)
+                return serializedPlayerSettings.FindProperty("enableNativePlatformBackendsForNewInputSystem").boolValue;
```
Rene-Damm: We could make this somewhat future proof by considering the backends enabled if the property isn't there.

Packages/com.unity.inputsystem/InputSystem/InputSystemObject.cs
```diff
@@ -40,6 +42,50 @@ public void ReviveAfterDomainReload()
             SetUpRemoting();
         }
 
+        private SerializedObject GetPlayerSerializedPlayerSettings()
+        {
+            PlayerSettings[] array = Resources.FindObjectsOfTypeAll<PlayerSettings>();
```
Rene-Damm: Use var. Could also use FirstOrDefault here. Okay to use Linq in editor code. With FirstOrDefault, this method becomes a single line of code.

Packages/com.unity.inputsystem/InputSystem/InputSystemObject.cs
```diff
@@ -40,6 +42,50 @@ public void ReviveAfterDomainReload()
...
+        {
+            var serializedPlayerSettings = GetPlayerSerializedPlayerSettings();
+
+            if (serializedPlayerSettings != null)
+                return serializedPlayerSettings.FindProperty("enableNativePlatformBackendsForNewInputSystem").boolValue;
+
+            return false;
+        }
+
+        public void DisplayOldInputSystemWarningDialog()
```
Rene-Damm: Won't really affect the old input system as such.

Rene-Damm (approved):
Nice!

PR #135, opened 2018-06-08
[XboxOne] Added ability to support multiple gamepads

martinh-unity:
Can read gamepadId and userId from each gamepad.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxOneGamepad.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxOneGamepad.cs
@@ -4,6 +4,7 @@ using UnityEngine.Experimental.Input.LowLevel;
 using UnityEngine.Experimental.Input.Utilities;
 using System.Runtime.InteropServices;
 using UnityEngine.Experimental.Input.Plugins.XInput.LowLevel;
+using System.Collections.Generic;
 
 ////TODO: player ID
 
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxOneGamepad.cs
```diff
@@ -161,6 +208,76 @@ protected override void FinishSetup(InputDeviceBuilder builder)
...
+        public int xboxOneUserId
+        {
+            get
+            {
+                UpdatePadSettings();
+                return m_XboxOneUserId;
+            }
+        }
+
+        public new static List<XboxOneGamepad> all
```
Rene-Damm: No lists. Use array for s_Devices and ReadOnlyArray as return type here.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxOneGamepad.cs
```diff
@@ -161,6 +208,76 @@ protected override void FinishSetup(InputDeviceBuilder builder)
...
+
+        protected override void OnRemoved()
+        {
+            base.OnRemoved();
+
+            for (int i = 0; i < s_Devices.Count; i++)
+            {
+                if (s_Devices[i] != null && s_Devices[i].gamepadId == gamepadId)
+                {
+                    s_Devices.RemoveAt(i);
```
Rene-Damm: With `OnAdded` doing `Add(this)`, seems like this should just be `Remove(this)`. When switched to array, use ArrayHelpers.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxOneGamepad.cs
```diff
@@ -161,6 +208,76 @@ protected override void FinishSetup(InputDeviceBuilder builder)
...
+                return m_GamepadId;
+            }
+        }
+
+        public int xboxOneUserId
+        {
+            get
+            {
+                UpdatePadSettings();
+                return m_XboxOneUserId;
```
Rene-Damm, line 226: This should be part of the device configuration. There's no way with the current API to know when the user has changed other than to constantly poll the value. Configuration getters should retrieve the current configuration via a device command / IOCTL and then cache it. When the configuration changes, the backend should send a configuration change event.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxOneGamepad.cs
martinh-unity: Really? The problem is it doesn't know how many there might be. Also returning an array from a list just allocates more memory and creates garbage. We do try to minimize that on the consoles as the garbage collect is more expensive on consoles. However I'll sort out this change.

Rene-Damm: You can use ArrayHelpers to allocate with capacity (same thing that list does internally). E.g. with a capacity of 10. With a list you get two heap allocations. For the price you pay for the List header object, you can easily pay for 5 extra entries in your array. And List.Append() is not guarantee at all to not create garbage. Internally, it just allocates with capacity as well (same thing you can do with ArrayHelpers).

Rene-Damm: Also, no returning mutable internal structures from public APIs.

Rene-Damm:
Please also add tests.

PR #148, opened 2018-06-19
Few touch tests + Xbox Controller mappings on Android

todi1856:
Android Game Controllers:
* Added correct map for Xbox Controller + test

Added two touch tests which are currently failing, so we can decide what to do:
* Devices_TouchesWithSameIdDontGetStuck
  two touches with same id happen during one frame, because of that touch is stuck in Stationary phase.
  Such behavior happened on Android, where it was legal for touches to have same id, when they happen one after another
  If we don't allow such behavior, can we add an exception or error in Development build?

* Devices_TouchesWithWrongTimestampCorrectlyRecognized
  Touch event which comes timestamp which is less than previous event is implicitly ignored. Happened on android due precision issues... Again, can input system in development build throw a warning that event is ignored?


Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -2032,6 +2032,78 @@ public void Devices_TouchesBecomeStationaryWhenNotMovedInFrame()
         Assert.That(device.allTouchControls[0].phase.ReadValue(), Is.EqualTo(PointerPhase.Stationary));
     }
 
+    [Test]
+    [Category("Devices")]
+    public void Devices_TouchesWithSameIdDontGetStuck()
```
Rene-Damm: Please name TODO_Devices_TouchesWithSameIdDontGetStuck. Regular tests should be green.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -2032,6 +2032,78 @@ public void Devices_TouchesBecomeStationaryWhenNotMovedInFrame()
...
+        InputSystem.Update();
+        InputSystem.Update();
+
+        Assert.That(device.activeTouches.Count, Is.EqualTo(0));
+        Assert.That(device.allTouchControls[0].phase.ReadValue(), Is.EqualTo(PointerPhase.None));
+    }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_TouchesWithWrongTimestampCorrectlyRecognized()
```
Rene-Damm: Same here. Prefix with TODO_.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
Rene-Damm: Alternatively we could also adopt a FIXME_ prefix for tests that isolate known bugs (unlike the more general "this still needs to be implement" kind of thing that the TODO_ tests generally revolve around).

todi1856:
@Rene-Damm prefixed with FIXME_

todi1856:
Closing for now, found more issues with gamepad

PR #150, opened 2018-06-20
Onscreen controller

douglas-piconi:
Here is the optimized version of the OnScreenController.

What could I be missing from the Unit Tests?   Once I changed this code to cache the buffers and remove the created device, the tests have blown up.    Everything works fine in the editor with a scene, but in the unit tests,  all kinds of crazy things are happening now.

If I run all the OnScreenTests  as a group, none of them pass, and I get all kinds of array out of bounds when it hits the InputSystem.RemoveDevice code as well as blows up the real devices and supported devices.   I get strange errors like "looks like a GamePad is trying to be removed but was never added"   even though it is in the device array.   Really strange.   Been debugging all day but can't seem figure out what is going on.   I am guessing I am missing some major set up code for the tests that I didnt need prior to the caching?   Or is it domain reload issues?


Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenControl.cs
```diff
@@ -25,18 +27,140 @@ namespace UnityEngine.Experimental.Input.Plugins.OnScreen
...
+        // it can clean up the memory and remove the device from the input system.
+        private static List<int> s_RegisteredOnScreenControls = new List<int>();
+
+        public string controlPath
+        {
+            get { return m_ControlPath; }
+            set
+            {
+                m_ControlPath = value;
+                if (enabled)
```
Rene-Damm: Check for whether m_Control has been initialized, not for enabled status.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenControl.cs
```diff
@@ -25,18 +27,140 @@ namespace UnityEngine.Experimental.Input.Plugins.OnScreen
...
+                    SetupInputControl();
+                }
+            }
+        }
+
+        [NonSerialized] internal InputControl m_Control;
+        [SerializeField] internal string m_ControlPath;
+
+        private int m_DeviceEventInfoIndex;
+        private string m_Layout;
```
Rene-Damm: Unused.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenControl.cs
```diff
@@ -25,18 +27,140 @@ namespace UnityEngine.Experimental.Input.Plugins.OnScreen
...
+
+        private static void RemoveOnScreenDevice(int id)
+        {
+            s_DeviceEventInfoArray[id].buffer.Dispose();
+            InputSystem.RemoveDevice(s_DeviceEventInfoArray[id].device);
+        }
+
+        private static InputControl RegisterInputControl(string controlPath, out int id)
+        {
+            var layout = InputControlPath.TryGetDeviceLayout(controlPath);
```
Rene-Damm: Need to check for success.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenControl.cs
```diff
@@ -25,18 +27,140 @@ namespace UnityEngine.Experimental.Input.Plugins.OnScreen
...
+            public InputDevice device;
+        }
+
+        private static InlinedArray<OnScreenDeviceEventInfo> s_DeviceEventInfoArray = new InlinedArray<OnScreenDeviceEventInfo>();
+
+        // We will have N devices mapped to X number of OnScreenControls
+        // Need to keep track and reference count how many controls are sharing
+        // the same device, so that when the last control is ready to be destroyed
+        // it can clean up the memory and remove the device from the input system.
+        private static List<int> s_RegisteredOnScreenControls = new List<int>();
```
Rene-Damm: This is more convoluted than it needs to be. Simply have a `public OnScreenControl firstControl` on `OnScreenDeviceEventInfo` and a `private OnScreenControl m_NextControlOnDevice;` in `OnScreenControl`. In `OnEnable` look up / create entry and add to list. In `OnDisable`, remove from list and nuke device if last OSC disappeared from entry.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenControl.cs
```diff
@@ -25,18 +27,140 @@ namespace UnityEngine.Experimental.Input.Plugins.OnScreen
...
+            s_DeviceEventInfoArray.Append(deviceEventInfo);
+
+            // Give the caller the index in the cache.
+            return s_DeviceEventInfoArray.length - 1;
+        }
+
+        private static void RemoveOnScreenDevice(int id)
+        {
+            s_DeviceEventInfoArray[id].buffer.Dispose();
+            InputSystem.RemoveDevice(s_DeviceEventInfoArray[id].device);
```
Rene-Damm, line 93: You're not actually removing the entry from s_DeviceEventInfoArray. This is why you're crashing. `buffer` remains initialized with the data you have disposed and the next test writes into the now invalid memory.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/OnScreenTests.cs
```diff
@@ -1,34 +1,113 @@
...
     [Test]
     [Category("Devices")]
-    public void TODO_Devices_CanCreateOnScreenDevice()
+    public void Devices_CanCreateOnScreenStick()
+    {
+        var gameObject = new GameObject();
+        var stick = gameObject.AddComponent<OnScreenStick>();
+        stick.controlPath = "/<Gamepad>/leftStick";
+        Assert.That(InputSystem.devices, Has.Exactly(1).TypeOf<Gamepad>());
+    }
```
Rene-Damm: Missing test that checks for cleanup when OSC is disabled.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenControl.cs
```diff
@@ -25,18 +29,160 @@ namespace UnityEngine.Experimental.Input.Plugins.OnScreen
...
+            {
+                if (s_DeviceEventInfoArray[deviceIndex].firstControl.m_NextControlOnDevice == null)
+                {
+                    RemoveOnScreenDevice(deviceIndex);
+                    s_DeviceEventInfoArray.RemoveAt(deviceIndex);
+                }
+                else
+                {
+                    // Not going to search entire list to match exact object.  Just remove head of list
+                    // Unit we get to no other devices left.
```
Rene-Damm, line 181: This isn't going to work.

PR #164, opened 2018-07-09
Switch develop

yebata:
- Changed interface name to `Switch`
- Removed redundant properties
- Made IOCTL commands in separated commands
- Added gyro support


Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -118,174 +110,351 @@ public NPadInputState WithButton(Button button, bool value = true)
...
+            var supportCommand = NpadDeviceIOCTLStopSixAxisSensor.Create();
 
-            base.FinishSetup(builder);
+            return ExecuteCommand(ref supportCommand);
         }
-    }
+
+        protected override void FinishSetup(InputDeviceBuilder builder)
+		{
+            base.FinishSetup(builder);
```
Rene-Damm: Please run format.sh or format.ps1 found in the repo root folder.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -40,29 +40,31 @@ public FourCC GetFormat()
...
-        ////FIXME: it's so counterintuitive that we have to set "layout" here even though we inherit the control from Gamepad
-        [InputControl(name = "leftStick", layout = "Stick")]
         [FieldOffset(4)]
         public Vector2 leftStick;
 
-        [InputControl(name = "rightStick", layout = "Stick")]
         [FieldOffset(12)]
         public Vector2 rightStick;
 
+        [InputControl(name = "acceleration")]
```
Rene-Damm, line 52: If the name matches the field name, no need to specify it on the attribute.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/SwitchTests.cs
```diff
@@ -14,25 +14,16 @@ public void Devices_SupportsSwitchNpad()
...
-        InputSystem.QueueStateEvent(controller,
-            new NPadInputState()
-            .WithLeftStick(new Vector2(0.123f, 0.456f))
-            .WithRightStick(new Vector2(0.789f, 0.987f)));
         InputSystem.Update();
 
-        Assert.That(controller.leftStick.x.ReadValue(), Is.EqualTo(0.123).Within(0.00001));
-        Assert.That(controller.leftStick.y.ReadValue(), Is.EqualTo(0.456).Within(0.00001));
-        Assert.That(controller.rightStick.x.ReadValue(), Is.EqualTo(0.789).Within(0.00001));
-        Assert.That(controller.rightStick.y.ReadValue(), Is.EqualTo(0.987).Within(0.00001));
```
Rene-Damm: Please add this back or add something equivalent. This tests whether the sticks come through on the gamepad.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -118,174 +110,351 @@ public NPadInputState WithButton(Button button, bool value = true)
...
+			}
+		}
+		public NpadId npadID
+		{
+			get
+			{
+                RefreshConfigurationIfNeeded();
+				return m_NpadId;
+			}
+		}
```
Rene-Damm: What's the difference between npadID and controllerID? Preferably add XML doc comments. Maybe there's a better name for controllerID?

Rene-Damm (changes requested):

Rene-Damm:
Please add tests.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
yebata: I imagined they can be different. NPad ID represents a port where a controller is connected to. For example, you can be a 1st player even if you use JoyCons which is connect as NPad ID 2, but it doesn't do anything for now. I will remove it.

Rene-Damm: That sounds useful. Maybe call it `portNumber` or `npadPortNumber`?

Also wondering, is there a relation between `userId` (coming from `InputDevice`) and `controllerID`? Might be worth looking at the Xbox One code for parallels.

Naming convention wise, think we should use `Id` instead of `ID` everywhere.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
yebata: Without them, test failed. "Cannot find input control 'acceleration'"

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
yebata: NpadId is a type name and it can be given to SDK via native plugins if developers want to do something with it, so it would be `npadId`.
I imagined a controller ID was just a number. `userId` seems to represent like a gamer tag? Anyway, I changed mind that a controller ID is unnecessary, so I removed it.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
yebata: Sorry forgot to do that. I hope it can be automated like Mercurial.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/SwitchTests.cs
yebata: Backed them, and added more tests.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
Rene-Damm: Interesting. That sounds like a bug. I'll give it a try after this PR is merged.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/SwitchTests.cs
```diff
@@ -49,16 +71,133 @@ public void Devices_SupportsSwitchNpad()
...
+    }
+
+    [Test]
+    [Category("Devices")]
+    public unsafe void Devices_CanStopSixAxisSensors()
+    {
+        var controller = InputSystem.AddDevice<NPad>();
+
+        NpadDeviceIOCTLStopSixAxisSensor? receivedCommand = null;
+        unsafe
```
Rene-Damm, line 182: Minor but this is redundant. If the method is unsafe, all code in the body can use unsafe features.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/SwitchTests.cs
```diff
@@ -23,16 +25,36 @@ public void Devices_SupportsSwitchNpad()
...
         Assert.That(controller.leftStick.x.ReadValue(), Is.EqualTo(0.123).Within(0.00001));
         Assert.That(controller.leftStick.y.ReadValue(), Is.EqualTo(0.456).Within(0.00001));
         Assert.That(controller.rightStick.x.ReadValue(), Is.EqualTo(0.789).Within(0.00001));
         Assert.That(controller.rightStick.y.ReadValue(), Is.EqualTo(0.987).Within(0.00001));
 
+        Assert.That(controller.acceleration.x.ReadValue(), Is.EqualTo(0.987).Within(0.00001));
+        Assert.That(controller.acceleration.y.ReadValue(), Is.EqualTo(0.654).Within(0.00001));
+        Assert.That(controller.acceleration.z.ReadValue(), Is.EqualTo(0.321).Within(0.00001));
+
+        Quaternion attitude = controller.attitude.ReadValue();
```
Rene-Damm, line 47: Please use `var` for local variables.

Rene-Damm (approved):

PR #179, opened 2018-07-19
Fix Devices_SupportsAndroidSonyDualShock test

todi1856:


Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/AndroidTests.cs
```diff
@@ -250,17 +250,22 @@ public void FIXME_Devices_SupportsAndroidSonyDualShock()
 
         Assert.That(gamepad.name, Is.EqualTo("AndroidGamepadDualShock"));
 
+        float rxValue = 0.123f;
+        float ryValue = 0.456f;
```
Rene-Damm, line 254: Use const.

PR #186, opened 2018-07-26
CHANGE: Enable HID/DualShock templates for UWP (WSA)

timkeo:


Rene-Damm (changes requested):
Looking good. Please add a test, though, that makes sure we're recognizing the DualShock as it looks on UWP. Something like Devices_SupportsDualShockOnUWP in DualShockTests.

PR #187, opened 2018-07-26
Remove XR tests.  These tests are migrating to https://gitlab.interna…

jackpr-unity:
Remove XR tests.

These tests are migrating to https://gitlab.interna…l.unity3d.com/xr/manualtests/input-system for the sake of Unity major release versioning.


Rene-Damm:
What's the reason for the move? Don't quite follow the "Unity major release versioning" thing.

As much as possible (i.e. wherever we're not prevented from doing so by NDAs), I'd like all tests (manual and automatic) to be part of the public repo and available together with the system for everyone to run. We had the situation before with the old system where no one outside of a couple people in QA really know how to run manual QA on the input system which not only led to devs not knowing how to manually verify but also led to unawareness of what was manually tested and what wasn't.

Put another way, separate repos for tests are shit.

jackpr-unity:
Why is this necessary? Backporting test projects for regression tests.

Opening Unity projects with earlier versions of the editor is not supported, and that has led to major issues for XR QA in the past.  Here are some things that have changed since 17.3, off the top of my head:
- The names of settings
- The name of the packages folder
- the default scripting API (3.5 -> 4)
- the way prefabs work (not talking about the new prefab flow.  Backporting a project with prefabs between certain versions causes all prefabs to become missing in all scenes).
- and on and on and on.  

The only way we have been able to run regression tests without having a big painful backport ramp-up every time is by maintaining separate branches for test projects tied to major releases.

Also, I have a constant communication stream with the XR devs.  They know how to use these projects, and if they have issues we solve them together.

Rene-Damm:
Jack and I talked it over and agreed to have the XR tests in a separate repo but keep a copy of them publicly available in the input system repo.

PR #191, opened 2018-07-31
FIX: Fix OnScreenStick to use proper coord space and have radial move…

douglas-piconi:
1. Fixed problem with Canvas using screen space overlay to screen space camera.
2. Fixed OnScreenStick class to use the proper coordinates.
3. OnScreenStick now has radial movement using the same logic that is in the StandardAssets package version of an on screen stick.


douglas-piconi:
Need to change the OnScreenStickTest

PR #192, opened 2018-07-31
Fix coordinate and rotation problem with OnScreenStick

douglas-piconi:
1. Fixed problem with Canvas using screen space overlay to screen space camera.
2. Fixed OnScreenStick class to use the proper coordinates.
3. OnScreenStick now has radial movement using the same logic that is in the StandardAssets package version of an on screen stick.
4. Added a camera and canvas to playmode test to support the coord changes.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
@@ -13,25 +13,22 @@ namespace UnityEngine.Experimental.Input.Plugins.OnScreen
     {
         private void Start()
         {
-            m_StartPos = (transform as RectTransform).anchoredPosition;
-            Canvas canvas = GetComponentInParent<Canvas>();
-            if (canvas)
-                m_Camera = canvas.worldCamera;
+            m_StartPos = ((RectTransform)transform).anchoredPosition;
         }
 
         public void OnPointerDown(PointerEventData data)
         {
-            RectTransformUtility.ScreenPointToLocalPointInRectangle(GetComponentInParent<RectTransform>(), data.position, m_Camera, out m_PointerDownPos);
+            RectTransformUtility.ScreenPointToLocalPointInRectangle(GetComponentInParent<RectTransform>(), data.position, data.pressEventCamera, out m_PointerDownPos);
         }
 
         public void OnDrag(PointerEventData data)
         {
             Vector2 position;
-            RectTransformUtility.ScreenPointToLocalPointInRectangle(GetComponentInParent<RectTransform>(), data.position, m_Camera, out position);
+            RectTransformUtility.ScreenPointToLocalPointInRectangle(GetComponentInParent<RectTransform>(), data.position, data.pressEventCamera, out position);
             Vector2 delta = position - m_PointerDownPos;
 
             delta = Vector2.ClampMagnitude(delta, movementRange);
-            (transform as RectTransform).anchoredPosition = m_StartPos + (Vector3)delta;
+            ((RectTransform)transform).anchoredPosition = m_StartPos + (Vector3)delta;
 
             var newPos = new Vector2(delta.x / movementRange, delta.y / movementRange);
             SendValueToControl(newPos);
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/OnScreenTests.cs
```diff
@@ -14,6 +14,8 @@ public void Devices_CanCreateOnScreenStick()
     {
         var gameObject = new GameObject();
         var eventSystem = gameObject.AddComponent<EventSystem>();
+        var camera = gameObject.AddComponent<Camera>();
+        var canvas = gameObject.AddComponent<Canvas>();
```
Rene-Damm: Variables don't seem to be used (will spill warnings).

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -13,47 +13,40 @@ public class OnScreenStick : OnScreenControl, IPointerDownHandler, IPointerUpHan
     {
         private void Start()
         {
-            m_StartPos = transform.position;
+            m_StartPos = (transform as RectTransform).anchoredPosition;
+            Canvas canvas = GetComponentInParent<Canvas>();
```
Rene-Damm: Use var for local variables.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -13,47 +13,40 @@ public class OnScreenStick : OnScreenControl, IPointerDownHandler, IPointerUpHan
...
+            var newPos = new Vector2(delta.x / movementRange, delta.y / movementRange);
             SendValueToControl(newPos);
-
-            transform.position = new Vector3(m_StartPos.x + newPos.x, m_StartPos.y + newPos.y, m_StartPos.z);
         }
 
         public void OnPointerUp(PointerEventData data)
         {
-            transform.position = m_StartPos;
+            (transform as RectTransform).anchoredPosition = m_StartPos;
```
Rene-Damm: Use direct cast `((RectTransform)transform)` when the cast mustn't fail. Otherwise you end up with an NullReferenceException instead of an InvalidCastException when things go wrong.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -13,47 +13,40 @@ public class OnScreenStick : OnScreenControl, IPointerDownHandler, IPointerUpHan
     {
         private void Start()
         {
-            m_StartPos = transform.position;
+            m_StartPos = (transform as RectTransform).anchoredPosition;
```
Rene-Damm: Probably better to do all this in `OnEnable()` rather than `Start()`.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -13,47 +13,40 @@ public class OnScreenStick : OnScreenControl, IPointerDownHandler, IPointerUpHan
...
+            Canvas canvas = GetComponentInParent<Canvas>();
+            if (canvas)
+                m_Camera = canvas.worldCamera;
```
Rene-Damm: Not sure caching this here is safe (especially in the current code that only looks things up once at start). Can we get this from `PointerEventData`.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/OnScreenTests.cs
```diff
@@ -23,8 +25,6 @@ public void Devices_CanCreateOnScreenStick()
         Assert.That(stick.control, Is.TypeOf<StickControl>());
         var stickControl = (StickControl)stick.control;
 
-        ////FIXME: OnScreenStick mixes up transform.position and 2D position space
-
```
Rene-Damm, line 27: Can we add a check for the transform?

douglas-piconi:
The only one of these items I didn't change is the Start() to OnEnable.   When I changed it to use OnEnable, everything broke, and it wasn't obvious why.  No events were sent to the input system, the stick moved up to a weird position on the screen.   I'll have to come back to that.

PR #216, opened 2018-08-09
Minor tweaks

todi1856:
* Remove touchType in TouchState, and introduce touchFlags instead, currently that will be only used for identifying Indirect touches
* Fix Devices_DualshockTriggersHaveCorrectDefaultValues, with introduction of defaultState, we can now properly initialize game controller values


Packages/com.unity.inputsystem/InputSystem/Controls/TouchFlagsControl.cs
```diff
@@ -4,23 +4,23 @@
 
 namespace UnityEngine.Experimental.Input.Controls
 {
-    public class TouchTypeControl : InputControl<TouchType>
+    public class TouchFlagsControl : InputControl<TouchFlags>
```
Rene-Damm: I wonder whether we shouldn't nuke this entirely. With this now being a bitfield, we could simply treat it as buttons.

The pen, for example, has an "in-range" bit. It's simply exposed as a button on the pen. Wondering whether we shouldn't just do the same thing here.

todi1856:
@Rene-Damm Did the way u suggested. Thx

PR #235, opened 2018-08-29
Noise filter and XR Device Changes

StayTalm:
General
-InputControl.ReadValueFrom(InputEventPtr) now was a ReadRawValueFrom(...) variation, and returns a bool to inform the caller whether or not a value was actually read from the event.  The value is now an out parameter.
-InputControl.GetStatePtrFromStateEvent now properly handles DeltaStateEvents
-Fixed an offset issue with DeltaStateEvents.  Offsets were not adding in the device offset on top of the event offset.
-Fixed issue with InlinedArrays and ToString() causing a crash.
-Test for delta event offsets

Noise Filter
-Input Devices now have NoiseFilters
    -Noise Filters prevent setting the current device, and udating the last device event time, when a noisy control is the only change, or when a change results in a no-operation after processors are applied
-Added simple bitmask helpers
-There is now a noisefilter buffer in the core InputStateBuffers block
-Tests for Noise Filters

XR
-Added simple support for buffered rumbling IOCTL commands
-Layout naming fixes for Vive, Oculus and WindowsMR
-More Accurate device matching for WindowsMR
-Removed duplicate parametric test cases for device matching.


Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -409,32 +409,53 @@ internal unsafe IntPtr GetStatePtrFromStateEvent(InputEventPtr eventPtr)
...
+                var stateEvent = StateEvent.From(eventPtr);
+
+                stateOffset = 0;
+                stateFormat = stateEvent->stateFormat;
+                stateSizeInBytes = stateEvent->stateSizeInBytes;
+                statePtr = stateEvent->state;
+            }
+            else
+            {
+                throw new ArgumentException("Event must be a state or delta state event", "eventPtr");
```
Rene-Damm, line 441: Remove respective throw above.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -515,24 +536,33 @@ public override void WriteValueFromObjectInto(IntPtr buffer, long bufferSize, ob
...
 
-            var value = ReadRawValueFrom(statePtr);
-            if (process)
-                value = Process(value);
+            value = ReadValueFrom(statePtr);
+            return true;
+        }
 
-            return value;
+        public bool ReadRawValueFrom(InputEventPtr inputEvent, out TValue value)
```
Rene-Damm: Maybe we should use "unprocessed" instead of "raw". Thinking that "raw" actually isn't all that clear after all.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -515,24 +536,33 @@ public override void WriteValueFromObjectInto(IntPtr buffer, long bufferSize, ob
             WriteRawValueInto(buffer, (TValue)value);
         }
 
-        ////REVIEW: this should return a bool and pass the value as an out parameter; bool should indicate
-        ////        whether value is actually coming from the event or just a default value
```
Rene-Damm: Hmm, now that I see this change implemented, looks quite ugly in the code. Will look better with C# 7 where we have `out var` so maybe it's okay. And yeah, API-wise probably still the right thing to do.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        /// <summary>
+        /// Simple cached value identifying how to filter this element (Bitmask, or individual type filtered).
+        /// </summary>
+        public enum EElementType
+        {
+            TypeUnknown = 0,
+            TypeFull,
+            TypeFloat,
+            TypeVec2
```
Rene-Damm: No prefixes on enum values. They aren't in the parent namespace like in C++.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+    /// <summary>
+    /// A filter for individual devices to check if events or device states contain significant, relevant changes.
+    /// Irrelevant changes are any updates controls that are tagged as 'noisy', and any state change that turns into a no-operation change once processors are applied (e.g. in deadzone joystick movements).
+    /// </summary>
+    public class NoiseFilter
+    {
+        /// <summary>
+        /// Simple cached value identifying how to filter this element (Bitmask, or individual type filtered).
+        /// </summary>
+        public enum EElementType
```
Rene-Damm: No "E" prefixes on enums.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+using UnityEngine.Experimental.Input.Processors;
+using UnityEngine.Experimental.Input.Utilities;
+
+namespace UnityEngine.Experimental.Input
+{
+    /// <summary>
+    /// A filter for individual devices to check if events or device states contain significant, relevant changes.
+    /// Irrelevant changes are any updates controls that are tagged as 'noisy', and any state change that turns into a no-operation change once processors are applied (e.g. in deadzone joystick movements).
+    /// </summary>
+    public class NoiseFilter
```
Rene-Damm: If we have this, has to be InputNoiseFilter.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -106,12 +106,22 @@ protected override void OnRemoved()
...
+            m_Rumble = new SimpleRumble(this);
+        }
+
+        protected override void OnAdded()
+        {
+            base.OnAdded();
+
+            m_BufferedRumble = new BufferedRumble(this);
+            HapticCapabilities capabilities = m_BufferedRumble.capabilities;
+            Debug.Log(string.Format("XR Device Found with Haptic Capabilites: NumChannels:[{0}] FrequencyHz:[{1}] MaxBufferSize:[{2}]", capabilities.numChannels, capabilities.frequencyHz, capabilities.maxBufferSize));
```
Rene-Damm: No debug logs like this.

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
```diff
@@ -0,0 +1,83 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine.Experimental.Input.LowLevel;
+using UnityEngine.Experimental.Input.Processors;
+using UnityEngine.Experimental.Input.Utilities;
+
+namespace UnityEngine.Experimental.Input
+{
+    public unsafe struct BitmaskHelpers
```
Rene-Damm: Make internal. Should be part of the public input system APi.

Rene-Damm: Preferably move to memory helpers which already has some bit-related functionality.

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            MarkInBuffer(filterBuffer, control.stateBlock.byteOffset, control.stateBlock.sizeInBits, true);
+        }
+
+        static public void Blacklist(IntPtr filterBuffer, InputControl control)
+        {
+            MarkInBuffer(filterBuffer, control.stateBlock.byteOffset, control.stateBlock.sizeInBits, false);
+        }
+
+        static public void MarkInBuffer(IntPtr filterBuffer, uint byteOffset, uint sizeInBits, bool state)
```
Rene-Damm: The above three methods need better names. Unclear what bit operations they actually perform. A general helper should not operate in the terminology of a higher subsystem (e.g. "filterBuffer").

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        /// <summary>
+        /// Checks an Input Event for any significant changes that would be considered user activity.
+        /// </summary>
+        /// <param name="inputEvent">The input event being checked for changes</param>
+        /// <param name="device">The input device being checked against </param>
+        /// <param name="offset">The offset into the device that the event is placed</param>
+        /// <param name="sizeInbytes">The size of the event in bytes</param>
+        /// <returns>True if any changes exist in the event once the device has been filtered through for noise and non-significant changes.  False otherwise.</returns>
+        public unsafe bool HasValidData(InputDevice device, InputEventPtr inputEvent, uint offset, uint sizeInbytes)
```
Rene-Damm: `sizeInBytes`

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+        {
+            if (elements.Length == 0)
+                return true;
+
+            if ((offset + sizeInbytes) * 8 > device.stateBlock.sizeInBits)
+                return false;
+
+            bool result = false;
```
Rene-Damm: Use `var` on all local variables.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+
+            IntPtr noiseFilterPtr = InputStateBuffers.s_NoiseFilterBuffer;
+            if (noiseFilterPtr != IntPtr.Zero)
```
Rene-Damm: Prefer early returns over nesting.

```
if (noiseFilterPtr == IntPtr.Zero)
    return false;
```

not only avoids deeper nesting, it also answers the question about how that case is handled much more concisely. The nesting requires jumping around in the code to find out what happens.

Rene-Damm: Incidentally, this seems weird. Why are we saying there's no data if there's no noise filter? Shouldn't that mean it's all data and no noise?

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+                    StateEvent* stateEvent = StateEvent.From(inputEvent);
+                    ptrToEventState = stateEvent->state;
+                }
+                else if (inputEvent.IsA<DeltaStateEvent>())
+                {
+                    DeltaStateEvent* stateEvent = DeltaStateEvent.From(inputEvent);
+                    ptrToEventState = stateEvent->deltaState;
+                }
+
+                if (ptrToEventState != IntPtr.Zero)
```
Rene-Damm: When is `ptrToEventState` being `IntPtr.Zero` a valid scenario? I assume this is to catch us getting an event with no state. If so, either explicitly handle that as an `else` above or prohibit that in general for the method and throw an exception from an argument check early in the method (probably preferable).

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+                {
+                    result = BitmaskHelpers.CheckForMaskedValues(ptrToEventState, noiseFilterPtr, offset, sizeInbytes * 8);
+                    for (int i = 0; i < elements.Length && !result; i++)
+                    {
+                        result = elements[i].HasValidData(inputEvent, device);
+                    }
```
Rene-Damm: I think code like this is much easier to follow like this

```
                    if (BitmaskHelpers.CheckForMaskedValues(ptrToEventState, noiseFilterPtr, offset, sizeInbytes * 8))
                        return true;
                    for (int i = 0; i < elements.Length; i++)
                    {
                        if (elements[i].HasValidData(inputEvent, device))
                            return true;
                    }
```

No `result` variable that is hard to tell meaning from and no having to decipher the boolean handling through the loop.

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                *filterIter |= mask;
+            }
+            else
+            {
+                *filterIter &= ~mask;
+            }
+        }
+
+        static public bool CheckForMaskedValues(IntPtr eventBuffer , IntPtr maskPtr, uint offsetBytes, uint sizeInBits)
```
Rene-Damm: Visibility comes first, then static, i.e. `public static bool`.

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            if (eventBuffer == IntPtr.Zero || maskPtr == IntPtr.Zero)
```
Rene-Damm: Accepting these as valid arguments seems odd. If code is screwing up, don't hide it. Make it fail.

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
Rene-Damm: Recommend using plain pointers instead of `IntPtr`. I want to do a pass to get rid of pretty much all use of IntPtr in the system. The ECS guys have already switched away from it (e.g. `UnsafeUtility` now takes `void*` instead of `IntPtr`). `IntPtr` just adds needless extra cost.

Rene-Damm: Think the name can be improved on. Maybe `HasAnyNonZeroBitsAfterMaskingWithBuffer` or something like that.

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            uint mask = (uint)((1 << (int)sizeRemaining) - 1);
+            if (state)
+            {
+                *filterIter |= mask;
+            }
+            else
+            {
+                *filterIter &= ~mask;
+            }
```
Rene-Damm: This is potentially writing out of bounds.

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
Rene-Damm: `MarkInBuffer` ==> `SetBits`? `state` => `value`

Packages/com.unity.inputsystem/InputSystem/Utilities/BitmaskHelpers.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        static public void Whitelist(IntPtr filterBuffer, InputControl control)
+        {
+            MarkInBuffer(filterBuffer, control.stateBlock.byteOffset, control.stateBlock.sizeInBits, true);
+        }
+
+        static public void Blacklist(IntPtr filterBuffer, InputControl control)
+        {
+            MarkInBuffer(filterBuffer, control.stateBlock.byteOffset, control.stateBlock.sizeInBits, false);
+        }
```
Rene-Damm: These two methods don't seem to add value.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -409,32 +409,53 @@ internal unsafe IntPtr GetStatePtrFromStateEvent(InputEventPtr eventPtr)
...
 
-            var stateEvent = StateEvent.From(eventPtr);
 
             // Make sure we have a state event compatible with our device. The event doesn't
             // have to be specifically for our device (we don't require device IDs to match) but
             // the formats have to match and the size must be within range of what we're trying
             // to read.
-            var stateFormat = stateEvent->stateFormat;
-            if (stateEvent->stateFormat != device.m_StateBlock.format)
+            if (stateFormat != device.m_StateBlock.format)
```
Rene-Damm, line 449: Exception below should say "DeltaStateEvent" instead of "StateEvent" if it's a delta event.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -409,32 +409,53 @@ internal unsafe IntPtr GetStatePtrFromStateEvent(InputEventPtr eventPtr)
             if (!eventPtr.IsA<StateEvent>() && !eventPtr.IsA<DeltaStateEvent>())
                 throw new ArgumentException("Event must be a state or delta state event", "eventPtr");
 
-            ////TODO: support delta events
```
Rene-Damm: Test please.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+        public enum EElementType
+        {
+            TypeUnknown = 0,
+            TypeFull,
+            TypeFloat,
+            TypeVec2
+        }
+
+        /// <summary>
+        /// A filter for a single InputControl that gets it's value filtered.
```
Rene-Damm: "its"

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+        /// </summary>
+        public struct FilteredElement
+        {
+            /// <summary>
+            /// The index in InputDevice.allControls for the filtered control.
```
Rene-Damm: Use `<see cref="...">` to link to elements.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_NoiseFilter.cs
```diff
@@ -0,0 +1,305 @@
+using NUnit.Framework;
```
Rene-Damm: Preferably move these tests to `CoreTests_Devices`.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_NoiseFilter.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        InputSystem.QueueStateEvent(device1, new GamepadState { leftStick = new Vector2(1.0f, 1.0f) });
+        InputSystem.Update();
+
+        Assert.AreEqual(Gamepad.current, device1);
+    }
+
+    [Test]
+    [Category("NoiseFilter")]
+    public void NoiseFilter_NoisyControlsDetectedOnDeltaStateEvents()
```
Rene-Damm: `NoiseFilter_CanDetectNoisyControlsOnDeltaStateEvents`

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_NoiseFilter.cs
```diff
@@ -0,0 +1,305 @@
...
+            return new FourCC('T', 'E', 'S', 'T');
+        }
+
+        [FieldOffset(0)] public short button1;
+        [FieldOffset(2)] public short button2;
+    }
+
+    [Test]
+    [Category("NoiseFilter")]
+    public void NoiseFilter_NoisyControlInLayoutAppliesToDevices()
```
Rene-Damm: `NoiseFilter_CanMarkControlAsNoisy`

Redundant with existing test `Layouts_CanMarkControlAsNoisy`.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -136,22 +169,6 @@ public unsafe void Events_CanInitializeStateEventFromDevice()
...
-        // Adding a device makes it current so add another one so that .current
-        // is not already set to the gamepad we just created.
-        InputSystem.AddDevice("Gamepad");
-
-        InputSystem.QueueStateEvent(gamepad, new GamepadState());
-        InputSystem.Update();
-
-        Assert.That(Gamepad.current, Is.SameAs(gamepad));
-    }
-
```
Rene-Damm, line 157: Nuke/replace test below.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -136,22 +169,6 @@ public unsafe void Events_CanInitializeStateEventFromDevice()
...
-        // Adding a device makes it current so add another one so that .current
-        // is not already set to the gamepad we just created.
-        InputSystem.AddDevice("Gamepad");
-
-        InputSystem.QueueStateEvent(gamepad, new GamepadState());
-        InputSystem.Update();
-
-        Assert.That(Gamepad.current, Is.SameAs(gamepad));
-    }
```
Rene-Damm, line 156: Not 100% sure about this one. Seems like just one more gotcha to take into account. `current` has certain aspects of `any`.

Either way, think we can revisit this as part of looking at `current` in relation to user management. I think there's still good changes for `current` to go away entirely.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -187,6 +187,26 @@ public override Type valueType
             get { return typeof(byte[]); }
         }
 
+        public NoiseFilter noiseFilter
+        {
+            get
+            {
+                return m_NoiseFilter;
+            }
+            set
```
Rene-Damm: If we want to allow setting custom noise filters on individual devices, it has to be an API in `InputSystem` and the information has to be persisted as part of `InputManager.SerializedState`. This also includes any modifications that are allowed to be made to a noise filter (like resetting it).

Rene-Damm: Overall, one thing that bugs me a little is that it seems like noise filtering should be control-level functionality rather than device-level functionality. It seems like it should be possible to take any arbitrary control, point it at some state data and ask "do you have meaningful state in there?". The restriction to devices seems to be imposed more by implementation constraints.

Packages/com.unity.inputsystem/InputSystem/State/InputStateBuffers.cs
```diff
@@ -79,6 +79,11 @@ internal struct InputStateBuffers
         /// </summary>
         public IntPtr defaultStateBuffer;
 
+        /// <summary>
+        /// Buffer that contains bitflags for noisy and non-noisy controls, to identify significant device changes.
+        /// </summary>
+        public IntPtr noiseFilterBuffer;
```
Rene-Damm: `noiseBitmaskBuffer`

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
Rene-Damm: Not fond of this being a class or this being separated out into a class. We add an additional reference to InputDevice (still hoping we can nuke all references on InputControl and InputDevice) and all this class contains is one reference to a C# array. Sort of the typical C# setup of tiny fragmented objects that end up all over the place.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+                filter.elements[j] = elementsToAdd[j];
+            }
+
+            return filter;
+        }
+
+        /// <summary>
+        /// The list of elements to be checked for.  Each element represents a single InputControl.
+        /// </summary>
+        public FilteredElement[] elements;
```
Rene-Damm: Shouldn't expose its raw array. Anyone can modify this and have it go out of sync with data elsewhere. Use ReadOnlyArray to expose arrays.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        /// <summary>
+        /// Called when the NoiseFilter gets applied to a device, calls down to any individual FilteredElements that need to do any work.
+        /// </summary>
+        /// <param name="device">The device you want to apply filtering to.</param>
+        internal void Apply(InputDevice device)
```
Rene-Damm, line 144: `ApplyToNoiseBitmask`?

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+                    elements[i].Apply(noiseFilterPtr, device);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Resets a device to unfiltered
+        /// </summary>
+        /// <param name="device">The device you want reset</param>
+        internal void Reset(InputDevice device)
```
Rene-Damm: `ResetNoiseBitmask`?

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+
+                for (int i = 0; i < elements.Length; i++)
+                {
+                    elements[i].Apply(noiseFilterPtr, device);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Resets a device to unfiltered
```
Rene-Damm: Doc comments are useful if they do explain things :) Preferable to leave things undocumented in that case.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -854,6 +854,8 @@ public void AddDevice(InputDevice device)
             if (device.updateBeforeRender)
                 updateMask |= InputUpdateType.BeforeRender;
 
+            device.noiseFilter = NoiseFilter.CreateDefaultNoiseFilter(device);
```
Rene-Damm: We shouldn't create noise filters for devices that don't need them.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2079,9 +2085,17 @@ internal unsafe void OnUpdate(InputUpdateType updateType, int eventCount, IntPtr
...
+                                new IntPtr(deviceBuffer.ToInt64() + stateBlockOfDevice.byteOffset),
                                 sizeOfStateToCopy, offsetInDeviceStateToCopyTo);
 
+
+                        var deviceStateOffset = device.m_StateBlock.byteOffset + offsetInDeviceStateToCopyTo;
+
+                        InputEventPtr eventPtr = new InputEventPtr(currentEventPtr);
+                        NoiseFilter filter = device.noiseFilter;
+                        bool hasSignificantControlChanges = filter != null ? filter.HasValidData(device, eventPtr, deviceStateOffset, sizeOfStateToCopy) : true;
+                        doNotMakeDeviceCurrent |= !hasSignificantControlChanges;
```
Rene-Damm: Comment for block.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+            TypeUnknown = 0,
+            TypeFull,
+            TypeFloat,
+            TypeVec2
+        }
+
+        /// <summary>
+        /// A filter for a single InputControl that gets it's value filtered.
+        /// </summary>
+        public struct FilteredElement
```
Rene-Damm: `FilterElement`?

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+    /// </summary>
+    public class NoiseFilter
+    {
+        /// <summary>
+        /// Simple cached value identifying how to filter this element (Bitmask, or individual type filtered).
+        /// </summary>
+        public enum EElementType
+        {
+            TypeUnknown = 0,
+            TypeFull,
```
Rene-Damm: `Full` as a name isn't very clear to me. Can't really tell the difference between `full` vs `float` etc.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: Also, should probably be set up by `InputDeviceBuilder`. It's odd behavior that calling `InputSystem.AddDevice(myDevice)` initializes its noise filter. Especially with the current API where the user could have set a custom noise filter in-between creating and adding the device.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -187,6 +187,26 @@ public override Type valueType
             get { return typeof(byte[]); }
         }
 
+        public NoiseFilter noiseFilter
```
Rene-Damm: I think if we have this property, the name needs to be refined to point out what the filter applies to. From the name, I would assume that it, well, filters out noise whereas in reality, it doesn't really filter anything but only impacts `current` and doesn't affect what data we store and return.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
Rene-Damm:
```
public enum FilterType
{
     EntireControl,
     FloatBelowEpsilon,
     Vector2MagnitudeBelowEpsilon
}
```

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
Rene-Damm: In general, no abbreviations. (some exceptions like "Ptr" are fine)

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
```diff
@@ -0,0 +1,246 @@
...
+                        {
+                            InputControl<Vector2> castedControl = control as InputControl<Vector2>;
+                            if (castedControl != null)
+                            {
+                                Vector2 value;
+                                if (castedControl.ReadValueFrom(inputEvent, out value) && Vector2.SqrMagnitude(value) > float.Epsilon)
+                                    return true;
+                            }
+                        }
+                        break;
```
Rene-Damm: This logic seems like it should be *on* the control rather than having type switches here. Maybe `InputControl.IsAtDefaultValue()` or something. Unfortunately, the virtual calls are starting to accumulate in there but oh well.

Rene-Damm (commented):
Great work! Hope my litany of comments isn't discouraging :)

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Events.cs
Rene-Damm: Actually, realizing now that I confused this with `Devices_AddingDevice_MakesItCurrent` and that you basically moved this test and updated it for noise. Ignore my comment.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
Rene-Damm: (Also, "flow comments" generally appreciated on blocks. `// Here-I'm-doing-what-because-of-why.`)

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
StayTalm: Going through the tests, I definitely needed some kind of 'Did Read' check, so I think it's a bit of an ugly necessity atm.

StayTalm:
Pinging, this should be unblocked now after the 2 commits at the end.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
StayTalm: This one I'm not so sure.  
I want to keep it open, in case future changes to the NoiseFilter need to do something other than just set the bitmask on apply.

Packages/com.unity.inputsystem/InputSystem/Devices/NoiseFilter.cs
StayTalm: Again, I'd like to keep it named more on the context it's called in, so that future edits don't need function naming changes.  The NoiseFilter Resets here, what it resets is private knowledge.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2471,6 +2488,12 @@ private bool FlipBuffersForDeviceIfNecessary(InputDevice device, InputUpdateType
             return flipped;
         }
 
+        internal struct NoiseFilterElementState
```
Rene-Damm, line 2491: Didn't catch this before. This won't work without `[Serializable]`. So essentially, the current domain reload survival logic doesn't really do anything for noise filter :)

Recommend adding a test that makes sure we can successfully take noise filters across.

Issue #234, opened 2018-08-29
Sensor orientations differ between iOS and Android


Rene-Damm:
Gyro attitude X and Y are inverse between iOS and Android, for example. As much as possible, we should standardize sensor orientations the same way we standardize gamepad stick and pointer axis orientations.

Chippit:
For some extra information as I have experienced the problem, the orientation of the devices seem to have a big impact here, and perhaps the system needs to compensate for the different orientation somehow, and also perhaps do some handedness conversions, because that seems to differ sometimes too?

My expectations are that the attitude's x represents the device's pitch in its current orientation, y its yaw, and z its roll. However, my experience is that z is always yaw, x is roll and y is pitch, and that the roll and pitch orientation changes depending on whether or not you hold the device home-left or home-right (which is maybe expected?).

For example, on a Sony Xperia Z1, when I have the device in landscape, home-right and facing up, if I tilt the device forward (away from me), I get a positive y rotation. On an iPad Air 2 in the same orientation, I get a negative y. The Air held home-left, however, has the same results.

On a third generation Fire HDX, I receive those same values if I hold the device in portrait with home-bottom.

todi1856:
New Input system does the compensation already, assuming InputConfiguration.CompensateSensorsForScreenOrientation is true (which is default value).

The returned values should be the same as in old input system.

renaldasrep:
@Chippit are you accessing the whole Quaternion or it's separate values?
There's a related known issue for it https://github.com/Unity-Technologies/InputSystem/issues/125

Fiddling around with this I found that on android the Quaternion values work solid, however on iOS there's a weird way to reproduce an issue resulting in 'incorrectly placed' values

1. Run app in portrait mode
2. Turn device -90 degrees (no repro, home button is on the right)
3. Turn device back to portrait orientation
4. Turn device +90 degrees (landscape, home button is on the left)
5. Turn device +90 degrees (reverse portrait)
6. Turn device +90 degrees (landscape, same as step 2)
7. REPRO

Expected value example: (0.0, 0.0, 0.0, -1.0)
Actual value example: (0.0, 0.0, -1.0, 0.0)

Notice that the last two values switched places

Chippit:
[...]

PR #240, opened 2018-08-31
Fix issue - https://github.com/Unity-Technologies/InputSystem/issues/234 . Target ARM64 iOS instead ARM7

todi1856:
Arm7 is basically obsolete


Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidGameController.cs
```diff
@@ -32,9 +32,8 @@ public unsafe struct AndroidGameControllerState : IInputStateTypeInfo
         [InputControl(name = "select", bit = (uint)AndroidKeyCode.ButtonSelect, variants = kVariantGamepad)]
         public fixed uint buttons[(kMaxAndroidButtons + 31) / 32];
 
-
-        [InputControl(name = "leftTrigger", offset = (uint)AndroidAxis.Ltrigger * sizeof(float) + kAxisOffset, variants = kVariantGamepad)]
-        [InputControl(name = "rightTrigger", offset = (uint)AndroidAxis.Rtrigger * sizeof(float) + kAxisOffset, variants = kVariantGamepad)]
+        [InputControl(name = "leftTrigger", offset = (uint)AndroidAxis.Brake * sizeof(float) + kAxisOffset, variants = kVariantGamepad)]
```
todi1856, line 35: Not sure why this is being shown as a change, because it matches the code in develop branch

PR #248, opened 2018-09-17
IME Support

StayTalm:
Support for IME in the New Input System.  Adds to the Keyboard, where you can turn IME on/off, get the composition string, check if it's currently active, and set the IME dialog position.

A windows implementation is here: https://ono.unity3d.com/unity/unity/pull-request/74110/_/input/new/ime


Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -829,7 +884,21 @@ public override void OnTextInput(char character)
...
+                for (var i = 0; i < m_ImeCompositionStringListeners.length; ++i)
+                    m_ImeCompositionStringListeners[i](imeString);
+            }
+        }
+
         internal InlinedArray<Action<char>> m_TextInputListeners;
         private string m_KeyboardLayoutName;
+
+        internal InlinedArray<Action<string>> m_ImeCompositionStringListeners;
+        bool m_imeEnabled;
```
todi1856: m_ImeEnabled

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2576,7 +2582,7 @@ internal SerializedState SaveState()
             {
                 var device = m_Devices[i];
                 string[] usages = null;
-                if (device.usages.Count > 0)
+                if(device.usages.Count > 0)
```
Rene-Damm: Please run format before committing.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -3010,7 +3010,7 @@ public void Devices_NoisyControlDoesNotUpdateCurrentDevice()
 
         var device1 = InputSystem.AddDevice("MyDevice");
         var device2 = InputSystem.AddDevice("MyDevice");
-        var lastUpdateTime = device1.lastUpdateTime;
+        double lastUpdateTime = device1.lastUpdateTime;
```
Rene-Damm: Use var.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -3109,12 +3109,12 @@ public void Devices_CanDetectNoisyControlsOnDeltaStateEvents()
 
         Assert.That(NoisyInputDevice.current == device2);
 
-        InputSystem.QueueDeltaStateEvent(device1["first"], short.MaxValue);
+        InputSystem.QueueDeltaStateEvent<short>(device1["first"], short.MaxValue);
```
Rene-Damm: Bad merge. Seems like there's a number of changes here that simply undo changes I've committed. The explicit type arg is redundant here.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -3177,7 +3177,7 @@ public void Devices_CanOverrideDefaultNoiseFilter()
         // Tag the entire device as noisy
         device1.userInteractionFilter = new InputNoiseFilter
         {
-            elements = new[]
+            elements = new InputNoiseFilter.FilterElement[]
```
Rene-Damm: Same here. Redundant.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -829,7 +884,21 @@ public override void OnTextInput(char character)
                 m_TextInputListeners[i](character);
         }
 
+        public override void OnIMEStringEvent(IMECompositionStringEvent imeEvent)
+        {
+            if (m_ImeCompositionStringListeners.length > 0)
+            {
+                string imeString = imeEvent.AsString();
```
Rene-Damm: Old system does this but we shouldn't. Using input must not create garbage. Our API should allow creating a `string` but only if the user requests. This API has to allow going through the IME input character-by-character without creating a `string`.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -333,6 +337,49 @@ public class Keyboard : InputDevice
...
+            set
+            {
+                if (m_ImePosition != value)
+                {
+                    SetIMECursorPositionCommand command = SetIMECursorPositionCommand.Create(value);
+                    if (ExecuteCommand(ref command) >= 0)
+                        m_ImePosition = value;
+                }
+            }
+            get { return m_ImePosition; }
```
Rene-Damm: IME state neither sits fully in device state nor fully in configuration state. Will get lost on domain reload and become inconsistent.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -515,6 +563,11 @@ public string keyboardLayout
         public KeyControl oem4Key { get; private set; }
         public KeyControl oem5Key { get; private set; }
 
+        /// <summary>
+        /// True when IME composition is enabled.  Requires imeEnabled to be set to true, and the user to enable it at the OS level.
+        /// </summary>
```
Rene-Damm: Use <see> to create links. Also good habit to link related APIs with <seealso>.

In general, the state of Unity API documentation is aggravating. You get one short blurb pretending to explain the API and then nothing. No links to related docs, no details, nothing.

For example, in the doc comments in this commit, I can't find a comment on what IME actually is. Where it's supported. Why I would use it. What the APIs that comprise the feature are.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -138,6 +138,7 @@ public static FourCC kFormat
         [InputControl(name = "OEM3", layout = "Key", bit = (int)Key.OEM3)]
         [InputControl(name = "OEM4", layout = "Key", bit = (int)Key.OEM4)]
         [InputControl(name = "OEM5", layout = "Key", bit = (int)Key.OEM5)]
+        [InputControl(name ="IsIMESelected", layout = "Button", bit = (int)Key.IsIMESelected)]
```
Rene-Damm: Let's drop the "Is".

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -264,6 +264,10 @@ public virtual void OnTextInput(char character)
         {
         }
 
+        public virtual void OnIMEStringEvent(IMECompositionStringEvent imeEvent)
```
Rene-Damm: If we really want to pass an event as an arg, this should be a ref arg or pointer. However, not sure we really should be passing the event along. Input events are really low-level memory buffers. Would prefer we pass higher-level data. Ideally a struct that wraps around the character-by-character data and also can be passed to listeners as is.

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/SetIMECursorPositionCommand.cs
```diff
@@ -0,0 +1,39 @@
...
+        public static FourCC Type { get { return new FourCC('I', 'M', 'E', 'P'); } }
+
+        [FieldOffset(0)]
+        public InputDeviceCommand baseCommand;
+
+        [FieldOffset(InputDeviceCommand.kBaseCommandSize)]
+        float xPosition;
+
+        [FieldOffset(InputDeviceCommand.kBaseCommandSize + sizeof(float))]
+        float yPosition;
```
Rene-Damm: Can just put Vector2 inside here. It's blittable.

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/SetIMECursorPositionCommand.cs
```diff
@@ -0,0 +1,39 @@
...
+        public FourCC GetTypeStatic()
+        {
+            return Type;
+        }
+
+        public static SetIMECursorPositionCommand Create(Vector2 position)
+        {
+            return new SetIMECursorPositionCommand
+            {
+                baseCommand = new InputDeviceCommand(Type, InputDeviceCommand.kBaseCommandSize + (sizeof(float) * 2)),
```
Rene-Damm: Can use `sizeof(Vector2)`.

Packages/com.unity.inputsystem/InputSystem/Events/IMECompositionStringEvent.cs
```diff
@@ -0,0 +1,49 @@
...
+        public int size;
+
+        [FieldOffset(InputEvent.kBaseEventSize + sizeof(int))]
+        public fixed char buffer[kIMECharBufferSize];
+
+        /// <summary>
+        /// Returns the composition as a string.  Call this only if needed, as it generates garbage.
+        /// </summary>
+        /// <returns> The composition string at the current point in time.</returns>
+        public unsafe string AsString()
```
Rene-Damm: Override `ToString` or rename. Confusing to have both `ToString` and `AsString`.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/SetIMECursorPositionCommand.cs
StayTalm: Used a kSize constant like other files, but I can't use sizeof(Vector2), because I can't use it for the [StructLayout(LayoutKind.Explicit, Size = kSize)] at the top.  I don't want to have 2 different ways of calculating size.

Packages/com.unity.inputsystem/InputSystem/Events/IMECompositionStringEvent.cs
StayTalm: Removed, fixed with new structure to make getting the Composition string as a fixed buffer

StayTalm:
All changes made

Packages/com.unity.inputsystem/InputSystem/Events/IMECompositionEvent.cs
```diff
@@ -0,0 +1,64 @@
...
+                }
+            }
+        }
+
+        [FieldOffset(0)]
+        int size;
+
+        [FieldOffset(sizeof(int))]
+        fixed char buffer[LowLevel.IMECompositionEvent.kIMECharBufferSize];
+    }
```
Rene-Damm, line 63: This needs a `ToString()` override and ideally should be `IEnumerable<char>`.

PR #259, opened 2018-09-25
IME PR Feedback

StayTalm:
Added a missing .meta file.
Also made IMEComposition override ToString and inherit from IEnumerable<char> for usability.


Rene-Damm (changes requested):
No tests :)

EDIT: Actually, I should check my eyesight. Cancel that comment...

Rene-Damm (approved):

PR #260, opened 2018-09-26
Initial ISX TPD Push

mattdatwork:
This adds the initial version of the ISX TPD which uses loose actions to track things.
Also includes initial version of unit tests.

documentation to follow once I figure out what format to put it in.


Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,237 @@
...
+using UnityEngine;
+using UnityEngine.Experimental.Input;
+using UnityEngine.Experimental.Input.Controls;
+using UnityEngine.XR.PoseProvider;
+using UnityEngine.XR;
+
+namespace UnityEngine.Experimental.Input.XR
+{
+
+    [DefaultExecutionOrder(-30000)]
```
StayTalm: Could be a good design to put this as a constant in the TrackedPoseDriver itself.
E.g. public const int kExecutionOrder = -30000;

That way someone who wants something to run before or after the TrackedPose driver could just do
[DefaultExecutionOrder(TrackedPoseDriver.kExecutionOrder+1)]

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,237 @@
...
+        UpdateType m_UpdateType = UpdateType.UpdateAndBeforeRender;
+        public UpdateType updateType
+        {
+            get { return m_UpdateType; }
+            set { m_UpdateType = value; }
+        }
+
+        [SerializeField]
+        InputAction m_PositionAction;
+        public InputAction positionAction {  get { return m_PositionAction;  } set { m_PositionAction = value; m_PositionBound = false;  BindActions(); } }
```
StayTalm: You enable the new action in BindActions(), but do you need to disable the previously assigned action?

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,237 @@
...
+            {
+                m_RotationAction.Disable();
+                m_RotationAction.performed -= OnRotationUpdate;
+                m_RotationBound = false;
+            }
+        }
+
+        void OnPositionUpdate(InputAction.CallbackContext context)
+        {
+            if (m_PositionBound)
```
StayTalm: This almost seems like an assert, if you get a callback but position is not bound, something went wrong internally.

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        protected virtual void Awake()
+        {           
+#if ENABLE_VR
+            if (HasStereoCamera())
+            {
+                XRDevice.DisableAutoXRCameraTracking(GetComponent<Camera>(), true);
+                Camera c = GetComponent<Camera>();
+                Debug.Log(c.fieldOfView);
```
StayTalm: This should be deleted

Packages/manifest.json
```diff
@@ -1,10 +1,14 @@
...
     "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "2.0.16",
-    "com.unity.package-manager-ui": "1.9.11",
+    "com.unity.analytics": "3.0.9",
+    "com.unity.collab-proxy": "1.2.9",
+    "com.unity.package-manager-ui": "2.0.0-preview.7",
     "com.unity.purchasing": "2.0.3",
-    "com.unity.textmeshpro": "1.2.4",
+    "com.unity.textmeshpro": "1.3.0",
+    "com.unity.xr.oculus.standalone": "1.28.0",
```
StayTalm: This got added the moment you went to try it out, but I think it should be removed for the PR.

StayTalm (changes requested):
Looks good could be a bug on switching actions at runtime though (not disabling the overwritten action).  If that's the case be sure to add a test for that too.

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
Rene-Damm: Would recommend staying away from execution order entirely. Looks like the component wants to process in sync with input so we should simply hook it in there. ATM there's no callback to process stuff after events have been processed, only one happening before. I think we should just add this callback and then the TPD would update as part of input updating.

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        void OnRotationUpdate(InputAction.CallbackContext context)
+        {
+            if (m_RotationBound)
+            {
+                var quatControl = context.control as QuaternionControl;
+                if (quatControl != null)
+                {
+                    m_CurrentRotation = quatControl.ReadValue();
```
Rene-Damm: Perform `ReadValue` on context, not on control.

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        void OnPositionUpdate(InputAction.CallbackContext context)
+        {
+            if (m_PositionBound)
+            {
+                var vec3Control = context.control as Vector3Control;
+                if (vec3Control != null)
+                {
+                    m_CurrentPosition = vec3Control.ReadValue();
```
Rene-Damm: `ReadValue` on context.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -1,6 +1,8 @@
 {
     "name": "Unity.InputSystem",
-    "references": [],
+    "references": [
+    	"UnityEngine.XR.PoseProvider"
```
Rene-Damm: What's in this and where does this come from?

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        public void Reset(GameObject go)
+        {
+            go.transform.position = new Vector3(0.0f, 0.0f, 0.0f);
+            go.transform.rotation = new Quaternion(0.0f, 0.0f, 0.0f, 0.0f);
+        }
+
+        [Test]
+        public void TPDUpdateOptionTest()
```
Rene-Damm: No `XXXTest()` tests. All tests should express the functionality that is being established or the behavior that is being verified. `XXXTest` is a bad pattern. You can't know what the test is really about until you've actually read it.

For guidance, take a look at the other tests. Good names are things like `CanUpdateStateOfDeviceWithEvent`. Out of a good name usually follows better content, too. `XXXTest()` tend to lack focus by never establishing what exactly is being verified.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
```diff
@@ -0,0 +1,155 @@
...
+            tpd.poseProviderComponent = tpp;
+            return tpp;
+        }
+
+
+        [TearDown]
+        public void TearDown()
+        {
+
+        }
```
Rene-Damm: Remove.

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,237 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Experimental.Input;
+using UnityEngine.Experimental.Input.Controls;
+using UnityEngine.XR.PoseProvider;
+using UnityEngine.XR;
+
+namespace UnityEngine.Experimental.Input.XR
```
Rene-Damm: Put in `UnityEngine.Experimental.Input.Plugins.XR`. Move file respectively.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
```diff
@@ -0,0 +1,155 @@
...
+                tmp.position = testpos;
+                tmp.rotation = testrot;
+                output = tmp;
+                return true;
+            }
+        }
+
+        internal static TestTrackedPoseDriverWrapper CreateGameObjectWithTPD()
+        {
+            GameObject go = new GameObject();
```
Rene-Damm: Use `var` for all local variables.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
```diff
@@ -0,0 +1,155 @@
+using UnityEngine;
+using UnityEngine.TestTools;
+using NUnit.Framework;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine.SpatialTracking;
+using System.Linq;
+using UnityEngine.XR.PoseProvider;
+
+namespace UnityEngine.Experimental.Input.XR
```
Rene-Damm: Move this file respectively to TrackedPoseDriver.cs. Kill namespace on tests. Just put them in the global namespace.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
```diff
@@ -0,0 +1,155 @@
...
+            Assert.That(tpd.positionAction.bindings[0].path, Is.EqualTo("<Gamepad>/buttonSouth"));
+            tpd.positionAction = action2;
+            Assert.That(tpd.positionAction.bindings[0].path, Is.EqualTo("<Gamepad>/buttonNorth"));
+
+            
+            tpd.rotationAction = action1;
+            Assert.That(tpd.rotationAction.bindings[0].path, Is.EqualTo("<Gamepad>/buttonSouth"));
+            tpd.rotationAction = action2;
+            Assert.That(tpd.rotationAction.bindings[0].path, Is.EqualTo("<Gamepad>/buttonNorth"));        
+        }
```
Rene-Damm: This test doesn't test anything meaningful about TrackedPoseDriver. The checks are on functionality sitting in the core input system and covered by tests elsewhere.

There should be a test that triggers an action and then checks that TrackedPostDeriver has set the resulting transform as expected.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
mattdatwork: which callback should i use here? this sounds _way_ better than hoping re: priority.

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
mattdatwork: good catch. doing that.

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
mattdatwork: agreed

Packages/manifest.json
mattdatwork: agreed

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
mattdatwork: ok!

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
mattdatwork: ok!

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
mattdatwork: single monobehaviour (BasePoseProvider) which is shared between the old/new TPD's. useful for arm models / etc.  also pretty handy for testing as i can create a pose provider and send data to the tpd and make sure it does the right thing. can potentially remove if we've got a better option for things like arm models and tracking random data.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
mattdatwork: On it.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
mattdatwork: Ok

Packages/com.unity.inputsystem/InputSystem/Monobehaviors/TrackedPoseDriver.cs
mattdatwork: On it.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
mattdatwork: this makes me sad, but doing it.

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
mattdatwork: Ok

Packages/com.unity.inputsystem/Tests/InputSystem/Monobehaviors/TrackedPoseDriverTests.cs
mattdatwork: I was trying to ensure that I correctly allowed changing the binding (as there's code in there which checks if the binding is bound and i want to make sure people dont screw that up). i think the better option is to do as tom suggests and check to ensure i unbind the option before i re-bind it.

Packages/manifest.json
```diff
@@ -1,10 +1,12 @@
...
   "dependencies": {
     "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "2.0.16",
-    "com.unity.package-manager-ui": "1.9.11",
+    "com.unity.analytics": "3.0.9",
+    "com.unity.collab-proxy": "1.2.9",
+    "com.unity.package-manager-ui": "2.0.0-preview.7",
     "com.unity.purchasing": "2.0.3",
-    "com.unity.textmeshpro": "1.2.4",
+    "com.unity.textmeshpro": "1.3.0",
```
Rene-Damm: Please revert changes to this file.

mattdatwork: on it.

mattdatwork: reverted to 492289b7d33dba5269844c73c2c958112463bf07

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,240 @@
...
+                case InputUpdateType.Dynamic:
+                    OnUpdate();
+                    break;
+                case InputUpdateType.BeforeRender:
+                    OnBeforeRender();
+                    break;                
+            }
+        }
+
+        protected virtual void FixedUpdate()
```
StayTalm: Do we really want to update in FixedUpdate?
At the moment it's no different than the Dynamic update data, and can run multiple times per frame, wasting time setting the same data repeatedly.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,240 @@
+using System;
+using UnityEngine.XR;
+
+namespace UnityEngine.Experimental.Input.Plugins.XR
+{   
+    [Serializable]
+    [AddComponentMenu("XR/Tracked Pose Driver (New Input System)")]
+    public class TrackedPoseDriver: MonoBehaviour
+    {
+
```
StayTalm: Extra Space in here.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -0,0 +1,240 @@
...
+        private bool HasStereoCamera()
+        {
+            var camera = GetComponent<Camera>();
+            return camera != null && camera.stereoEnabled;
+        }
+
+        protected virtual void PerformUpdate()
+        {
+            if (!enabled)
```
StayTalm: This shouldn't be necessary.  You use OnEnable/OnDisable to hook up the update loop.  That means it should always be true when hitting this.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/TrackedPoseDriverTests.cs
```diff
@@ -0,0 +1,116 @@
...
+       
+        public void Reset(GameObject go)
+        {
+            go.transform.position = new Vector3(0.0f, 0.0f, 0.0f);
+            go.transform.rotation = new Quaternion(0.0f, 0.0f, 0.0f, 0.0f);
+        }
+
+        [Test]
+        [Category("TrackedPoseDriver")]
+        public void TPD_UpdateOptions_EffectProcessedData()
```
StayTalm: Affect instead of effect.

StayTalm (commented):
Small notes, all just preference or formatting, so it looks good to me.

StayTalm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
mattdatwork: removed

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
mattdatwork: removed

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
mattdatwork: removed

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/TrackedPoseDriverTests.cs
mattdatwork: changed

PR #268, opened 2018-10-03
Fix AxisComposite returning 1 if no buttons are pressed

jkdhn:
Return 0 if both values are equal (both pressed or none pressed)


Rene-Damm:
Thank you for the fix :)

PR #276, opened 2018-10-04
New asset editor iteration

tomekpaszek:


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -238,6 +238,14 @@ public int GetControlSchemeIndex(string name)
             return m_ControlSchemes[index];
         }
 
+        public void RemoveControlScheme(string name)
+        {
+            if (string.IsNullOrEmpty(name))
+                throw new ArgumentNullException("name");
+
+            ArrayHelpers.Erase(ref m_ControlSchemes, GetControlScheme(name));
```
Rene-Damm: Use `GetControlScheme` index and `ArrayHelpers.EraseAt`. The current code first looks up the control scheme and then does an `Equals` comparison of each scheme by going through the default equality comparer. All stuff that can be avoided.

Also, please extend `Actions_CanAddControlSchemeToAsset` to `Actions_CanAddAndRemoveControlSchemeToAsset` test to cover this method.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/ActionInspectorContextMenu.cs
```diff
@@ -0,0 +1,171 @@
...
+                selectedRow = selectedRow.parent;
+            }
+            while (selectedRow.parent != null);
+
+            return null;
+        }
+
+        private ActionMapTreeItem GetSelectedActionMapLine()
+        {
+            TreeViewItem selectedRow = m_ActionMapsTree.GetSelectedRow();
```
Rene-Damm, line 158: Please use `var` for all local variables.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/ActionInspectorContextMenu.cs
```diff
@@ -0,0 +1,171 @@
+using System;
```
Rene-Damm, line 2: Missing `#if UNITY_EDITOR`.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/AddControlSchemePopup.cs
```diff
@@ -0,0 +1,367 @@
+using System;
```
Rene-Damm, line 2: Missing `#if UNITY_EDITOR`.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/InputActionTreeBase.cs
```diff
@@ -0,0 +1,243 @@
+using System;
```
Rene-Damm, line 2: Missing `#if UNITY_EDITOR`.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/InputActionWindowToolbar.cs
```diff
@@ -0,0 +1,223 @@
+using System;
```
Rene-Damm, line 2: Missing `#if UNITY_EDITOR`.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/Editor/OnScreenControlEditor.cs
```diff
@@ -1,4 +1,5 @@
 #if UNITY_EDITOR
+using System;
```
Rene-Damm: Doesn't seem to be used.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/InputActionAssetManager.cs
```diff
@@ -0,0 +1,148 @@
...
+            }
+            // Duplicate the asset along 1:1. Unlike calling Clone(), this will also preserve
+            // GUIDs.
+            m_AssetObjectForEditing = Object.Instantiate(importedAsset);
+            m_AssetObjectForEditing.hideFlags = HideFlags.HideAndDontSave;
+            m_AssetObjectForEditing.name = importedAsset.name;
+            m_SerializedObject = new SerializedObject(m_AssetObjectForEditing);
+        }
+
+        public void UnloadAssets()
```
Rene-Damm: Minor but this isn't really unloading something. It's just destroying a transient object.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/AddControlSchemePopup.cs
```diff
@@ -0,0 +1,367 @@
...
+            m_DevicesReorderableList.onRemoveCallback += OnDeviceRemove;
+        }
+
+        void OnDeviceRemove(ReorderableList list)
+        {
+            list.list.RemoveAt(list.index);
+            list.index = -1;
+        }
+
+        void OnDeviceAdd(ReorderableList list)
```
Rene-Damm, line 102: No way we can re-use the control picker here and just have an option on it to only show devices and not controls? I'm just thinking that with all the improvements the control picker is still mean to receive, it'd be great to see them here as well. And to also have stuff like filtering here. Plus, I think later we want to have some global preferences that affect what we even consider for the control picker and it'd be great to have that automatically come through here as well.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/ActionInspectorContextMenu.cs
tomekpaszek: It needs to be specific type in this case.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionAsset/AddControlSchemePopup.cs
tomekpaszek: We could but in don't think it makes sense in the current form. When we switch to AdvancedDropdown, the control will have to be (partially) rewritten anyway

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1985,10 +1985,14 @@ public void Actions_CanFindControlSchemeInAssetByName()
 
     [Test]
     [Category("Actions")]
-    [Ignore("TODO")]
-    public void TODO_Actions_CanRemoveControlSchemeFromAsset()
+    public void Actions_CanRemoveControlSchemeFromAsset()
```
Rene-Damm, line 1988: Ah nice, didn't even remember that one was there.

PR #277, opened 2018-10-04
NEW: Scripts for running automated tests on 3 versions

douglas-piconi:
Using Henricks unity-downloader tool, and testrunner scripts.   Can now run various versions of tests
on different OS's.   Right now this will run in GitLab CI.   Having this work is a prerequisite to converting the the script to run on Yamato.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,11 +1,51 @@
 stages:
-- test
+  - test
 
-Windows-2018.2-default:
+test:windows:2018-2:
   stage: test
-  image: input-system/win-2018.2:stable
+  image: 'build-pipeline/addressables_win:stable'
+  artifacts:
+    paths:
+      - "TestArtifacts"
+    expire_in: 1 week
+    when: always
   tags:
-    - bokken-job
+   - bokken-runner
   script:
-  - cd "C:\Program Files\Unity\Editor"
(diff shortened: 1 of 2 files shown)


testrunner.py
```diff
@@ -0,0 +1,105 @@
+import sys
```
Rene-Damm, line 1: As we're adding more stuff to the root folder, as some point we probably want to start moving some stuff to a Tools/ folder or something but for I think this is just fine.

.gitlab-ci.yml
```diff
@@ -1,11 +1,51 @@
...
+    expire_in: 1 week
+    when: always
   tags:
-    - bokken-job
+   - bokken-runner
   script:
-  - cd "C:\Program Files\Unity\Editor"
-  - unity.exe -batchmode -automated -quit -nographics -runTests -projectPath "C:\users\bokken\builds\vm\InputSystem" -testPlatform playmode
\ No newline at end of file
+    - python.exe -u testrunner.py StandaloneWindows64 StandaloneWindows Editor --version=2018.2
```
Rene-Damm, line 15: Can we add Android to the list?

Rene-Damm: Ah doh, just realized we're not just building here but running tests. Can we add a command to just perform an Android build? Worrying about actually running on a device IMO is fine to keep for later but even just having an Android build on the Windows machine and an iOS build on Mac would be tremendously useful I think.

Rene-Damm:
Doug, I love you.

PR #278, opened 2018-10-05
Switch develop

yebata:
Swapped struct values which were in incorrect order. The Main color should be first, and then the sub color.


Rene-Damm (changes requested):
Please add tests.

yebata:
Added a test for the controller colors.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/SwitchTests.cs
```diff
@@ -107,6 +112,17 @@ public unsafe void Devices_CanUpdateStatus()
         Assert.That(controller.npadId, Is.EqualTo(NPad.NpadId.Handheld));
         Assert.That(controller.orientation, Is.EqualTo(NPad.Orientation.Vertical));
         Assert.That(controller.styleMask, Is.EqualTo(NPad.NpadStyle.Handheld));
+        Assert.That(controller.leftControllerColor.Main, Is.EqualTo((Color32)Color.red));
+        Assert.That(controller.leftControllerColor.Sub, Is.EqualTo((Color32)Color.black));
+        Assert.That(controller.rightControllerColor.Main, Is.EqualTo((Color32)Color.cyan));
+        Assert.That(controller.rightControllerColor.Sub, Is.EqualTo((Color32)Color.gray));
+    }
+
+    private int ColorToNNColor(Color color)
```
Rene-Damm, line 121: Would prefer for this to sit next to the conversion function in NPad.cs that goes the opposite way.

yebata: Right. I will move it into the NPad class.

PR #282, opened 2018-10-09
Simple, Action-Based uGUI Implementation

StayTalm:
This gets uGUI functionality almost up to par with the Standalone Input Module (Minus Touch Support) including testing and XML comments.

The design is a little bit unusual, but I think it will help solve a lot of extensibility problems, and problems with uGUI itself.  I've moved the event sending to the UIInputModule, and wrapped it up in helpers that map to a mouse and joystick model.  I'm still tied to replicating mice and joysticks, because that's what uGUI responders expect, but now you can simply update the state of the model, and the event ordering and internal data is done for you consistently.  Since each touch is a new replication of mouse behaviour, this should be easily extensible, and should we need a VR model we can add support there too.

I can't truly get rid of the uGUI guts, but I CAN make the model more easily extensible for support for explicit devices or controls, and more consistent from module to module.


Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -94,17 +104,163 @@ public void TDO_Actions_CanDriveUI()
         uiModule.leftClick = new InputActionProperty(leftClickAction);
         uiModule.middleClick = new InputActionProperty(middleClickAction);
         uiModule.rightClick = new InputActionProperty(rightClickAction);
+        uiModule.scrollWheel = new InputActionProperty(scrollAction);
 
         // Enable the whole thing.
         map.Enable();
 
+        yield return null;
```
Rene-Damm, line 112: Why do we need to wait a frame? Think this warrants a comment.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIActionInputModule.cs
```diff
@@ -199,6 +326,29 @@ private void UnhookActions()
...
+
+            oldProperty = newProperty;
+
+            if (oldProperty != null)
+            {
+                if (m_ActionsHooked)
+                    oldProperty.action.performed += m_ActionCallback;
+
+                if (m_ActionsEnabled)
+                    oldProperty.action.Enable();
```
Rene-Damm: Enabling and disabling actions should not be under the control of the UI input module. The input module should solely be a consumer.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MouseModel.cs
```diff
@@ -0,0 +1,317 @@
...
+using UnityEngine.EventSystems;
+
+namespace UnityEngine.Experimental.Input.Plugins.UI
+{
+    /// <summary>
+    /// A series of flags to determine if a button has been pressed or released since the last time checked.
+    /// Useful for identifying press/release events that occur in a single frame or sample.
+    /// </summary>
+    [Flags]
+    public enum ButtonDeltaState
```
Rene-Damm: What's the case for exposing the APIs in here? Seems like everything in here should be internal and the input modules being the only thing that the plugin exports.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/JoystickModel.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.EventSystems;
+
+namespace UnityEngine.Experimental.Input.Plugins.UI
+{
+    /// <summary>
+    /// Represents the state of a joystick in the uGUI system. Keeps track of various book-keeping regarding UI selection, and move and button states.
+    /// </summary>
+    public struct JoystickModel
```
Rene-Damm: Same question as for the things in MouseModel.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MouseModel.cs
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    public struct MouseModel
+    {
+        public struct InternalData
+        {
+            /// <summary>
+            /// This tracks the current GUI targets being hovered over.  Syncs up to <see cref="PointerEventData.hovered"/>.
+            /// </summary>
+            public List<GameObject> hoverTargets { get; set; }
```
Rene-Damm: No Lists please. Either InlinedArray or plain array.

(Doesn't seem to be used, though)

Packages/com.unity.inputsystem/InputSystem/Events/IMECompositionEvent.cs
```diff
@@ -34,7 +34,7 @@ public FourCC GetTypeStatic()
 namespace UnityEngine.Experimental.Input
 {
     [StructLayout(LayoutKind.Explicit, Size = sizeof(int) + sizeof(char) * IMECompositionEvent.kIMECharBufferSize)]
-    public unsafe struct IMECompositionString : IEnumerable<char>
+    public unsafe struct IMEComposition : IEnumerable<char>
```
Rene-Damm: Is the revert intentional?

Rene-Damm (changes requested):
Nice!!

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIInputModule.cs
```diff
@@ -16,30 +16,360 @@ namespace UnityEngine.Experimental.Input.Plugins.UI
...
+                }
+            }
+
+            // if we have not changed hover target
+            if (eventData.pointerEnter == currentPointerTarget && currentPointerTarget)
+                return;
+
+            var commonRoot = FindCommonRoot(eventData.pointerEnter, currentPointerTarget);
+
+            // and we already an entered object from last time
```
Rene-Damm: Minor thing but please use proper English orthography in comments. Full sentences starting upper and ended with a dot.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIInputModule.cs
```diff
@@ -16,30 +16,360 @@ namespace UnityEngine.Experimental.Input.Plugins.UI
...
+                return;
+
+            Vector2 movement = joystickState.move;
+            if (!usedSelectionChange && (!Mathf.Approximately(movement.x, 0f) || !Mathf.Approximately(movement.y, 0f)))
+            {
+                float time = Time.unscaledTime;
+
+                Vector2 moveVector = joystickState.move;
+
+                MoveDirection moveDirection = MoveDirection.None;
```
Rene-Damm: Use `var` for all local variables.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MouseModel.cs
StayTalm: It's used with in HandleEnterAndExit.  Will see what I can do with inlined array, the number of entries in the list is really hard to predict or cap.

Packages/com.unity.inputsystem/InputSystem/Events/IMECompositionEvent.cs
StayTalm: Yes, since it's not a string under the hood or in how it's accessed, I removed all references to the word string, except in ToString.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MouseModel.cs
Rene-Damm: If InlinedArray isn't appropriate, going with just a plain array plus capacity is fine.

BTW note re mutability. Doesn't matter for this API if it goes internal but no external APIs should expose mutable data structures (think we'll have to fix that in NoiseFilter, too).

Packages/com.unity.inputsystem/InputSystem/Events/IMECompositionEvent.cs
Rene-Damm: Heh in that case, please revert :) It's an IEnumerable<char> which makes it a string.

Packages/com.unity.inputsystem/InputSystem/Events/InputEventBuffer.cs
```diff
@@ -59,7 +59,7 @@ public long capacityInBytes
                 if (!m_Buffer.IsCreated || m_BufferEnd == kBufferSizeUnknown)
                     return 0;
 
-                return m_BufferEnd - m_Buffer.Length;
+                return m_Buffer.Length - m_BufferEnd;
```
mattdatwork, line 62: I cant figure out how to get a larger diff, is this a correct change? nothing else in this file changed, so im assuming this fixes a bug somewhere.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIActionInputModule.cs
StayTalm: I don't agree with this one.

UIActionInputModule is a standalone component, and therefore shouldn't require extra code or scripts to just drop and use.  And that means that it has to Enable/Disable it's properties on OnEnable/OnDisable.

And if it's activating it's controls on Enable/Disable, it should also manage those same controls when swapping.

Rene-Damm: >And if it's activating it's controls on Enable/Disable, it should also manage those same controls when swapping.

Yup agree. But I think in this setup it shouldn't activate them in OnEnable or deactivate them in OnDisable either.

I think you have a point though and IMO it's something to revisit. Especially if we allow *defining* actions directly on the module, it seems inconsistent for actions not to activate by themselves along with the UI.

For now, though, to fit into the bigger and still only emerging picture of action management, I'd like to go with action orchestration happening entirely outside of the UI module and leave the module as just a consumer.

Packages/com.unity.inputsystem/InputSystem/Events/InputEventBuffer.cs
StayTalm: This is returning the amount available in a semi-used buffer.  
BufferEnd is the current in-use amount, and Buffer.Length is the entire container.
This means that buffer end is always smaller then (or equal to) the total potential space, so it would result in a negative capacity.

We want the gap between the total capacity and the current end of the buffer.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/UITests.cs
StayTalm: Added a comment.
The canvas doesn't iterate over the graphics objects to determine depth ordering until the next frame.  So returning null let's the graphics get ordered so the raycaster can hit them.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/JoystickModel.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            get
+            {
+                return m_CancelButtonDown;
+            }
+            set
+            {
+                if (cancelButtonDown != value)
+                {
+                    cancelButtonDelta = value ? ButtonDeltaState.Pressed : ButtonDeltaState.Released;
```
mattdatwork: assume this is the return value?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MouseModel.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// Call this at the end of polling for per-frame changes.  This resets delta values, such as <see cref="deltaPosition"/>, <see cref="scrollDelta"/>, and <see cref="MouseButtonModel.lastFrameDelta"/>.
+        /// </summary>
+        public void OnFrameFinished()
+        {
+            changedThisFrame = false;
+            deltaPosition = scrollDelta = Vector2.zero;
+            m_LeftButton.OnFrameFinished();
+            m_RightButton.OnFrameFinished();
+            m_MiddleButton.OnFrameFinished();
+        }
```
mattdatwork, line 310: is there a way to make sure this gets called? as things will break if they dont

mattdatwork (approved):
couple of questions, and rene had some changes. otherwise it looks decent!

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIActionInputModuleEnabler.cs
```diff
@@ -0,0 +1,79 @@
+using UnityEngine;
```
Rene-Damm, line 1: This one shouldn't be part of the plugin.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -94,17 +103,164 @@ public void TDO_Actions_CanDriveUI()
...
+        InputSystem.Update();
+        eventSystem.InvokeUpdate();
+
+        Assert.That(leftChildReceiver.events, Has.Count.EqualTo(1));
+        Assert.That(leftChildReceiver.events[0].type, Is.EqualTo(EventType.Move));
+        leftChildReceiver.Reset();
+        Assert.That(rightChildReceiver.events, Has.Count.EqualTo(0));
+
+        // Check Submit
+        InputSystem.QueueStateEvent(gamepad, new GamepadState { buttons = (uint)(1 << (int)GamepadState.Button.South) });
```
Rene-Damm, line 226: There's a helper for buttons. `new GamepadState().WithButton(GamepadState.Button.South)`. I'm pulling GamepadState.Button out to the toplevel so that it bcomes `new GamepadState().WithButton(GamepadButton.South)`.

PR #283, opened 2018-10-10
Move controller and Aim controller support

martinh-unity:


Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadPS4.cs
```diff
@@ -1,506 +1,602 @@
...
-                case 3:
-                    return Color.magenta;
-                default:
-                    return Color.black;
-            }
-        }
-    }
-}
-#endif // UNITY_EDITOR || UNITY_PS4
+#if UNITY_EDITOR || UNITY_PS4
```
Rene-Damm: Please run format.sh/format.ps1 found in root of repo.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/MoveControllerSupport.cs
```diff
@@ -0,0 +1,20 @@
+using UnityEngine.Experimental.Input.Layouts;
```
Rene-Damm: Please move PS4 stuff to new Plugins/PS4 folder.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/PS4Tests.cs
```diff
@@ -0,0 +1,105 @@
...
+using UnityEngine.Experimental.Input.Plugins.DualShock;
+using UnityEngine.Experimental.Input.Plugins.DualShock.LowLevel;
+using UnityEngine.Experimental.Input.Processors;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.Experimental.Input.Layouts;
+using UnityEngine.TestTools.Utils;
+
+
+public class PS4Tests : InputTestFixture
```
Rene-Damm, line 12: As part of moving PS4 stuff to its own plugin, please move PS4-specific tests over from DualShockSupport.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/MoveControllerPS4.cs
```diff
@@ -0,0 +1,380 @@
...
+            var capabilities = description.capabilities;
+            var deviceDescriptor = PS4MoveDeviceDescriptor.FromJson(capabilities);
+
+            if (deviceDescriptor != null)
+            {
+                m_SlotId = (int)deviceDescriptor.slotId;
+                m_MoveIndex = (int)deviceDescriptor.moveIndex;
+                m_DefaultColorId = (int)deviceDescriptor.defaultColorId;
+                m_PS4UserId = (int)deviceDescriptor.userId;
+            }
```
Rene-Damm: I think for now this is fine. There's a change that has happened recently which I think will require us to more cleanly separate certain pieces of device setup/configuration. In order to not create garbage when the user unplugs devices and plugs them back in, the system now retains devices that have been reported by the runtime and have been removed through device remove events. So we may end up re-using InputDevice instances in certain cases.

Right now, the descriptor actually has to match 100% so there shouldn't be a problem in this particular case.

At some point, I think we probably want to make sure that all one-time configuration on devices gets reset when they are removed.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/PS4/DualShockGamepadPS4.cs
```diff
@@ -306,7 +332,10 @@ public int slotIndex
         {
             get
             {
-                UpdatePadSettingsIfNeeded();
+#if !UNITY_2019_1_OR_NEWER
+                UpdatePadSettingsIfNeeded();  // 2018.3 uses IOCTL to read this. not required in later versions
```
Rene-Damm, line 336: Probably meant 2018.2?

Packages/com.unity.inputsystem/InputSystem/Plugins/PS4/DualShockGamepadPS4.cs
```diff
@@ -315,11 +344,22 @@ public int ps4UserId
         {
             get
             {
-                UpdatePadSettingsIfNeeded();
+#if !UNITY_2019_1_OR_NEWER
+                UpdatePadSettingsIfNeeded(); // 2018.2 uses IOCTL to read this. not required in later versions
```
Rene-Damm, line 348: Comments like this make me happy. It's small details but nice to see explanations for things.

Packages/com.unity.inputsystem/InputSystem/Plugins/PS4/DualShockGamepadPS4.cs
```diff
@@ -315,11 +344,22 @@ public int ps4UserId
...
+#endif
                 return m_PS4UserId;
             }
         }
 
+        public bool IsAimController
```
Rene-Damm, line 354: Properties need to follow the UnityEngine convention, i.e. `lowerCaseStart`. We'll clean it up in the consistency pass.

PR #288, opened 2018-10-25
Support for Touch and Tracked devices for New Input System in uGUI.

StayTalm:
This creates support for Tracked devices and Touch pointers in the UIActionInputModule.  It also supports multiple of both, all with unique pointer Ids in the UI system.  

This adds a new Raycaster for raycasts without using a fake camera or coming from the camera.

Tests and XML Docs are in too.

*** I'm really not sure about the delegate pattern used to map callbacks to the mutiple devices (e.g. getting a TouchPosition action callback could be meant to be sent to ANY touch model, so I needed to use an intermediate delegate to manage it).


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/TrackedDeviceRaycaster.cs
```diff
@@ -0,0 +1,240 @@
...
+            TrackedPointerEventData trackedEventData = eventData as TrackedPointerEventData;
+            if (trackedEventData != null)
+            {
+                PerformRaycast(trackedEventData, resultAppendList);
+            }
+        }
+
+        // Use this list on each raycast to avoid continually allocating.
+        [NonSerialized]
+        private List<RaycastHitData> m_RaycastResultsCache = new List<RaycastHitData>();
```
Rene-Damm, line 73: Preferably use arrays (using the xxxWithCapacity routines) instead of `List`. `List` is kind of stupid. What could be a single heap object turns into two and every single access turns into a two pointer hop. Except for public APIs for user convenience, I'd prefer not to have code in the input system use `List`.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/TrackedDeviceRaycaster.cs
```diff
@@ -0,0 +1,240 @@
...
+
+                if (hits.Length > 0 && hits[0].fraction * raycastDistance < hitDistance)
+                {
+                    hitDistance = hits[0].fraction * raycastDistance;
+                }
+            }
+
+            m_RaycastResultsCache.Clear();
+            SortedRaycastGraphics(canvas, ray, m_RaycastResultsCache);
+            Vector2 screenPosition = Vector2.zero;
```
Rene-Damm: Remember to use `var`.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIActionInputModule.cs
```diff
@@ -18,69 +18,320 @@ namespace UnityEngine.Experimental.Input.Plugins.UI
...
+            private InputActionProperty m_Orientation;
+            [SerializeField]
+            private InputActionProperty m_Select;
+        }
+
         /// <summary>
-        /// An <see cref="InputAction"/> delivering a <see cref="Vector2">2D screen position
+        /// This enables background event processing, so that the Input Module can continue sending UI events even in the background.
+        /// </summary>
+        public bool ForceEventsWithoutFocus
```
Rene-Damm: Lowercase start `forceEventsWithoutFocus`.

Name seems unclear. Comment is much clearer than name. `sendEventsWhenInBackground`?

Doesn't seem to be used.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIActionInputModule.cs
```diff
@@ -18,69 +18,320 @@ namespace UnityEngine.Experimental.Input.Plugins.UI
     /// </remarks>
     public class UIActionInputModule : UIInputModule
     {
+        private static void SwapProperty(ref InputActionProperty oldProperty, InputActionProperty newProperty, bool actionsHooked, Action<InputAction.CallbackContext> actionCallback)
```
Rene-Damm: `SwapAction`?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIActionInputModule.cs
```diff
@@ -121,7 +375,60 @@ private bool ShouldIgnoreEventsOnNoFocus()
             }
         }
 
-        public void OnAction(InputAction.CallbackContext context)
+        /// <summary>
+        /// Adds Touch UI responses based the Actions provided.
```
Rene-Damm, line 383: No uppercasing of nouns except if uppercased in normal English.

jackpr-unity:
In the GUITest scene, the tracked device mappings need to be enabled.

jackpr-unity:
In the GUITest scene, I recommend changing the highlighted and pressed colors to yellow/red to be more obvious.  I also recommend making the navigation between interactables explicit and set so that you can use up/down and left/right to move between items.

jackpr-unity:
In the GUITest scene, the button and dropdown stay "highlighted" when you navigate to a different interactable using a controller or the mouse.

jackpr-unity:
The GUITest scene needs one of the elements to be selected on Start() so that you can easily test with a gamepad.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIActionInputModule.cs
```diff
@@ -18,69 +18,320 @@ namespace UnityEngine.Experimental.Input.Plugins.UI
...
+                set { SwapProperty(ref m_Orientation, value, m_ActionsHooked, actionCallback); }
+            }
+
+            public InputActionProperty select
+            {
+                get { return m_Select; }
+                set { SwapProperty(ref m_Select, value, m_ActionsHooked, actionCallback); }
+            }
+
+            public bool ActionsHooked
```
Rene-Damm: `actionsHooked`

Assets/QA/Input_Test/GUITest.cs
```diff
@@ -1,77 +1,22 @@
...
+        {
+            EventSystem current = EventSystem.current;
+            if (current != null)
+                EventSystem.current.SetSelectedGameObject(initialSelectedGameObject);
+        }      
     }
 
     public void OnButtonClick()
     {
         Debug.Log("Button Click Recieved");
```
Rene-Damm: Typo

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MouseModel.cs
```diff
@@ -309,10 +303,36 @@ public void OnFrameFinished()
...
+            m_InternalData.pointerTarget = eventData.pointerEnter;
+        }
+
         private Vector2 m_Position;
         private Vector2 m_ScrollPosition;
         private MouseButtonModel m_LeftButton;
         private MouseButtonModel m_RightButton;
         private MouseButtonModel m_MiddleButton;
+
+        private InternalData m_InternalData;
```
Rene-Damm, line 336: What's the advantage of keeping that in a struct rather than just making it member fields?

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -192,16 +217,45 @@ public IEnumerator Actions_CanDriveUI()
...
         rightChildReceiver.Reset();
+    }
 
-        // Reset Scroll
-        InputSystem.QueueDeltaStateEvent(mouse.scroll, Vector2.zero);
-        InputSystem.Update();
-        eventSystem.InvokeUpdate();
+    [UnityTest]
+    [Category("Actions")]
+    public IEnumerator JoystickActions_CanDriveUI()
```
Rene-Damm, line 224: Prefix with category. e.g. `Actions_CanDriveUI_FromGamepad`.

Rene-Damm (commented):
Think I need a walkthrough for the touch and tracked responder structs on the UI module.

jackpr-unity:
The Left/Right navigation on the slider interactable should be removed again - right now you can't test changing its value with a controller because pressing left/right does ui navigation instead of an interaction.

jackpr-unity:
The TrackedDevices actions still aren't working for me in the GUITest scene.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MouseModel.cs
StayTalm: Organizational.
If I expect others to use this, I think it's easier to get people to use it if I split 'Things you set' and 'Things uGUI sets' so that it's nor overwhelming.  Compiler side, it's the same, since it's a struct in a struct, but for users new to it, it increases legibility.

PR #289, opened 2018-10-25
Improved IME test (Manual and Automatic)

StayTalm:
Added a UI to the IMETest scene, so that it can be tested in standalone builds.  Added Automated tests as well.  Also fixed issues with setting Cursor Position in the manual tests.


Packages/com.unity.inputsystem/InputSystem/Events/IMECompositionEvent.cs
```diff
@@ -132,5 +95,62 @@ object IEnumerator.Current
...
+
+        [FieldOffset(0)]
+        int size;
+
+        [FieldOffset(sizeof(int))]
+        fixed char buffer[LowLevel.IMECompositionEvent.kIMECharBufferSize];
+
+        public IMECompositionString(string characters)
+        {
+            Debug.Assert(characters.Length < LowLevel.IMECompositionEvent.kIMECharBufferSize);
```
Rene-Damm, line 129: If it's a public API, it has to throw. `Debug.Asserts` are only okay for when we want to guard against our own code messing up.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -3369,4 +3369,90 @@ public void Devices_CanUseNoiseFilterWithMultipleProcessors()
 
         Assert.AreEqual(NoisyInputDevice.current, device1);
     }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_IMECompositionEventsGoThroughKeyboard()
+    {
+        const string imeCompositionCharacters = "CompositionTestCharacters! ɝ";
+        bool callbackWasCalled = false;
```
Rene-Damm, line 3378: Use `var`.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -3369,4 +3369,90 @@ public void Devices_CanUseNoiseFilterWithMultipleProcessors()
 
         Assert.AreEqual(NoisyInputDevice.current, device1);
     }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_IMECompositionEventsGoThroughKeyboard()
```
Rene-Damm, line 3375: These tests all show a focus on "XYZ works". This test for example, focuses on a mechanic that is irrelevant to the user. Same for `Devices_IMEEnableSendsCorrectIOCTLCommand`, for example.

Each test should broadly go in one of three categories:
- Highlight something that *can* be done (`Devices_CanBeReadded`)
- Highlight something that *cannot* be done (`Devices_CannotChangeSetupOfDeviceWhileAddedToSystem`)
- Highlight a notable behavior when something *is* done (`Devices_WhenCreationFails_SystemRecoversGracefully`)

I recommend as thinking of the verification aspect of tests entirely as a side effect and focusing entirely on the "this is how I do" aspect in tests.

PR #292, opened 2018-10-29
FIX: IndexOutOfRangeException on handling multiple action maps.

kuriharaan:
I had IndexOutOfRangeException on runtime on making some input trigger with using Input Action asset that have two or more Action maps.
It was caused by accessing minus value to array of map.m_Actions.
It looks like using actionIndex is fine to map.m_Actions.
Instead, triggerStates seems to need to be accessed with value adjusted by startActionIndex. 
I'm not sure there are any other bugs related the way to access map.m_Actions, It's clear that exception I had, for now.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMapState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMapState.cs
@@ -764,7 +764,7 @@ namespace UnityEngine.Experimental.Input
 
             // Let listeners know.
             var map = maps[trigger.mapIndex];
-            var action = map.m_Actions[actionIndex - mapIndices[trigger.mapIndex].actionStartIndex];
+            var action = map.m_Actions[actionIndex];
             switch (newPhase)
             {
                 case InputActionPhase.Started:


Rene-Damm:
Will have a closer look. The fix doesn't look right (`actionIndex` is a combined index for all actions coming from all maps added to the state so it has to be adjusted to be able to index the action array of an individual map) but the exception clearly hints at a bug somewhere around there.

Rene-Damm:
I think the bug was in `InputBindingResolver` and should be fixed by https://github.com/Unity-Technologies/InputSystem/commit/e63a33d6d4ff381e321accf949c429cd248f26f6#diff-3482826fdfa6c3d8851221dfb4669e2eR210.

Could you give latest head a try and see if you're still seeing the problem?

kuriharaan:
There's still exception problem.
I'm looking into it further.

danielheim:
I'm experiencing this issue too, with 0.0.12-preview. If there's hasn't been an update, I'll do some investigation.

Rene-Damm:
Yup, ran into the same thing. There was a confusion in the code about whether `actionIndex` is relative to the map (bug) or relative to the combined state (correct). Should be fixed in https://github.com/Unity-Technologies/InputSystem/commit/8ac817b906adcc9ae7acad9f5b9c628e977e3729.

danielheim:
I think you may have missed one... InputActionMapState, ChangePhaseOfAction, line 765
`var action = map.m_Actions[actionIndex - mapIndices[trigger.mapIndex].actionStartIndex];`
Removing the start index seems to fix the issue for me, i.e.:
`var action = map.m_Actions[actionIndex];`
Thanks for the fix!

Edit: I was wrong - Although the fix seems to help, the issue is still present.

Rene-Damm:
Ok, will take another look. The first version of the line is in fact the correct one (`actionIndex` needs to be adjusted when indexing the actions on the map instead of indexing the actions on the state) but given the issue still pops up, somewhere things must still get confused.

danielheim:
As a temporary work around, I've been creating a separate Input Action Control asset for each Map. The maps seem to work fine if there's only one in the asset.

Rene-Damm:
Ok, this dreaded bug should finally be resolved for good. Added a [test](https://github.com/Unity-Technologies/InputSystem/blob/30b85c3ed7c9b24814742f83a02fbff837992619/Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs#L453) that should keep this thing permanently out. Hopefully.

PR #301, opened 2018-11-07
NEW: Add the ability for automated test runner to choose runtime.

douglas-piconi:
defaulting all tests to latest.  legacy is deprecated as of 2018.3.


douglas-piconi:
Actually going to run both so we can determine if specific runtime breaks code

PR #306, opened 2018-11-10
Reset Device and Run in Background

StayTalm:
This allows creates a ResetDevice command, uses it when focus is changed, and adds a few small hooks to allow the input events to ignore focus.

Relies on https://ono.unity3d.com/unity/unity/pull-request/77815/_/input/new/resetDevice


Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -107,6 +107,19 @@ public Action onShutdown
             }
         }
 
+        public Action<bool> onFocusChanged
+        {
+            set
+            {
+                if (value == null)
+                    Application.focusChanged -= OnFocusChanged;
```
Rene-Damm, line 116: We can't merge things that break the build for users that don't have the respective changes. This needs to either use #if based on Unity version where the native change has landed or use reflection to hook into the functionality when available.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -107,6 +107,19 @@ public Action onShutdown
...
+                else if (m_ShutdownMethod == null)
```
Rene-Damm: This looks like a copypaste mistake. `m_FocusChangedMethod`?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1905,7 +1918,8 @@ internal unsafe void OnUpdate(InputUpdateType updateType, int eventCount, IntPtr
             var buffersToUseForUpdate = updateType;
 #if UNITY_EDITOR
             gameIsPlayingAndHasFocus = InputConfiguration.LockInputToGame ||
-                (UnityEditor.EditorApplication.isPlaying && Application.isFocused);
+                (UnityEditor.EditorApplication.isPlaying && Application.isFocused) ||
+                ((updateMask & InputUpdateType.IgnoreFocus) != 0);
```
Rene-Damm: The added check seems like it should be outside the `#if UNITY_EDITOR` block.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -773,11 +773,10 @@ public static void DisableDevice(InputDevice device)
             s_Manager.EnableOrDisableDevice(device, false);
         }
 
-        ////REVIEW: should this be a device-level reset along with sending the reset IOCTL
-        ////        or a control-level reset on just the memory state?
-        public static void ResetDevice(InputDevice device)
+        public static unsafe void ResetDevice(InputDevice device)
         {
-            throw new NotImplementedException();
+            var resetCommand = RequestResetCommand.Create();
```
Rene-Damm, line 791: Null-check device and throw `ArgumentNullException` if necessary.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1217,6 +1216,27 @@ public static InputUpdateType updateMask
             set { s_Manager.updateMask = value; }
         }
 
+        /// <summary>
+        /// Tells the Input System to ignore focus, and continue to trigger events and actions regardless of current focus state
+        /// </summary>
+        /// /// <remarks>
+        /// Off by default, this does not work on all platforms and devices, only those that can recieve their own input data while not in focus.
+        /// </remarks>
+        public static bool ignoreFocus
```
Rene-Damm: Wondering if we should `#if` this one to make it clearer that this isn't a universal capability?

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
```diff
@@ -71,6 +71,14 @@ public enum InputUpdateType
         /// </remarks>
         ManualThreaded = Manual | 1 << 5,
 
+        /// <summary>
+        /// Turning this flag on allowed for updates to run even when Unity is running in the background.
+        /// </summary>
+        /// <remarks>
+        /// This doesn't guarantee that all devices will update properly while in the background, and so you will still only recieve updates for devices that support running in the background natively.
+        /// </remarks>
+        IgnoreFocus = 1 << 31,
```
Rene-Damm: This shouldn't be exposed. I think it's an okay-ish thing for the native runtime to roll it into the update mask, but `InputUpdateType` should only have proper update types.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: Also wondering now whether it's clearer naming this explicitly after the run-in-background ability rather than relating it to focus. Maybe `receiveInputWhenInBackground` or `receiveInputWhenInBackgroundIfSupported`?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
StayTalm: It's good inside the UNITY_EDITOR block, because outside the editor we rely on the native updates being called.  It's by default running.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
StayTalm: I went with `runInBackground`.  Those property names are way too wordy.

Rene-Damm: Names that leave no doubt about what they do in general are good. Wordy or not doesn't matter overly much. There's no tax on characters.

jackpr-unity:
Windows Editor: keyboard, mouse, and XInput work fine

jackpr-unity (changes requested):
(1) Windows Editor: HID does not work in background.

I recommend using Jia's nice scene "02_Input" to try this out.  You can use CTRL+# to change modes.  CTRL+1 is kb+mouse.  CTRL+2 is xinput. CTRL+3 is hid.

(2) Windows Editor: Setting InputSystem.runInBackground = true; in Start() does not work.  Setting the value in Update() seems to work fine, but this setting should function in Awake() or Start()

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/QueryCanRunInBackground.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Runtime.InteropServices;
+using UnityEngine.Experimental.Input.Utilities;
+
+namespace UnityEngine.Experimental.Input.LowLevel
+{
+    [StructLayout(LayoutKind.Explicit, Size = InputDeviceCommand.kBaseCommandSize + sizeof(bool))]
+    internal unsafe struct QueryCanRunInBackground : IInputDeviceCommandInfo
```
Rene-Damm: Should be public so that other device implementors can respond to it.

Rene-Damm: Also wondering whether we should turn this into a QueryCapabilities command that returns a flag bitfield instead of just a single bool.

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/RequestResetCommand.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace UnityEngine.Experimental.Input.LowLevel
+{
+    /// <summary>
+    /// A command to tell the runtime to reset the device to it's default state.
+    /// </summary>
+    /// <remarks>
+    /// This triggers an event being sent from the device that represents an empty, or untouched device
+    /// </remarks>
+    [StructLayout(LayoutKind.Explicit, Size = InputDeviceCommand.kBaseCommandSize)]
+    internal unsafe struct RequestResetCommand : IInputDeviceCommandInfo
```
Rene-Damm: Public.

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/RequestSyncCommand.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace UnityEngine.Experimental.Input.LowLevel
+{
+    /// <summary>
+    /// A command to tell the runtime to sync the device to it's last known state.
+    /// </summary>
+    /// <remarks>
+    /// This triggers an event from the underlying device that represents the whole, current state.
+    /// </remarks>
+    [StructLayout(LayoutKind.Explicit, Size = InputDeviceCommand.kBaseCommandSize)]
+    internal unsafe struct RequestSyncCommand : IInputDeviceCommandInfo
```
Rene-Damm: Public.

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/RequestSyncCommand.cs
```diff
@@ -0,0 +1,36 @@
...
+using UnityEngine.Experimental.Input.Utilities;
+
+namespace UnityEngine.Experimental.Input.LowLevel
+{
+    /// <summary>
+    /// A command to tell the runtime to sync the device to it's last known state.
+    /// </summary>
+    /// <remarks>
+    /// This triggers an event from the underlying device that represents the whole, current state.
+    /// </remarks>
```
Rene-Damm, line 12: Sidenote, make copious use of `<seealso ...>` to link related docs so it's easy to find which APIs belong together.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1217,6 +1230,34 @@ public static InputUpdateType updateMask
             set { s_Manager.updateMask = value; }
         }
 
+#if UNITY_2018_3_OR_NEWER
+        private const InputUpdateType ms_runInBackgroundMask = (InputUpdateType)(1 << 31);
```
Rene-Damm: `s_RunInBackgroundMask`.

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
```diff
@@ -71,6 +71,8 @@ public enum InputUpdateType
         /// </remarks>
         ManualThreaded = Manual | 1 << 5,
 
+        RunInBackground = 1 << 31,
```
Rene-Damm: Let's keep this out of InputUpdateType.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1217,6 +1230,34 @@ public static InputUpdateType updateMask
...
+
+        /// <summary>
+        /// Tells the Input System to run even when the application is in the backgrond, and continue to trigger events and actions regardless of current focus state
+        /// </summary>
+        /// /// <remarks>
```
Rene-Damm: Double `///`

Rene-Damm:
Semi-related to this, how are we going to handle the XR focus thing?

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/QueryCanRunInBackground.cs
StayTalm: If it's Query Capabilities it would change how it's used.
Right now I call it when I care about if it can run in background, direct from the property, but if I'm filling in a whole bunch of settings it should be more of a query on connect and cache.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
StayTalm: I do

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
StayTalm: I need to add this back in every time I want to debug, because C#'s debugging and watch window info only shows known enumeration values.
So: Default | (1 << 31)
will display in the watch window as 'Default', which makes debugging and reading the flags a real pain.

jackpr-unity (changes requested):
the canRunInBackground attributes don't seem to work for me on Windows.  They return false in both the editor and in the player.  I'm calling the following in an Update() function:

Debug.Log("Keyboard " + InputSystem.GetDevice<Keyboard>().canRunInBackground);
Debug.Log("Mouse " + InputSystem.GetDevice<Mouse>().canRunInBackground);
Debug.Log("XInput " + InputSystem.GetDevice<XInputController>().canRunInBackground);

Packages/com.unity.inputsystem/InputSystem/Devices/Commands/QueryCanRunInBackground.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public static FourCC Type { get { return new FourCC('Q', 'R', 'I', 'B'); } }
+
+        public const int kSize = InputDeviceCommand.kBaseCommandSize;
+
+        [FieldOffset(0)]
+        public InputDeviceCommand baseCommand;
+
+        [FieldOffset(InputDeviceCommand.kBaseCommandSize)]
+        public bool canRunInBackground;
```
jackpr-unity, line 22: the canRunInBackground attributes don't seem to work for me on Windows.  They return false in both the editor and in the player.  I'm calling the following in an Update() function:

Debug.Log("Keyboard " + InputSystem.GetDevice<Keyboard>().canRunInBackground);
Debug.Log("Mouse " + InputSystem.GetDevice<Mouse>().canRunInBackground);
Debug.Log("XInput " + InputSystem.GetDevice<XInputController>().canRunInBackground);

jackpr-unity (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/ActionButtonTracker.cs
```diff
@@ -0,0 +1,48 @@
+using UnityEngine;
+using UnityEngine.Experimental.Input;
+using UnityEngine.UI;
+
+public class ActionButtonTracker : MonoBehaviour
```
Rene-Damm, line 5: What's this one about? This doesn't seem like it should go in the input system.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1905,7 +1919,8 @@ internal unsafe void OnUpdate(InputUpdateType updateType, int eventCount, IntPtr
             var buffersToUseForUpdate = updateType;
 #if UNITY_EDITOR
             gameIsPlayingAndHasFocus = InputConfiguration.LockInputToGame ||
-                (UnityEditor.EditorApplication.isPlaying && Application.isFocused);
+                (UnityEditor.EditorApplication.isPlaying && Application.isFocused) ||
+                ((updateMask & (InputUpdateType)(1 << 31)) != 0);
```
Rene-Damm, line 1923: Now realize this won't work correctly. What it will effectively do is take input away from the editor (like `LockInputToGame` does) whereas the intention would be that the input is visible in both when `runInBackground` is enabled. With the current event processing, that'd take more extensive changes. Think for now I'll just leave it be and go back to how it was before.

PR #316, opened 2018-11-27
Warning tests

douglas-piconi:
New test for checking compile and warning errors on gitlab CI.  Separated into its own Unity project and workspace in the inputsystem repo to be separate form our other tests.    Due to big changes and instabilities between versions of unity with compiler classes and how the editor handles internal, only running the test on 2018.3 and OSX.  This should be good enough to determine if there are compiler errors or warnings in our package.


douglas-piconi:
going to refactor and move project

PR #327, opened 2018-12-04
Update Input_Test

JiaUnity:
- Add Menu Item "QA Tools" for easy Input Manager asset swap
- Update scripts 
- Add joystick device (incomplete)


Assets/QA/Input_Test/Editor/ImportInputManagerAsset.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.IO;
+using UnityEditor;
+using UnityEngine;
+
+[InitializeOnLoad]
+public class ImportInputManagerAsset {
```
Rene-Damm, line 6: Please run `format.sh` or `format.ps1` before committing. Otherwise the next time someone does, your files will get picked up, too.

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.2.5f1
+m_EditorVersion: 2018.2.18f1
```
Rene-Damm, line 1: In general, this one shouldn't be committed except we explicitly want to bump our minimum version requirement.

PR #329, opened 2018-12-09
Joystick

StayTalm:
HID devices that are either joystick or gamepad type get picked up as Joysticks.  This also combines the X & Y axes if available into a 'Stick' control, as well as maps the twist and buttons properly.  And it allows variable sized bit arrays to work.


Packages/com.unity.inputsystem/InputSystem/Devices/Joystick.cs
```diff
@@ -45,7 +46,7 @@ namespace UnityEngine.Experimental.Input
     // By default comes with just a trigger, a potentially twistable
     // stick and an optional single hatswitch.
     [InputControlLayout(stateType = typeof(JoystickState))]
-    public class Joystick : InputDevice
+    public class Joystick : HID
```
Rene-Damm: We can't do this. Joystick needs to be independent of any interface.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputControlTreeView.cs
```diff
@@ -324,10 +324,27 @@ private unsafe string ReadRawValueAsString(InputControl control, byte[] state)
...
-                    if (MemoryHelpers.ReadSingleBit(ptr, control.m_StateBlock.bitOffset))
-                        value = "1";
+                    if (control.valueSizeInBytes == 1)
+                    {
+                        value = MemoryHelpers.ReadSingleBit(ptr, control.m_StateBlock.bitOffset) ? "1" : "0";
+                    }
                     else
-                        value = "0";
+                    {
+                        value = (float)(MemoryHelpers.ReadIntFromMultipleBits(ptr, control.m_StateBlock.bitOffset, control.m_StateBlock.sizeInBits));
```
Rene-Damm: Why the float conversion?

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
```diff
@@ -113,15 +113,22 @@ internal static string OnFindLayoutForDevice(int deviceId, ref InputDeviceDescri
...
+                {
                     baseLayout = "Joystick";
+                    baseType = typeof(Joystick);
+                }
                 else if (hidDeviceDescriptor.usage == (int)GenericDesktop.Gamepad)
-                    baseLayout = "Gamepad";
+                {
+                    baseLayout = "Joystick";
+                    baseType = typeof(Joystick);
+                }
```
Rene-Damm, line 124: Minor... use `||`.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
```diff
@@ -319,16 +326,88 @@ public static bool UsageToString(UsagePage usagePage, int usage, out string usag
...
                 var builder = new InputControlLayout.Builder
                 {
-                    type = typeof(HID),
+                    type = deviceType,
+                    extendsLayout = parentLayout,
                     stateFormat = new FourCC('H', 'I', 'D'),
                 };
 
                 ////TODO: for joysticks, set up stick from X and Y
+                var xElement = Array.Find(hidDescriptor.elements, element => element.usagePage == UsagePage.GenericDesktop && element.usage == (int)GenericDesktop.X);
```
Rene-Damm, line 336: `////TODO` seems to be obsolete.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
```diff
@@ -319,16 +326,88 @@ public static bool UsageToString(UsagePage usagePage, int usage, out string usag
...
+                        .WithParameters("clamp,clampMin=-1,clampMax=0,invert")
+                        .WithBitOffset(0)
+                        .WithSizeInBits((uint)xElement.reportSizeInBits);
+
+                    builder.AddControl(stickName + "/right")
+                        .WithFormat(InputStateBlock.kTypeSBit)
+                        .WithLayout("Button")
+                        .WithParameters("clampMin=0,clampMax=1")
+                        .WithBitOffset(0)
+                        .WithSizeInBits((uint)xElement.reportSizeInBits);
```
Rene-Damm, line 402: Are left/right/up/down actually needed here? The default setup in stick seems like it should do the trick. It's wired to use the state from X and Y and should thus automatically get the format, offsets, and sizes and such (assuming there's no bug there, of course :)).

Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
```diff
@@ -230,12 +244,30 @@ public float ReadFloat(void* statePtr)
                 Debug.Assert(bitOffset == 0, "FLT state must be byte-aligned");
                 value = *(float*)valuePtr;
             }
-            else if (format == kTypeBit)
+            else if (format == kTypeBit || format == kTypeSBit)
             {
-                if (sizeInBits != 1)
-                    throw new NotImplementedException("Cannot yet convert multi-bit fields to floats");
```
Rene-Damm, line 236: Thanks for implementing this!

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
StayTalm: It was not getting the proper state.  It was expecting 32 bits for each X and Y, and working off that.  Which meant it was overreadding.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputControlTreeView.cs
StayTalm: No idea, copy-paste I suspect

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
Rene-Damm: Crap, means that there's a problem in the `useStateFrom` logic. Probably something with it not working in combination with another layout retroactively changing the state layout. I'll add a ////TODO.

Packages/com.unity.inputsystem/InputSystem/Utilities/MemoryHelpers.cs
```diff
@@ -209,6 +209,15 @@ public static int ReadIntFromMultipleBits(void* ptr, uint bitOffset, uint bitCou
             if (bitCount >= sizeof(int) * 8)
                 throw new ArgumentException("Trying to read more than 32 bits as int", "bitCount");
 
+            //Shift the pointer up on larger bitmasks and retry
+            if (bitOffset > 32)
+            {
+                int newBitOffset = (int)bitOffset % 32;
+                int intOffset = ((int)bitOffset - newBitOffset) / 32;
+                void* offsetPtr = (byte*)ptr + (intOffset * 4);
+                return ReadIntFromMultipleBits(offsetPtr, (uint)newBitOffset, bitCount);
```
Rene-Damm: No need to recurse. Can just adjust values.

Packages/com.unity.inputsystem/InputSystem/Devices/Joystick.cs
```diff
@@ -4,6 +4,7 @@
 using UnityEngine.Experimental.Input.Layouts;
 using UnityEngine.Experimental.Input.LowLevel;
 using UnityEngine.Experimental.Input.Utilities;
+using UnityEngine.Experimental.Input.Plugins.HID;
```
Rene-Damm: Obsolete

PR #331, opened 2018-12-10
Bug 295: Fixed device count not ready in Start/Awake for editor.

douglas-piconi:
Was already fixed in player.  Was previously working in editor for other reasons, but code changes in Editor most likely took away that functionality.  Fixed now in editor with same fix that was in player.


Rene-Damm (approved):
Please add test.

PR #379, opened 2019-01-15
[FIX] Temporilary turn off time slicing.

douglas-piconi:
Not sure if you want to merge this, but turns off the problem section that causes the dropped events.


Rene-Damm:
I've moved all the code to #ifs for 2019.2. Hope we can fix the stuff for the next package release.

PR #390, opened 2019-01-24
Fixes naming confusion with Oculus Go and GearVR

StayTalm:
I've renamed the two Oculus standalone Layouts such that they use Base and Extended instead of simple, as well, I've made sure those layouts appear in the Editor UI as Oculus Go and GearVR so users don't have to understand what type of Oculus headset is base and what is extended.


jackpr-unity:
I get a compile error:

`D:\Git\InputSystem_StayTalmFork\Packages\com.unity.inputsystem\InputSystem\Plugins\XR\GearVR.cs(65,48): error CS0246: The type or namespace name 'OculusStandaloneHMDBase' could not be found (are you missing a using directive or an assembly reference?)
`

It looks like the definition of GearVRHMD in GearVR.cs just needs to be renamed to OculusStandaloneHMDBase

jackpr-unity:
Another compile error reveals itself after the GearVR.cs one is fixed:

`D:\Git\InputSystem_StayTalmFork\Packages\com.unity.inputsystem\InputSystem\NativeInputRuntime.cs(69,25): error CS1593: Delegate 'Action<NativeInputUpdateType, int, IntPtr>' does not take 2 arguments
`

StayTalm:
@jackpr-unity Once you get back and want to look at this one, sync with me.  There were some native changes and you need the correct version of 2019.2 in order to pass by that error.  The defines don't work so well on changes while working on a version.

Rene-Damm:
> Another compile error reveals itself after the GearVR.cs one is fixed:

That one's on me. The native changes have landed in 2019.1 and by moving the managed changes to 2019.2 because they weren't working right yet, I broke 2019.1. Will fix.

Rene-Damm (approved):

StayTalm:
@jackpr-unity 
Can you try again with 2019.2.0a4 or greater and the current version of this branch?
2019.1 and 2018.3 work as well.

jackpr-unity:
StayTalm:develop has unity project compile issues related to steamworks things

jackpr-unity (commented):
There are unrelated Steamworks errors in the StayTalm:develop branch.  Otherwise, looks good to me.

PR #394, opened 2019-01-28
Android fixes

todi1856:


Rene-Damm:
Resolved conflict, enabled test, and landed this as #464.

PR #402, opened 2019-01-30
FIX: Native Event Buffer changes have landed in 2019_1.

douglas-piconi:
Fix 2019.1.0 beta1  compile issue now that native event has landed.


douglas-piconi:
nevermind lol

Issue #252, opened 2018-09-19
[Mac Editor] Error on first play


iBicha:
Editor on macOS, Unity 2018.3.0b1

On first play, exceptions are thrown:
```
Could not re-recreate input device 'Apple Inc. Apple Internal Keyboard / Trackpad (HID)' with layout 'HID::Apple Inc. Apple Internal Keyboard / Trackpad' and variants 'Default' after domain reload
UnityEngine.Experimental.Input.LowLevel.<>c__DisplayClass6_0:<set_onUpdate>b__0(NativeInputUpdateType, Int32, IntPtr)
UnityEngineInternal.Input.NativeInputSystem:NotifyUpdate(NativeInputUpdateType, Int32, IntPtr) (at /Users/builduser/buildslave/unity/build/Modules/Input/Private/Input.cs:91)
```
```
ArgumentException: Cannot find layout matching device description 'Apple Inc. Apple Internal Keyboard / Trackpad (HID)'
Parameter name: description
UnityEngine.Experimental.Input.InputManager.AddDevice (UnityEngine.Experimental.Input.Layouts.InputDeviceDescription description, System.Boolean throwIfNoLayoutFound, System.Int32 deviceId, UnityEngine.Experimental.Input.InputDevice+DeviceFlags deviceFlags) (at Packages/com.unity.inputsystem/InputSystem/InputManager.cs:887)
UnityEngine.Experimental.Input.InputManager.RestoreDevicesAfterDomainReload () (at Packages/com.unity.inputsystem/InputSystem/InputManager.cs:2680)
UnityEngine.Experimental.Input.LowLevel.<>c__DisplayClass6_0:<set_onUpdate>b__0(NativeInputUpdateType, Int32, IntPtr)
UnityEngineInternal.Input.NativeInputSystem:NotifyUpdate(NativeInputUpdateType, Int32, IntPtr) (at /Users/builduser/buildslave/unity/build/Modules/Input/Private/Input.cs:91)
```

iBicha:
This is probably on domain unload, to be more precise

Rene-Damm:
Yup, that one has starting popping recently and indeed, happens on domain reload. Functionally, other than being annoying, it should not have an impact on the usability of the system. The HID here isn't actively used by the system and should probably be suppressed entirely. Still, the problem itself needs fixing. From what I've gather so far, some HIDs end up sometimes failing system calls asking for their descriptors.

JiaUnity:
[...]

PR #405, opened 2019-02-04
Fix exceptions on domain reload

jechter:
Currently, we add newly discovered devices with `throwIfNoLayoutFound: false`, but then, on domain reload, we re-add the same devices with `throwIfNoLayoutFound: true`, which causes exceptions in the console on domain reload (for instance when entering play mode on the mac, which reports some devices we don't not know how to deal with: https://github.com/Unity-Technologies/InputSystem/issues/252). We should handle both cases equally.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -3269,7 +3269,7 @@ namespace UnityEngine.Experimental.Input
                         // to create a device with the same layout.
                         if (!deviceState.description.empty)
                         {
-                            device = AddDevice(deviceState.description, throwIfNoLayoutFound: true,
+                            device = AddDevice(deviceState.description, throwIfNoLayoutFound: false,
                                 deviceId: deviceState.deviceId, deviceFlags: deviceState.flags);
                         }
                         else


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3269,7 +3269,7 @@ internal void RestoreDevicesAfterDomainReload()
                         // to create a device with the same layout.
                         if (!deviceState.description.empty)
                         {
-                            device = AddDevice(deviceState.description, throwIfNoLayoutFound: true,
+                            device = AddDevice(deviceState.description, throwIfNoLayoutFound: false,
```
Rene-Damm, line 3272: On the fence about this.

Overall, I think we shouldn't silently swallow this. If a device was there before but we're dropping it now, seems like we should point give an indication of what we did. The same way that the other code path in the `else` clause does.

jechter: I don't understand this: 
>If a device was there before but we're dropping it now

As I understand the code, when we first saw the device (in `OnNativeDeviceDiscovered`, we could not create a device for it, but did not throw (`OnNativeDeviceDiscovered` calls `AddDevice` with `throwIfNoLayoutFound: false`), and added it to `m_AvailableDevices`.

Now, on domain reload, we are processing the saved available devices again - and we _still_ cannot create a device for it. Only this time, we'd throw (`RestoreDevicesAfterDomainReload` calls AddDevice` with `throwIfNoLayoutFound: true`) - which did not make sense to me. So I changed it to not throw in this case either. It should still stay in `m_AvailableDevices`.

Also not sure what you mean by 
>The same way that the other code path in the `else` clause does.

what other code, what `else` clause?

Rene-Damm: >when we first saw the device (in `OnNativeDeviceDiscovered`, we could not create a device for it, but did not throw (`OnNativeDeviceDiscovered` calls `AddDevice` with `throwIfNoLayoutFound: false`), and added it to `m_AvailableDevices`

`m_SavedDeviceStates` works differently. It actually *only* contains devices we successfully created before. It is `m_SavedAvailableDevices` which has the list of *all* devices that were reported.

So, when we fail here we are failing on something that we actually did *not* fail on before. Which is why I'm thinking it's good to give an indication of "hey user, we just dropped something that was there before".

For the HIDs in particular, these messages have actually successfully pointed out a bug. And even in cases where we rightfully dropped something because we no longer have a device layout matching the device, I found it useful to have feedback (not sure a user would say the same, though).

>what other code, what `else` clause?

Line 3275 and on. On that path, we check explicitly *before* trying to create a device and log a message if the layout is no longer there.

jechter: Ok, got it. Will investigate further.

jechter:
Better fix here: https://github.com/Unity-Technologies/InputSystem/pull/429

PR #406, opened 2019-02-04
Add additional profiler markers useful for profiling initialization/d…

jechter:
We need to improve Input System performance. I did not get into optimization, but I added some profiler markers I considered useful to find out where time goes during Initialization/Domain reload, so I figured it might make sense to keep these.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1322,6 +1325,7 @@ public void Update(InputUpdateType updateType)
 
         internal void Initialize(IInputRuntime runtime, InputSettings settings)
         {
+            Debug.Log("Initialize");
```
Rene-Damm: Please remove.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jechter: Oops. Fixed.

PR #407, opened 2019-02-04
Remoting fixes

jechter:
I tested the InputDebugger, especially wrt remote debugging, and ran into several issues. This PR fixes them.

1. Makes the player send devices on first connection. I could not get the setup to work at all without this. How was this meant to work?

2. Make RemoteInputPlayerConnection a SerializedSingleton in the editor so the PlayerConnection callbacks we register stay valid through domain reloads.

3. Restore the serialized device names on device recreation during domain reload, as otherwise, remote devices will be listed under the "Local" foldout after domain reload.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3271,6 +3271,10 @@ internal void RestoreDevicesAfterDomainReload()
                         {
                             device = AddDevice(deviceState.description, throwIfNoLayoutFound: true,
                                 deviceId: deviceState.deviceId, deviceFlags: deviceState.flags);
+
+                            // Re-set the name to the stored name. Otherwise we lose the sorting of 
+                            // devices to remote players, which is encoded in the name.
+                            device.m_Name = new InternedString(deviceState.name);
```
Rene-Damm: Wondering if we should put this inside `if ((deviceState.flags & InputDevice.DeviceFlags.Remote) != 0)`.

Overall, the logic here is more a workaround. Think the proper solution would be to store a local device ID to remote device ID mapping and get rid of the name dependency.

Rene-Damm (commented):
Not-losing-callbacks change LGTM. Other change I'd prefer for us to go all the way and fix the name dependency.

Are tests passing? Any test we can add to make sure we don't regress?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jechter: What do you exactly mean by "get rid of the name dependency" and "proper solution would be to store a local device ID to remote device ID mapping and get rid of the name dependency".

-Use the stored mapping to restore the name correctly?
-Don't encode player id in the name at all, but change the input debugger code to sort devices in the UI using some other information?

jechter: What do you think of adding a participant Id int member to the InputDevice class instead of encoding it in the name, like this:

```
diff --git a/Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs b/Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
index 3fb4b3c2..8b807f40 100644
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
@@ -39,6 +39,7 @@ namespace UnityEngine.Experimental.Input
     public class InputDevice : InputControl
     {
         public const int kInvalidDeviceId = 0;
+        public const int kLocalParticipantId = 0;
         internal const int kInvalidDeviceIndex = -1;
 
         /// <summary>
@@ -264,6 +265,7 @@ namespace UnityEngine.Experimental.Input
         public InputDevice()
         {
             m_Id = kInvalidDeviceId;
+            m_ParticipantId = kLocalParticipantId;
             m_DeviceIndex = kInvalidDeviceIndex;
         }
 
@@ -402,6 +404,7 @@ namespace UnityEngine.Experimental.Input
 
         internal DeviceFlags m_DeviceFlags;
         internal int m_Id;
+        internal int m_ParticipantId;
         internal int m_DeviceIndex; // Index in InputManager.m_Devices.
         internal InputDeviceDescription m_Description;
 
diff --git a/Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs b/Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
index d699afd6..c0e6648b 100644
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
@@ -516,8 +516,9 @@ namespace UnityEngine.Experimental.Input
                 InputDevice device;
                 try
                 {
-                    device = receiver.m_LocalManager.AddDevice(data.layout,
-                        string.Format("Remote{0}::{1}", msg.participantId, data.name));
+                    device = receiver.m_LocalManager.AddDevice(data.layout);
+                    device.m_ParticipantId = msg.participantId;
                 }
                 catch (Exception exception)
                 {
diff --git a/Packages/com.unity.inputsystem/InputSystem/Editor/InputDebuggerWindow.cs b/Packages/com.unity.inputsystem/InputSystem/Editor/InputDebuggerWindow.cs
index 13f417df..871c0183 100644
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/InputDebuggerWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/InputDebuggerWindow.cs
@@ -341,7 +341,7 @@ namespace UnityEngine.Experimental.Input.Editor
                     foreach (var player in EditorConnection.instance.ConnectedPlayers)
                     {
                         var playerNode = AddChild(remoteDevicesNode, player.name, ref id);
-                        AddDevices(playerNode, devices, ref id, "Remote" + player.playerId + InputControlLayout.kNamespaceQualifier);
+                        AddDevices(playerNode, devices, ref id, player.playerId);
                     }
                 }
                 else
@@ -424,23 +424,18 @@ namespace UnityEngine.Experimental.Input.Editor
                 return root;
             }
 
-            private void AddDevices(TreeViewItem parent, IEnumerable<InputDevice> devices, ref int id, string namePrefix = null)
+            private void AddDevices(TreeViewItem parent, IEnumerable<InputDevice> devices, ref int id, int participantId = InputDevice.kLocalParticipantId)
             {
                 foreach (var device in devices)
                 {
-                    if (namePrefix != null)
-                    {
-                        if (!device.name.StartsWith(namePrefix))
-                            continue;
-                    }
-                    else if (device.name.Contains(InputControlLayout.kNamespaceQualifier))
-                        continue;
+                    if (device.m_ParticipantId != participantId)
 
                     var item = new DeviceItem
                     {
                         id = id++,
                         depth = parent.depth + 1,
-                        displayName = namePrefix != null ? device.name.Substring(namePrefix.Length) : device.name,
+                        displayName = device.name,
                         device = device,
                         icon = EditorInputControlLayoutCache.GetIconForLayout(device.layout),
                     };
diff --git a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
index 8e49e63e..cd3dd662 100644
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -3120,6 +3120,7 @@ namespace UnityEngine.Experimental.Input
             public string variants;
             public string[] usages;
             public int deviceId;
+            public int participantId;
             public InputDevice.DeviceFlags flags;
             public InputDeviceDescription description;
 
@@ -3179,6 +3180,7 @@ namespace UnityEngine.Experimental.Input
                     layout = device.layout,
                     variants = device.variants,
                     deviceId = device.id,
+                    participantId = device.m_ParticipantId,
                     usages = usages,
                     description = device.m_Description,
                     flags = device.m_DeviceFlags
@@ -3271,10 +3273,7 @@ namespace UnityEngine.Experimental.Input
                         {
                             device = AddDevice(deviceState.description, throwIfNoLayoutFound: true,
                                 deviceId: deviceState.deviceId, deviceFlags: deviceState.flags);
-
-                            // Re-set the name to the stored name. Otherwise we lose the sorting of 
-                            // devices to remote players, which is encoded in the name.
-                            device.m_Name = new InternedString(deviceState.name);
+                            device.m_ParticipantId = deviceState.participantId;
                         }
                         else
                         {
```

jechter:
I updated this PR with changes to:
-No longer encode player ids in device names
-Don't immediately start sending upon player connection connect, instead you have to manually enable sending
-Add UI to the input debugger to connect to devices and to enable/disable remote devices.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Remoting.cs
```diff
@@ -128,23 +134,40 @@ public void Remote_CanConnectInputSystemsOverEditorPlayerConnection()
...
-        ////TODO: make sure that we also get the connection sequence right and send our initial layouts and devices
-        Assert.That(observer.messages, Has.Count.EqualTo(4));
-        Assert.That(observer.messages[0].type, Is.EqualTo(InputRemoting.MessageType.Connect));
-        Assert.That(observer.messages[1].type, Is.EqualTo(InputRemoting.MessageType.NewDevice));
-        Assert.That(observer.messages[2].type, Is.EqualTo(InputRemoting.MessageType.NewEvents));
-        Assert.That(observer.messages[3].type, Is.EqualTo(InputRemoting.MessageType.RemoveDevice));
+        fakeEditorConnection.Send(RemoteInputPlayerConnection.kStopSendingMsg, null);
+
+        // We should not obseve any messages for these, as we stopped sending!
+        device = InputSystem.AddDevice("Gamepad");
```
Rene-Damm: Preferably use `InputSystem.AddDevice<Gamepad>();`. The `AddDevice("Gamepad")` thing that's still in some tests is mainly a leftover.

Packages/com.unity.inputsystem/InputSystem/Editor/InputDebuggerWindow.cs
```diff
@@ -214,11 +215,72 @@ private void ToggleDiagnosticMode()
...
+                    InputSystem.remoting.RemoveRemoteDevices(player.playerId);
+
+            }
+        }
+
+        private void DrawConnectionGUI()
+        {
+            if (GUILayout.Button("Remote Devices…", EditorStyles.toolbarDropDown))
+            {
+                GenericMenu menu = new GenericMenu();
```
Rene-Damm: Use `var` for all local variables.

Rene-Damm (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Editor beeping or triggering menu commands when binding keys interactively.
 - "Add Action" button being active and triggering exceptions when no action map had been added yet.
+- Input Debugger: Make remote connections still work after domain reloads
+- Input Debugger: Add UI to connect to remote devices and to enable/disable remote input debugging.
```
Rene-Damm: Put in separate `### Changes` section. Something like.

```

### Changes

- The input debugger will no longer automatically connect to players when the profiler is connected. Use "Options" menu in debugger toolbar to connect.
```

PR #422, opened 2019-02-19
Action importer fixes

jechter:
I was looking at https://fogbugz.unity3d.com/f/cases/1091478, and an across some more issues while fixing that:

- InputActions editor: Fixed warnings in Console when using InputActions editor
- InputActions editor: Fixed Assert when generating C# class and make sure it gets imported correctly
- InputActions editor: Generate directories as needed when generating C# class, and allow path names without "Assets/" path prefix.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/InputActionCodeGenerator.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/InputActionCodeGenerator.cs
@@ -472,13 +472,6 @@ namespace UnityEngine.Experimental.Input.Editor
                     return false;
             }
 
-            if (!filePath.ToLower().StartsWith("assets/"))
-                filePath = Path.Combine("Assets", filePath);
-
-            var dir = Path.GetDirectoryName(filePath);
-            if (!Directory.Exists(dir))
-                Directory.CreateDirectory(dir);
-
             // Write.
             File.WriteAllText(filePath, code);
             return true;
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/InputActionCodeGenerator.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/InputActionCodeGenerator.cs
@@ -472,6 +472,13 @@ namespace UnityEngine.Experimental.Input.Editor
                     return false;
             }
 
+            if (!filePath.ToLower().StartsWith("assets/"))
+                filePath = Path.Combine("Assets", filePath);
+
+            var dir = Path.GetDirectoryName(filePath);
+            if (!Directory.Exists(dir))
+                Directory.CreateDirectory(dir);
+
             // Write.
             File.WriteAllText(filePath, code);
             return true;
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/InputActionCodeGenerator.cs
```diff
@@ -472,6 +472,13 @@ public static bool GenerateWrapperCode(string filePath, IEnumerable<InputActionM
                     return false;
             }
 
+            if (!filePath.ToLower().StartsWith("assets/"))
+                filePath = Path.Combine("Assets", filePath);
+
+            var dir = Path.GetDirectoryName(filePath);
+            if (!Directory.Exists(dir))
+                Directory.CreateDirectory(dir);
```
Rene-Damm: Do these steps in importer instead of here. No need to hardcode the code generator in terms of where it writes the file to.

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionImporter.cs
```diff
@@ -203,10 +213,7 @@ public override void OnImportAsset(AssetImportContext ctx)
                 };
 
                 if (InputActionCodeGenerator.GenerateWrapperCode(wrapperFilePath, maps, asset.controlSchemes, options))
-                {
-                    // Inform database that we modified a source asset *during* import.
-                    AssetDatabase.ImportAsset(wrapperFilePath);
-                }
+                    EditorApplication.update += DelayedRefresh;
```
Rene-Damm: Can we replace this with `EditorApplication.delayCall += AssetDatabase.Refresh;` ?

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionImporterEditor.cs
```diff
@@ -17,6 +17,7 @@ internal class InputActionImporterEditor : ScriptedImporterEditor
     {
         public override void OnInspectorGUI()
         {
+            serializedObject.Update();
```
Rene-Damm, line 20: Explain why we need this here. Counterintuitive as ScriptedImporterEditor should be expected to handle this for us.

Add blank line after statement.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Editor/InputActionImporter.cs
jechter: No. The point of the wrapper is to remove itself from `EditorApplication.update` once invoked once.

Rene-Damm: delayCall does that automatically.

jechter: Oh. I did not see that you replaced `update` with `delayCall`, which I did not know about. Then this is much nicer indeed!

Rene-Damm (approved):

PR #427, opened 2019-02-20
NEW: NPad vibration support

unity-cwells:


Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/INPadRumble.cs
```diff
@@ -0,0 +1,25 @@
+using UnityEngine.Experimental.Input.Haptics;
+
+namespace UnityEngine.Experimental.Input.Plugins.Switch
+{
+    /// <summary>
+    /// Extended dual motor gamepad that adds seperate controls for rumble frequency and amplitued.
```
Rene-Damm: Typo "amplitued".

Rene-Damm (changes requested):
Please add tests to `SwitchTests`.

PR #429, opened 2019-02-21
 Don't incorrectly create non-functioning devices

jechter:
Don't incorrectly create non-functioning devices if a physical device implements multiple incompatible logical HID devices.

This is the proper fix for the issue where we get errors recreating devices on domain reload. It also gets rid of the unusable garbage duplicate devices.

What was happening is that some devices on the mac report as multiple incompatible HID devices, most of which we have no use for. Then, if we'd detect one instance which we could use, we'd create a device, and a device matcher, which would then go back and create devices for all the HID devices we had no use for, as they have the same product and vendor ids. Then when domain reloading, depending on the order of devices, we may have not recreated that matcher yet, so we get an exception when trying to create them again.


Rene-Damm (changes requested):
Change looks good to me. Please add test.

jechter:
Test added.

PR #437, opened 2019-03-01
Allow binding dpad controls to actions of type "Vector2"

jechter:
This is an attempt to fix https://fogbugz.unity3d.com/f/cases/1129673/

I hope that this is the right level to fix it at.

Note that this assumes that assigning dpads to actions of type Vector2 _makes sense_. I don't have any gamepad at home to sanity-check this setup.


Packages/com.unity.inputsystem/InputSystem/Editor/InputControlPicker/InputControlPickerDropdown.cs
```diff
@@ -219,7 +219,16 @@ public void SetDeviceFilter(string[] deviceFilter)
...
+            switch (expectedControlLayout)
+            {
+                // Special case for Vector2, where we use the InputControl<Vector2> base type, to match both DpadControl and Vector2Control
+                case "Vector2":
+                    m_ExpectedControlLayoutFilterType = typeof(InputControl<Vector2>);
+                    break;
+                default:
+                    m_ExpectedControlLayoutFilterType = InputSystem.s_Manager.m_Layouts.GetControlTypeForLayout(new InternedString(expectedControlLayout));
+                    break;
+            }
```
Rene-Damm: It's an approach I definitely haven't thought of but I think it's not the right level to solve the problem at.

To give a bit more background, at the `InputControl` level, a dpad and a stick really should be interchangeable. For the most part they are except for the one fact that `DpadControl` has no `x` and `y` controls. The reason being that the state/control system requires jumping through some hoops to set those up such that they work regardless of how the dpad is stored in memory. It's not that difficult to solve, just hasn't been done yet.

If that was solved, `DpadControl` could simply derive straight from `Vector2Control` and that would implicitly solve the problem here without any further changes. I believe this is the proper fix for the problem.

There's also an alternative one-line change that would probably also solve the problem by changing

```
            RegisterControlLayout("Vector2", typeof(Vector2Control));
```

to

```
            RegisterControlLayout("Vector2", typeof(InputControl<Vector2>));
```

However, that has the downside of making the `x` and `y` controls disappear from the `Vector2` layout. Probably not a good idea :)

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Editor/InputControlPicker/InputControlPickerDropdown.cs
jechter: Yeah, I have considered both those approaches (changing 
RegisterControlLayout, and making DpadControl derive from Vector2Control). Hence my comment "I hope that this is the right level to fix it at.".

I'll look into making them derive, but will do that next week when I can test with an actual device in the office.

jechter: >The reason being that the state/control system requires jumping through some hoops to set those up such that they work regardless of how the dpad is stored in memory.

where would you implement that in the code?

Rene-Damm: Might be a good opportunity to spread some familiarity with the layout and control stuff, so bear with me for a bit of a roundabout explanation. Some of it is not directly relevant but I think it'll help add some contextual info.

The root of the problem is that a) controls do not store state themselves and b) usually controls have little say in how exactly their associated state is stored. Some make specific assumptions but for the most part, controls support whatever can be converted to the kind of values they need.

For dpads in particular, we may be looking at a bitfield (e.g. XInput), an enum (e.g. PS4/HID), or 4 floats (e.g. iOS). This means that dpad/up, dpad/down, dpad/left, and dpad/right may or may not share state and that the conversion to floats is usually quite specific to the device we're looking at.

So for adding dpad/x and dpad/y we need to somehow synthesize a value from the state of both dpad/up&down and dpad/left&right. 

The layout system directly supports controls *sharing* state via `useStateFrom` but that doesn't help here as we need the state from two controls, not just one. The controls that do this kind of stuff usually do it by adding a new InputControl-derived class. In fact, all the controls that are made up of other controls work this way by synthesizing their own values from the values they read from their children. DpadControl itself is that kind of thing.

The problem here is that e.g. dpad/x can't have dpad/left and dpad/right as children. So the control has to either somehow refer to siblings and read from them or has to refer to and read the underlying state from both directions itself. Either way would be doable.

Not sure which one would work out better. My guess is the second option will work better. It's basically the same thing that StickControl does. Probably will still require a new type of control to read out the two state values and combine them.

Packages/com.unity.inputsystem/InputSystem/Controls/DpadAxisControl.cs
```diff
@@ -0,0 +1,21 @@
...
+            component = name == "x" ? 0 : 1;
+        }
+
+        public override unsafe float ReadUnprocessedValueFromState(void* statePtr)
+        {
+            var value = (m_Parent as DpadControl).ReadUnprocessedValueFromState(statePtr);
+            return value[component];
+        }
+    }
+}
```
Rene-Damm: There's a complication with this I didn't think about. For various things such as actions to work correctly with controls, a control must have a proper state record. I.e. there must be a block of state memory associated with the control. Actions, for example, will monitor the memory block for changes and drive change processing based on that.

Packages/com.unity.inputsystem/InputSystem/Controls/DpadAxisControl.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace UnityEngine.Experimental.Input.Controls
+{
+    public class DpadAxisControl : AxisControl
+    {
+        public int component;
+
+        protected override void FinishSetup(InputDeviceBuilder builder)
+        {
+            base.FinishSetup(builder);
+            component = name == "x" ? 0 : 1;
```
Rene-Damm: Hmm, this seems rather hacky. Wonder if there is a more elegant solution.

Rene-Damm (changes requested):

PR #439, opened 2019-03-04
CHANGE: Disable GitLab CI runners, Enable GitHub Yamato CI runners

douglas-piconi:
With Gitlab CI we have basically lost ability to test on OSX and they are no longer maintaining the licenses.
We need to move over to Yamato and the GitHub CI system.

This PR accomplishes the following:

1. Removes our compiler/warnings tests.  These cause issues with GitHub Package CI editor runners , and I never really liked these tests, because at any given time we always have a compiler or warning error in the demo project as well as regular, and they hid test failures that we needed to see, because we were always red across the board.   The new GitHub package CI runner will test and fail on compiler or warnings errors on the package verification step.

2. Removes the GitLab CI tests scripts.  

3. Enables GitHub Package CI on yamato, will verify package as well as run all of our playmode and editmode tests on 

Windows, OSX, 2018.3, 2019.1, trunk


jechter (approved):
lgtm, but not familiar with CI setup, so that is not saying much.

PR #440, opened 2019-03-05
Removed non-functioning sort triangles in event list in Input Debugger device windows.

jechter:
Currently, in the event lists in the Input Debugger device windows, we show clickable sort triangles for each column. But these don't work, as we have not implemented the logic to sort by the different data types. I don't think that in our case there is any actual user need for sorting this data by different types, so I removed the disfunctional search triangle widgets.


Rene-Damm (approved):
The thought behind InputEventTreeView is to allow showing an arbitrary InputEventTrace. In that use, I think sorting would actually be quite useful as a trace could contain data from arbitrary devices.

However, given for now the only use we have is in the input device debugger where we only show a trace for a single device, I think the change is fine for now. Maybe add `////TODO: implement sorting` or something at the top of the file.

PR #445, opened 2019-03-06
Make parsing of parameters more robust.

jechter:
Our argument parsing failed to detect some valid float formats, resulting in exceptions would would then break the actions UI: https://fogbugz.unity3d.com/f/cases/1133702/ .

This PR fixes that, and also uses TryParse to avoid throwing exceptions for unexpected formats.

Also, added tests.


Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Layouts.cs
```diff
@@ -1853,6 +1853,63 @@ public void Layouts_CanSetUpMultipleLayoutsFromSingleState_UsingVariants()
...
+        var parameters = InputControlLayout.ParseParameters(input);
+        Assert.AreEqual(expectedValues.Length, parameters.Length, "Number of parameters does not match.");
+        unsafe
+        {
+            for (int i = 0; i < expectedValues.Length; i++)
+            {
+                var p = parameters[i];
+                Assert.AreEqual(parameterType, p.type, $"Unexpected type in parameter {i}");
+                Assert.AreEqual($"p{i}", p.name, $"Unexpected name in parameter {i}");
+                Assert.AreEqual(expectedValues[i], *(T*)p.value, $"Unexpected value in parameter {i}");
```
Rene-Damm: Use `Assert.That` syntax. See other tests.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Layouts.cs
```diff
@@ -1853,6 +1853,63 @@ public void Layouts_CanSetUpMultipleLayoutsFromSingleState_UsingVariants()
...
+        Layouts_ParseParameters_GenericWorks(
+            new int[] { 0, 0, 1234, -5678, 0 },
+            InputControlLayout.ParameterType.Integer,
+            "p0=0,p1=-0,p2=1234,p3=-5678,p4=12345678901234567890"
+        );
+    }
+
+    [Test]
+    [Category("Layouts")]
+    public void Layouts_ParseParameters_FloatWorks()
```
Rene-Damm: No `XYZWorks` style tests. Try to always formulate the test as a requirement where verifying that things work correctly is more or less a side effect and the test focuses on specifying an ability the system should have or trait it should exhibit.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Layouts.cs
```diff
@@ -1853,6 +1853,63 @@ public void Layouts_CanSetUpMultipleLayoutsFromSingleState_UsingVariants()
...
+        {
+            for (int i = 0; i < expectedValues.Length; i++)
+            {
+                var p = parameters[i];
+                Assert.AreEqual(parameterType, p.type, $"Unexpected type in parameter {i}");
+                Assert.AreEqual($"p{i}", p.name, $"Unexpected name in parameter {i}");
+                Assert.AreEqual(expectedValues[i], *(T*)p.value, $"Unexpected value in parameter {i}");
+            }
+        }
+    }
```
Rene-Damm, line 1880: Remove helper and replace the tests with a single test that takes all the test permutations as TestCase attributes.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -1243,7 +1243,7 @@ private static NameAndParameters ParseNameAndParameters(string text, ref int ind
             return new NameAndParameters {name = name, parameters = new ReadOnlyArray<ParameterValue>(parameters)};
         }
 
-        private static ParameterValue[] ParseParameters(string parameterString)
+        internal static ParameterValue[] ParseParameters(string parameterString)
```
Rene-Damm: In general, as much as possible, avoid isolating arbitrary pieces from a system for the sake of testing. The tests in `CoreTests_Layouts` should be focused on using layouts as a complete feature. All the CoreTests are integration tests and should not do unit testing.

In this case, I think it's okay-ish. The parameter parsing really should be lifted out of `InputControlLayout` entirely as it has long stopped being specific to just layouts. When that is done, having a unit test suite for just the parameter parsing probably makes sense.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Don't incorrectly create non-functioning devices if a physical device implements multiple incompatible logical HID devices (such as the MacBook keyboard/touch pad and touch bar).
 - Removed non-functioning sort triangles in event list in Input Debugger device windows.
 - Sort events in input debugger window by id rather then by timestamp.
+- Make parsing of parameters more robust.
```
Rene-Damm: Bit fuzzy :)

Rene-Damm (changes requested):
Change itself LGTM. Added comments re the tests.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Layouts.cs
jechter: I don't think that works with the generic argument.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
jechter: I agree that it might make sense to move this logic out, but for now, I changed the test code to use the non-private ParseNameAndParameters method.

PR #446, opened 2019-03-07
Lazy register analytics events only when they are being used

jechter:
As part of the discussion on what to do to make domain reloads faster, I suggested not registering the analytics events on every domain reload (as that seems to take ~15ms of time for some reason I don't understand).

This PR registers analytics events only when they are first sent, saving those 15ms on every reload.

It does keep the original API where you register events using a separate method, specifying maxPerHour & maxPropertiesPerEvent, and will remember those values, in order to later use them when actually registering the events - and it tracks if an event has been registered so it will only register it once. Currently, this is not really needed and we could simplify this code further, because 1. we always specify 10 & 100 for maxPerHour & maxPropertiesPerEvent, and 2. we never send an event twice. So we could simply always call register event on sending the event, which would serve the current use with less code. But I don't know how much flexibility for future use we want to keep, so I decided to keep the API unchanged for now. Let me know if I should change this.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using Unity.Collections.LowLevel.Unsafe;
 using UnityEngineInternal.Input;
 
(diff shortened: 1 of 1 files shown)


Rene-Damm:
Would prefer the latter change, i.e. solving it directly in InputAnalytics and keeping NativeInputRuntime as is. Basically, nuking InputAnalytics.Initialize entirely and just adding some bools that govern event registration right before we actually send an event.

```
        public static void OnStartup(InputManager manager)
        {
            if (!s_StartupEventRegistered)
            {
                manager.m_Runtime.RegisterAnalyticsEvent(kEventStartup, 10, 100);
                s_StartupEventRegistered = true;
            }
````

Rene-Damm (changes requested):
See comment.

Rene-Damm:
Looks like this one has been obsoleted by #454.

PR #447, opened 2019-03-07
Make keyboard Initialization faster

jechter:
One of the things  I looks  at when investigating Domain reload performance (and player startup times) is that `Keyboard.FinishSetup` seems to be rather slow (~11ms on my MBP). I believe that to be due to the compiler inlining the creation of each InputControl, making this a very long function, resulting in a lot of time being spent in JIT. This is wasted optimization here, of course, since this function is not called often. So instead I changed the InputControl generation to be a loop, reducing the time it takes for this method to run to ~4ms. Not a great saving, but still worth it IMO.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
@@ -423,7 +423,7 @@ namespace UnityEngine.Experimental.Input
         /// <summary>
         /// The space bar key.
         /// </summary>
-        public KeyControl spaceKey { get; private set; }
+        public KeyControl spaceKey => this[Key.Space];
 
         /// <summary>
         /// The enter/return key in the main key block.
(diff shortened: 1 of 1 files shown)


Rene-Damm:
Was wondering, can we also solve the problem using `MethodImplOptions.NoInlining` ([link](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimploptions?redirectedfrom=MSDN&view=netframework-4.7.2))?

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -596,149 +596,10 @@ public string keyboardLayout
...
-                    case Key.OEM1: return oem1Key;
-                    case Key.OEM2: return oem2Key;
-                    case Key.OEM3: return oem3Key;
-                    case Key.OEM4: return oem4Key;
-                    case Key.OEM5: return oem5Key;
-                }
-
-                throw new ArgumentException("key");
+                var index = (int)key;
+                if (index <= 0 || index > m_Keys.Length)
```
Rene-Damm: Range check looks like it's off by one. `index >= m_Keys.Length`

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -596,149 +596,10 @@ public string keyboardLayout
...
+                    throw new ArgumentException("key");
```
Rene-Damm: Throw ArgumentOutOfRangeException. Use nameof.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -596,149 +596,10 @@ public string keyboardLayout
...
+                return m_Keys[(int)key];
```
Rene-Damm, line 602: Seems like we can drop the first entry in the array altogether and just subtract one here.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -757,135 +618,134 @@ protected override void OnRemoved()
                 current = null;
         }
 
+        KeyControl[] m_Keys;
```
Rene-Damm: Use explicit visibility modifiers.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -757,135 +618,134 @@ protected override void OnRemoved()
...
-            oem1Key = builder.GetControl<KeyControl>("OEM1");
-            oem2Key = builder.GetControl<KeyControl>("OEM2");
-            oem3Key = builder.GetControl<KeyControl>("OEM3");
-            oem4Key = builder.GetControl<KeyControl>("OEM4");
-            oem5Key = builder.GetControl<KeyControl>("OEM5");
-
             imeSelected = builder.GetControl<ButtonControl>("IMESelected");
 
-            ////REVIEW: Ideally, we'd have a way to do this through layouts; this way nested key controls could work, too,
-            ////        and it just seems somewhat dirty to jam the data into the control here
```
Rene-Damm, line 883: Please move REVIEW comment to keyCode assignment above.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -757,135 +618,134 @@ protected override void OnRemoved()
...
+                "f11",
+                "f12",
+                "oem1",
+                "oem2",
+                "oem3",
+                "oem4",
+                "oem5",
+            };
+            m_Keys = new KeyControl[keyStrings.Length];
+            for (int i = 1; i < keyStrings.Length; i++)
```
Rene-Damm: Use var.

Rene-Damm (changes requested):
Some minor stuff. Otherwise looks good to me.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -757,135 +618,134 @@ protected override void OnRemoved()
...
+            {
+                m_Keys[i] = builder.GetControl<KeyControl>(keyStrings[i]);
+                m_Keys[i].keyCode = (Key)i;
+            }
+            Debug.Assert(keyStrings[(int)Key.OEM5] == "oem5");
```
Rene-Damm: Always use the message arg on Debug.Assert. I made the mistake early on of not doing so consistently and then found it's often really hard to track down the asserts as unlike in native, it does not print the expression when an assert fails.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
jechter: Do we have any code style guidelines we follow? I'm asking because Unity's C# conventions specify dropping this (see https://ono.unity3d.com/unity-extra/unity-meta/raw/@/ReferenceSource/CSharp/Assets/CSharpReference.cs : 
`        // |    - Drop redundant access specifiers (leave off 'private' at type scope).` ).

jechter:
>Was wondering, can we also solve the problem using MethodImplOptions.NoInlining (link)?

I tried that, but it did not seem to speed things up.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
jechter: I changed it anyways, but it would be nice to clarify what guidelines (if any) we follow?

Rene-Damm (approved):

PR #448, opened 2019-03-07
FIX: Kept around array after a disposal of the memory.

douglas-piconi:
One of the array is holding a pointer that was disposed of, need to reassign it to the new index of the binding arrays.


jechter (approved):

Rene-Damm (approved):
Good catch Doug!

PR #453, opened 2019-03-11
Use native InputEvent struct from unity

jechter:
As discussed, we'd like to be able to change the struct layout of the Input Events to avoid the misaligned double. To do so, I am changing the InputEvent struct to encapsulate UnityEngineInternal.Input.NativeInputEvent instead of duplicating it's content. That way, we hand control over the struct layout back to Unity, where we can then change it without breaking the package.

Since this change is internal to the InputEvent class, no API or functionality should be changed by this.

Caveat: We can now change the native layout of NativeInputEvent, but we cannot change it's size, as that is hardcoded in InputEvent.kBaseEventSize. This is because other structs containing InputEvents need this constant for their explicit layouts, and C# does not allow compile-time sizeof of structs (as the size could change at runtime if the struct contained pointers or references, which is not the case here, but still a language restriction).

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Events/InputEvent.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Events/InputEvent.cs
@@ -11,7 +11,7 @@ namespace UnityEngine.Experimental.Input.LowLevel
     /// A chunk of memory signaling a data transfer in the input system.
     /// </summary>
     // NOTE: This has to be layout compatible with native events.
-    [StructLayout(LayoutKind.Explicit)]
+    [StructLayout(LayoutKind.Explicit, Size = kBaseEventSize)]
     public struct InputEvent
     {
         private const uint kHandledMask = 0x80000000;
@@ -120,6 +120,14 @@ namespace UnityEngine.Experimental.Input.LowLevel
         {
             get => m_Event.time;
             set => m_Event.time = value;
+        }
+
+        static InputEvent()
+        {
+            unsafe
+            {
+                Debug.Assert(kBaseEventSize == sizeof(NativeInputEvent), "kBaseEventSize sizemust match NativeInputEvent struct size.");
+            }
         }
 
         public InputEvent(FourCC type, int sizeInBytes, int deviceId, double time = -1)
--- a/Packages/com.unity.inputsystem/InputSystem/Events/InputEvent.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Events/InputEvent.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Runtime.InteropServices;
 using UnityEngine.Experimental.Input.Utilities;
+using UnityEngineInternal.Input;
 
 ////REVIEW: can we get rid of the timestamp offsetting in the player and leave that complication for the editor only?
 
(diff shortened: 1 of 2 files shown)


jechter:
Regarding `InputEvent.kBaseEventSize` -  I think having that hardcoded is problematic, as it means we don't have real flexibility of changing this struct on the native side. What if we did this for any event extending InputEvent (using `DeltaStateEvent` as a sample):

instead of:
```
    [StructLayout(LayoutKind.Explicit, Pack = 1, Size = InputEvent.kBaseEventSize + 9)]
    public unsafe struct DeltaStateEvent : IInputEventTypeInfo
    {
        public const int Type = 0x444C5441; // 'DLTA'

        [FieldOffset(0)] public InputEvent baseEvent;
        [FieldOffset(InputEvent.kBaseEventSize)] public FourCC stateFormat;
        [FieldOffset(InputEvent.kBaseEventSize + 4)] public uint stateOffset;
        [FieldOffset(InputEvent.kBaseEventSize + 8)] public fixed byte stateData[1]; // Variable-sized.
```
we'd do
```
    [StructLayout(LayoutKind.Explicit, Pack = 1, Size = InputEvent.kBaseEventSize + 9)]
    public unsafe struct DeltaState
    {
        [FieldOffset(0)] public FourCC stateFormat;
        [FieldOffset(4)] public uint stateOffset;
        [FieldOffset(8)] public fixed byte stateData[1]; // Variable-sized.
    }

    [StructLayout(LayoutKind.Sequential)]
    public unsafe struct DeltaStateEvent : IInputEventTypeInfo
    {
        public const int Type = 0x444C5441; // 'DLTA'

        public InputEvent baseEvent;
        public DeltaState state;
```

Does that make sense?

Rene-Damm:
@jechter I'm good with that. In the example `DeltaState` would be in `UnityEngineInternal.Input`, right?

Would prefer to completely hide the separate struct we have from the API. Guess a point can be made, though, that someone who also accesses `UnityEngineInternal.Input` would want to have it accessible. Still, all the xxxState and xxxEvent stuff would make for a confusing API IMO.

Another thing I'm wondering about is whether DeltaStateEvent and the like will now come out with the wrong sizeof. That may break some code.

jechter:
@Rene-Damm 
>In the example DeltaState would be in UnityEngineInternal.Input, right?

Yes, but - I have been thinking a bit more. I made a Unity branch where I added exposed structs for _all_ event types, state types and command types, the thinking here being that a dependency to an explicit API is better then having the same dependency, just on the lower binary level - as the API would let use fix things like we have today with the double alignment after the fact.

But the problem with that thought is:
API compatibility means that whenever we want to support a new event/state/command format, the input system package would then only compile with a unity version which exposes that command struct in the API - thus making package compatibility with different Unity versions very hard. Without requiring an API, it is easy: older versions will simply never send that Event type, or simply ignore such commands. So I think this is a big advantage, and not being able to break binary compatibility is a valid price to pay (if we _have_ to change binary compatibility, we would then just have to make a new version of the event/command with a different FourCC). 

So I think let's keep things as they are - beyond the scope of this PR, which I still think makes sense. The base Input Event is fundamental, and allowing us to change/expand that is valuable, even if we cannot change the layout of individual event types. The only think I'd like to add on top of this is to move the size constant to Unity API, so we can change the size of the struct as well (But that requires landing such a const in Unity first).

So I propose landing this PR as is.

Rene-Damm:
>So I think this is a big advantage, and not being able to break binary compatibility is a valid price to pay

Yup, agree. That was exactly what the stuff was designed for. E.g. the whole IOCTL stuff was born out of the desire to have the native input system work as a pure transport layer that has no understanding of the structure of the data it transports. So you can extend it at both ends and not break the API.

>So I propose landing this PR as is.

Yup, makes sense to me.

Rene-Damm (approved):

PR #455, opened 2019-03-11
Expose mouse clickCount control

jechter:
Added a `clickCount` control to the `Mouse` class, which specifies the click count for the last mouse click (to allow distinguishing between single-, double- and multi-clicks).


Rene-Damm:
Hmm, one thing I didn't realize before and which, looking at the new control, makes me wonder... how will this work? Specifically, what duration or reference point does the count refer to?

`delta` is kind of the same deal. It's one of these controls that don't really give just a state of a control but rather present a value that is relative to something in else. In the case of `delta`, it's even quite confusing. At the event level, `delta` is the amount of motion relative to the previous event whereas when looking at state, it's the amount of motion in the current frame/update (another confusing thing that I'd like to get rid of).

What's clickCount relative to? When is it reset?

jechter:
`clickCount` is relative to the last mouseDown event. It is updated whenever a button is pressed down. And, yes - I agree that this, like delta, does not really fit the concept of a "state" as it is an event property which is not meaningful without context. But I am not sure if there is any better way to convey this information (and I think that this information _is_ needed). I guess we could take delta and clickCount entirely out of the state, and use custom events for them (like we handle keyboard text input currently) - but I guess that would require a lot of extra logic to deal with them (and at least for the `delta`s (scroll + wheel), they are useful even without context, as you can use them as input axes.

Rene-Damm (approved):

PR #458, opened 2019-03-11
FIX: Native landed, enable on 2018.3.

douglas-piconi:
NativeEvent changes are now in public Unity 2018.3.8.f1   Need to enable for package compilation.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
@@ -98,7 +98,7 @@ namespace UnityEngine.Experimental.Input.LowLevel
                             eventBufferPtr->sizeInBytes = 0;
                         }
                     };
-                    #elif UNITY_2019_1
+                    #elif UNITY_2018_3_OR_NEWER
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -98,7 +98,7 @@ public unsafe InputUpdateDelegate onUpdate
                             eventBufferPtr->sizeInBytes = 0;
                         }
                     };
-                    #elif UNITY_2019_1
+                    #elif UNITY_2018_3_OR_NEWER
```
Rene-Damm: We can simply nuke the other codepath. 2018.3 now being our min version, we don't need explicit UNITY_2018_3_OR_NEWER blocks anymore.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
douglas-piconi: @Rene-Damm fixed.

PR #459, opened 2019-03-11
Add support for joysticks on Linux

dbuckley3d:
handle linux joystick: create hatswitches and sticks appropriately


StayTalm (approved):
Cleaned up unwanted changes, formatter doesn't seem to be run enough.

Packages/com.unity.inputsystem/InputSystem/Plugins/Linux/SDLDeviceBuilder.cs
```diff
@@ -0,0 +1,329 @@
+using System;
+using System.Collections.Generic;
+using UnityEngine.Experimental.Input.LowLevel;
+using UnityEngine.Experimental.Input.Utilities;
+using System.Text;
+using UnityEngine.Experimental.Input.Layouts;
+
+namespace UnityEngine.Experimental.Input.Plugins.Linux
+{
+    [Serializable]
```
Rene-Damm, line 10: Oh, that makes me realize that this requirement is actually no longer necessary as layout information is no longer persisted but rather re-registered from scratch on every domain reload. Something to fix on the input system side.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/LinuxTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        Assert.That(generatedLayout.controls, Has.Count.EqualTo(16));
+
+        var triggerControl = generatedLayout.controls[0];
+        Assert.That(triggerControl.name, Is.EqualTo(SDLSupport.GetButtonNameFromUsage(SDLButtonUsage.Trigger)));
+        Assert.That(triggerControl.offset, Is.EqualTo(0));
+        Assert.That(triggerControl.layout, Is.EqualTo(new InternedString("Button")));
+    }
+
+    [Test]
+    public void State_LinuxDeviceTest()
```
Rene-Damm, line 152: Each test should read as and be a requirement. E.g. stuff like `Devices_HIDsIgnoreUnusedExcessElements()` or `Actions_CanDetermineIfMapIsUsableWithGivenDevice`.

Assets/Plugins/x86_64/libsteam_api.so.meta
```diff
@@ -56,7 +58,7 @@ PluginImporter:
     second:
       enabled: 1
       settings:
-        CPU: x86_64
+        CPU: AnyCPU
```
Rene-Damm, line 61: Why that?

Rene-Damm (approved):

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/LinuxTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        var generatedLayout = InputSystem.TryLoadLayout("Linux::TestManufacturer::TestProduct");
+
+        Assert.That(generatedLayout, Is.Not.Null);
+        Assert.That(generatedLayout.controls, Has.Count.EqualTo(16));
+
+        var triggerControl = generatedLayout.controls[0];
+        Assert.That(triggerControl.name, Is.EqualTo(SDLSupport.GetButtonNameFromUsage(SDLButtonUsage.Trigger)));
+        Assert.That(triggerControl.offset, Is.EqualTo(0));
+        Assert.That(triggerControl.layout, Is.EqualTo(new InternedString("Button")));
+    }
```
Rene-Damm, line 149: @dbuckley3d A test like this is really hard to debug. Right now, it's failing because the count is 32 and not 16. It's hard to see why as it's not very clear what the test is actually testing. Instead of having a very long descriptor string with repetitive data, a better setup would be to have very specific and minimal test data that makes it very obvious what functionality is being established.

Packages/com.unity.inputsystem/InputSystem/Plugins/Linux/SDLDeviceBuilder.cs
```diff
@@ -0,0 +1,329 @@
...
+            }
+
+            if (deviceDescriptor == null)
+            {
+                return null;
+            }
+
+            if (string.IsNullOrEmpty(matchedLayout))
+            {
+                //matchedLayout = "Joystick";
```
Rene-Damm, line 66: Why commented out?

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/LinuxTests.cs
```diff
@@ -0,0 +1,302 @@
...
+        Assert.That(thumb2Control, Is.TypeOf(typeof(ButtonControl)));
+
+        InputControl topControl = createdDevice[SDLSupport.GetButtonNameFromUsage(SDLButtonUsage.Top)];;
+        Assert.That(topControl, Is.TypeOf(typeof(ButtonControl)));
+
+        InputControl top2Control = createdDevice[SDLSupport.GetButtonNameFromUsage(SDLButtonUsage.Top2)];
+        Assert.That(top2Control, Is.TypeOf(typeof(ButtonControl)));
+
+        Assert.That(createdDevice[SDLSupport.GetButtonNameFromUsage(SDLButtonUsage.Top)], Is.Not.Null);
+        Assert.That(createdDevice[SDLSupport.GetButtonNameFromUsage(SDLButtonUsage.Top2)], Is.Not.Null);
```
Rene-Damm, line 179: Tests for null what was already tested for type.

Packages/com.unity.inputsystem/InputSystem/Plugins/Linux/SDLDeviceBuilder.cs
```diff
@@ -0,0 +1,329 @@
...
+            if (xFeature.offset <= yFeature.offset)
+            {
+                byteOffset = xFeature.offset;
+            }
+            else
+            {
+                byteOffset = yFeature.offset;
+            }
+
+            var stickName = "Stick";
```
Rene-Damm, line 110: What if there is more than one stick?

Packages/com.unity.inputsystem/InputSystem/Plugins/Linux/SDLDeviceBuilder.cs
```diff
@@ -0,0 +1,329 @@
...
+                    {
+                        SDLAxisUsage usage = (SDLAxisUsage)feature.usageHint;
+                        string featureName = SDLSupport.GetAxisNameFromUsage(usage);
+                        string parameters = "scale,scaleFactor=65538,clamp,clampMin=-1,clampMax=1";
+
+                        if (IsAxisX(feature) && i + 1 < descriptor.controls.Count)
+                        {
+                            SDLFeatureDescriptor nextFeature = descriptor.controls[i + 1];
+                            if (IsAxisY(nextFeature))
+                                BuildStickFeature(ref builder, feature, nextFeature);
```
Rene-Damm, line 264: This seems to build a stick *plus* X and a Y controls at the same level. That doesn't seem useful given sticks already have X and Y axes.

Packages/com.unity.inputsystem/InputSystem/Plugins/Linux/SDLDeviceBuilder.cs
Rene-Damm: Wondering in general, is every axis and button usage guaranteed to be unique? If so, that point is moot.

Packages/com.unity.inputsystem/Tests/InputSystem/Plugins/LinuxTests.cs
```diff
@@ -0,0 +1,302 @@
...
+
+        public void SetButtonValue(SDLButtonUsage usage, bool value)
+        {
+            bool useButtonMask2 = false;
+            byte bitflag = 0;
+            switch (usage)
+            {
+                case SDLButtonUsage.Trigger:
+                {
+                    bitflag = 1 >> 0;
```
Rene-Damm, line 31: Isn't this shifting the wrong way?

PR #461, opened 2019-03-12
Made Vector2 composite actions respect the press points of button controls used to compose the value.

jechter:
Fixes https://fogbugz.unity3d.com/f/cases/1131844/


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -2089,10 +2091,14 @@ internal TValue ReadCompositePartValue<TValue>(int bindingIndex, int partNumber)
                     {
                         result = value;
                         isFirstValue = false;
+                        if (controls[controlIndex] is Controls.ButtonControl)
```
Rene-Damm, line 2117: Pattern matching obviates the need for the second cast.

`if (controls[controlIndex] is ButtonControl button)`

Rene-Damm (commented):
TBH, on the fence about this.

The thing that is bothering me is that I'm more and more thinking that `pressPoint` on ButtonControl is probably a bad design and should most likely just be nuked and forgotten. I mean, how is `pressPoint` different to deadzoning we do on sticks? If it isn't, why don't we just deadzone certain buttons that are actually true axes underneath. And if we do that, there's no need to treat a button any different than any other float control here. So yeah, I'm thinking I simply went down a wrong path and introduced something that shouldn't be there in the first place.

What do you think @jechter?

jechter:
I guess technically no, this is not different from a dead zone. Can we set up a dead zone on the control so that values are automatically processed by it without requiring further manual setup? Then that might be a more consistent approach indeed.

Rene-Damm:
@jechter and I talked this one over quite a bit and in the end agreed to keep button press points.

Rene-Damm (approved):

jechter:
FIX: Made Vector2 composite actions respect the press points of button controls used to compose the value.

PR #462, opened 2019-03-12
CHANGE: Rewrite action editor to address a number of issues.

### LINKED ISSUE #393: OnScreenButtonEditor script causes errors as soon as you add OnScreenButton component
making `public class OnScreenButtonEditor` not derive from `OnScreenControlEditor` fixes all of them

Rene-Damm:
Sorry for the massive PR guys...

Large refactor/rewrite of almost all action editor UI code to get rid of large amounts of copypaste code and address numerous issues.

- Rewrote tree views to collapse four different tree view classes with lots of copypaste code into one single, much smaller implementation.
  * Can now flexibly decide what to include in a tree.
  * Inspector and asset based action editing now largely sits on the same code paths. This will avoid the kind of breakage we've frequently seen where the less often used inspector-based code path breaks after updates to the asset-based code path.
- Rewrote toolbar code to get rid of of a bunch of unnecessary code and to get rid of spaghetti dependencies.
- Rewrote search/item filtering code
  * Side-effect: can now enter queries directly into search box (e.g. "d:Gamepad" to filter for gamepad bindings)
- Refactored control picker code to properly expose picker APIs to users.
- Redid how the tree refers to serialized data.
  * Everything has GUIDs now; maps and actions already did, bindings have them, too, now.
  * Data is deleted by ID not by index; fixes problems where the wrong data was deleted because indices were shifting around.
- Item IDs are now based on GUIDs of the data.
  * This means that they are stable; before, given that IDs depended on array indices, when things moved around (such as when deleting items), IDs ended up referring to the wrong items.
- Replaced custom OnScreenControl inspectors with simple custom property drawer for string properties that denote input control paths.
  * Reusable by any user code. Simply put `InputControlAttribute` on your `string`-type property and you'll get an proper inspector that allows to pick a control path (from either the picker or interactively).
  * Fixes #393.
  * As a side-effect, was able to nuke all code in Plugins/OnScreen/Editor.
- Changed how editing control paths as plain text is used
  * Button is now labelled "T" instead of "...".
  * It's a toggle now that toggles between picking using the control picker and entering paths manually.
- Rewrote copypaste code for action editing
  * Fixed copypasting not preserving action properties other than name.
  * Can now copypaste multiple items of different types.
- Drag&drop code has been rewritten to fix a number of issues and allow many more operations
  * Can now alt-drag to dupliate data instead of moving it.
  * Can drag&drop items *between* different parents (e.g. drag a binding from one action to another).
  * Can drag&drop multiple items now.
- Can now re-assign which part of a composite a binding belongs to.
  * Can also copypaste/duplicate part bindings now.
  * Means it's now possible not just in the data but also in the UI to set up composites where a particular part is bound multiple times. This is useful, for example, to setup WASD controls that are bound both to WASD as well as arrow keys.
- Replaced binding overlay ("Waiting for input...") with a progress bar.
  * Works the same for wherever interactive picking is used.
  * Can cancel by clicking cancel button.
  * Interactive rebinds in editor UI now time out automatically after 4 seconds if no suitable input is detected.
  * Replace "O" character on interactive binding button with a button icon.
- Fixed old name of action or map being visible in background while renaming.
- Fixed property drawers for actions and action maps hanging around on the global undo callback.
- Fixed deleting multiple items (e.g. multiple action maps) at the same time leading to either the wrong entries getting deleting or indices being out of bounds.
- Rewrote undo handling with a much simpler solution based on checking object dirty counts.
  * Also fixes property drawers leaking global Undo.undoRedoPerformed hooks.
  * Also protects against *all* outside SerializedObject changes, not just undo/redo.
- Rearranged the Editor/ folder for a clearer organization.
- Undo now works for changes made to control schemes.
- Control schemes are now sorted alphabetically.
- Setting a device filter now takes layout inheritance into account (e.g. if you select "Mouse" bindings to "Pointer" are included).
- Can now access action properties when editing action maps embedded in components.
- When a control scheme is selected, bindings that are not in any control scheme are now also shown but tagged with "{GLOBAL}".
- Added a number of tests.
- Bunch of stuff I forgot I fixed or changed along the way...

Known Issues:
- Sometimes when using right-click context menus in the action editor, there's an error about mismatched layout groups
- Regresses the editor beeping behavior that @jechter fixed in #419.
- When binding interactively, sometimes there is an error about a native collection that has not been disposed. Haven't been able to track it down but suspect that some unmanaged memory held on to by `RebindingOperation` is not getting released when it should.


Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeViewItems.cs
```diff
@@ -0,0 +1,530 @@
...
+            var assetObject = property.serializedObject;
+            if (!(assetObject.targetObject is InputActionAsset))
+                throw new InvalidOperationException(
+                    $"Action map must be part of InputActionAsset but is in {assetObject.targetObject} instead");
+
+            InputActionSerializationHelpers.DeleteActionMap(assetObject, guid);
+
+            // When we delete an entry from an array of action maps, the SerializedProperties
+            // of all action map items following the one we deleted will now have incorrect array
+            // indices. Account for that
```
Rene-Damm: Leftover comment. No longer applies. Will delete.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
```diff
@@ -0,0 +1,1486 @@
...
+                menu.AddItem(s_RenameLabel, false, () => BeginRename(GetSelectedItems().First()));
+            }
+            else if (canRename)
+            {
+                menu.AddDisabledItem(s_RenameLabel);
+            }
+            menu.AddItem(s_DuplicateLabel, false, () =>
+            {
+                CopySelectedItemsToClipboard();
+                ////REVIEW: would be nice for duplicate to insert right after the current selection
```
Rene-Damm: Outdated comment. Will remove.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -32,19 +32,6 @@ public class InputActionImporter : ScriptedImporter
...
-        [SerializeField] private RememberedGuid[] m_ActionGuids;
-        [SerializeField] private RememberedGuid[] m_ActionMapGuids;
-
-        [Serializable]
-        internal struct RememberedGuid
-        {
-            public string name;
-            public string guid;
-        }
-
```
Rene-Damm, line 47: I nuked this mechanism as there isn't a good way to do the same for bindings. Plus the code was even wrong and it didn't work. Think overall it just added complication for something that is niche functionality.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
```diff
@@ -0,0 +1,277 @@
...
+                    });
+            }
+
+            m_PickerDropdown.SetControlPathsToMatch(m_ControlPathsToMatch);
+            m_PickerDropdown.SetExpectedControlLayout(m_ExpectedControlLayout);
+
+            m_PickerDropdown.Show(rect);
+        }
+
+        private void DrawInteractivePickingProgressBar()
```
Rene-Damm, line 237: I found that the progress bar is a vastly superior solution to the overlay I programmed before.

For one, it works everywhere regardless of where we have a pick button. So even when actions are embedded in components and shown inline in the inspector (where we have no access to the EditorWindow), we can give proper visual feedback.

Also, it make sense to have interactive picks time out, I think, and progress bars give very nice visual feedback about that.

The downside is that this change makes it harder to swallow keyboard input in the UI. We have to find a new solution for that.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
```diff
@@ -0,0 +1,1486 @@
...
+        {
+            serializedObject.ApplyModifiedProperties();
+            UpdateSerializedObjectDirtyCount();
+            Reload();
+            onSerializedObjectModified?.Invoke();
+        }
+
+        private void UpdateSerializedObjectDirtyCount()
+        {
+            m_SerializedObjectDirtyCount = EditorUtility.GetDirtyCount(serializedObject.targetObject);
```
douglas-piconi, line 1197: Looks like a compile issue here.  no definition for GetDirtyCount  on 2018.3

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeViewItems.cs
```diff
@@ -0,0 +1,530 @@
...
+                array = arrayProperty;
+                arrayIndex = this.arrayIndex + 1 +
+                    InputActionSerializationHelpers.GetCompositePartCount(array, this.arrayIndex);
+                return true;
+            }
+
+            return false;
+        }
+
+        public static CompositeBindingTreeItem AddTo(TreeViewItem parent, SerializedProperty bindingProperty)
```
douglas-piconi: new warning  warning CS0108: 'CompositeBindingTreeItem.AddTo(TreeViewItem, SerializedProperty)' hides inherited member

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeViewItems.cs
```diff
@@ -0,0 +1,530 @@
...
+                    m_ExpectedControlLayout = layoutName ?? "";
+                }
+
+                return m_ExpectedControlLayout;
+            }
+        }
+
+        private string m_ExpectedControlLayout;
+
+        public static PartOfCompositeBindingTreeItem AddTo(TreeViewItem parent, SerializedProperty bindingProperty)
```
douglas-piconi: new warning   warning CS0108: 'PartOfCompositeBindingTreeItem.AddTo(TreeViewItem, SerializedProperty)' hides inherited member 'BindingTreeItem.AddTo(TreeViewItem, SerializedProperty)'. Use the new keyword if hiding was intended.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeViewItems.cs
```diff
@@ -0,0 +1,530 @@
...
+        /// </summary>
+        public void AddBindingsTo(TreeViewItem parent)
+        {
+            var isSingleton = actionMapProperty == null;
+            var bindingsArrayProperty = isSingleton
+                ? property.FindPropertyRelative("m_SingletonActionBindings")
+                : actionMapProperty.FindPropertyRelative("m_Bindings");
+
+            var bindingsCountInMap = bindingsArrayProperty.arraySize;
+            var bindingCountForAction = 0;
```
douglas-piconi: new warning warning CS0219: The variable 'bindingCountForAction' is assigned but its value is never used

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
Rene-Damm: Crap...

Good find. Didn't realize this was newly introduced. What if we just take the plunge at this point and with the next package, move on to 2019.1?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlAttribute.cs
```diff
@@ -1,20 +1,36 @@
...
+    /// attribute to fields in a state struct (<see cref="IInputStateTypeInfo"/>, or <see cref="GamepadState"/>
+    /// for an example) or to properties in an input device (<see cref="InputDevice"/>), will cause an
+    /// <see cref="InputControl"/> to be created from the field or property at runtime. The attribute
+    /// can be applied multiple times to create multiple input controls (e.g. when having an int field
+    /// that represents a bitfield where each bit is a separate button).
+    ///
+    /// Another use is for marking <c>string</c> type fields that represent input control paths. Applying
+    /// the attribute to them will cause them to automatically use <see cref="InputControlPathDrawer"/>
+    /// when edited in inspectors.
     ///
```
jechter, line 27: As I see it this attribute is currently being used for two completely unrelated purposes. Wouldn't it be nicer to have different attributes for this?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorToolbar.cs
```diff
@@ -0,0 +1,661 @@
...
+    {
+        public void OnGUI()
+        {
+            EditorGUILayout.BeginHorizontal(EditorStyles.toolbar);
+            DrawSchemeSelection();
+            DrawDeviceFilterSelection();
+            if (!InputEditorUserSettings.autoSaveInputActionAssets)
+                DrawSaveButton();
+            GUILayout.FlexibleSpace();
+            DrawAutoSaveToggle();
```
jechter, line 35: Not something which should be decided for the scope of this already big PR, but:

Why are we handling these assets different from all other assets in Unity wrt saving state?

If I edit Materials, Prefabs, PhysicMaterials, GUISkins, etc, for any of these changes will immediately take effect, and be saved to disk the next time the project is saved. I have found the fact that action map assets behave differently confusing me a lot already (making changes; hitting play; not understanding why my changes don't work). 

Why are we inconsistent with all of Unity here? (Not saying that the way unity handles saving of assets is the most reasonable, but unless we want to change it for all of unity, I don't see what makes action maps different).

And if we think we have a good reason to be inconsistent with all of Unity, can we add a (very) visual dirty indicator to the window?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorToolbar.cs
```diff
@@ -0,0 +1,661 @@
...
+            GUILayout.Space(5);
+            DrawSearchField();
+            GUILayout.Space(5);
+            EditorGUILayout.EndHorizontal();
+        }
+
+        private void DrawSchemeSelection()
+        {
+            var buttonGUI = new GUIContent(selectedControlScheme?.name ?? "No Control Scheme");
```
jechter: "No Control Scheme" seems misleading - what we really seem to be doing is showing all controls regardless of scheme. Maybe it should be "Show all Control Schemes"?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
```diff
@@ -0,0 +1,753 @@
...
+
+            m_ActionAssetManager.CreateWorkingCopyAsset();
+            InitializeTrees();
+            LoadPropertiesForSelection();
+            Repaint();
+        }
+
+        #if UNITY_2019_1_OR_NEWER
+        ////FIXME: the shortcuts seem to have focus problems; often requires clicking away and then back to the window
+        [Shortcut("Input Action Editor/Save", typeof(InputActionEditorWindow), KeyCode.S, ShortcutModifiers.Alt)]
```
jechter, line 638: I'd suggest adding these keyboard shortcuts to the tooltips of the corresponding buttons to make them discoverable.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
```diff
@@ -0,0 +1,1466 @@
...
+        public void HandleCopyPasteCommandEvent(Event uiEvent)
+        {
+            var validate = uiEvent.type == EventType.ValidateCommand;
+            if (!validate && uiEvent.type != EventType.ExecuteCommand)
+                return;
+
+            switch (uiEvent.commandName)
+            {
+                case k_CopyCommand:
+                    if (!validate)
```
jechter: If we want to "properly" implement UI conventions here, then validate should actually perform validation:

For Paste, check if the paste buffer contains pasteable data (check for `INPUTASSET`).

For Copy/Cut/Delete/Duplicate, check if anything is selected.

jechter: (To be fair, unity is not very proper about this elsewhere, but that doesn't mean we cannot do better)

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
```diff
@@ -0,0 +1,1466 @@
...
+                    {
+                        CopySelectedItemsToClipboard();
+                        DeleteDataOfSelectedItems();
+                    }
+                    break;
+                case k_DuplicateCommand:
+                    if (!validate)
+                    {
+                        CopySelectedItemsToClipboard();
```
jechter: Duplicate should not change clipboard contents. Either use different code, or backup and restore clipboard contents around this block.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
jechter: I hate to disagree, but having tried it I don't think that a progress bar is an improvement here at all. I think a progress bar is the wrong UX here, and what we had before was better. Here's why:

1. A progress bar is a convention for "The computer is doing something, wait until it is over". If you don't _know_ what this is about (for instance because you just wanted to try what that little bullseye button does), and you see a progress bar, all your trained user instincts will tell you "Ok, here's a progress bar, so whatever it's doing, i'll sit here and wait for it". Followed by "Ok, so that button did something, taking some time. I wonder what it was?".

2. I don't understand what we need the timeout for here at all?

3. The progress bar breaks the contextual association of the binding with the window you are interacting with - the old UI made it clear that you are interacting with content in the actions window, the progress bar is detached from the window and thus breaks that association.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
```diff
@@ -333,7 +333,7 @@ void ISerializationCallbackReceiver.OnAfterDeserialize()
             s_Instance = this;
         }
 
-        class InputSystemTreeView : TreeView
+        private class InputSystemTreeView : TreeView
```
jechter, line 336: I asked before on slack, but what code convention are we following? (Because Unity's C# conventions says to drop these)

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/SerializedPropertyHelpers.cs
```diff
@@ -0,0 +1,365 @@
...
+using System.Text;
+using UnityEditor;
+using UnityEngine.Experimental.Input.Utilities;
+
+namespace UnityEngine.Experimental.Input.Editor
+{
+    /// <summary>
+    /// Helpers for working with <see cref="SerializedProperty"/> in the editor.
+    /// </summary>
+    internal static class SerializedPropertyHelpers
```
jechter, line 14: Can we have tests for these helpers?

Packages/com.unity.inputsystem/InputSystem/Utilities/StringHelpers.cs
```diff
@@ -7,8 +7,43 @@ namespace UnityEngine.Experimental.Input.Utilities
 {
     internal static class StringHelpers
     {
+        public static string Escape(this string str)
```
jechter, line 10: test?

jechter (changes requested):

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
Rene-Damm: Let's bounce it to @shawnmcclelland. This is his home turf.

Think you have a point about the role of progress bars and I'm probably not the best to judge how isn't coming in with clear expectations. Personally, I find it much nicer than the previous UI which always seemed a bit of a kludge to me and looked even more a kludge when adapted to other places where control picking was used.

The timeout is born more out of a hate for modal things that you can't get out of except through one specific UI element. Have found myself stuck in UIs that way plenty times.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlAttribute.cs
Rene-Damm: Nicer how? :) Think it'd mainly add more stuff to the API.

As it is, the attribute basically says "the following property or field represents an input control in some way". This is true for all three uses we currently have. Granted, not all the properties of the attribute apply for every use but IMO that's no big deal.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/SerializedPropertyHelpers.cs
Rene-Damm: Hah, accidentally hit resolve...

TBH I generally think there isn't much of a point to those tests. Sometimes unit tests can be a great benefit but for the most part, I don't care much what an internal API somewhere deep in the system does. To me, this stuff here is equivalent to private methods somewhere.

What I want the tests to do is specify how the user facing parts (public API and UI) are to behave and ensure that's what they do. So for example here, `InputActionTreeView` is the user-facing piece of code (in this case, it's not public API but it is "public" UI) so that's where I want to test. `SerializedPropertyHelpers` and `InputActionSerializedHelpers` get extensively exercised by the tests for the tree view. Writing tests for them separately to me is largely equivalent to writing tests for some private method in `InputActionTreeView`.

We do have a few isolated unit tests for some stuff that proved tricky to get right but IMO wherever we can get away with not having them, we're usually better off.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
jechter: Modal UIs are never great, and should be avoided if there is a better solution. But I'm not convinced that dismissing the modal UI after an arbitrary timeout is that better solution.

But maybe, we would not need a modal UI at all for this. I think ultimately, what this is, is a glorified input field. So what if we represent it as such? Just put in a text field, which instead of handling normal text input, process all device input as long as it is selected - and dismisses itself as soon as you click somewhere else? I'm not sure that this would work out well, but it is worth a try IMO.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorToolbar.cs
Rene-Damm: Yeah, I think that could have been done better.

It evolved out of a decision to have JSON text formats for all data consumed by the system. Which ended up putting the editor on a track where it saved out in that format.

I think there's also some other complications that I have mostly forgotten about. The data you mention above is all Unity native data. But we're sitting in a package and IIRC there were some problems there. But regardless of what it was, I'm thinking it probably would have been solvable.

FWIW I think the new shader graph stuff works in a similar way.

>can we add a (very) visual dirty indicator to the window?

What do you suggest? The current one IMO is pretty visible but could well be it should be more prominent.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
Rene-Damm: Hmm, good point. Makes me wonder, though... all this is remappable by the user, so we should probably look it up dynamically. Need to research this one more. I'm still quite unfamiliar with the new shortcut system.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
Rene-Damm: Fixed

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
Rene-Damm: Fixed

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
Rene-Damm: Added a TODO. Will investigate how to properly do this.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlAttribute.cs
jechter: I'm not sure. Usually when I learn of some API, I remember "this does that", and I find that more difficult to understand if something does very different things in different contexts.

I understand that for an attribute you can read it more as a label, ie "this _is_ that", then an instruction to _do_ something specific, so I see your point, but I'm not sure I find it intuitive. So let's leave this for now, unless anyone else has an opinion to add to the discussion.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
Rene-Damm: Could be worth a try. @shawnmcclelland What do you think?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorToolbar.cs
jechter: >The current one IMO is pretty visible 

Heh, I did not notice that was there, but I guess that's because I use the development editor build, which draws the UI update widget over the corner of the window, partially covering it. So maybe that's fine (I guess ideally, I'd want a dot in the red window close button as per macOS conventions, but i guess we cannot do that without editor-level support).

But maybe we should reconsider just always saving (and removing the explicit save UI) to match unity conventions. The fact that all the examples I mentioned "is all Unity native data." should not make any difference to the user here.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorToolbar.cs
Rene-Damm: Good point. Changed it so that

* If you have no control schemes in the asset, it shows "No Control Schemes" and the popup also doesn't list any (previously it presented "No Control Scheme" like a control scheme).
* If you do have control schemes, then what used to be "No Control Scheme" is now "All Control Schemes" (including "Show " thing is a bit long).

Works?

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
Rene-Damm: In the end, I guess part of the answer is "mine".

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorToolbar.cs
jechter: Works, but: 
>"All Control Schemes" (including "Show " thing is a bit long).

I included the word "Show" there on purpose. Because when I first tried this menu, it was not clear to me that it acted as a "filter" controlling which parts of the data I see (as opposed to, eg, performing some reformatting of the data to make it suitable for different things, which could be an action actually changing the data). I was thinking about making a change to communicate that better. I considered suggesting a "Filter: " prefix label for this menu, but I figured that that might be unnecessary UI clutter, so I thought that adding "Show" to the label might be a lightweight but sufficient clue to the user.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
jechter: Yeah, I thought so. The problem with that is, it makes it hard for contributors to follow these guidelines if they exist only in your head. And from what I've seen in PRs, you are rather eager in having everyone follow these guidelines.

As a non-you committer, I'd prefer it if we could either:
-adopt a written down guideline.
or
-agree that this is not important enough to discuss in PRs.

jechter (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
douglas-piconi: Sounds good to me, I would LOVE to stop supporting 2018.3!

PR #468, opened 2019-03-14
"Press and Release" interactions will now invoke the `performed` callback twice

jechter:
"Press and Release" interactions will now invoke the `performed` callback on both press and release (instead of invoking `performed` and `cancel`, which was inconsistent with other behaviors).

This is based on https://fogbugz.unity3d.com/f/cases/1134081/, and the related discussion on Slack were we decided that this would be more intuitive.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/Interactions/PressInteraction.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/Interactions/PressInteraction.cs
@@ -93,12 +93,13 @@ namespace UnityEngine.Experimental.Input.Interactions
                             if (isActuated)
                                 context.PerformedAndStayPerformed();
                             else
-                                context.Cancelled();
+                            {
+                                context.PerformedAndGoBackToWaiting();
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/Interactions/PressInteraction.cs
```diff
@@ -93,12 +93,13 @@ public void Process(ref InputInteractionContext context)
                             if (isActuated)
                                 context.PerformedAndStayPerformed();
                             else
-                                context.Cancelled();
+                            {
+                                context.PerformedAndGoBackToWaiting();
+                                context.Waiting();
```
Rene-Damm: separate `Waiting()` call should not be necessary. `PerformedAndGoBackToWaiting()` triggers a "current -> Peformed -> Waiting" state change cycle.

Rene-Damm (changes requested):
Minor thing. Otherwise LGTM.

PR #443, opened 2019-03-06
Fixed Xbox Controller on macOS reporting negative values for the stic…

jechter:
…ks when represented as dpad buttons.

PR #471, opened 2019-03-15
Fixed Xbox Controller on macOS reporting negative values for the stic…

jechter:
…ks when represented as dpad buttons.

Same as https://github.com/Unity-Technologies/InputSystem/pull/443, but changed to work with @douglas-piconi 's latest changes, and added testing.


douglas-piconi (approved):

douglas-piconi:
Ok , at first I was trying to get my head around this change.  I think it makes sense now.   Does everyone agree up/down/left/right are "directional" on or off values?   and as long as the DPad can go -1 when pressing left and 1 when pressing right.

jechter:
>Does everyone agree up/down/left/right are "directional" on or off values?

Well, they are not on/off with this change, but [0.0,1.0] - rather than inconsistently [-1.0,1.0] for up/down and [0.0,1.0] for left right.

Rene-Damm (approved):

PR #475, opened 2019-03-18
CHANGE: Drop support for Unity 2018.3.

douglas-piconi:
Drop support for Unity 2018.3

NOTE: I pulled trunk out of Yamato config so that we can stabilize tests on 2019.1


.yamato/upm-ci.yml
```diff
@@ -1,7 +1,5 @@
 editors:
-  - version: 2018.3
   - version: 2019.1
-  - version: trunk
```
douglas-piconi, line 4: NOTE: I pulled trunk out of Yamato config so that we can stabilize tests on 2019.1

Rene-Damm (commented):
Could we also purge all outdated #ifs as part of this change?

Think we can even kill off the entire AdvancedDropdown code. AFAIK that stuff has landed in 2019.1 and we can just use the built-in version. This way we'll get any fixes, too.

douglas-piconi:
@Rene-Damm ok, pulled out the ifdef's     I didn't want to touch AdvancedDropdown in this PR though.

Rene-Damm:
@douglas-piconi Ok, I'll rip out AdvancedDropdown in a follow-up PR. Should be as simple as nuking the folder and having Rider update some usings.

Rene-Damm (approved):

PR #476, opened 2019-03-19
Fix incompatible postprocessors

jechter:
This is to fix https://fogbugz.unity3d.com/f/cases/1135152/

Currently, it is possible to assign incompatible processors to bindings/actions, which will then throw an exception when trying to read the value. In the example given in that fogbugz case, a processor is assigned to a composite axis making up a dpad. It is then correctly applied on the axis, but the input system will try to apply it again on the composite dpad, failing, because it does not match the type. In this case, it should be applied the first time, but not the second type.

So, I decided that we should always ignore incompatible processors. Also, I changed the Action editor UI to
1. Make it impossible to assign incompatible processors to actions (before it would only filter on bindings)
2. If there are incompatible processors (which can happen when changing action type after adding a processor), label those as "(Ignored)".
3. Also, I added a header to the parameters UI for processors and interactions to make the association of the parameters with the list view item more clear.
4. Fixed reloading the Action Asset editing UI when editing parameters, thus clearing your selection and no longer showing the parameters (https://fogbugz.unity3d.com/f/cases/1138244/)

### DIFF
--- a/Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
@@ -2584,6 +2584,33 @@ partial class CoreTests
         Assert.That(receivedVector.Value.y, Is.EqualTo(0.5678).Within(0.00001));
     }
 
+    [Test]
+    [Category("Actions")]
+    public void Actions_IncompatibleProcessorIsIgnored()
+    {
+        var gamepad = InputSystem.AddDevice<Gamepad>();
+
+        InputSystem.RegisterControlProcessor<ConstantVector2TestProcessor>();
+        var action = new InputAction(processors: "ConstantVector2Test");
+        action.AddBinding("<Gamepad>/leftStick/x");
+        action.Enable();
+
+        float? receivedFloat = null;
+        action.performed +=
+            ctx =>
+            {
+                Assert.That(receivedFloat, Is.Null);
+                // ConstantVector2TestProcessor processes Vector2s. It would throw an exception when 
+                // trying to use it reading a float if not ignored.
+                receivedFloat = ctx.ReadValue<float>();
+            };
+
+        Set(gamepad.leftStick, Vector2.one);
+
+        Assert.That(receivedFloat, Is.Not.Null);
+        Assert.That(receivedFloat.Value, Is.EqualTo(1).Within(0.00001));
+    }
+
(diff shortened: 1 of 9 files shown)


douglas-piconi:
@jechter  I'm getting a compile error on this branch.

`Editor\AssetEditor\InputActionEditorWindow.cs(510,27): error CS0122: 'InputActionTreeView.UpdateSerializedObjectDirtyCount()' is inaccessible due to its protection level`

douglas-piconi (approved):
Other than compile error, looks good to me.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -2584,6 +2584,33 @@ public void Actions_CanAddProcessorsToActions()
         Assert.That(receivedVector.Value.y, Is.EqualTo(0.5678).Within(0.00001));
     }
 
+    [Test]
+    [Category("Actions")]
+    public void Actions_IncompatibleProcessorIsIgnored()
```
Rene-Damm, line 2589: Just wondering, is this actually desirable behavior? If there's a processor that doesn't apply, isn't that a problem in the setup that should be diagnosed?

jechter:
@douglas-piconi good catch - missed committing a file again. Fixed.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
jechter: I've been thinking about the same, and considered just making it generate more meaningful errors instead. But I chose this solution instead. Some reasons:

-In the case reported in the error where we have a dpad composite binding, the processor seems to be applied on both the composite elements (where it makes sense), and the result (where it does not). Ignoring it where it does not does the right thing here, which throwing errors would not solve.

-I think  it may be convenient to not invalidate everything when changing bindings. This way, you can quickly switch types, try out different things, without having to change all the processors. Instead I chose to mark incompatible processors as "(ignored)" in the UI.

Rene-Damm: >the processor seems to be applied on both the composite elements (where it makes sense)

Oh, that one would be a bug. The composite and its parts are each individual bindings and the processors applied to one should not affect the other.

>I think it may be convenient to not invalidate everything when changing bindings. This way, you can quickly switch types, try out different things, without having to change all the processors.

Good point I think.

PR #478, opened 2019-03-19
Repopulate WMR scene action mappings

jackpr-unity:
The actions for this scene are unpopulated.  I am repopulating them.


jackpr-unity:
I guess this needs to be merged to develop first?

jackpr-unity:
I'm still going to try.

PR #488, opened 2019-03-21
Add support for Bluetooth Xbox One controllers on macOS.

jechter:
https://fogbugz.unity3d.com/f/cases/1091502/

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Support gamepad vibration on Switch.
+- Support for Bluetooth Xbox One controllers on macOS.
 
 #### Actions
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [0.3-preview] - TBD
 
+### Added
+
+- Support for Bluetooth Xbox One controllers on macOS.
+
 ### Changed
 
 - `Joystick.axes` and `Joystick.buttons` have been removed.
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Support gamepad vibration on Switch.
+- Support for Bluetooth Xbox One controllers on macOS.
```
Rene-Damm: The usual conflict around CHANGELOG.md...

Rene-Damm (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Needs to go up there in the 0.3-preview section.

jechter: Ah, yes. I just merged in github and did not even notice this.

Rene-Damm (approved):

douglas-piconi (approved):

Rene-Damm:
@jechter Does this also work with 360 controllers?

jechter:
@Rene-Damm This only affects wireless controllers on mac. AFAIK the only xbox controllers which can be wirelessly connected to macs are the latest generation of Xbox One controllers (which support Bluetooth, unlike the earlier models). Xbox (360 or other) controllers connected through a wire should still use the old code path.

FRickReich:
as a sidenote, I would like to know how to actually use the xbox controller on Mac in the new system, because the default mapping is really strange (north button is horizontal axis?!)

jechter:
@FRickReich can you tell me:
-which xbox controller (360 or xbox one), 
-how is it connected (usb or bluetooth)
-which version of the Input System package?

FRickReich:
I am sorry, could have given these infos in the first run...
- Xbox One Controller from 2018
- bluetooth, but I tried both
- 0.2.6

jechter:
Ok - so in that case, the problem you are describing is exactly the thing that this PR has fixed. Update to 0.2.8, and it should work as expected :)

FRickReich:
sounds great! 0.2.8 is not in the package manager, so I guess I will have to manually download on GitHub?

Edit: ...excuse me. THE SECOND I wrote this, unity told me I can update the input system.

PR #490, opened 2019-03-21
Remove runInBackground

jechter:
In discussion following up https://fogbugz.unity3d.com/f/cases/1136414/ , we decided that having a global "run in background" does not make much sense (and will never really work people would expect anyways). So I removed it. Which also fixes the bug about it not working :)

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -238,7 +238,12 @@ namespace UnityEngine.Experimental.Input
             }
         }
 
-        private bool gameIsPlayingAndHasFocus => EditorApplication.isPlaying && !EditorApplication.isPaused && (m_HasFocus || InputEditorUserSettings.lockInputToGameView);
+        private bool gameIsPlayingAndHasFocus =>
+#if UNITY_EDITOR
+            EditorApplication.isPlaying && !EditorApplication.isPaused && (m_HasFocus || InputEditorUserSettings.lockInputToGameView);
+#else
+            true;
+#endif
 
         ////TODO: when registering a layout that exists as a layout of a different type (type vs string vs constructor),
         ////      remove the existing registration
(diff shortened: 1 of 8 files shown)


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2200,26 +2194,11 @@ private void OnFocusChanged(bool focus)
...
-                case InputUpdateType.Editor:
-                    // If we're in play mode and the player has focus (or ignores focus), don't run editor updates.
-                    if (Application.isPlaying && !EditorApplication.isPaused && (m_HasFocus || m_Settings.runInBackground))
-                        return false;
-                    break;
+            // If we're in play mode and the player has focus (or ignores focus), don't run editor updates.
+            if (updateType == InputUpdateType.Editor && Application.isPlaying && !EditorApplication.isPaused)
+                return false;
 #endif
-            }
```
Rene-Damm, line 2222: This seems to be breaking focus behavior. For the player it should be fine but for the editor it would mean that the game view is eating input targeted at the editor.

jechter: I don't understand. The behavior is the same as before _if runInBackground was enabled_? Should we maybe always run editor Updates?

Rene-Damm: Think I misread what m_HasFocus *actually* means. However, that makes me wonder whether we had a regression here before. The code in native used to take *player* focus into account. The current code seems to no longer do that. The code says

```
            // If we're in play mode and the player has focus (or ignores focus), don't run editor updates.
```

which I think is still what it should do in order to be able to have input in EditorWindow.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -238,6 +238,8 @@ public float pollingFrequency
             }
         }
 
+        private bool gameIsPlayingAndHasFocus => EditorApplication.isPlaying && !EditorApplication.isPaused && (m_HasFocus || InputEditorUserSettings.lockInputToGameView);
```
Rene-Damm: This needs to be `#if UNITY_EDITOR` to not break the player.

Rene-Damm (changes requested):

Rene-Damm (approved):

PR #496, opened 2019-03-25
Generate input action map in wrapper

jechter:
https://fogbugz.unity3d.com/f/cases/1136337/ started to get me thinking why we need to manually assign a reference to the InputActions asset to scripts using the generated wrappers in the first place. After all, we'd expect that only one asset matches the layout of the generated wrapper, so there should be no need to manually assign that one asset.

Following discussion with @Rene-Damm on how to best solve that, we figured we should try putting the information from the InputActions asset _into_ the generated code - that way a reference to the asset is no longer needed at all.

This is my attempt at doing so.

Usage changes from 
```
// reference assigned in inspector
public SimpleControls controls; 
```

to 

```
private SimpleControls controls = new SimpleControls();
```

This instance can be shared with other users, or you can create new separate instances using `new SimpleControls()`.

To do:
-I guess I should try to  add a test to make sure that the `InputActionMap` produced by the generated code is equal to the one in the asset.

So I wonder:
-Does this approach make sense?
-With this, is there any use in keeping `InputActionReference` at all?
-Anything else we'd need to add to support this?


Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
```diff
@@ -130,72 +157,25 @@ public static string GenerateWrapperCode(IEnumerable<InputActionMap> maps, IEnum
...
-            writer.WriteLine("private void Uninitialize()");
-            writer.BeginBlock();
-            foreach (var map in maps)
+            writer.WriteLine("public void Enable()");
+			writer.BeginBlock();
+            foreach (var set in maps)
             {
-                var mapName = CSharpCodeHelpers.MakeIdentifier(map.name);
-
-                if (options.generateInterfaces)
```
Rene-Damm, line 143: I think this path we would generally want to preserve. But we probably can get rid of `SetCallbacks()` and simply add a constructor that takes an instance of the interface as an instance.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
```diff
@@ -130,72 +157,25 @@ public static string GenerateWrapperCode(IEnumerable<InputActionMap> maps, IEnum
...
-                }
-
-                writer.WriteLine($"m_{mapName} = null;");
-
-                foreach (var action in map.actions)
-                {
-                    var actionName = CSharpCodeHelpers.MakeIdentifier(action.name);
-                    writer.WriteLine($"m_{mapName}_{actionName} = null;");
-
-                    if (options.generateEvents)
```
Rene-Damm, line 159: Loosing this path I'm good with. Think it was never a good solution and `PlayerInput` has a more elegant version of it.

Assets/Demo/SimpleDemo/SimpleControls.cs
```diff
@@ -5,52 +5,35 @@
 using UnityEngine.Experimental.Input;
 
 
-[Serializable]
-public class SimpleControls : InputActionAssetReference
+public class SimpleControls
```
Rene-Damm: I think this needs to implement `IInputActionCollection`. Otherwise we will lose some functionality and information from the asset (e.g. access to control schemes).

Rene-Damm (commented):
Overall, I like the approach and think it could make for a better solution than the previous one.

For the tests, now that we have moved to the new Mono, I wonder whether `CSharpCodeProvider` now works. If so, I think that would be a far superior solution to those text match tests I have written.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
jechter: But this path is _uninitialization_ of interfaces (`SetCallbacks(null)`). The old code would call this when changing the asset attached to the wrapper. When would we need or do uninitialization today?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
jechter: Again, this is just the uninitalization part. We still generate the events if requested on initialization. I could remove that, though.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
jechter: Also, I wonder: is there a reason to ever not generate this? Ie, should we remove `options.generateInterfaces` and always generate them? After all, just because we generate the code doesn't mean you have to use it.

jechter:
> For the tests, now that we have moved to the new Mono, I wonder whether `CSharpCodeProvider` now works. If so, I think that would be a far superior solution to those text match tests I have written.

It appears that, no, we can't. That stuff is in a System.CodeDom.dll, which we don't package with the .NET system libraries we distribute with Unity.

jechter:
> It appears that, no, we can't. That stuff is in a System.CodeDom.dll, which we don't package with the .NET system libraries we distribute with Unity.

Ah, seems that when using .NET 4.0 API Compat level that works.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
Rene-Damm: >But this path is uninitialization of interfaces (SetCallbacks(null)). The old code would call this when changing the asset attached to the wrapper. When would we need or do uninitialization today?

Ah right, I read the diff incorrectly. Didn't realize the rest of the codepaths were still there.

Resetting the callbacks may still be useful in some situations but certainly doesn't seem to be a reason to do that automatically.

>is there a reason to ever not generate this? Ie, should we remove options.generateInterfaces and always generate them? After all, just because we generate the code doesn't mean you have to use it.

True. I think it's fine to do that. The current switches came more out of an experimentation with different approaches. generateInterfaces proved to be one of the more successful ones.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
Rene-Damm: Yup, nvm.

Assets/Demo/SimpleDemo/SimpleControls.cs
```diff
@@ -26,6 +29,47 @@ public SimpleControls()
...
+        }
+    }
+    private InputControlScheme[] s_controlSchemes;
+    public ReadOnlyArray<InputControlScheme> controlSchemes
+    {
+        get
+        {
+            if (s_controlSchemes == null)
+            {
+                s_controlSchemes = new InputControlScheme[] {
```
Rene-Damm: No need to allocate here. ReadOnlyArray is fine with a null array.

douglas-piconi (approved):
Nice all checks pass!  First code review with real CI!

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
```diff
@@ -265,7 +208,7 @@ public static string GenerateWrapperCode(IEnumerable<InputActionMap> maps, IEnum
                     $"public static implicit operator InputActionMap({mapTypeName} set) {{ return set.Get(); }}");
 
                 // SetCallbacks method.
-                if (options.generateInterfaces)
+                //if (options.generateInterfaces)
```
Rene-Damm: Please remove commented out code.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
```diff
@@ -335,20 +277,8 @@ public static string GenerateWrapperCode(IEnumerable<InputActionMap> maps, IEnum
...
-                writer.BeginBlock();
-                writer.EndBlock();
-            }
-
-            // End class.
-            writer.EndBlock();
-
             // Generate interfaces.
-            if (options.generateInterfaces)
+            //if (options.generateInterfaces)
```
Rene-Damm: Please remove.

Assets/Demo/DemoControls.cs
```diff
@@ -34,50 +473,45 @@ private void Initialize()
...
-            menu.SetCallbacks(null);
-        }
-        m_menu = null;
-        m_menu_navigate = null;
-        m_menu_click = null;
-        m_menu_steamExitMenu = null;
-        m_menu_submit = null;
-        m_menu_point = null;
-        m_Initialized = false;
+        UnityEngine.Object.Destroy(asset);
```
Rene-Damm, line 478: Do we even need this?

Assets/Demo/DemoControls.cs
```diff
@@ -235,6 +668,7 @@ public InputControlScheme KeyboardMouseScheme
     public InputControlScheme GamepadScheme
     {
         get
+
```
Rene-Damm: Generator seems to emit an extra newline here. Think we already have/had some whitespace issue but overall, would be great if the generator spits out code that results in zero formatting changes when running the Unity formatting tools.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
```diff
@@ -208,21 +166,14 @@ public static string GenerateWrapperCode(IEnumerable<InputActionMap> maps, IEnum
 
                 // Caching field for action map.
                 writer.WriteLine($"private InputActionMap m_{mapName};");
-                if (options.generateInterfaces)
-                    writer.WriteLine(string.Format("private I{0} m_{0}CallbackInterface;", mapTypeName));
+                //if (options.generateInterfaces)
```
Rene-Damm: Please remove.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporterEditor.cs
```diff
@@ -15,6 +15,25 @@ namespace UnityEngine.Experimental.Input.Editor
     [CustomEditor(typeof(InputActionImporter))]
     internal class InputActionImporterEditor : ScriptedImporterEditor
     {
+        // Show a PropertyField with a greyed-out default text if the field is empty and not being edited.
+        // This is meant to communicate the fact that filling these properties is optional and that Unity will
+        // use reasonable defaults if left empty.
+        private void PropertyFieldWithDefaultText(SerializedProperty prop, GUIContent label, string defaultText)
```
Rene-Damm: Can we move this to SerializedPropertyHelpers or somewhere like that? Having utility methods hidden inside classes like this makes them hard to find.

Assets/Demo/DemoControls.cs
jechter: Since we generate the asset when creating the wrapper, I think we should destroy it. Otherwise it would eventually be cleaned up by `GarbageCollectSharedAssets` when loading a new scene, or when someone calls `Resources.UnloadUnusedAssets`, but I thought it would make sense to free it right away if we know we won't use it any more.

Assets/Tests/DemoGameTestFixture.cs
```diff
@@ -253,18 +253,6 @@ public void Click(string button, int playerIndex = 0)
...
-    {
-        // Look up action.
-        var controls = game.players[playerIndex].controls;
-        var actionInstance = controls.asset.FindAction(action);
-        if (actionInstance == null)
-            throw new ArgumentException("action");
-
-        // And trigger it.
-        Trigger(actionInstance);
-    }
```
Rene-Damm, line 266: What's the reason for killing this?

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -1612,65 +1613,36 @@ public void Editor_CanGenerateCodeWrapperForInputAsset()
...
-        Assert.That(code, Contains.Substring("public interface IMap1Actions"));
-        Assert.That(code, Contains.Substring("public interface IMap2Actions"));
-        Assert.That(code, Contains.Substring("private IMap1Actions m_Map1ActionsCallbackInterface;"));
-        Assert.That(code, Contains.Substring("private IMap2Actions m_Map2ActionsCallbackInterface;"));
-        Assert.That(code, Contains.Substring("public void SetCallbacks(IMap1Actions instance)"));
-        Assert.That(code, Contains.Substring("public void SetCallbacks(IMap2Actions instance)"));
-        Assert.That(code, Contains.Substring("void OnAction1(InputAction.CallbackContext context)"));
-        Assert.That(code, Contains.Substring("void OnAction2(InputAction.CallbackContext context)"));
-        Assert.That(code, Contains.Substring("void OnAction3(InputAction.CallbackContext context)"));
+#if NET_4_6
```
Rene-Damm: Should probably go for the entire test.

Rene-Damm (approved):

Assets/Tests/DemoGameTestFixture.cs
jechter: `controls.asset` is no longer public.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -154,7 +150,7 @@ public override void OnImportAsset(AssetImportContext ctx)
                 if (!Directory.Exists(dir))
                     Directory.CreateDirectory(dir);
 
-                if (InputActionCodeGenerator.GenerateWrapperCode(wrapperFilePath, maps, asset.controlSchemes, options))
+                if (InputActionCodeGenerator.GenerateWrapperCode(wrapperFilePath, asset, options))
```
Rene-Damm, line 153: Just realized that we forgot to bump the importer version. I'll bump it.

PR #499, opened 2019-03-26
FIX: When we went from field to property, we lost bit reset.

douglas-piconi:
In the process of changing to use the native InputEvent struct, the act of resetting the ID in the test runners InputTestRuntime lost its ability to completely reset the value , the handled bit.  It went from a field to a property meaning only the setter for event ID was getting processed.

on another note, we do "overflow" while doing this "handled" bit, but the way C# handles the overflow, it still works.  Wouldn't it be cleaner and less error prone, if we just had handled as a bool on the struct instead of the "bit" on the id field?

### DIFF
--- a/Packages/com.unity.inputsystem/Tests/InputSystem/InputTestRuntime.cs
+++ b/Packages/com.unity.inputsystem/Tests/InputSystem/InputTestRuntime.cs
@@ -94,7 +94,8 @@ namespace UnityEngine.Experimental.Input
 
             lock (m_Lock)
             {
-                eventPtr->eventId = m_NextEventId;
+                eventPtr->eventId = m_NextEventId; 
+                eventPtr->handled = false;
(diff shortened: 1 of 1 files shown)


jechter (approved):

jechter:
As for
>Wouldn't it be cleaner and less error prone, if we just had handled as a bool on the struct instead of the "bit" on the id field?

Maybe. I assume that @Rene-Damm would not like it, because he likes all the data tightly packed :) I don't care as much about losing some bytes, but handling that change would be scary, as then InputEvent would have an odd size, and we currently handle all the FieldOffsets in all the state types as "[Size of InputEvent] + x", meaning it would all go unaligned. So I think even if we agreed that that was better, it is probably not worth changing it.

Rene-Damm (approved):

Issue #500, opened 2019-03-27
PlayerInput with UnityEvents broken


zadirion:
PlayerInput line 688 tries to find action by id
`var action = m_Actions.FindAction(id);`
id is a guid without being enclosed in curly braces
FindAction fails because of this. It calls TryGetActionIndex inside, which figures out whether the passed parameter is an action name or an id, distinguishing them by whether they start with a curly brace. See InputActionMap.cs line 213:
`var isReferenceById = nameOrId[0] == '{';`

Thus FindAction thinks an action name has been passed, not an id, and thus fails the find, and initialization of PlayerInput with UnityEvents breaks because of this

The PlayerInputEditor also suffers from this issue in PlayerInputEditor.cs line 395:
`var action = asset.FindAction(guid);`
resulting in the actions not being saved

Also PlayerInputEditor line 407 
`newActionEvents.Add(new PlayerInput.ActionEvent(action.id));`
Adds actions that were already added from the asset (on line 398), a second time, resulting in an ever growing list of actions in the inspector, every time the editor re-executes/updates
Code should check whether it's already been added, before adding it again, to avoid duplicates.

Rene-Damm:
Should be fixed in https://github.com/Unity-Technologies/InputSystem/pull/509.

PR #501, opened 2019-03-27
Fixed PlayerInput not initializing UnityEvents, fixed PlayerInputEditor not saving UnityEvents

zadirion:
Attempting to fix issue:
https://github.com/Unity-Technologies/InputSystem/issues/500

Tested locally, fix seems to work.


zadirion:
@Rene-Damm I think you're the owner of this system. I wasn't sure if unity action guids are supposed to be stored with or without {} inside the string, I assumed without, and just added the braces at call sites. Not sure what convention you had in mind. Adding them at call site is kind of ugly, so perhaps this should be abstractized somehow

zadirion:
on an unrelated note, I noticed the system is using the old SendMessage API, I noticed there's a new messaging api out, perhaps the implementation should use that, to avoid a premature obsolescence of this code?

Rene-Damm:
Hey @zadirion,

Thanks for the pull request!

This was indeed buggy. Looking at it, I think the whole `{...}` idea wasn't the best one and really quite unnecessary. I've created a PR that removes the entire thing (but keeps it working in existing data): https://github.com/Unity-Technologies/InputSystem/pull/509. This should hopefully fix the problem in #500.

>I noticed the system is using the old SendMessage API, I noticed there's a new messaging api out, perhaps the implementation should use that, to avoid a premature obsolescence of this code?

Which new messaging API do you have in mind?

zadirion:
@Rene-Damm 
I'm thinking of this, although I haven't personally worked with it so far
https://docs.unity3d.com/Manual/MessagingSystem.html

Rene-Damm:
>I'm thinking of this, although I haven't personally worked with it so far
https://docs.unity3d.com/Manual/MessagingSystem.html

Ah right. We could add support for that one as well but it would require code generation so not sure. I'm thinking that for simple prototyping, the SendMessage/Broadcast message options are good as they require zero setup. And for getting something more robust, I think the UnityEvents based notifications are probably an easier and more robust alternative to ExecuteEvents.

zadirion:
Right. Indeed, UnityEvents also works. I have personal beef with SendMessage, as it has the nasty habit of calling into unintended methods due to name coincidence. While ok for prototyping, I don't think it should remain there, or at least not enabled by default.

Issue #494, opened 2019-03-24
If PS4 DualShock is plugged into PC, read input action path assignment is always angularVelocity


periodyctom:
If I have a PS4 DualShock plugged in and use the input reading assignment feature, it always assigns to angularVelocity within a second and overrides any other input that may occur, including keyboard presses.

Win10 64, PS4 DualShock, 0.2.6-preview

Rene-Damm:
Should be fixed as part of #508. The DualShock stuff shouldn't have been in that PR but oh well, there it is :)

There were a couple problems here.

1) The DualShock sensor controls weren't been marked as noisy. Thus the rebinding stuff thought them good data and picked them up.
2) The DualShock sensors used to be defined all the way up at `DualShockGamepad`. Theoretically that makes sense but in practice we're not at this point supporting the sensors other than on the PS4 itself. What you got on the PC was all random garbage. I moved the controls to being confined to `DualShockGamepadPS4` for now until we have proper support across platforms.
3) Finally, I realized that being noisy didn't propagate down the hierarchy. So marking a Vector3 control noisy didn't make X, Y, and Z noisy, too. That, too, is changed now.

PR #508, opened 2019-04-02
CHANGE: Revamp control picker UX for clarity.

Rene-Damm:
- Control picker UI has been cycled through several iterations of changes
image
[video](https://youtu.be/gUECGxRKMWQ)
   * Interactive pick button has moved inside control picker and is now a “Listen” toggle that functions like a special kind of search mode
   * Controls have nice names
   * Controls have icons
   * Child controls are indented instead of using the more programmery “/” notation
   * Keyboard has extra level to decide between binding by location and binding by character
   * Added separators between usages, controls, and devices
   * Cleaned up hierarchy of devices more
- Can now bind by display name of a control: `<Keyboard>/#(ö)`
  * Doesn't yet react to when display name changes (i.e. no re-resolving of action bindings when keyboard layout changes)
- Fixed bug where spaces in control scheme names lead to all actions and bindings disappearing in the tree view.
- Removed `acceleration`, `orientation`, and `angularVelocity` from the `DualShockGamepad` base class as we do not yet support these controls other than on PS4.
- Marked `acceleration`, `orientation`, and `angularVelocity` on PS4 controller as noisy (#494).
- Make `noisy` a recursive property; i.e. if the parent is noisy, all its children are marked as noisy as well (devices are exempt from the rule, of course).


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -2037,6 +2082,34 @@ public InternedString GetRootLayoutName(InternedString layoutName)
                 return layoutName;
             }
 
+            public InternedString FindLayoutThatIntroducesControl(InputControl control, Cache cache)
+            {
+                // Find the topmost child control on the device. A device layout can only
+                // add children that sit directly underneath it (e.g. "leftStick"). Children of children
+                // are indirectly added by other layouts (e.g. "leftStick/x" which is added by "Stick").
+                // To determine which device contributes the control has a whole, we have to be looking
```
jechter: >which device contributes the control has a whole

_as_ a whole?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -2037,6 +2082,34 @@ public InternedString GetRootLayoutName(InternedString layoutName)
                 return layoutName;
             }
 
+            public InternedString FindLayoutThatIntroducesControl(InputControl control, Cache cache)
```
jechter, line 1628: Should we have a unit test for this method?

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -779,4 +777,72 @@ public void OnIMECompositionChanged(IMECompositionString compositionString)
...
+                case Key.RightAlt:
+                case Key.LeftShift:
+                case Key.RightShift:
+                case Key.LeftMeta:
+                case Key.RightMeta:
+                case Key.LeftCtrl:
+                case Key.RightCtrl:
+                case Key.ContextMenu:
+                case Key.Escape:
+                case Key.Delete:
```
jechter: Why is delete not a "text input key"? are the arrow keys "text input keys"?

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -779,4 +777,72 @@ public void OnIMECompositionChanged(IMECompositionString compositionString)
...
+            switch (key)
+            {
+                case Key.LeftAlt:
+                case Key.RightAlt:
+                case Key.LeftShift:
+                case Key.RightShift:
+                case Key.LeftMeta:
+                case Key.RightMeta:
+                case Key.LeftCtrl:
+                case Key.RightCtrl:
```
jechter, line 794: Is Caps Lock a modifier key? Is "Fn" (guess we don't actually support Fn atm. Is there a reason for this)?

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
```diff
@@ -248,4 +236,217 @@ protected override void FinishSetup(InputDeviceBuilder builder)
...
     }
+
+    [InputControlLayout(displayName = "Magnetic Field")]
+    public class MagneticFieldSensor : Sensor
+    {
+        /// <summary>
+        /// TODO
+        /// </summary>
+        /// <remarks>
+        /// Values are in micro-Tesla (uT) and measure the ambient magnetic field in the X, Y and Z axis.
```
jechter, line 247: I'm surprised to see specific physical units for all these added sensors. Can we actually assume to get specific physical measurements from all/most implementations (instead of axes of arbitrary scale)?

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPickerDropdown.cs
```diff
@@ -140,86 +355,191 @@ private void AddDeviceTreeItem(InputControlLayout layout, AdvancedDropdownItem p
...
+                .WithMagnitudeHavingToBeGreaterThan(0.15f)
+                ////REVIEW: should we exclude only the system's active pointing device?
+                // With the mouse operating the UI, its cursor control is too fickle a thing to
+                // bind to. Ignore mouse position and delta and clicks.
+                // NOTE: We go for all types of pointers here, not just mice.
+                .WithControlsExcluding("<Pointer>/position")
+                .WithControlsExcluding("<Pointer>/delta")
+                .WithControlsExcluding("<Pointer>/button")
+                .WithControlsExcluding("<Mouse>/leftButton")
+                .WithControlsExcluding("<Mouse>/scroll")
```
jechter, line 378: On my Mac, this will exclude the generic Mouse device from showing up in listening queries, but mouse interactions will still show up from the more specific HID device implementing my Touchpad. I think this is wrong. I don't think you would ever want to bind to the specific HID implementation of the MacBook touchpad. So we should hide that as well (somehow?).

jechter (changes requested):
I tried this locally. I think it is a great UX improvement, good job! I did run into some minor issues with the listen mode, however, which should be fixed first:

-When I selected listen mode, and then turn off listen mode, and then select listen mode again, it won't work anymore (I have to close and reopen the popup to make it work again).
-It seems that when in listen mode, the window will randomly select items being listened to.
-See my comment about the HID mouse device not being filtered (I think this was broken before this PR already, so we might leave this to fix for later, but then we should file a bug so we don't forget).

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
Rene-Damm: It's an internal API. We have a test that covers the public API that uses this (only one so far).

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
Rene-Damm: Not in the context of this function. I think the problem may be the name of the function. Maybe `GeneratesPrintableTextCharacter` is better?

What the function tries to establish is whether a key is considered to have a printable character representation or is considered a "function" key. Bit hard to establish unambiguously. Enter/return, for example, clearly generates text character input -- just not the kind being looked for here.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
Rene-Damm: Caps Lock is not usually considered a modifier key as far as I'm aware.

AFAIK Fn is not reported by OSs. My assumption is that in many cases the key is actually handled by the keyboard's HW itself. In any way, responding to Fn doesn't seem like a useful thing for Unity apps.

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
Rene-Damm: Don't know TBH. So far, these are all taken directly from Android which is the only platform where we currently support all these sensors. When extending sensor support, this needs to be looked at.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPickerDropdown.cs
Rene-Damm: Yes, just not here IMO. Getting rid of undesirable HIDs should happen at a system-wide level. ATM we don't yet have good mechanisms for reigning in the fishnet-like behavior of the HID layout builder.

Rene-Damm:
>-When I selected listen mode, and then turn off listen mode, and then select listen mode again, it won't work anymore (I have to close and reopen the popup to make it work again).

Can't repro that so far. My guess is this is actually a problem with UI refreshes that I've seen happen sometimes and haven't been able to track down yet. Basically, the rebind stuff all works just fine and gathers the right data and triggers `Repaint()` on the UI just fine but the UI never updates -- except when you move the mouse around. Then suddenly all the stuff pops up. I'll have another look at that and see if I can track it down.

Rene-Damm:
>-It seems that when in listen mode, the window will randomly select items being listened to.

Hmm, I'm not seeing that one either. Let's find a repro and fix that separately as it sounds like something that pertains directly to `RebindOperation`.

Rene-Damm:
>-See my comment about the HID mouse device not being filtered (I think this was broken before this PR already, so we might leave this to fix for later, but then we should file a bug so we don't forget).

Yup, let's find a good solution for this and land it separately.

For now, I think it'd be fine to be quite aggressive and just radically clamp down on the HID usage pages we will automatically create a device for. We could even go all the way and just stick to the same set that the old input system uses.

ATM the HID fallback will go and create a device from *any* HID that has at least one usable control. So yeah, that's where all the garbage is coming from. This kind of behavior is useful for developing but not very useful for normal operation.

Packages/com.unity.inputsystem/InputSystem/Devices/Sensor.cs
jechter: Well, guess since it's only comments in the code, doesn't really matter anyways. Was just curious.

Rene-Damm:
>-When I selected listen mode, and then turn off listen mode, and then select listen mode again, it won't work anymore (I have to close and reopen the popup to make it work again).

Was indeed broken. Digging in deeper I found that `RebindOperation` did not properly allow to be reused. Fixed in https://github.com/Unity-Technologies/InputSystem/pull/508/commits/fd7859aafa4fb44fc9309981ab2ccc1c74313ca7.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
Rene-Damm: Finally got it :) Fixed in https://github.com/Unity-Technologies/InputSystem/pull/508/commits/91a4115cd7d7d8add591985c4799d942e2e317be.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceBuilder.cs
```diff
@@ -857,44 +856,6 @@ private static void AddProcessors(InputControl control, ref InputControlLayout.C
...
-                        case InputControlLayout.ParameterType.Float:
-                            value = *(float*)parameter.value;
-                            break;
-                    }
-                }
-
-                field.SetValue(onObject, value);
-            }
-        }
-
```
Rene-Damm: Uhm, no clue what this is about. Looks like a messed up merge.

jechter (approved):

PR #509, opened 2019-04-02
FIX: PlayerInput not finding actions when using Unity Events (#500).

Rene-Damm:
Also fixes drag&dropping between items sometimes being off-by-one.

Finally, also attempts a fix at the recent test instabilities with the ActionTree tests.

### DIFF
--- a/Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
@@ -2007,10 +2007,55 @@ partial class CoreTests
 
         Assert.That(map.TryGetAction("action1"), Is.SameAs(action1));
         Assert.That(map.TryGetAction("action2"), Is.SameAs(action2));
+        Assert.That(map.TryGetAction("action3"), Is.Null);
 
         // Lookup is case-insensitive.
         Assert.That(map.TryGetAction("Action1"), Is.SameAs(action1));
         Assert.That(map.TryGetAction("Action2"), Is.SameAs(action2));
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_CanLookUpActionsInMapById()
+    {
+        var map = new InputActionMap();
+
+        var action1 = map.AddAction("action1");
+        var action2 = map.AddAction("action2");
+
+        Assert.That(map.TryGetAction(action1.id), Is.SameAs(action1));
+        Assert.That(map.TryGetAction(action2.id), Is.SameAs(action2));
+        Assert.That(map.TryGetAction(Guid.NewGuid()), Is.Null);
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_CanLookUpActionsInMapByStringId()
+    {
+        var map = new InputActionMap();
+
+        var action1 = map.AddAction("action1");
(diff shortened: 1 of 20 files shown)


douglas-piconi (approved):

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -2007,12 +2007,57 @@ public void Actions_CanLookUpActionInMap()
...
+        var action1 = map.AddAction("action1");
+        var action2 = map.AddAction("action2");
+
+        Assert.That(map.TryGetAction(action1.id.ToString()), Is.SameAs(action1));
+        Assert.That(map.TryGetAction(action2.id.ToString()), Is.SameAs(action2));
+        Assert.That(map.TryGetAction(Guid.NewGuid().ToString()), Is.Null);
+    }
+
+    // We used to require string GUIDs to be using a "{...}" format when looking up actions. We no
+    // longer do this but there's still data that may be using this format so make sure it works for now.
```
jechter, line 2046: Do we want to keep supporting this "forever"? Otherwise, we should somehow message that it is deprecated.

jechter (approved):

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: IMO the couple backwards compatible things we have already accumulated by now we could/should all drop when we go 1.0 final. I'm thinking that in this phase, we don't really need to worry about a deprecation path any more format than that.

For this particular behavior, the .inputactions editor "luckily" doesn't use action IDs yet but rather references actions by name. So there should be few instances of this data actually out in the wild. Not even sure I should have bothered preserving backwards compatibility here.

jechter: Agreed we need nothing formal to guide the few users. My main point is about marking it somehow so we don't forget about it.

jechter: Alternatively, can we just drop it _now_?

Rene-Damm: Bit hesitant to do that. Been doing pretty badly in the past at saying what impact a breaking change has.

I do think that .inputactions asset created in our editor would be unaffected and that it would only have a minor impact on PlayerInput. But could be there's stuff I'm not taking into account.

jechter: I don't see your point, though. OTOH, you say deprecation is not necessary because we will drop this soon anyways, OTOH, you say dropping it now is too scary :) When will it not be too scary then?

Rene-Damm: Hehe, yeah, it is a bit backwards-through-the-nose. My thinking is kind of... instead of removing backwards-compatibility every now and then, let's batch it all up to 1.0 and then remove it all in one fell swoop. Whether that makes more sense is a good question of course :)

PR #511, opened 2019-04-03
NEW: Add APIs to change or erase existing bindings on actions.

Rene-Damm:
This is in response to https://forum.unity.com/threads/interaction-adding-interaction-on-an-action-during-runtime.651187/.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
@@ -2,10 +2,6 @@ using System;
 using UnityEngine.Experimental.Input.Layouts;
 using UnityEngine.Experimental.Input.Utilities;
 
-////TODO: support for removing bindings
-
-////TODO: rename AddBinding to just AddBinding
-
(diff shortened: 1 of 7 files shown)


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
```diff
@@ -225,6 +225,49 @@ private static int AddBindingInternal(InputActionMap map, InputBinding binding)
             return bindingIndex;
         }
 
+        public static BindingSyntax ChangeBinding(this InputAction action, int index)
```
jechter, line 228: What's the use case for using indexes? Iterating over all bindings?

Rene-Damm: Stuff like

```
for (var i = 0; i < myAction.bindings.Count; ++i)
    if (myAction.bindings[i].SomeKindOfCheck)
        myAction.ChangeBinding(i)....;
```

jechter (approved):

PR #513, opened 2019-04-04
FIX: Floating-point parameters being sensitive to current culture.

Rene-Damm:
Other than the fix, this also includes a long-overdue refactor that pulls the `ParameterValue` code out of `InputControlLayout` (which is why the diff is large for such a simple fix). That was where the concept of "parameters" started but in the end, several other places that had nothing to do with layouts started reusing the code.

Other than pulling the code of out `InputControlLayout`, it also unifies it with `PrimitiveValue` (which was introduced to eventually merge with `ParameterValue`) and reuses .NET `TypeCode` instead of the previous custom `ParameterValueType`.

A side benefit of the refactor is that we need much fewer `switch (whatTypeDoWeHave) { ... }` statements scattered around the code as `PrimitiveValue` can easily be converted between various formats.


jechter (approved):

douglas-piconi (approved):

mrrhetorical:
I don't believe this has been fixed, or it has been reintroduced. On Unity 2020.1.0f1 and input system package version 1.2.0 the issue persists where decimals on systems with locales using commas to separate the decimal are being stored improperly. I've included a stack trace that shows the error when trying to load "0,5" from the asset as a decimal, which it shouldn't have been saved as in the first place. 

https://pastebin.com/xAb29iuQ

PR #514, opened 2019-04-04
CHANGE: Make initial state check on actions optional.

Rene-Damm:
This reverts the default behavior of actions back to *not* doing an initial state check when enabled. Instead, this behavior is now optional and can be toggled on separately.

The reason is that while the initial state check is useful in some cases, it can get in the way for others. An example for when it is problematic can be found [here on the forum](https://forum.unity.com/threads/input-system-update.508660/page-10#post-4275154).

The change should make the behavior available while hopefully providing a good default in not performing the check by default.


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -188,6 +188,22 @@ public ReadOnlyArray<InputControl> controls
...
         }
 
+        public bool initialStateCheck
+        {
+            get => (m_Flags & ActionFlags.InitialStateCheck) != 0;
+            set
+            {
+                if (enabled)
+                    throw new InvalidOperationException(
+                        $"Cannot change the 'initialStateCheck' flag of action '{this} while the action is enabled");
```
jechter, line 198: Why not?

jechter (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
Rene-Damm: We'd have to sync `InputActionState` somehow. Either with a separate codepath or by re-resolving. I'm thinking it's not worth the trouble. In general, the action APIs make the assumption that configuration APIs require actions to be disabled.

PR #516, opened 2019-04-05
FIX: Saving .inputactions causing selection loss on parameters&interactions.

Rene-Damm:


douglas-piconi (approved):

jechter:
I take it that changing the file on disk still properly causes a reload?

Rene-Damm:
> I take it that changing the file on disk still properly causes a reload?

Yup tried that and as long as the serialized data is different in any way, it'll reload as before.

jechter (approved):

Issue #472, opened 2019-03-16
enabling events are inconsistent 


gotenxds:
Working on macos, unity 2018.3.7f1 input system 0.2.1

This may be related to #261 but a bit different.

If you keep presing a button while enabling/disabling it you will get different results and different times:

1. press button (and hold it)
2. disable event
3. enable event => event is called (button is still pressed)
4.disable event
5. enable even => event is not called this time (button is not pressed.)

Edit, so I made more checks And wrote a simple as I could make it code example and found even more weird stuff here:

`            controls.Player.first.performed += _ => Debug.Log("Entered First");
            controls.Player.first.cancelled += _ => Debug.Log("cancelled First");
            controls.Player.second.performed += _ =>
`

`

            {
                if (controls.Player.first.enabled)
                {
                    Debug.Log("Disable");
                    controls.Player.first.Disable();
                }
                else
                {
                    Debug.Log("Enable");
                    controls.Player.first.Enable();
                }
            };`

So using this code in Awake of a game Object I get the following results.

![image](https://user-images.githubusercontent.com/3519520/54476140-508c0980-4802-11e9-90ea-50e3c6546c7b.png)

This is the log I got from preforming the following key strokes: 
1. first
2. letgo of first
3. first
4. let go of first
5. second and let go of second
6. first and let go of first
7. second and let go (enabling first again)
8. first and let go (no event fired)

Rene-Damm:
Confirmed. Looking into it.

Rene-Damm:
Should be fixed by #523.

PR #523, opened 2019-04-08
FIX: Calling Enable/Disable from action callbacks corrupting internal state (#472).

Rene-Damm:
What I found was that adding and removing state monitors while in state monitor callbacks wasn't safe. This led to actions seemingly not triggering anymore even though `Enable` had been called on them. Changed the code such that calling Enable/Disable from callbacks is now safe.

Also fixes
* Blurry icons on .inputactions assets and actions in them
* Failing warnings tests.
* Auto-Save toggle in action editor resetting on domain reload (also every other local user input setting).

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
@@ -20,10 +20,10 @@ namespace UnityEngine.Experimental.Input.Editor
     {
         private const int kVersion = 6;
 
-        private const string kActionIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/Add Action.png";
-        private const string kAssetIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/Add ActionMap.png";
-        private const string kActionIconDark = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/d_Add Action.png";
-        private const string kAssetIconDark = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/d_Add ActionMap.png";
+        private const string kActionIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/Add Action@4x.png";
+        private const string kAssetIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/Add ActionMap@4x.png";
+        private const string kActionIconDark = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/d_Add Action@4x.png";
+        private const string kAssetIconDark = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/d_Add ActionMap@4x.png";
 
         [SerializeField] private bool m_GenerateWrapperCode;
         [SerializeField] private string m_WrapperCodePath;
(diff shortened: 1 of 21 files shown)


douglas-piconi:
this is causing something weird now on project startup, the package verification failing because this is logged in console     

  Found logs in the console on project startup.

`
Logs: Cancelling DisplayDialog: Warning This project is using the new input system package but the native platform backends for the new input system are not enabled in the player settings. This means that no input from native devices will come through.

Do you want to enable the backends. Doing so requires a restart of the editor.
This should not be called in batch mode.
UnityEditor.EditorAssemblies:ProcessInitializeOnLoadAttributes(Type[])
 (Library/PackageCache/com.unity.inputsystem@0.2.6-preview/InputSystem/InputSystem.cs:1651)

  Expected: <empty>
  But was:  "Cancelling DisplayDialog: Warning This project is using the new input system package but the native platform backends for the new input system are not enabled in the player settings. This means that no input from native devices will come through.

Do you want to enable the backends. Doing so requires a restart of the editor.
This should not be called in batch mode.
UnityEditor.EditorAssemblies:ProcessInitializeOnLoadAttributes(Type[])
 (Library/PackageCache/com.unity.inputsystem@0.2.6-preview/InputSystem/InputSystem.cs:1651)`

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -20,10 +20,10 @@ public class InputActionImporter : ScriptedImporter
     {
         private const int kVersion = 6;
 
-        private const string kActionIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/Add Action.png";
-        private const string kAssetIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/Add ActionMap.png";
-        private const string kActionIconDark = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/d_Add Action.png";
-        private const string kAssetIconDark = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/d_Add ActionMap.png";
+        private const string kActionIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/Add Action@4x.png";
```
jechter, line 23: Shouldn't we actually get the _right_ version for the current resolution instead?

jechter (commented):
Changes look fine (except for the minor comment about picking the actual icon resolution). But, it seems that CI is still red with this. Let's not land anything else before it is green again, so we don't regress further.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
Rene-Damm: What's the right version, though? The icons can be zoomed in and out via the slider in the Project Browser. So in a way, our largest available option at @4x version is already the wrong version that quickly looks quite blurry. Going smaller doesn't make sense IMO.

jechter: Oh. I see. These icons are being baked into the generated assets. Then, yeah, 4x is fine. But... baking the skin into the icon? So it looks different when using light or dark skin on import? That makes no sense, IMO. Do other asset icons change when choosing skins? I think we should pick one which works decently in both skins, and if neither does, design a new one.

Rene-Damm: Oh good point. I think designing new icons would be best. What I found is that even with the right skin, the icons only look good and proper when in non-preview mode. As soon as you go and display previews, the action icons nested inside the asset become unreadable against the background (the root asset icon does look decent, though).

Not sure what others do. My assumption is they choose icons that work equally well or badly :)

jechter: I think that for asset icons what we really need are custom designed 128x128 (at least) icons (so it looks correct even in preview with high scale), which work well on both backgrounds.

jechter (approved):

douglas-piconi (approved):

PR #525, opened 2019-04-09
FIX: In the action map editor window, switching from renaming an acti…

jechter:
…on to renaming an action map will no longer break the UI.

This is to fix https://fogbugz.unity3d.com/f/cases/1140869/

The fix works, but it is not particularly elegant. I cannot think of a better solution, but if someone can, let me know. The problem is that the TreeView does not expose all it's functionality in public APIs, so we are limited in what we know about it's state and in being able to tell it what to do.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
@@ -561,9 +561,7 @@ namespace UnityEngine.Experimental.Input.Editor
                 // Now, our fix to this problem is to force-end and accept any renaming session on the action column if we see 
                 // that the action map column had processed the current event. This is not particularly elegant, but I cannot think
                 // of a better solution as we are limited by the public APIs exposed by TreeView.
-                m_ActionsTree.forceAcceptRename = true;
-                m_ActionsTree.EndRename();
-                m_ActionsTree.forceAcceptRename = false;
+                m_ActionsTree.EndRename(forceAccept: true);
             }
             DrawActionsColumn(columnAreaWidth * 0.38f);
             DrawPropertiesColumn(columnAreaWidth * 0.40f);
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
@@ -546,7 +546,21 @@ namespace UnityEngine.Experimental.Input.Editor
             var columnAreaWidth = position.width - InputActionTreeView.Styles.backgroundWithBorder.margin.left -
                 InputActionTreeView.Styles.backgroundWithBorder.margin.left -
                 InputActionTreeView.Styles.backgroundWithBorder.margin.right;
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
```diff
@@ -546,7 +546,25 @@ private void OnGUI()
...
+                // get to use the mouse event before the action collumn gets to see it, which would cause the action map column
+                // to enter rename mode and use the event, before the action column gets a chance to see it and exit rename mode.
+                // Then we end up with two active renaming sessions, which does not work correctly. 
+                // (See https://fogbugz.unity3d.com/f/cases/1140869/). 
+                // Now, our fix to this problem is to force-end and accept any renaming session on the action column if we see 
+                // that the action map column had processed the current event. This is not particularly elegant, but I cannot think
+                // of a better solution as we are limited by the public APIs exposed by TreeView.
+                m_ActionsTree.forceAcceptRename = true;
+                m_ActionsTree.EndRename();
+                m_ActionsTree.forceAcceptRename = false;
```
Rene-Damm: The public property seems a bit awkward. Maybe have `ForceEndRename` method that leaves the rest to `InputActionTreeView`?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
```diff
@@ -546,7 +546,25 @@ private void OnGUI()
...
+            }
```
Rene-Damm, line 565: What about just hooking into `onBeginRename` on both of the trees and simply call `ForceEndRename` on the other tree? (making `ForceEndRename` do nothing if no rename is in progress)

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
jechter: Hmm, yeah, that's better.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
jechter: Nope, does not work. Because we cannot guarantee that `onBeginRename` is called. That's one of the parts of what I meant by "The problem is that the TreeView does not expose all it's functionality in public APIs, so we are limited in what we know about it's state and in being able to tell it what to do." . `onBeginRename` is called by our `BeginRename` method, but that is only called if we explicitly call that. If the TreeView internally initiates the rename, our `BeginRename` is not called, because `BeginRename` is not virtual.

Rene-Damm: Ah right, that problem...

Thinking I should have worked around this properly instead of the `public new void BeginRename(...)` thing. `RowGUI` can detect when a new rename has been started and trigger the callback instead. Not perfect but seems like a more thorough solution than what `InputActionTreeView` currently does.

Rene-Damm (approved):

douglas-piconi (approved):

PR #526, opened 2019-04-09
FIX: ReleaseOnly not triggering on actions with multiple bound controls (#492).

### LINKED ISSUE #492: Release-Only Button Interaction Produces no events
[PressButtonGamepadMinRepro.zip](https://github.com/Unity-Technologies/InputSystem/files/3000846/PressButtonGamepadMinRepro.zip)

This updated sample shows that buttonNorth/keyboard a produce no events when the action is set to "Release" and not "Press" or "PressAndRelease".

Win10 64, tested with keyboard, PS4 DualShock, and XBOX One Gamepads. 0.2.6-preview

Rene-Damm:
Fixes #492.


Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1108,13 +1108,74 @@ public void Actions_WithMultipleBoundControls_DriveInteractionsFromControlWithGr
...
+            Press(gamepad.buttonNorth);
+
+            Assert.That(trace, Is.Empty);
+
+            Release(gamepad.buttonNorth);
+
+            actions = trace.ToArray();
+            Assert.That(actions, Has.Length.EqualTo(1));
+            Assert.That(actions[0].phase, Is.EqualTo(InputActionPhase.Performed));
+            Assert.That(actions[0].control, Is.SameAs(gamepad.buttonNorth));
```
jechter: Why do we test this case twice?

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1108,13 +1108,74 @@ public void Actions_WithMultipleBoundControls_DriveInteractionsFromControlWithGr
...
+            Press(keyboard.aKey);
+
+            Assert.That(trace, Is.Empty);
+
+            Release(keyboard.aKey);
+
+            actions = trace.ToArray();
+            Assert.That(actions, Has.Length.EqualTo(1));
+            Assert.That(actions[0].phase, Is.EqualTo(InputActionPhase.Performed));
+            Assert.That(actions[0].control, Is.SameAs(keyboard.aKey));
```
jechter: What happens (or what should happen), if I press buttonNorth, then a, and then release buttonNorth, and then a? or  press buttonNorth, then a, and then release a, and then buttonNorth? On which releases will it trigger? Can we make the test cover that behavior?

jechter (changes requested):

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: Probably warrants a comment. I had the case where the code wouldn't trigger correctly again for the same button so wanted to make sure that the second button press of the same button comes through correctly.

Packages/com.unity.inputsystem/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: >What happens (or what should happen), if I press buttonNorth, then a, and then release buttonNorth, and then a?

1. Reacts to press of `buttonNorth`.
2. Does not react to press of `a` (same level of actuation as `buttonNorth` so does not result in change on the action).
3. It switches from `buttonNorth` to `a` when `buttonNorth` is released (sticks to control with greatest level of actuation).
4. It reacts to the release of `a`.

>or press buttonNorth, then a, and then release a, and then buttonNorth?

The in-between press and release of `a` has no effect. It has the same level of actuation as `buttonNorth` and thus will not get to drive the action as `buttonNorth` is already driving it. If it had *greater* actuation than `buttonNorth`, it would take over and temporarily drive the action until it is released (as then `buttonNorth` has the greater level of actuation).

Sounds complicated this way. Put simply, when there are multiple controls driving an action, the code attempts to always stick to the control with the greatest level of actuation. So as long as any control is still actuated, the action keeps going.

>Can we make the test cover that behavior?

Might be worth having a dedicated test for this. Not sure. ATM the test that covers the disambiguation described above is tested by `Actions_WithMultipleBoundControls_DriveInteractionsFromControlWithGreatestActuation`.

Rene-Damm: Ok, fleshed out the test and covered the "two buttons being pressed and then released one after the other" scenario.

jechter (approved):

douglas-piconi:
Looks like 1 test failure

Test Name: CoreTests.Actions_CanQueryLastTrigger

Test Fixture:

Test Status: Failure

Test Execution Time: 00:00:00.048

Message:
  Expected: same as <Button:/Gamepad/rightTrigger>
  But was:  null

Rene-Damm:
@douglas-piconi @jechter Fixed the test failure and tweaked the interaction stuff some more. Also removes the `InputAction.lastTriggerXXX` APIs.

Assets/Demo/DemoGame.cs
```diff
@@ -263,6 +263,7 @@ public void OnStartGameButton()
         Debug.Assert(state == State.InMainMenu);
 
         ////TODO: this should be something that UIActionInputModule can give us without us querying the action directly
+        /*
```
jechter, line 266: So what do we do with this commented out block?

jechter:
Should we have test coverage for the new press points for the interactions?

Assets/Demo/DemoGame.cs
Rene-Damm: Depends a bit on where the demo goes. Likely will get deleted.

Most of the code here has been obsoleted by `PlayerInput`. The functionality commented out is something that `PlayerInputManager` is meant to handle. When the UI stuff gets a pass, IMO makes sense to revisit that too and add the functionality to PlayerInput.

Rene-Damm:
>Should we have test coverage for the new press points for the interactions?

Probably :) Writing.

Rene-Damm:
@jechter Test added here https://github.com/Unity-Technologies/InputSystem/pull/526/commits/63ff9fe90874a9082e3e43dce55ef64e1057b949

Also added a helper to test action sequences in `InputActionTrace` to get rid of the verbosity and cut down on the code needed to do this kind of stuff.

jechter (approved):

douglas-piconi (approved):

PR #527, opened 2019-04-09
FIX: Disabling and enabling actions in action callbacks should work correctly now.

jechter:
Fixes https://fogbugz.unity3d.com/f/cases/1140861/

This did not work, because if a callback changed the enabled actions, it would change `m_StateChangeMonitors[deviceIndex].signalled`, but we would then later overwrite that change when `FireStateChangeNotifications` is done. Now we clear all bits instead of overwriting, which fixes this problem.


jechter:
Redundant by https://github.com/Unity-Technologies/InputSystem/pull/523

PR #530, opened 2019-04-10
FIX: Fix choosing "Add Action" in action map context menu throwing an…

jechter:
… exception.

Not particularly elegant, but cannot think of a better solution (other then maybe dropping that menu item, or making it not select the newly created object?). The problem is that you choose "Create Action" on the context menu in the _Action Map_ Tree View, but the item which needs to be selected is in the _Action_ Tree view, which is different.

Fixes https://fogbugz.unity3d.com/f/cases/1141331/

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
@@ -825,7 +825,7 @@ namespace UnityEngine.Experimental.Input.Editor
             var canRename = false;
             if (itemType == typeof(ActionMapTreeItem))
             {
-                menu.AddItem(s_AddActionLabel, false, () => {(viewToAddActions ?? this).AddNewAction(); });
+                menu.AddItem(s_AddActionLabel, false, AddNewAction);
             }
             else if (itemType == typeof(ActionTreeItem))
             {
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
```diff
@@ -1232,6 +1232,7 @@ private bool ReloadIfSerializedObjectHasBeenChanged()
         public bool drawMinusButton { get; set; }
         public float foldoutOffset { get; set; }
 
+        public InputActionTreeView viewToAddActions { get; set; }
```
Rene-Damm: Would prefer for `InputActionTreeView` to have an `onHandleAddNewAction` delegate property or something like that.

```
        public void AddNewAction(SerializedProperty actionMapProperty)
        {
            if (onHandleAddNewAction != null)
                onHandleAddNewAction(actionMapProperty);
            else
            {
                //...
            }
        }
```

douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionTreeView.cs
jechter: Agreed, that's much better. Changed.

Rene-Damm (approved):

PR #533, opened 2019-04-11
CHANGE: Move tests to separate package

jechter:
This PR moves the tests to a separate com.unity.inputsystem.tests package. That way, we don't distribute test code to end users. Yet, the InputTestFixture class is still in the com.unity.inputsystem package, and will be shipped to end users (so they can test their own input code). This way, they can use the test fixture without dragging our tests into their project (case https://fogbugz.unity3d.com/f/cases/1143798/).

Getting upm-ci to pick up both packages was a bit tricky as it is normally used to test packages _in isolation_, and I had to work around it a bit to make it pick up both packages together. That solution took me some amount of tweaking to figure out, and it is not particularly elegant, but works. Suggestions for improving the solution are welcome!


Rene-Damm (approved):

Packages/com.unity.inputsystem.tests/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [0.2.6-preview] - 2019-04-14
+
+Test split off into separate package
```
Rene-Damm, line 8: Do we need a separate CHANGELOG for this one? I assume it's mainly a technicality required by package verification but what if we just add a link to the main CHANGELOG and leave it at that?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   * They are still on `DualShockGamepadPS4`.
   * The reason is that ATM we do not yet support these controls other than on the PS4. The previous setup pretended that these controls work when in fact they don't.
 - Marking a control as noisy now also marks all child controls as noisy.
+- Tests are now no longer part of the com.unity.inputsystem package. The `InputTestFixture` class still is for when you want to write input-related tests for your project. You can reference the `Unity.InputSystem.TestFixture` assmembly when you need to do that.
```
Rene-Damm, line 45: Typo `assmembly` :)

Rene-Damm (approved):

Packages/com.unity.inputsystem.tests/CHANGELOG.md
jechter: Yes, I had to add it due to the verification. But, good point, if we can do this using a symlink, then that would make it easier to maintain. Will give it a try.

jechter:
Actually, @Rene-Damm suggested to not use a package for the tests at all, but to move them to the project folder. That way, we don't need to hack upm-ci to get both packages in isolation, _and_ we also pick up the demo game tests in CI. Seems to be a better approach, so will close this PR in favor of https://github.com/Unity-Technologies/InputSystem/pull/543

PR #534, opened 2019-04-11
CHANGE: Input settings UX improvements

jechter:
I was not happy with the handling of assets in the Input Settings window. @Rene-Damm convinced me that having settings stored in dynamic locations is worth having (maybe ;) ) , but in most cases, the user should not have to think about it. So I have made this mostly automatic, and rolled it into a gear menu. I also exposed it in the inspector when selecting an Input Settings asset, so you can activate it from there (and we no longer show an ugly debug version of the settings inspector there instead).

- In the Input Settings window, asset selection has now been moved to the "gear" popup menu. If no asset is created, we now automatically create one.
- In the inspector for Input Settings assets, we now show a button to go to the Input Settings window, and a button to make the asset active if it isn't.


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -185,11 +166,16 @@ private void InitializeWithCurrentSettings()
...
-                EditorBuildSettings.RemoveConfigObject(kEditorBuildSettingsConfigKey);
+                if (m_AvailableInputSettingsAssets.Length != 0)
+                {
+                    m_CurrentSelectedInputSettingsAsset = 0;
+                    m_Settings = AssetDatabase.LoadAssetAtPath<InputSettings>(m_AvailableInputSettingsAssets[0]);
+                    InputSystem.settings = m_Settings;
+                }
+                else
+                {
+                    CreateNewSettingsAsset("Assets/InputSystem.inputsettings.asset");
```
Rene-Damm: Hmm, this one I'm not so sure about. To me it seems that just *looking* at something should not have the side effect of creating an asset. But *if* I read the changes correctly, then just clicking "Input (NEW)" in the project settings now creates an asset. As a user, that would bother me.

Rene-Damm: In a similar vein, it seems pointless to create an asset if all it does is have the default settings. To me, it seemed nice to basically make the input settings asset thing only something you needed to bother with if you *actually* needed it.

Rene-Damm (commented):
Overall change LGTM. Not sure sure about always creating an asset.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
jechter: Ok, added a button as you suggested :)

Rene-Damm (approved):

PR #545, opened 2019-04-16
CHANGE: The input system now defaults to ignoring any HID devices wit…

jechter:
…h usage types not known to map to game controllers. You can use `HIDSupport.supportedUsages` to enable specific usage types.

This cleans up the device list, and significantly speeds up startup/domain reload on my mac. 

Code wise the change is straightforward, but I'm not sure the Dictionary/HashMap is an API we want to use?


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HIDSupport.cs
```diff
@@ -1,6 +1,8 @@
+using System.Collections.Generic;
 #if UNITY_EDITOR
 using UnityEditor;
 using UnityEngine.Experimental.Input.Editor;
+using UnityEngine.Experimental.Input.Plugins.HID;
```
Rene-Damm: Redundant using.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HIDSupport.cs
```diff
@@ -50,6 +62,8 @@ public static void Initialize()
             #endif
         }
 
+        public static Dictionary<HID.UsagePage, HashSet<int>> supportedUsages { get; set; }
```
Rene-Damm: Hmm, the dictionary pains me a little :) A dictionary with a nested hashset for the case here... that's four heap objects for a relatively trivial lookup. Why not just a flat array of tuples?

Looking at the API, I'm also wondering now whether a callback wouldn't be better. Just puts the whole decision into the user's hands instead of just having a whitelist.

```
    using ShouldCreateHIDCallback = Func<HID.HIDDeviceDescriptor, bool?>;

//...

public static class HIDSupport
{
        public static event ShouldCreateHIDCallback shouldCreateHID
        {
            add => s_ShouldCreateHID.Append(value);
            remove => s_ShouldCreateHID.Remove(value);
        }

        internal static InlinedArray<ShouldCreateHIDCallback> s_ShouldCreateHID;

        private static bool? DefaultShouldCreateHIDCallback(HID.HIDDeviceDescriptor descriptor)
        {
            if (descriptor.usagePage == HID.UsagePage.GenericDesktop)
            {
                switch (descriptor.usage)
                {
                    case (int)HID.GenericDesktop.Joystick:
                    case (int)HID.GenericDesktop.Gamepad:
                    case (int)HID.GenericDesktop.MultiAxisController:
                        return true;
                }
            }
            return null;
        }
        
        /// <summary>
        /// Add support for generic HIDs to InputSystem.
        /// </summary>
        public static void Initialize()
        {
            s_ShouldCreateHID.Append(DefaultShouldCreateHIDCallback);

            //...
```

And then in the HID class.

```
            // Check callbacks to see whether we should actually create a device for this specific HID.
            // If no callback says yes or one says no, we ignore the device.
            if (HIDSupport.s_ShouldCreateHID.All(f => f(hidDeviceDescriptor) != true) ||
                HIDSupport.s_ShouldCreateHID.Any(f => f(hidDeviceDescriptor) == false))
                return null;
```

Seems more flexible. But also a bit more code.

What do you think?

jechter: I like it. (As i wrote, wasn't really sure about the Dictionary/HashSet myself, that felt kind of heavy).

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   * They are still on `DualShockGamepadPS4`.
   * The reason is that ATM we do not yet support these controls other than on the PS4. The previous setup pretended that these controls work when in fact they don't.
 - Marking a control as noisy now also marks all child controls as noisy.
+- The input system now defaults to ignoring any HID devices with usage types not known to map to game controllers. You can use `HIDSupport.supportedUsages` to enable specific usage types.
```
Rene-Damm, line 45: This looks to be outdated now.

Rene-Damm (approved):

PR #542, opened 2019-04-15
Switch develop

yebata:
To implement ISX-255, at first added device orientation as a usage to a device.
I improved it to be able to add/remove multiple usages in commit d608040, but it seemed to be in an another PR so I backed out it once.

PR #548, opened 2019-04-17
Improved to allow a device to have a multiple usages

yebata:
I improved a device to be able to have a multiple usages (especially for Switch Joy-Con). I'd like to ask you some points to make it to follow an ISX specification and its design concept more.

- To give a multiple usages to a control is now implemented in an another way. Should we change them more an integrated way to work with both controls and devices?

- Now giving a multiple usages to a device means you can find it with any usage it has.
e.g. If a game pad has "Vertical" and "LeftHand" as its usages, now both path `<Gamepad>{Vertical}` and `<Gamepad>{LeftHand}` match for it. 
Should we make path like `<Gamepad>{Vertical}{LeftHand}` to be acceptable to find a device which matches to all usages you give? And a one like `<Gamepad>{Vertical|Horizontal}` to find any devices which match to the one of usages you give?


yebata:
Don't know how to add members as reviewers to a draft PR XD
looking for the way....

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -446,10 +446,53 @@ internal bool hasControlsWithDefaultState
         internal void SetUsage(InternedString usage)
         {
             // Make last entry in m_UsagesForEachControl be our device usage string.
+            SetUsage(m_UsageToControl != null ? m_UsageToControl.Length : 0, usage);
+        }
+
+        internal void AddUsage(InternedString usage)
+        {
+            // Check if the usage already exists
+            if (Array.Exists(m_UsagesForEachControl, o => o == usage))
```
Rene-Damm: This also checks the usages for each control, not just for the device. Best way is probably to just iterate over `usages` on the device.

Also, the delegate likely allocates garbage. Would prefer to iterate directly here.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -446,10 +446,53 @@ internal bool hasControlsWithDefaultState
...
+                return;
+
+            SetUsage(m_UsagesForEachControl != null ? m_UsagesForEachControl.Length : 0, usage);
```
Rene-Damm: Minor thing... simpler way: `SetUsage(m_UsagesForEachControl.LengthSafe(), usage);`

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -446,10 +446,53 @@ internal bool hasControlsWithDefaultState
...
+        }
+
+        internal void RemoveUsage(InternedString usage)
+        {
             var numControlUsages = m_UsageToControl != null ? m_UsageToControl.Length : 0;
-            Array.Resize(ref m_UsagesForEachControl, numControlUsages + 1);
-            m_UsagesForEachControl[numControlUsages] = usage;
-            m_UsagesReadOnly = new ReadOnlyArray<InternedString>(m_UsagesForEachControl, numControlUsages, 1);
+            var usageIndex = Array.IndexOf(m_UsagesForEachControl, usage, numControlUsages);
```
Rene-Damm: Would prefer not to even iterate over the usages of the controls and only iterate over the ones for the device instead.

```
var index = -1;
var usages = this.usages;
for (var i = 0; i < usages.Count; ++i)
    if (usages[i] == usage)
    {
        index = i;
        break;
    }
if (index == -1)
    return;
```

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -446,10 +446,53 @@ internal bool hasControlsWithDefaultState
...
+            if (usageIndex < numControlUsages)
+                return;
+
+            var usageList = new System.Collections.ArrayList(m_UsagesForEachControl);
+            var commonPart = Array.FindAll<InternedString>((InternedString[])usageList.GetRange(numControlUsages, usageList.Count - numControlUsages).ToArray(typeof(InternedString)), e => e != usage);
+
+            Array.Resize(ref m_UsagesForEachControl, numControlUsages); 
+            ArrayHelpers.Append(ref m_UsagesForEachControl, commonPart);
```
Rene-Damm: Use `ArrayHelpers.EraseAt` instead of these four lines. The current code needless allocate several temporary objects.

```
ArrayHelpers.EraseAt(ref m_UsagesForEachControl, usageIndex);
```

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -871,6 +871,34 @@ public void SetUsage(InputDevice device, InternedString usage)
...
+            if (device == null)
+                throw new ArgumentNullException(nameof(device));
+            device.RemoveUsage(usage);
+
+            // Notify listeners.
+            for (var i = 0; i < m_DeviceChangeListeners.length; ++i)
+                m_DeviceChangeListeners[i](device, InputDeviceChange.UsageChanged);
+
+            // Usage may affect current device so update.
+            device.MakeCurrent();
```
Rene-Damm: Move 849-899 to private helper method

```
        private void NotifyUsageChanged(InputDevice device)
        {
            // Notify listeners.
            for (var i = 0; i < m_DeviceChangeListeners.length; ++i)
                m_DeviceChangeListeners[i](device, InputDeviceChange.UsageChanged);

            // Usage may affect current device so update.
            device.MakeCurrent();
        }
```

Rene-Damm (changes requested):
Thanks for tackling this! Left a couple comments.

yebata:
Thanks @Rene-Damm . Right, I didn't much pay attention to prevent causing unnecessary temporal memory allocation. I'll fix them and other points you pointed out.

Is there any future perspective about to extend path to allow describing a multiple usage?

Rene-Damm:
>Is there any future perspective about to extend path to allow describing a multiple usage?

Ah right, forgot to reply to that.

>Should we make path like <Gamepad>{Vertical}{LeftHand} to be acceptable to find a device which matches to all usages you give? And a one like <Gamepad>{Vertical|Horizontal} to find any devices which match to the one of usages you give?

Yup, I think we absolutely should. Not sure yet about the the syntax. But at least `<Gamepad>{Vertical}{LeftHand}` seems like a straightforward thing that should be supported.

yebata:
Thanks. I'll take a look it later. I'll keep that in my mind to not allocate garbage memory in a screening process.

yebata:
Applied feedback and added to support a multiple usages syntax like `<Gamepad>{Vertical}{LeftHand}`
I expect this should be merged into the dev branch after PR #542 is merged.

Rene-Damm:
Sorry @yebata, lost track of this along the way. I've pulled your changes and incorporated them into #734.

PR #549, opened 2019-04-17
CHANGE: Add a bunch of documentation; still incomplete.

Rene-Damm:
This is a batch of documentation updates as well as a few minor code cleanups/changes.

Many of the pages are still incomplete but I think it's a good step forward. Would like to merge this, go back to fixing a couple things and maybe doing the touch work, and then coming back to the docs for a second round.


Packages/com.unity.inputsystem/Documentation~/ActionAssets.md
```diff
@@ -0,0 +1,135 @@
+# Action Assets
```
jechter: Let's try for some more consistent naming here. In the docs we call it "Action Asset" now, but in the editor UI we call it "Input Actions". Yes, I get that "Input" may feel redundant in the input system docs and that it is important that this is an asset, but still, using a different wording is very confusing (I had to read the page to make sure it was talking about the same things). And we don't call PhysicMaterials "Physic Material Assets" in our docs either.

jechter: I guess the problem here is that "Input Actions" also refers to the InputAction class, but the asset actually contains more of that. But then, maybe we should rename the asset (also in the editor and API)? "Input Action Collection" or something?

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -0,0 +1,223 @@
+# Action Bindings
+
+----mention/explain disambiguation on this page
+
+An `InputBinding` represents a connection between an action and one or more controls identified by a [control path](Controls.md#control-paths). An action can have arbitrary many bindings pointed at it and the same control may be referenced by multiple bindings.
```
jechter, line 7: Again, why is this page called Action Binding, when the actual class name is `InputBinding`?

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -0,0 +1,223 @@
...
+        EditorGUILayout.Label("Custom stuff");
+        target.floatParameter = EditorGUILayout.FloatField("Some Parameter", target.floatParameter);
+    }
+}
+#endif
+```
+
+## Binding Resolution
+
+...
```
jechter, line 200: Should also talk about which binding drives an action if multiple are used here

jechter: Also, "Pass Through"

jechter: And "Initial State Check"

Packages/com.unity.inputsystem/Documentation~/Debugging.md
```diff
@@ -2,13 +2,22 @@
...
+|----|-----------|
+|Devices|[Input devices](Devices.md) that are currently added to the system as well as a list of unsupported/unrecognized devices.|
+|Layouts||
+|Actions|Only visible in play mode and only if there are [actions](Actions.md) that are currently enabled.<br><br>Shows the list of all currently enabled actions and the controls they are bound to.|
+|Users|Only visible when there is one or more `InputUser` instances. See [documentation](UserManagement.md).<br><br>Lists all currently active uers along with their active control schemes and devices, all their associated actions as well as the controls they are bound to.<br><br>Note that `PlayerInput` uses `InputUser` underneath. This means that when using `PlayerInput` components, each player will have an entry here.|
+|Settings||
+|Metrics||
 
->NOTE: The input debugger connection to the player is still being worked on. At the moment, it is still unstable and unreliable.
+## Connecting a Player
```
jechter, line 21: Todo: document "Remote Devices…" popup menu

Packages/com.unity.inputsystem/Documentation~/Interactions.md
```diff
@@ -0,0 +1,133 @@
...
+|*Press* (PressAndRelease)|<br>|Control crosses button press threshold|Control goes back below button threshold|
+|*Press* (PressOnly; continuous)|<br>|Control crosses button press threshold; then any time the control changes value or at least every frame|<br>|
+|*Press* (ReleaseOnly; continuous)|<br>|No difference to non-continuous mode| |
+|*Press* (PressAndRelease; continuous)|<br>|Control cross button press threshold; then any time the control changes value or at least every frame|Control goes back below button threshold|
+|*Hold*|Control Actuated|Held for >= `duration`|
+|*Hold* (continuous)|Control Actuated|Held for >= `duration`; after that, every frame regardless of whether the bound control receives input in the frame or not.|
+|*Tap*|Control Actuated|Control Released within `duration` (defaults to `InputSettings.defaultTapTime`) seconds|Control Released before `duration` seconds|
+|*SlowTap*|Control Actuated|Control Released within `duration` (defaults to `InputSettings.defaultSlowTapTime`) seconds|Control Released before `duration` seconds|
+|*MultiTap*|||
+
```
jechter, line 99: This is great to have documented!

Packages/com.unity.inputsystem/Documentation~/ActionAssets.md
Rene-Damm: I'm fine calling it something like "Input Action Assets" here. IMO focusing on the fact that these assets *principally* contain actions and ignoring the other stuff is okay.

"Input Action Collection" to me is a bit too generic. Action maps are basically collections of actions, too. And in fact, in the API we have `IInputActionCollection` which in fact both `InputActionAsset` and `InputActionMap` implement.

jechter: So, yeah, let's use `InputActionAsset` - and not drop the `Input` part in the docs.

PR #551, opened 2019-04-17
FIX: Fixed a bug where the event buffer used by `InputEventTrace` cou…

jechter:
…ld get corrupted.

Fixes https://fogbugz.unity3d.com/f/cases/1145616/


douglas-piconi (approved):
oh man that is a weird one.

Rene-Damm (approved):

PR #564, opened 2019-04-25
FIX: No deadzoning on stick.x/y causing noise.

Rene-Damm:
This adds deadzones to `StickControl.x` and `StickControl.y`.

* Unlike the `StickDeadzoneProcessor` on the stick itself, which has a radial deadzone, the deadzones on `x` and `y` are linear. Means they will not necessary match the `x` and `y` components on the `Vector2` of the stick.
* Fixes the problem with noise coming in through stick axes even though it is suppressed properly on the stick itself.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Utilities/TypeTable.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Utilities/TypeTable.cs
@@ -1,6 +1,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
+
+#if UNITY_EDITOR
+using System.ComponentModel;
+#endif
 
 namespace UnityEngine.Experimental.Input.Utilities
 {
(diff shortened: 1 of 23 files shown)


Packages/com.unity.inputsystem/InputSystem/Utilities/TypeTable.cs
```diff
@@ -65,5 +70,20 @@ public Type LookupTypeRegistration(string name)
                 return type;
             return null;
         }
+
+        #if UNITY_EDITOR
+        public bool ShouldHideInUI(string name)
```
Rene-Damm, line 75: This isn't used ATM. Was from an aborted attempt that added a new type of processor for internal use. Left the code in here because I think the functionality is actually generally useful (for layouts, we already have and use something similar).

douglas-piconi (approved):
Nice fix, this one had been eluding us for quite a while!

douglas-piconi:
This PR shouldn't affect standalone, but you might want to merge latest develop into this branch to get a standalone run just in case.

jechter (approved):

PR #571, opened 2019-04-30
Check formatting in CI

jechter:
Add a CI check to run our formatting utilities and to fail a PR if code is not properly formatted.

### DIFF
--- a/Assets/Demo/DemoControls.cs
+++ b/Assets/Demo/DemoControls.cs
@@ -473,10 +473,12 @@ public class DemoControls : IInputActionCollection
         m_menu_submit = m_menu.GetAction("submit");
         m_menu_point = m_menu.GetAction("point");
     }
+
     ~DemoControls()
     {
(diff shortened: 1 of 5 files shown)


douglas-piconi (approved):

Assets/Demo/DemoControls.cs
```diff
@@ -495,22 +497,27 @@ public bool Contains(InputAction action)
...
     public void Enable()
     {
         asset.Enable();
     }
+
     public void Disable()
     {
         asset.Disable();
     }
+
```
Rene-Damm, line 520: For generated code, I think it'd be great to generate it such that it already complies with formatting.

Rene-Damm (approved):

Assets/Demo/DemoControls.cs
jechter: Agreed - was planning to do that in a separate PR

PR #579, opened 2019-05-02
Multiplayer ui

jechter:
Added a `MultiplayerEventSystem` class, which allows you use multiple UI event systems to control different parts of the UI by different players. That way, each player can keep separate selections and state.

You can use the `playerRoot` property on MultiPlayerEventSystem to specify a part of the hierarchy belonging to this player. Mouse selection will ignore any game objects not within this hierarchy. For gamepad/keyboard selection, you need to make sure that the navigation links stay within the player's hierarchy.

Q: `playerRoot` is implemented by our input module, and will not work with e.g.  StandaloneInputModule. So it might be more logical to have this property on UIInputModule and not on the MultiplayerEventSystem. *But*: this is very similar to (and related to) EventSystem.firstSelected, which also needs to be implemented by the input modules (it actually did not work with `UIInputModule`, which this PR also fixes). So I figured it would be more consistent to have both on the EventSystem. Any opinions on this?

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.cs
@@ -297,7 +297,7 @@ internal class UITests : InputTestFixture
         foreach (var player in players)
             player.eventSystem.InvokeUpdate();
 
-        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.EqualTo(players[0].leftGameObject));
+        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.SameAs(players[0].leftGameObject));
         Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.Null);
 
         // Click right gameObject of player 1
@@ -309,8 +309,8 @@ internal class UITests : InputTestFixture
         foreach (var player in players)
             player.eventSystem.InvokeUpdate();
 
-        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.EqualTo(players[0].leftGameObject));
-        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.EqualTo(players[1].rightGameObject));
+        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.SameAs(players[0].leftGameObject));
+        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.SameAs(players[1].rightGameObject));
 
         // Click right gameObject of player 0
         InputSystem.QueueStateEvent(mouse, new MouseState { position = new Vector2(400, 100), buttons = 1 << (int)MouseButton.Left });
@@ -320,8 +320,8 @@ internal class UITests : InputTestFixture
         foreach (var player in players)
             player.eventSystem.InvokeUpdate();
 
-        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.EqualTo(players[0].rightGameObject));
-        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.EqualTo(players[1].rightGameObject));
+        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.SameAs(players[0].rightGameObject));
+        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.SameAs(players[1].rightGameObject));
     }
 
     [UnityTest]
@@ -360,8 +360,8 @@ internal class UITests : InputTestFixture
         // We need to wait a frame to let the underlying canvas update and properly order the graphics images for raycasting.
         yield return null;
 
-        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.EqualTo(players[0].leftGameObject));
-        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.EqualTo(players[1].leftGameObject));
+        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.SameAs(players[0].leftGameObject));
+        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.SameAs(players[1].leftGameObject));
 
         // Reset initial selection
         players[0].leftChildReceiver.Reset();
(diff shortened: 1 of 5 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MultiplayerEventSystem.cs
```diff
@@ -0,0 +1,16 @@
+using UnityEngine;
+using UnityEngine.EventSystems;
+
+public class MultiplayerEventSystem : EventSystem
```
jechter: Should this be in our namespace?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/UIInputModule.cs
```diff
@@ -87,6 +98,21 @@ internal void ProcessMouse(ref MouseModel mouseState)
             mouseState.OnFrameFinished();
         }
 
+        // if we are using a MultiplayerEventSystem, ignore any transforms
+        // not under the current MultiplayerEventSystem's root.
+        private bool MouseShouldIgnoreTransform(Transform t)
```
Rene-Damm: Thins one should probably be `PointerShouldIgnoreTransform`. Could be pen or touch, too.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MultiplayerEventSystem.cs
Rene-Damm: Yup. `UnityEngine.Experimental.Input.Plugins.UI`.

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -223,6 +245,194 @@ public IEnumerator MouseActions_CanDriveUI()
...
+        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.EqualTo(players[0].leftGameObject));
+        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.EqualTo(players[1].leftGameObject));
+
+        // Reset initial selection
+        players[0].leftChildReceiver.Reset();
+        players[1].leftChildReceiver.Reset();
+
+        // Check Player 0 Move Axes
+        InputSystem.QueueDeltaStateEvent(gamepads[0].leftStick, new Vector2(1.0f, 0.0f));
+        InputSystem.Update();
```
Rene-Damm, line 372: These days, `InputTestFixture` has some nice wrappers around this kind of stuff. Can just do

```
Set(gamepads[0].leftStick, new Vector2(1, 0));
```

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -223,6 +245,194 @@ public IEnumerator MouseActions_CanDriveUI()
...
+
+        foreach (var player in players)
+        {
+            Assert.That(player.leftChildReceiver.events, Has.Count.EqualTo(0));
+            Assert.That(player.rightChildReceiver.events, Has.Count.EqualTo(0));
+            player.eventSystem.InvokeUpdate();
+        }
+
+        Assert.That(players[0].eventSystem.currentSelectedGameObject, Is.EqualTo(players[0].rightGameObject));
+        Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.EqualTo(players[1].leftGameObject));
```
Rene-Damm: Minor... `SameAs` may be more appropriate here.

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -223,6 +245,194 @@ public IEnumerator MouseActions_CanDriveUI()
...
+        Assert.That(players[1].leftChildReceiver.events, Has.Count.EqualTo(1));
+        Assert.That(players[1].leftChildReceiver.events[0].type, Is.EqualTo(EventType.Submit));
+        players[1].leftChildReceiver.Reset();
+
+        foreach (var player in players)
+        {
+            Assert.That(player.leftChildReceiver.events, Has.Count.EqualTo(0));
+            Assert.That(player.rightChildReceiver.events, Has.Count.EqualTo(0));
+        }
+    }
```
Rene-Damm, line 434: Don't remember what the test framework does with the scene. Will it clean up the scene? I vaguely remember it resets the scene only once and then whatever each test spills into the scene will stick around. Wondering if we can get side-effects between the tests here?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MultiplayerEventSystem.cs
Rene-Damm: Please add some XML docs.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MultiplayerEventSystem.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    [Tooltip("If set, only process mouse events for any game objects which are children of this game object.")]
+    public GameObject playerRoot;
+
+    protected override void Update()
+    {
+        EventSystem originalCurrent = EventSystem.current;
+        current = this; // in order to avoid reimplementing half of the EventSystem class, just temporarily assign this EventSystem to be the globally current one
+        base.Update();
+        current = originalCurrent;
```
Rene-Damm: Might want to use `try...finally` to protect against exceptions.

Rene-Damm (approved):

douglas-piconi (approved):

Assets/Tests/InputSystem/Plugins/UITests.cs
jechter: This throws a NullReferenceException.

Assets/Tests/InputSystem/Plugins/UITests.cs
jechter: Yes, it seems to clean this  up between tests.

PR #582, opened 2019-05-03
Make PlayerInput synch with InputSystemUIInputModule

jechter:
FYI: PR is against `rename-uiactioninputmodule` to have a clean diff. Will target `develop` once `rename-uiactioninputmodule` is landed.

`InputSystemUIInputModule` now lets you specify an `InputActionAsset` in the `actionsAsset` property. If this is set, the inspector will populate all actions from this asset. If you have a `PlayerInput` component on the same game object, referencing the same  `InputActionAsset`, the `PlayerInput` component will keep the actions on the `InputSystemUIInputModule` in synch, allowing easy setup of multiplayer UI systems.

This setup seems to work quite nice for controlling multiple UI instances by multiple players. But I still have a few things I'm unsure about:

-now the InputSystemUIInputModule inspector only lets you choose actions from the `InputActionAsset` if one is provided. If non is provided, we use the same UI as before. This works, but do we need  such flexibility for configuration? Should we maybe only allow configuration via `InputActionAsset` for a less confusing variety of options? Or is there a use case for which we have to  support actions serialized into the component?

-To have  this work together with  PlayerInput, the PlayerInput and the InputSystemUIInputModule need to be on the same game object and reference the same `InputActionAsset`. Then it just works. Is this too much magic? Should we somehow make this configuration more explicit?


jechter:
Branch target updated to `develop`, as `rename-uiactioninputmodule` has landed.

Rene-Damm:
>If you have a PlayerInput component on the same game object

Not sure about this one. Seems restrictive. Think I would prefer having an explicit reference like for the camera instead of an implicit search.

Rene-Damm:
>Or is there a use case for which we have to support actions serialized into the component?

For this particular component, could be just fine simply requiring an asset.

Rene-Damm:
>Should we somehow make this configuration more explicit?

Yeah, IMO requiring specific GO setups that we implicitly rely on is the inferior solution in most cases. Linking objects with references is quick enough IMO and generally more robust and flexible.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -658,15 +666,22 @@ private void InitializeActions()
...
+
             ////REVIEW: should we *always* Instantiate()?
             // Check if we need to duplicate our actions by looking at all other players. If any
             // has the same actions, duplicate.
             for (var i = 0; i < s_AllActivePlayersCount; ++i)
                 if (s_AllActivePlayers[i].m_Actions == m_Actions && s_AllActivePlayers[i] != this)
-                {
                     m_Actions = Instantiate(m_Actions);
-                    break;
-                }
```
Rene-Damm: Why remove the break? The first `Instantiate(m_Actions)` would yield a new instance distinct from all other instances so going on with the iteration seems pointless.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -864,6 +864,57 @@ public void OnAfterDeserialize()
             m_ScrollWheelAction = OnAfterSerializeActionProperty(m_ScrollWheelActionReference, m_ScrollWheelActionData);
         }
 
+        [SerializeField, HideInInspector] private InputActionAsset m_ActionsAsset;
+        internal InputActionAsset actionsAssetNoEnable
```
Rene-Damm: This seems a bit unnatural as a property. A simple method with just the setter code seems more appropriate.

Rene-Damm (changes requested):
Tests?

jechter:
> > If you have a PlayerInput component on the same game object
> 
> Not sure about this one. Seems restrictive. Think I would prefer having an explicit reference like for the camera instead of an implicit search.

Yeah, was unsure as well, hence I asked - will try how an explicit reference works.

> > Or is there a use case for which we have to support actions serialized into the component?
> 
> For this particular component, could be just fine simply requiring an asset.

Cool, I thought so, but wanted to make sure. That should make the logic of both the asset and the inspector quite a bit simpler (don't need to care about handling both reference and action cases and serializing the data correctly for each).

> Tests?

Sure. As I wrote on slack "Will also add some tests for this once we agree that this is the design we actually want.". Didn't want to bother if I'd end up rewriting it all anyways.

jechter:
Another problem may be who owns the enabling of actions, if PlayerInput and InputSystemUIInputModule share the same action asset instance. Currently my idea is that PlayerInput enables the actions for it's selected map, and InputSystemUIInputModule enables the actions it cares about. But sometimes PlayerInput turns off all the actions in the asset, so in those cases, it needs to tell InputSystemUIInputModule to reenable its actions. Which is what I'm doing now, and it works - but it seems a little fragile. Would it make more sense if they both had separate instances instead?

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
jechter: Yeah, this was unintentional (I refactored this code a little back and forth). Will fix!

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jechter: Makes sense.

Rene-Damm:
>Another problem may be who owns the enabling of actions, if PlayerInput and InputSystemUIInputModule share the same action asset instance.

Hmm, good question.

>Currently my idea is that PlayerInput enables the actions for it's selected map, and InputSystemUIInputModule enables the actions it cares about. But sometimes PlayerInput turns off all the actions in the asset, so in those cases, it needs to tell InputSystemUIInputModule to reenable its actions.

Not sure. Seems like the question lurking there is what determines whether UI gets input or not. Think we may be lacking a decider mechanism there.

Overall, seems like PlayerInput should generally leave the UI actions alone.

>Would it make more sense if they both had separate instances instead?

I think that might lead to more problems than it solves. Generally would prefer if PlayerInput is smart enough to properly take UI input into account. Both when it comes to individual actions under its domain as well as when it comes to respecting the switching back and forth between UI and gameplay (plus the mixed scenario when *both* in-game UI and gameplay are active).

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
```diff
@@ -110,16 +111,24 @@ public override void OnInspectorGUI()
...
+            EditorGUILayout.PropertyField(uiModuleProperty, m_UIPropertyText);
+            if (EditorGUI.EndChangeCheck())
+                serializedObject.ApplyModifiedProperties();
+
+            if (uiModuleProperty.objectReferenceValue != null)
+            {
+                var uiModule = uiModuleProperty.objectReferenceValue as InputSystemUIInputModule;
+                if (actionsProperty.objectReferenceValue != null && uiModule.actionsAsset != actionsProperty.objectReferenceValue)
+                {
+                    EditorGUILayout.HelpBox("The referenced InputSystemUIInputModule is configured using differnet input actions then this PlayerInput. They should match if you want to synch PlayerInput actions to the UI input.", MessageType.Warning);
```
Rene-Damm: "sync" or "synchronize"?

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
```diff
@@ -110,16 +111,24 @@ public override void OnInspectorGUI()
...
+                    if (GUILayout.Button(m_FixInputModuleText))
+                        InputSystemUIInputModuleEditor.ReassignActions(uiModule, actionsProperty.objectReferenceValue as InputActionAsset);
+                }
+            }
```
Rene-Damm, line 131: Think it would be great to also detect the case here where the UI event system isn't set up for the new input system at all.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -402,14 +402,13 @@ public new Camera camera
             set => m_Camera = value;
         }
 
-        //nuke?
         /// <summary>
         /// The event system that should be fed with UI events from the player's inputs.
         /// </summary>
-        public EventSystem uiEventSystem
+        public InputSystemUIInputModule uiInputModule
```
Rene-Damm, line 408: Doc comment needs updating.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -459,19 +458,19 @@ public void ActivateInput()
...
-
             m_InputActive = true;
         }
 
         public void PassivateInput()
         {
-            // Disable all enabled action maps.
-            if (m_Actions != null)
-                m_Actions.Disable();
+            m_EnabledActionMap?.Disable();
```
Rene-Damm, line 473: Seems that we basically we enable two scenarios now:

1) You set a default action map, maybe switch around between a couple maps with `SwitchActions` but only ever use one action map at a time.
2) You set no default action map and handle enabling/disabling maps as needed entirely yourself.

Seems good and simple enough to me for what we have ATM. Could be we need something more eventually (especially WRT handling UI input).

Docs will need updating based on this change. They reflect the old behavior.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -668,6 +670,9 @@ private void InitializeActions()
                     break;
                 }
 
+            if (uiInputModule != null)
+                uiInputModule.actionsAsset = m_Actions;
```
Rene-Damm, line 683: This seems to create an initialization ordering issue between the two systems. If the UI module happens to initialize first, it will go do its thing based on the data it has only to then see the stuff it has set up get stomped when PlayerInput initializes.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -629,6 +629,8 @@ private static PlayerInput DoInstantiate(GameObject prefab)
         // by the receiver. We reuse the same object over and over to avoid allocating garbage.
         [NonSerialized] private InputValue m_InputValueObject;
 
+        [NonSerialized] internal InputActionMap m_EnabledActionMap;
```
Rene-Damm, line 630: Do we want to expose this? Maybe a `currentActionMap` property? The way this works now makes me think we have the concept of a "current" action map now. "Default Action Map" is the one that becomes current by default and `SwitchActions` (which I think should then probably be called `SwitchCurrentActionMap`) changes the currently active one.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
jechter: I don't understand. _If_ you have an `InputSystemUIInputModule` reference assigned here, you also have the UI Event system set up for the new input system. If you don't, then you are not connecting PlayerInput to any UI, so what's the point?

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
jechter: It should still work correctly, though - it just means that it does some unnecessary work. I guess I can change the execution order on PlayerInput to -1, that should fix it, right?

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
jechter: Where are the docs for this?

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
Rene-Damm: The workflow I have in mind here is something like

1) User adds a UI canvas and some UI elements. This implicitly adds an EventSystem configured for the old input system.
2) User adds a PlayerInput.

Now the user has a conflicting setup but without going to the implicitly added EventSystem and reconfiguring the system, the user won't know.

Think it would be great when in step 2 PlayerInput immediately tells the user "you know, there's something I see that I know won't work. want me to fix it?"

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
Rene-Damm: `Components.md` where it mentions "Default Action Map" and "SwichActions".

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
Rene-Damm: Yup, could be that's good enough for now.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
jechter: As we discussed, I'll add a general warning for using "old input" when it is not enabled.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
jechter: I'm not sure about renaming `SwitchCurrentActionMap`. We have a test called `PlayerInput_CanSwitchActionsWithMessage` which shows behavior which hardcodes the name of this function as a string, so I'm scared to change that. But I'd like to understand: what's the use case of this? It will likely cause problems on il2cpp with stripping. Also, should we keep `SwitchActions` at all? The `currentActionMap` is settable, making it redundant (but maybe still convenient  because it uses the action map name?).

Rene-Damm (approved):

PR #583, opened 2019-05-03
NEW: Add tanks demo to new input repo

douglas-piconi:
This is a big checkin, but a very small code review.   Only 3 things I'd like you guys to review.

1. Are you good with where this project sits?  InputSystem\Assets\TanksDemo

Using a Player Input component,...

2. InputSystem\Assets\TanksDemo\Assets\Scripts\Tank\TankMovement.cs
3. InputSystem\Assets\TanksDemo\Assets\Scripts\Tank\TankShooting.cs

This is the initial check-in of the project, with actions working for shooting and movement with gamepads.   More PR's will follow with UI and button mapping in game, and more advanced features.


jechter:
Only looked at the files you mentioned - assuming the rest is unmodified tanks code?

I guess `OnFire` & friends are messages sent by a PlayerInput  component, so no code is needed to call them. Q: Is the message sending player input config our preferred one to demo? Or should we use the Event based setup?

WRT location: IMO InputSystem\Assets\TanksDemo is fine - having it in the project means that we will make sure any changes we do will not break demo compilation. But I guess this project is also only used for Input System developers. I guess the bigger question is, how will we distribute this demo to users? I guess for that purpose it should be a standalone project folder.

jechter (approved):

Assets/TanksDemo/Assets/Editor/MoveSceneViewCamera.cs
```diff
@@ -0,0 +1,15 @@
...
+    [MenuItem("Window/Position Scene View Camera")]
+    static void PositionCamera()
+    {
+        SceneView.lastActiveSceneView.pivot = new Vector3(-147f, 23.5f, 237f);
+        SceneView.lastActiveSceneView.rotation = Quaternion.Euler(0f, 150f, 0f);
+        SceneView.lastActiveSceneView.orthographic = true;
+        SceneView.lastActiveSceneView.size = 100f;
+        Selection.activeGameObject = Camera.main.gameObject;
+    }
+}
```
Rene-Damm, line 16: Do we need this?

Rene-Damm (commented):
Any chance we can stuff TanksDemo into its own .asmdef? Would require adding a bunch of `#if UNITY_EDITOR` to the stuff in the Editor/ folder but would be great to have all the demo stuff in one single chunk.

jechter:
> Any chance we can stuff TanksDemo into its own .asmdef? Would require adding a bunch of `#if UNITY_EDITOR` to the stuff in the Editor/ folder but would be great to have all the demo stuff in one single chunk.

How would we benefit from this? I think there is value in having a minimal diff to the source project.

Rene-Damm:
>How would we benefit from this?

In the generated solution, all the sources would disappear in their own project. Easy to include/exclude.

Also, I'm still hoping that whatever we settle on as a demo, it will have at least some automated tests. Doing that will need the DLL.

>I think there is value in having a minimal diff to the source project.

Seems quite unlikely at this point for Tanks to change. My guess is it's not been touched since Unity 5.

jechter:
> > How would we benefit from this?
> 
> In the generated solution, all the sources would disappear in their own project. Easy to include/exclude.
> 
> Also, I'm still hoping that whatever we settle on as a demo, it will have at least some automated tests. Doing that will need the DLL.
> 
> > I think there is value in having a minimal diff to the source project.
> 
> Seems quite unlikely at this point for Tanks to change. My guess is it's not been touched since Unity 5.

ok.

douglas-piconi:
@Rene-Damm @jechter   This is now an asmdf project with TanksDemo.dll     All the #IF_UNITY_EDITORS have been added  and tested with editor and standalone.  Also has merged in develop branch with all the new namespaces.

PR #589, opened 2019-05-06
Namespace rename

jechter:
https://unity3d.atlassian.net/browse/ISX-56

Changed namespace from `UnityEngine.Experimental.Input` to `UnityEngine.InputSystem`.


Rene-Damm (approved):
Let's see what @JonathanHUnity says about `UnityEngine.InputSystem.InputSystem`. PR looks good to me. Think we've gone for the least of all evils available for us to choose from.

JonathanHUnity:
As I told @jechter , `InputSystem.InputSystem` is not preferred, but there are no rules against it and C# handles this pattern fairly well, so if it's the least of all evils then I'd say go with it.

PR #590, opened 2019-05-06
Change constants to match Unity conventions

jechter:
Addresses https://unity3d.atlassian.net/browse/ISX-209

In many cases, I changed the names of the constants, but in other cases I made the constants internal instead, as I do not believe that these values provide any use to users. These cases should obviously be reviewed, aa in some cases my judgement might be off here. In particular:

-I removed all constants specifying struct byte sizes from public API
-I removed the interface name constants from public API. Stuff like `AndroidSupport.kAndroidInterface`. Now for these, I _can_ see a use in having constant names to compare device description interfaces names against, to tell if it's some specific kind of device. _But_: these names were currently so bad that I'd rather not have them in public API, then to set these bad names in stone. Ie. `AndroidSupport.kAndroidInterface` vs `HID.kHIDInterface` vs `LinuxSupport.kInterfaceName`. If we think these should be public, let's make sure we have consistent names first, and that we have constants for all the interfaces.


Assets/Tests/InputSystem/APIVerification.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    [Test]
+    [Category("API")]
+    public void API_TypesHaveAnAppropriateNamespace()
+    {
+        var incorrectlyNamespacedTypes = GetInputSystemTypes().Where(t => !TypeHasValidNamespace(t));
+        Assert.That(incorrectlyNamespacedTypes, Is.Empty);
+    }
+}
+#endif
```
Rene-Damm: Awesome idea!

Assets/Tests/InputSystem/APIVerification.cs
```diff
@@ -0,0 +1,46 @@
+#if UNITY_EDITOR
+using System;
+using System.Linq;
+using NUnit.Framework;
+using Mono.Cecil;
+using Mono.Collections.Generic;
+
+partial class APIVerification
```
Rene-Damm: Preferably tag "Tests" as a postfix to the class (and file name). IMO nice to immediately spot in file and type requesters what's test stuff and what's "real" stuff.

Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLGamepad.cs
```diff
@@ -10,9 +10,9 @@ namespace UnityEngine.Experimental.Input.Plugins.WebGL.LowLevel
 {
     public unsafe struct WebGLGamepadState : IInputStateTypeInfo
     {
-        public const int kNumAxes = 4;
-        public const int kNumButtons = 16;
-        public const int kButtonOffset = kNumAxes * 4;
+        public const int NumAxes = 4;
+        public const int NumButtons = 16;
+        public const int ButtonOffset = NumAxes * 4;
```
Rene-Damm: `ButtonOffset` probably can be internal.

Rene-Damm (approved):

PR #592, opened 2019-05-07
Remove intptr api

jechter:
Takes care of https://unity3d.atlassian.net/browse/ISX-210

Note that I removed InputHistory.cs instead of updating it, as it's functionality was not implemented. We should not have unimplemented APIs in 1.0. If we think we need it, we should fix it, but not leave it as an empty stub (and I don't believe in keeping an unimplemented version around internally or commented either, as that would just rot).


Rene-Damm (approved):
I have InputHistory mostly implemented on my branch. Fine to delete it for now. I'll bring it back on a merge.

PR #595, opened 2019-05-08
Reduce public fields

jechter:
I took a look at https://unity3d.atlassian.net/browse/ISX-208 "Remove all public fields"

This PR changes all public fields into properties _for which there is no good reason to be a field_. However, it turns out that that is not the case for most fields. We have types with fields for the following reasons:

1. const fields or static readonly fields. 
I believe that these should be allowed. MS uses these in their own public APIs as well.

2. Fields used by serialization logic.
We could make these private and add a property. But then the property and field cannot have the same name, so we either have to change the serialized data or the API. There are two different types of cases of this:
* MonoBehaviours
* Types to match the json reported from native input backends

3. Fields populated by the input system through text parameters
We could make these private and add a property. But then the property and field cannot have the same name, so we either have to change the string format or the API. A better solution would be to make NamedValue handle properties.

4. Input state structures.
These need to be fields for the data layout. We could make these private and add a property. But then the property and field cannot have the same name, so we either have to change the API or the generated layout names.

5. IInputDeviceCommandInfo & IInputEventTypeInfo types
These need to be fields for the data layout. We could make these private and add a property. The downside in this case is that both these types always have a first field for the base struct (as structs have no inheritance). If we made that a property, you could not access single values of the base structure without always copying the whole struct, which seems inefficient.

So, my take is: let's land this PR (which is handling all the easy cases), and make decisions for all the cases outlines above.

I think we should probably change MonoBehaviours (part of 2.) - which would break serialized data in current scenes, but that is probably acceptable at this point and look at supporting properties for 3.. For all the others, my take is that it is not worth trying to change these to properties somehow. Fields in public API are not great, but OTOH, Unity API is already full of them, so I don't think we should bend backwards to avoid it. But we should look at all these APIs to see if there are any which could be internal.


Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
```diff
@@ -42,27 +42,27 @@ public unsafe struct InputStateBlock
...
-        public static FourCC kTypeUInt = new FourCC('U', 'I', 'N', 'T');
-        public static FourCC kTypeShort = new FourCC('S', 'H', 'R', 'T');
-        public static FourCC kTypeUShort = new FourCC('U', 'S', 'H', 'T');
-        public static FourCC kTypeByte = new FourCC('B', 'Y', 'T', 'E');
-        public static FourCC kTypeSByte = new FourCC('S', 'B', 'Y', 'T');
-        public static FourCC kTypeLong = new FourCC('L', 'N', 'G');
-        public static FourCC kTypeULong = new FourCC('U', 'L', 'N', 'G');
-        public static FourCC kTypeFloat = new FourCC('F', 'L', 'T');
-        public static FourCC kTypeDouble = new FourCC('D', 'B', 'L');
+        public static readonly FourCC kTypeBit = new FourCC('B', 'I', 'T');
```
jechter, line 45: Need to rename all these to remove the `k`. Will do that in a follow up PR where I'll change the const naming test coverage to include `static readonly` fields.

Assets/Tests/InputSystem/APIVerificationTests.cs
```diff
@@ -70,5 +121,13 @@ public void API_MethodParameterTypesAreNotIntPtr()
         var intptrMethods = GetInputSystemPublicMethods().Where(m => m.DeclaringType.BaseType?.FullName != "System.MulticastDelegate" && m.Parameters.Any(p => p.ParameterType.FullName == "System.IntPtr"));
         Assert.That(intptrMethods, Is.Empty);
     }
+
+    [Test]
+    [Category("API")]
+    public void API_NoDisallowedPublicFields()
```
Rene-Damm: Nitpick... `APU_DoesNotHaveDisallowedPublicFields`.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlAttribute.cs
```diff
@@ -32,27 +32,33 @@ namespace UnityEngine.InputSystem.Layouts
...
+        public string format { get; set; }
+        public string usage { get; set; }
+        public string[] usages { get; set; }
+        public string parameters { get; set; }
+        public string processors { get; set; }
+        public string alias { get; set; }
+        public string[] aliases { get; set; }
+        public string useStateFrom { get; set; }
+        public uint bit { get; set; }// = InputStateBlock.InvalidOffset
+        public uint offset { get; set; }// = InputStateBlock.InvalidOffset
```
Rene-Damm: What's the reason the defaults are commented out?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayoutAttribute.cs
```diff
@@ -19,16 +19,16 @@ public class InputControlLayoutAttribute : Attribute
...
 
         ////TODO: rename this to just "usages"; "commonUsages" is such a weird name
-        public string[] commonUsages;
+        public string[] commonUsages { get; set; }
 
-        public string variants;
+        public string variants { get; set; }
 
-        public bool? updateBeforeRender;
+        internal bool? updateBeforeRender;
```
Rene-Damm: Why internal?

jechter: This is a weird one - we have two public APIs for this, `updateBeforeRender`, and `beforeRender` - one is nullable and one is not. We should have only one. But I guess keeping `beforeRender` as I did also won't work, as it would throw if `updateBeforeRender` is null. What is this used for? Which ones do we need?

Rene-Damm: Gah, took me a moment to decipher my mess there. Definitely missing a comment.

Looks like I added `beforeRender` because you can't initialize a nullable property through an attribute initializer. So yup, `internal` makes sense and I think it's on the right one.

The only thing I'm thinking is that it'd probably be better to swap the naming `beforeRender` <-> `updateBeforeRender` and let `beforeRender` be internal. And maybe actually call the latter one `updateBeforeRenderInternal` or something like that.

Rene-Damm (approved):

PR #597, opened 2019-05-08
First iteration of moving TouchScreenKeyboard implementation inside input package

todi1856:
What is this?

https://docs.google.com/document/d/1iDsV8WRumTDap23TAyBcVRT7RS3oAslEKtVnBgayivA/edit

About 90% done.

Contains Screen Keyboard implementation for iOS/Android/UWP.

Would be nice to have it in official for for awareness.

Also, feedback is very welcomed


Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+        public bool inputFieldHidden;
+        ////TODO: no characterLimit here, because the logic for characterLimit is too complex when IME composition occurs, instead let user manage the text from OnTextChanged callbac
+    }
+
+
+    public class InputFieldEventArgs
+    {
+        public string text { set; get; }
+    }
+    ////TODO: probably need a better name, so not to collide with com.unity.inputsystem\InputSystem\Plugins\OnScreen\OnScreenKeyboard.cs
```
Rene-Damm, line 54: OnScreenKeyboard was nuked. Do we want this class to be `OnScreenKeyboard` and be part of the `OnScreen` plugin?

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+    public class ScreenKeyboard : Keyboard
+    {
+        private static ScreenKeyboard m_ScreenKeyboard;
```
Rene-Damm: `s_ScreenKeyboard` or just `s_Instance`.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+            m_ScreenKeyboard = InputSystem.AddDevice<UnityEngine.InputSystem.Plugins.Android.AndroidScreenKeyboard>();
+#elif UNITY_WSA
+            m_ScreenKeyboard = InputSystem.AddDevice<UnityEngine.InputSystem.Plugins.WSA.WSAScreenKeyboard>();
+#elif UNITY_IOS || UNITY_TVOS
+            m_ScreenKeyboard = InputSystem.AddDevice<UnityEngine.InputSystem.Plugins.iOS.iOSScreenKeyboard>();
+#elif UNITY_EDITOR
+            // ToDo: Should we show something for Editor?
+            m_ScreenKeyboard = new ScreenKeyboard();
+#else
+            throw new NotImplementedException("ScreenKeyboard is not implemented for this platform.");
```
Rene-Damm: `NotSupportedException` is probably a better fit here.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+        public static ScreenKeyboard GetInstance()
+        {
+            if (m_ScreenKeyboard != null)
+                return m_ScreenKeyboard;
+#if UNITY_ANDROID
+            m_ScreenKeyboard = InputSystem.AddDevice<UnityEngine.InputSystem.Plugins.Android.AndroidScreenKeyboard>();
+#elif UNITY_WSA
+            m_ScreenKeyboard = InputSystem.AddDevice<UnityEngine.InputSystem.Plugins.WSA.WSAScreenKeyboard>();
+#elif UNITY_IOS || UNITY_TVOS
+            m_ScreenKeyboard = InputSystem.AddDevice<UnityEngine.InputSystem.Plugins.iOS.iOSScreenKeyboard>();
```
Rene-Damm: Seems like these should be inside `#if !UNITY_EDITOR`? Or is it intentional that these are created in the editor, too, when the respective platform is selected?

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+#elif UNITY_EDITOR
+            // ToDo: Should we show something for Editor?
+            m_ScreenKeyboard = new ScreenKeyboard();
```
Rene-Damm: Probably find to just have nothing in the editor. Especially since the the editor platform may have an on-screen keyboard, too. At least on Windows, OSK support seems like something that may eventually exist.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+        URL = 3,
+        NumberPad = 4,
+        PhonePad = 5,
+        NamePhonePad = 6,
+        EmailAddress = 7,
+        Social = 8,
+        Search = 9
+    }
+
+    public enum ScreenKeyboardStatus
```
Rene-Damm: `State` rather than `Status`?

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// Show keyboard without input field?
+        /// Only supported on iOS and Android.
+        /// Note: TODO, review this If input field is hidden, you won't receive inputFieldTextChanged callback, instead you'll be receiving onTextInput
+        /// </summary>
+        public bool inputFieldHidden;
+        ////TODO: no characterLimit here, because the logic for characterLimit is too complex when IME composition occurs, instead let user manage the text from OnTextChanged callbac
+    }
+
+
+    public class InputFieldEventArgs
```
Rene-Damm: internal?

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+    {
+        Visible,
+        Done,
+        Canceled,
+        LostFocus
+    }
+
+
+    [StructLayout(LayoutKind.Sequential)]
+    public struct ScreenKeyboardShowParams
```
Rene-Damm, line 33: Would prefer this as a protected struct hidden inside ScreenKeyboard.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+    ////TODO: probably need a better name, so not to collide with com.unity.inputsystem\InputSystem\Plugins\OnScreen\OnScreenKeyboard.cs
+    public class ScreenKeyboard : Keyboard
+    {
+        private static ScreenKeyboard m_ScreenKeyboard;
+
+        internal ScreenKeyboardStatus m_Status;
+        internal InlinedArray<Action<ScreenKeyboardStatus>> m_StatusChangedListeners;
+        internal InlinedArray<Action<InputFieldEventArgs>> m_InputFieldTextListeners;
+
+        public static ScreenKeyboard GetInstance()
```
Rene-Damm: Just `Get()`? `ScreenKeyboard.Get()` seems nicer than `ScreenKeyboard.GetInstance()`.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,155 @@
...
+        ////TODO: no characterLimit here, because the logic for characterLimit is too complex when IME composition occurs, instead let user manage the text from OnTextChanged callbac
+    }
+
+
+    public class InputFieldEventArgs
+    {
+        public string text { set; get; }
+    }
+    ////TODO: probably need a better name, so not to collide with com.unity.inputsystem\InputSystem\Plugins\OnScreen\OnScreenKeyboard.cs
+    public class ScreenKeyboard : Keyboard
```
Rene-Damm: Hmm, wondering... what's the usage we get out of basing this on `Keyboard`? Is any of the `Keyboard` functionality actually functional on `ScreenKeyboard`? If not, doesn't seem like it should be a device. Input devices are all based on a state and event model that doesn't really seem to apply to `ScreenKeyboard`.

Rene-Damm:
I really like that we have a first piece of native input functionality that is moving into the package. Hoping this is just the start.

Think we still need to massage the API a bit but otherwise it's looking good to me.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Not familiar with OnScreen plugin, but looks like it's a set of screen controls.  So maybe, seems like ScreenKeyboard would be a good fit

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Ok

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Ok

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Not intentional, good catch

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: ok

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: This is actually a struct used by user

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: ok

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: That's a good question,  need to think about it.

But there are some shared pieces:
- Keyboard.onTextInput, it will be called when ScreenKeyboard doesn't have an input pane
- Most of mobile keyboard don't support key events even though you click "keys"  on screen keyboard (you only receive text event), but I think there might be instances of screen keyboard which will support key events

Maybe I should try to rewrite the implementation with states, then it will be closer to other input devices.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Remove the whole code with singleton. User will simply do InputSystem.GetDevice<ScreenKeyboard>

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Removed the code

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Removed the whole code

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Removed this code

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: remove singletone pattern

PR #603, opened 2019-05-10
Add explicit dependencies to ugui, now that it is a package

jechter:
Unity 19.2 is changing UGUI to become a package. That means that we now need to explicitly reference it. They have been foresighted enough to add an empty stub ugui package to 19.1, so this should not be a backwards-compatibility problem for us.

The question is if we should try to do without the hard-coded dependency on the ugui package, so you could use input without dragging UI into your project (which itself will drag more things into the project), which is relevant for making small mobile or web games. I believe we can do that using the "conditional version define" feature in asmdef files, where we could add a define to be set for our code if the ugui package is present, and we could thus conditionally not compile any parts of our code base depending on ugui if it is not used in the project.

The question is, do we want to that? It should not be too hard to set it up, but:
- Some functionality in the input system (on screen buttons and sticks) depend on ugui. So you would  lose this functionality if you don't add ugui to your project, which might be confusing.
- UGUI use is scattered around our project enough that it seems likely we'd break the conditional compilation at some point if we don't specifically test this setup. Guess we'd ideally need some setup in our CI to test with and without ugui. This would be some work.

So, with those open questions, I decided to keep it simple first, and keep this in mind as an optional next step. Thoughts?

### DIFF
--- a/Assets/Demo/Demo.asmdef
+++ b/Assets/Demo/Demo.asmdef
@@ -3,7 +3,8 @@
     "references": [
         "Unity.InputSystem",
         "Unity.TextMeshPro",
-        "Steamworks.NET"
+        "Steamworks.NET",
+        "Unity.ugui"
     ],
     "optionalUnityReferences": [],
     "includePlatforms": [],
--- a/Packages/com.unity.inputsystem/package.json
+++ b/Packages/com.unity.inputsystem/package.json
@@ -19,5 +19,7 @@
 		"vr",
 		"xr"
 	],
-	"dependencies": {}
+	"dependencies": {
+		"com.unity.ugui": "1.0.0"
+	}
 }
--- a/Assets/Tests/DemoTests.asmdef
+++ b/Assets/Tests/DemoTests.asmdef
@@ -3,7 +3,8 @@
     "references": [
         "Demo",
         "Unity.InputSystem",
-        "Unity.InputSystem.TestFramework"
+        "Unity.InputSystem.TestFramework",
+        "Unity.ugui"
     ],
     "optionalUnityReferences": [
         "TestAssemblies"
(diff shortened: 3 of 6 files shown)


Assets/Demo/Demo.asmdef
```diff
@@ -3,7 +3,8 @@
     "references": [
         "Unity.InputSystem",
         "Unity.TextMeshPro",
-        "Steamworks.NET"
+        "Steamworks.NET",
+        "Unity.ugui"
```
Rene-Damm, line 7: Ugh... makes me think we *really* want that separation of Unity.InputSystem and UI-related parts...

Packages/com.unity.inputsystem/package.json
```diff
@@ -19,5 +19,7 @@
 		"vr",
 		"xr"
 	],
-	"dependencies": {}
+	"dependencies": {
+		"com.unity.ugui": "1.0.0"
```
Rene-Damm, line 23: Does this affect the min Unity 2019.1 version we support? Which public version has the change with the empty uGUI package?

Assets/Demo/Demo.asmdef
jechter: Not for this. The Demo code references UI code directly. So now it needs to reference the UI assembly. This has _nothing_  to do with the input system.

jechter: We might still want it, but then we'd have to do the work outlined in the PR description.

Packages/com.unity.inputsystem/package.json
jechter: It took me some digging, but it seems that this was added  in 2019.1.0b2 (making that our minimum supported version)

Assets/Demo/Demo.asmdef
Rene-Damm: >The Demo code references UI code directly.

Ah right, forgot about that.

Rene-Damm (approved):

PR #604, opened 2019-05-10
More const name fixes

jechter:
This prior PR updated the names of all `const` fields to match Unity conventions: https://github.com/Unity-Technologies/InputSystem/commit/3a3988db691ececdbc646a9ac269423bcc867df0

However, there are some types of constants which are not `const` fields, and were thus not caught by the API test:
-`static readonly` fields
-`enum` values

This PR adds tests for those cases, and renames the ones which did not match conventions (was only the case for some `static readonly` fields).

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
@@ -32,7 +32,7 @@ namespace UnityEngine.InputSystem.LowLevel
     ///
     /// State memory is bit-addressable, meaning that it can be offset from a byte address in bits (<see cref="bitOffset"/>)
     /// and is sized in bits instead of bytes (<see cref="sizeInBits"/>). However, in practice, bit-addressing
-    /// memory reads and writes are only supported on the <see cref="kTypeBit">bitfield primitive format</see>.
+    /// memory reads and writes are only supported on the <see cref="TypeBit">bitfield primitive format</see>.
     ///
     /// Input state memory is restricted to a maximum of 4GB in size. Offsets are recorded in 32 bits.
     /// </remarks>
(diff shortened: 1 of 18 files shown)


Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
```diff
@@ -42,89 +42,89 @@ public unsafe struct InputStateBlock
...
+        public static readonly FourCC TypeInt = new FourCC('I', 'N', 'T');
+        public static readonly FourCC TypeUInt = new FourCC('U', 'I', 'N', 'T');
+        public static readonly FourCC TypeShort = new FourCC('S', 'H', 'R', 'T');
+        public static readonly FourCC TypeUShort = new FourCC('U', 'S', 'H', 'T');
+        public static readonly FourCC TypeByte = new FourCC('B', 'Y', 'T', 'E');
+        public static readonly FourCC TypeSByte = new FourCC('S', 'B', 'Y', 'T');
+        public static readonly FourCC TypeLong = new FourCC('L', 'N', 'G');
+        public static readonly FourCC TypeULong = new FourCC('U', 'L', 'N', 'G');
+        public static readonly FourCC TypeFloat = new FourCC('F', 'L', 'T');
+        public static readonly FourCC TypeDouble = new FourCC('D', 'B', 'L');
```
Rene-Damm: Looking at the use of these, I think `Type` was a misnomer here. They are format identifiers, not type identifiers. The respective API in `IInputStateTypeInformat` is `GetFormat` and the respective InputStateBlock field is `format`. Should we rename all of these to be `FormatBit` etc?

jechter: Makes sense. Fixed.

Rene-Damm (approved):

PR #608, opened 2019-05-13
Make Input System package compile with Unity 19.3

jechter:
1. IEditorPlayerConnection has been changed to add Unregister methods
2. UnityEngine.Experimental.PlayerLoop is removed - and it was not used in this file anyways

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Remoting.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Remoting.cs
@@ -243,7 +243,7 @@ partial class CoreTests
         {
             m_DisconnectionListeners.RemoveListener(callback);
         }
-        
+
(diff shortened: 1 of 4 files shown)


douglas-piconi:
Do you guys think I should add 19.2, 19.3  to the CI runs?  I only had it running 19.1 at first to keep it simple.

jechter:
>Do you guys think I should add 19.2, 19.3 to the CI runs? I only had it running 19.1 at first to keep it simple.

I guess it's a question of resources. I think adding il2cpp is a higher priority for now. But the, yes, I think having at least oldest & newest supported version is probably a good idea. Maybe not for all permutations (win/mac, player/editor)?

Rene-Damm (approved):

douglas-piconi (approved):

PR #614, opened 2019-05-15
CHANGE: Assigned icons for scripts

jechter:
This PR assigns default icons for all the MonoBehaviours we implement.

These are not newly designed icons. PlayerInput and PlayerInputManager use the InputAction icon (which I feel represents the Input System as a whole well, so makes sense to have these associated with the input system).

MultiplayerEventSystem and InputSystemUIInputModule use the icons of their respective cousins in UGUI module.

It would probably be nice to design custom icons for all of these, but I still feel that this is an improvement over using the generic "script" icons.


douglas-piconi (approved):
all binary!  easy review!

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Change all public API using `IntPtr` to use unsafe pointer types instead.
 - `PlayerInput` will no longer disable any actions not in the currently active action map when disabling input or switching action maps.
 - Change some public fields into properties.
+- Assigned icons for scripts.
```
Rene-Damm: Bit cryptic. "Added icons for `PlayerInput`, `PlayerInputManager`, `InputSystemUIInputModule` and `MultiplayerEventSystem` components."

Rene-Damm (approved):

PR #618, opened 2019-05-16
 Rename cancelled -> canceled (US spelling)

jechter:
This was pointed out by the Code analyzer. "Cancelled" is the UK spelling, "Canceled" is the preferred US spelling.

We don't want to have another "MonoBehaviour" stuck in our API.


douglas-piconi (approved):
Interesting, I have always spelled it Cancelled in the U.S.

Rene-Damm (approved):

jechter:
> Interesting, I have always spelled it Cancelled in the U.S.

Yeah, apparently, both spellings are valid in the US (with one "l" being the more common one). MS API guidelines do however suggest using one "l" (See https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1726-use-preferred-terms)

PR #622, opened 2019-05-17
Remove all unused parameters

jechter:
Fixes `CA1801: Review unused parameters` Code Analysis warning.

Many of these are unimplemented functions. I removed all of these. We should not have unimplemented functions in public API, that provides no value to users. If we want these as "TODO" items, let's put TODO comments into the code instead.


Rene-Damm (commented):
While cleaning up is definitely desirable, with some of the APIs in here I'd prefer the approach I have pursued so far with making unfinished things internal rather than nuking them outright. There's a good number of deletions here in the PR of stuff that we won't be missing but some stuff is genuinely unfinished things. Simply killing off the unfinished items IMO leaves things worse off as now the indication of what needs work disappears.

douglas-piconi:
You can cherry pick d6a21d85d24140f5cd8fef4787aebc27cd04f71a    to get these tests green again.

jechter:
@Rene-Damm:
>There's a good number of deletions here in the PR of stuff that we won't be missing but some stuff is genuinely unfinished things. Simply killing off the unfinished items IMO leaves things worse off as now the indication of what needs work disappears.

Ok. So then let's figure out which is which, and I'll update the PR.

But, for the ones we want to keep as "indication of what needs work", I'd still prefer that in the form of comment (can be commented out code), then keeping the code around, plastered with suppression attributes to make the code analyzer shut up (most of these fail both `CA1801: Review unused parameters` as well as `CA1065: Do not raise exceptions in unexpected locations`).

douglas-piconi (approved):

Rene-Damm (approved):

PR #627, opened 2019-05-20
FIX: Fix all exception-related code analyizer warnings

jechter:
Fixes the following warnings from the code analyzer:

`CA1032: Implement standard exception constructors`
`CA1064: Exceptions should be public`
`CA1065: Do not raise exceptions in unexpected locations` (except for APIs already removed by https://github.com/Unity-Technologies/InputSystem/pull/622) Note: commented out  some unimplemented API for this.
`CA2200: Rethrow to preserve stack details`
`CA2201: Do not raise reserved exception types`
`CA2208: Instantiate argument exceptions correctly`


Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
```diff
@@ -574,12 +574,12 @@ private int ResolveInteractions(string interactionString)
                 // Look up interaction.
                 var type = InputInteraction.s_Interactions.LookupTypeRegistration(m_Parameters[i].name);
                 if (type == null)
-                    throw new Exception(
-                        $"No interaction with name '{m_Parameters[i].name}' (mentioned in '{interactionString}') has been registered");
+                    throw new ArgumentException(
```
Rene-Damm: The remaining exceptions in this file don't seem like a good fit for ArgumentException. Seems to indicate misuse of an API which, given these are internal APIs, seems to point at a bug in the input system. However, the exceptions here are more about incorrect data. What do you think about throwing InvalidDataException?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -1173,7 +1174,7 @@ internal static bool VariantsMatch(string expected, string actual)
             foreach (var existing in controlLayouts)
                 if (string.Compare(name, existing.name, StringComparison.OrdinalIgnoreCase) == 0 &&
                     existing.variants == controlItem.variants)
-                    throw new Exception($"Duplicate control '{name}' in layout '{layoutName}'");
+                    throw new ArgumentException($"Duplicate control '{name}' in layout '{layoutName}'", nameof(layoutName));
```
Rene-Damm: InvalidDataException?

Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceBuilder.cs
```diff
@@ -115,8 +115,8 @@ public TControl TryGetControl<TControl>(InputControl parent, string path)
 
             var controlOfType = control as TControl;
             if (controlOfType == null)
-                throw new Exception(
-                    $"Expected control '{path}' to be of type '{typeof(TControl).Name}' but is of type '{control.GetType().Name}' instead!");
+                throw new ArgumentException(
```
Rene-Damm: Same thoughts here and for the following internal methods about ArgumentException hinting at an internal bug rather than a data issue.

Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
jechter: `InvalidDataException` is from the `System.IO` namespace, and specifically defined as being related to Streams ("The exception that is thrown when a data stream is in an invalid format."). I don't think it makes sense here. The exception I have been using for generic cases of "our data/state is not valid" is `InvalidOperationException` ("The exception that is thrown when a method call is invalid for the object's current state."). We could change this, but I don't see much gain or loss of clarity either way. I think in this case the data which is invalid is coming from the argument and I don't see any guideline saying that `ArgumentException` should only be used on arguments in public APIs.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
jechter: No, same as above. Could change it to `InvalidOperationException`, don't see much of a win (or loss) by doing so.

douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
Rene-Damm: The docs for ArgumentException state it "indicates developer error".

I don't see the docs for `InvalidDataException` mentioning it's for streams only. To me seems the most logical choice.

Fine with InvalidOperationException, too.

jechter: >I don't see the docs for InvalidDataException mentioning it's for streams only. To me seems the most logical choice.

https://docs.microsoft.com/en-us/dotnet/api/system.io.invaliddataexception?view=netframework-4.8
"The exception that is thrown when a data stream is in an invalid format."

Also, the `System.IO` namespace IMO clearly  associates this with IO.

Will go with `InvalidOperationException`.

Rene-Damm (approved):

PR #631, opened 2019-05-21
Fix IDisposable-related anaylizer warnings

jechter:
Fixes:

`CA1001: Types that own disposable fields should be disposable`
`CA1063: Implement IDisposable correctly` - This one is about making sure that classes which implement `IDisposable` and can be derived from also implement `protected void Dispose(bool)` for the deriving class to use in its `Dispose` implementation. Now, I "solved" this the easy way by sealing our disposable classes, simply because I don't see a need for users to inherit from these, and this leaves less margin for mistakes. However, if you think that this is too restrictive, then I can fix this by implementing `protected void Dispose(bool)`. In general, should we default to sealing our classes if we don't specifically develop them with inheritance in mind?

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
@@ -67,7 +67,6 @@ namespace UnityEngine.InputSystem
     /// </remarks>
     [Serializable]
     public sealed class InputAction : ICloneable, IDisposable
-        ////REVIEW: should this class be IDisposable? how do we guarantee that actions are disabled in time?
     {
         /// <summary>
         /// Name of the action.
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
@@ -66,7 +66,7 @@ namespace UnityEngine.InputSystem
     /// Actions are not supported in edit mode.
     /// </remarks>
     [Serializable]
-    public class InputAction : ICloneable, IDisposable
+    public sealed class InputAction : ICloneable, IDisposable
         ////REVIEW: should this class be IDisposable? how do we guarantee that actions are disabled in time?
     {
         /// <summary>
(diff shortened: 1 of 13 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -66,7 +66,7 @@ namespace UnityEngine.InputSystem
     /// Actions are not supported in edit mode.
     /// </remarks>
     [Serializable]
-    public class InputAction : ICloneable, IDisposable
+    public sealed class InputAction : ICloneable, IDisposable
```
Rene-Damm, line 69: Makes me realize the ////REVIEW comment below is outdated.

Rene-Damm:
>Now, I "solved" this the easy way by sealing our disposable classes, simply because I don't see a need for users to inherit from these, and this leaves less margin for mistakes. 

Sounds fine to me.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPickerDropdown.cs
```diff
@@ -52,6 +52,11 @@ protected override void OnDestroy()
             m_RebindingOperation = null;
         }
 
+        public void Dispose()
+        {
+            m_RebindingOperation?.Dispose();
+        }
```
Rene-Damm, line 58: Should these ones that now have IDisposable on them implement a full dispose pattern with a finalizer?

Rene-Damm (approved):

douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPickerDropdown.cs
jechter: Good question.

I _think_ we don't need to. Reading https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose, it seems that for the IDisposable pattern we should implement a Dispose(bool), where the bool indicates if Dispose was called from a finalizer or directly. If it was called from a finalizer, we should only dispose our own native data, but not call dispose on other managed objects (as they would get their own finalizer called) - which in this case would mean the finalizer would do nothing.

PR #634, opened 2019-05-22
Misc analyzer fixes

jechter:
Fix or suppress various issues reported by the code analyzer:

`CA1012: Abstract types should not have constructors`
`CA1036: Override methods on comparable types`
`CA1040: Avoid empty interfaces`
`CA1504: Review misleading field names`
`CA1707: Identifiers should not contain underscores`
`CA1812: Avoid uninstantiated internal classes`
`CA1813: Avoid unsealed attributes`
`CA1820: Test for empty strings using string length`
`CA2217: Do not mark enums with FlagsAttribute`
`CA2225: Operator overloads have named alternates`
`CA2241: Provide correct arguments to formatting methods`


Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -1040,12 +1040,7 @@ public void TODO_Layouts_CanBuildCustomLayoutsBasedOnTheHIDLayoutBuilder()
...
     {
-        string usagePage;
-        string usage;
-
-        HID.UsageToString((HID.UsagePage) 0xff01, 0x33, out usagePage, out usage);
-
-        Assert.That(usagePage, Is.EqualTo("Vendor-Defined"));
-        Assert.That(usage, Is.EqualTo("Vendor-Defined"));
+        Assert.That(HID.UsagePageToString((HID.UsagePage) 0xff01), Is.EqualTo("Vendor-Defined"));
+        Assert.That(HID.UsageToString((HID.UsagePage) 0xff01, 0x33), Is.Null);
```
jechter, line 1044: Note this change in behavior and API. I figured that returning null would be more consistent than returning "Vendor-Defined", when we return null in all other cases of "we don't know the usages for this page".

Rene-Damm: (Crap, hit resolve accidentally)

Rene-Damm (approved):

douglas-piconi (approved):

PR #638, opened 2019-05-22
CHANGE: Replace some `GetXxx` methods in our API with `xxx`  properties.

jechter:
Fixes `CA1024: Use properties where appropriate`.

I think API wise this analyzer recommendation makes sense, these _are_ better exposed as getters, IMO. but it will cause a disruption to current users. We can also chose to not do this and suppress the warnings, or to offer both (and deprecate the old one?).

### DIFF
--- a/Assets/Demo/SimpleDevice.cs
+++ b/Assets/Demo/SimpleDevice.cs
@@ -13,9 +13,9 @@ public struct MyDeviceState : IInputStateTypeInfo
     [InputControl(layout = "Axis")]
     public float axis1;
 
-    public FourCC GetFormat()
+    public FourCC format
     {
-        return new FourCC('M', 'Y', 'D', 'V');
+        get { return new FourCC('M', 'Y', 'D', 'V'); }
     }
 }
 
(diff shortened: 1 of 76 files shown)


douglas-piconi:
Looks like there is a windows only compile error on this one with  XInputControllerWindowsState

douglas-piconi (approved):

Rene-Damm (approved):
Think I'd prefer `GetFormat()` (which probably should be `GetFormatStatic()` for consistency) and `GetTypeStatic()`  as these aren't really meant to be properties of the resulting structure but are meant as a solution to retrieve these things through the type. But don't feel too strongly about it so fine either way.

jechter:
> Think I'd prefer `GetFormat()` (which probably should be `GetFormatStatic()` for consistency) and `GetTypeStatic()` as these aren't really meant to be properties of the resulting structure but are meant as a solution to retrieve these things through the type. But don't feel too strongly about it so fine either way.

I don't feel strongly either, so we can suppress and revert if you prefer. But I guess the reasoning  of the analyzer makes sense - a simple value like this with no big cost associated to it is more commonly associated with a property then a method. And I don't think being static on the type changes much at  all about that.

PR #639, opened 2019-05-22
CHANGE: no set-only properties

jechter:
Fixes `CA1044: Properties should not be write only`


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -198,14 +215,43 @@ private void OnFocusChanged(bool focus)
...
 
         public Action onProjectChange
         {
-            set => EditorApplication.projectChanged += value;
+            get => m_OnProjectChanged;
+            set
+            {
+                if (value == null)
+                    EditorApplication.projectChanged -= OnProjectChanged;
+                else if (m_FocusChangedMethod == null)
```
Rene-Damm: What's the m_FocusChangedMethod check here and in the setter above about?

jechter: oops, copy-paste mistake - will fix

Rene-Damm (approved):

PR #642, opened 2019-05-23
Make CI run code analyzer

jechter:
Run the roslyn code analyzers on the input system package in our CI.

This performs a lot of checks on the code and API we have to check for API conventions and for potential bugs in the code. By enforcing this in CI, we are committing to a higher code quality and API standard (but also adding a burden of potentially having to spend extra time on PRs to fix or suppress warnings).

I currently still suppress three warning types I think we should possibly fix, which only showed up in CI (which I think uses a different version of the analyzer package then my VS install did). I will review these and add PRs to fix after this, but rather wanted to get the CI in first, so we don't regress elsewhere.

I am committing our code analyzers from the private repo https://github.cds.internal.unity3d.com/unity/api-analyzers to this public repo. I'm assuming that's ok, as there is nothing apparently secret in them. @JonathanHUnity , please let me know if not, then I will delete this branch.


CodeAnalyzerTestProject/Assets/Tests/CleanConsoleTest.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace PackageTestSuite
+{
+    public class CleanConsoleTest
+    {
+        // Make sure it is run before anything else that can affect the console
+        [Test, Order(1)]
+        public void TestCleanConsole()
+        {
+            string logs = Utilities.GetLogs();
+			Assert.That(logs, Is.Empty, string.Format("Found logs in the console on project startup.\n\nLogs: {0}",logs));
```
Rene-Damm: Isn't this one already performed automatically by the package checks?

Rene-Damm (commented):
Can we put this in Tools/CodeAnalyzerTestProject or something? Or maybe even .yamato/CodeAnalyzerTestProject?

CodeAnalyzerTestProject/Assets/Tests/CleanConsoleTest.cs
jechter: Yes (this test is copied from the package validation suite), _but_ in this project we enabled the code analyzer (using the csc.rsp file), so we will also get warnings from the code analyzer this way (and, in case you want to ask next: yes, it would be nicer if we would not copy the code, but use it from the suite, but no, doing that is not at all easy).

jechter:
> Can we put this in Tools/CodeAnalyzerTestProject or something? Or maybe even .yamato/CodeAnalyzerTestProject?

`Tools/CodeAnalyzerTestProject` is fine with me (we don't have a Tools folder today, so that won't make the hierarchy any cleaner, but maybe in the future). I don't like `.yamato/CodeAnalyzerTestProject`. You may actually want to open the project to locally check analyzer issues, and making that harder is not helping.

PR #645, opened 2019-05-23
NEW: in game button remapping screen Tanks demo

douglas-piconi:
@Rene-Damm one issue in this review.  I could not get the PerformInteractiveRebinding()  to apply the override no matter how I configured the rebind.  Spent hours trying to get it to work but it just would not apply the override when I enabled the map.   The only way I could get this to work was using WithRebindAddingNewBinding()  and removing the original binding.   Either I am missing something really obvious,  or there is a problem with the override becoming the active binding.


Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
```diff
@@ -0,0 +1,86 @@
...
+        uiButton.enabled = false;
+        okButtonMessageText.text = "Press button/stick for " + name;
+
+        var action = playerActionMap.actions[mapIndex];
+        var oldMappingPath = action.bindings[0].path;
+
+        ///REVIEW I could only get this to work by having the rebind add, and delete the old one.
+        /// the behavior of apply override does nto seem to work
+        // Remove the old binding
+        action.ChangeBindingWithPath(oldMappingPath).Erase();
```
jechter: So if you cancel the rebinding, does that mean, you will have lost your old binding?

jechter:
I guess one thing which might be nice to add in a follow up PR is storing the bindings as preferences. Because this will be something everyone who does rebindings will want to do, and would likely look for pointers on what's the best way to do that.

Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
```diff
@@ -0,0 +1,86 @@
...
+        //remap and add a new binding
+        rebindOperation = action.PerformInteractiveRebinding().WithRebindAddingNewBinding().OnComplete(operation => ButtonRebindCompleted(uiButton)).Start();
+    }
+
+    private void ResetButtonMappingTextValues()
+    {
+        gasMappingValueText.text = playerActionMap.actions[gasMapIndex].bindings[0].path;
+        brakeMappingValueText.text = playerActionMap.actions[brakeMapIndex].bindings[0].path;
+        fireMappingValueText.text = playerActionMap.actions[fireMapIndex].bindings[0].path;
+        turretMappingValueText.text = playerActionMap.actions[turretMapIndex].bindings[0].path;
```
Rene-Damm: Using `InputControlPath.ToHumanReadableString()` should give very slightly more readable output.

Assets/TanksDemo/Assets/TanksInputActions.inputactions
```diff
@@ -48,21 +48,20 @@
                     "processors": "",
                     "interactions": "Press(behavior=2)",
                     "bindings": []
+                },
+                {
+                    "name": "Pause",
+                    "id": "020abfe6-6083-46ec-9487-3aa2ecb330f3",
+                    "expectedControlLayout": "",
```
Rene-Damm: Strongly recommend setting action types / expectedControlLayouts on actions. Without them, the rebinding apparatus has no idea what it's looking for and just blindly filters for the most likely input.

douglas-piconi:
@jechter @Rene-Damm   Ok, issues have been addressed.  Turns out I never needed the Erase(),  there is a bug with the second controller does not get the rebind, however the first one actually does.  In a separate PR I will address that issue along with more unit tests.      Rene,  effectivePath is an internal member, hence the change I made to have the binding ToString()  which resolves the effectivePath.

Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
douglas-piconi: removed the Erase, no longer an issue.

Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
douglas-piconi: fixed

Assets/TanksDemo/Assets/TanksInputActions.inputactions
douglas-piconi: fixed

jechter (approved):

Rene-Damm:
>Rene, effectivePath is an internal member

Doh, didn't realize. That shouldn't be the case. I'll add a TODO. I'm thinking that `path` should be what `effectivePath` is today. Would make the API more obvious I think.

Rene-Damm (approved):

PR #651, opened 2019-05-29
NEW: Added support for generic joysticks on WebGL (which don't use th…

jechter:
…e standard gamepad mapping).

I was testing gamepads on WebGL, and found that many devices (depening on browser) don't get the standard mapping and were not supported. This PR supports these by building a custom layout to match the device.

Note that I made `WebGLJoystick` not derive from `Joystick` any more. This is because Joystick has a `trigger` and `hat` control, which I could not assign in a reasonable way (as we have no understanding of what button is the main trigger). I'm not clear what issues this might cause?


Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLSupport.cs
```diff
@@ -10,22 +14,126 @@ namespace UnityEngine.InputSystem.WebGL
...
+                return null;
+
+            var deviceMatcher = InputDeviceMatcher.FromDeviceDescription(description);
+
+            var layout = new WebGLLayoutBuilder {capabilities = WebGLDeviceCapabilities.FromJson(description.capabilities)};
+            InputSystem.RegisterLayoutBuilder(() => layout.Build(),
+                description.product, null, deviceMatcher);
+
+            return description.product;
+        }
```
Rene-Damm, line 48: Wondering, how does the fallback logic work now? Previously, the matcher for gamepads with standard mappings was "more specific" (i.e. matched on more fields) than the one for joystick, so it took precedence. Now, the fallback is hooked into onFindLayoutForDevice and it seems to kick in for every WebGL device so it seems like it is overriding the WebGLGamepad case. But I assume the tests are passing so there must be something I'm not seeing.

jechter: The answer is quite simple: It didn't (and the test was failing). Fixed.

Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLJoystick.cs
```diff
@@ -1,10 +1,8 @@
 #if UNITY_WEBGL || UNITY_EDITOR
 
-////TODO
-
 namespace UnityEngine.InputSystem.WebGL
 {
-    public class WebGLJoystick : Joystick
+    public class WebGLJoystick : InputDevice
```
Rene-Damm: What's the reason for this one?

Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLSupport.cs
```diff
@@ -10,22 +14,130 @@ namespace UnityEngine.InputSystem.WebGL
...
+
+        internal static string OnFindLayoutForDevice(int deviceId, ref InputDeviceDescription description,
+            string matchedLayout, IInputRuntime runtime)
+        {
+            // If the device isn't a WebGL device, we're not interested.
+            if (string.Compare(description.interfaceName, InterfaceName, StringComparison.InvariantCultureIgnoreCase) != 0)
+                return null;
+
+            // If it was matched by the standard mapping, we don't need to fall back to generating a layout.
+            if (matchedLayout == typeof(WebGLGamepad).Name)
```
Rene-Damm: I think 

```
if (!string.IsNullOrEmpty(matchedLayout))
```

would be a better solution. This way we can go and add support for other WebGL gamepads/joysticks without this breaking.

Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLJoystick.cs
jechter: See PR description.

Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLSupport.cs
jechter: I tried that first, but it does not work, because all WebGL devices are matched by a generic "Gamepad" layout otherwise. Any idea why?

Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLJoystick.cs
jechter: I specifically asked you about this on slack two minutes after you posted this? :)

Packages/com.unity.inputsystem/InputSystem/Plugins/WebGL/WebGLSupport.cs
Rene-Damm: Argh, that is annoying. Didn't consider that. But maybe 

```
if (!string.IsNullOrEmpty(matchedLayout) && matchedLayout != "Gamepad")
```

then?

jechter: ok

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -6464,8 +6464,8 @@ public void Actions_InteractiveRebinding_CanBeRestrictedToSpecificControlPaths()
             Assert.That(rebind.candidates, Has.Count.EqualTo(4));
             Assert.That(rebind.candidates, Has.Exactly(1).SameAs(keyboard.spaceKey));
             Assert.That(rebind.candidates, Has.Exactly(1).SameAs(mouse.leftButton));
-            Assert.That(rebind.candidates[2], Is.SameAs(mouse.button));
-            Assert.That(rebind.candidates[3], Is.SameAs(keyboard.anyKey)); // Last place for AnyKey.
+            Assert.That(rebind.candidates[3], Is.SameAs(mouse.button));
+            Assert.That(rebind.candidates[2], Is.SameAs(keyboard.anyKey)); // Last place for AnyKey.
```
jechter: Note: turning AnyKeyControl into a ButtonControl flipped the order here. I think that should be fine, as long as space key comes first?

jechter:
Update: WebGLJoystick now derives from Joystick, with trigger being configured to trigger on all buttons.

Rene-Damm (approved):

Assets/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: Yup, should be fine. Neither of the two synthetic buttons seems a better fit than the other.

douglas-piconi (approved):

PR #655, opened 2019-06-01
CHANGE: Rework Touchscreen and add a new touch API.

Rene-Damm:
- `Touchscreen` is back to 10 touches and always keeps the Nth touch on the Nth TouchControl.
- Touches now also record start position and start time (`TouchControl.startPosition` and `TouchControl.startTime`).
- Add primary touch recognition (`Touchscreen.primaryTouch`).
- Add tap recognition (`TouchControl.tap` and `TouchControl.tapCount`)
- Phases are no longer a `Pointer`-level concept but are specific to touches now.
- `Pointer.button` is now `Pointer.press`, a button that indicates "engagement" of the pointer.
- Each individual TouchControl can now be bound from the action editor GUI.
- New `EnhancedTouch` plugin adds `Touch` and `Finger` API which provide a better, more flexible and robust API than the old `Touchscreen.activeTouches` getter. Also automatically records touch history.
- TouchInputSamples are now integrated into the repository.


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -23,9 +24,64 @@
 [SuppressMessage("ReSharper", "AccessToStaticMemberViaDerivedType")]
 
 // As should be obvious from the number of tests in here, the action system rivals the entire combined rest of the system
-// in terms of complexity.
+// in terms of complexity :(
```
jechter, line 27: :(

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1730,1484 +1849,996 @@ public void Actions_PressingAndReleasingButtonInSameUpdate_StillTriggersAction()
...
-        };
-        action.canceled +=
-            ctx =>
-        {
-            ++canceledReceivedCalls;
-            canceledAction = ctx.action;
-            canceledControl = ctx.control;
+    // We used to require string GUIDs to be using a "{...}" format when looking up actions. We no
+    // longer do this but there's still data that may be using this format so make sure it works for now.
+    [Test]
```
jechter, line 1979: I think we discussed this before, but do we want to carry this into 1.0?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -7072,4 +5788,194 @@ public void Actions_CanIterateOverActionsInMap()
 
         Assert.That(map.ToList(), Is.EquivalentTo(new[] { action1, action2, action3 }));
     }
+
```
jechter, line 5791: The refactor of this file above makes it incredibly hard to review (and to find the interesting bits below). Can you try to do refactors like that in separate PRs in the future for the sake of reviewability?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -7072,4 +5788,194 @@ public void Actions_CanIterateOverActionsInMap()
...
+            Assert.That(trace,
+                Started(touch1Action, touchscreen.touches[1].position, new Vector2(0.234f, 0.345f))
+                    .AndThen(Performed(touch1Action, touchscreen.touches[1].position, new Vector2(0.234f, 0.345f))));
+
+            trace.Clear();
+
+            EndTouch(1, new Vector2(0.345f, 0.456f));
+
+            Assert.That(trace,
```
jechter, line 5844: Q: is this what should happen? If I lift a finger, the secondary touch should drive the UI? Looking at native apps on my iPhone, that doesn't seem to be how it works.

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -2241,124 +2273,96 @@ public void Devices_CanUsePenAsPointer()
...
-            });
-        InputSystem.QueueDeltaStateEvent(device.allTouchControls[1],
-            new TouchState
-            {
-                phase = PointerPhase.Canceled,
-                touchId = 5,
-            });
-        InputSystem.Update();
+        // First finger goes up. Second finger becomes primary.
+        EndTouch(4, new Vector2(0.345f, 0.456f), time: 0.3);
```
jechter, line 2331: Same question as before, is this the behavior we actually want?

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
```diff
@@ -20,7 +18,8 @@ namespace UnityEngine.InputSystem.Controls
     public class ButtonControl : AxisControl
     {
         public float pressPoint;
-        public float pressPointOrDefault => pressPoint > 0.0f ? pressPoint : InputSystem.settings.defaultButtonPressPoint;
+
+        internal float pressPointOrDefault => pressPoint > 0.0f ? pressPoint : InputSystem.settings.defaultButtonPressPoint;
```
jechter: was this what we decided on? Didn't we end up concluding that press points were the more desirable solution after all, as they might be device/control specific? (Not sure if that means they should be public or not, just want to make sure this is what we want).

Packages/com.unity.inputsystem/InputSystem/Controls/TouchControl.cs
```diff
@@ -27,42 +28,67 @@ public class TouchControl : InputControl<TouchState>
         public Vector2Control delta { get; private set; }
         public AxisControl pressure { get; private set; }
         public Vector2Control radius { get; private set; }
-        public PointerPhaseControl phase { get; private set; }
+        public InputControl<TouchPhase> phase { get; private set; }
```
jechter: Why not TouchPhaseControl?

Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs
```diff
@@ -11,53 +10,173 @@
...
+// - Lots of tests to cover all this
+// - Ideally, make touch not depend on hardcoded state format
 
 namespace UnityEngine.InputSystem.LowLevel
 {
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1726:UsePreferredTerms", MessageId = "Flags", Justification = "Fix this after landing Touch refactor")]
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1717:OnlyFlagsEnumsShouldHavePluralNames", Justification = "Fix this after landing Touch refactor")]
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1028:EnumStorageShouldBeInt32", Justification = "byte to correspond to TouchState layout.")]
+    [Flags]
+    public enum TouchFlags : byte
```
jechter: Maybe just make this internal? Then the name does not matter - users have nicer properties to access the individual bits anyways, and should not bother with how it's stored.

Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs
```diff
@@ -126,75 +273,53 @@ public FourCC format
...
-                        ++touchCount;
-                    }
-                }
+        public IntegerControl tapCount { get; private set; }
 
-                return new ReadOnlyArray<TouchControl>(m_ActiveTouchesArray, 0, touchCount);
-            }
-        }
+        /// <summary>
+        /// Syntheti
```
jechter: garbage comment?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -147,7 +151,7 @@ public override void OnImportAsset(AssetImportContext ctx)
                     className = m_WrapperClassName,
                 };
 
-                if (!wrapperFilePath.ToLower().StartsWith("assets/"))
+                if (!wrapperFilePath.ToLower().StartsWith("assets/") && !wrapperFilePath.ToLower().StartsWith("assets\\"))
```
jechter, line 154: 🤦🏻‍♀️

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -116,6 +115,8 @@ public override void OnGUI(string searchContext)
                 EditorGUILayout.PropertyField(m_DefaultTapTime);
                 EditorGUILayout.PropertyField(m_DefaultSlowTapTime);
                 EditorGUILayout.PropertyField(m_DefaultHoldTime);
+                EditorGUILayout.PropertyField(m_TapRadius);
```
jechter, line 118: Do we actually want to expose these settings at all? If yes, do we need to make them visible in the UI?

I'm wondering because "what constitutes a double tap" is a gesture clearly defined by the OS. A well designed app should not say "I have my own rules for what's a double tap, because the developer of this app thinks that this is the coolest double tap definition". So we should just use the OS standard here. The only valid reason I see an app developer wanting to change this is that we fucked up using the right defaults. Unfortunately, at least on iOS, this stuff is not really exposed - you should use their gesture detection APIs instead, which is not something we can easily do (or would want to). But i guess we can experiment to find the values they use (and do the same for android and windows touch), and then use those values?

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/EnhancedTouchSupport.cs
```diff
@@ -0,0 +1,120 @@
...
+            InputSystem.onDeviceChange += OnDeviceChange;
+            InputSystem.onBeforeUpdate += Touch.BeginUpdate;
+            InputSystem.onSettingsChange += OnSettingsChange;
+
+            SetUpState();
+        }
+
+        public static void Disable()
+        {
+            if (!enabled)
```
jechter, line 44: nitpick: I'd rather use `s_Enabled > 0` then overriding the `enabled` property, which has a defined use for MonoBehaviours - I found it confusing to read this, and it's only used in this one place anyways.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/EnhancedTouchSupport.cs
```diff
@@ -0,0 +1,120 @@
...
+////TODO: gesture support
+////TODO: mouse/touch simulation support
+////TODO: high-frequency touch support
+
+////REVIEW: have TouchTap, TouchSwipe, etc. wrapper MonoBehaviours like LeanTouch?
+
+namespace UnityEngine.InputSystem.Touch
+{
+    [AddComponentMenu("Input/Enhanced Touch Support")]
+    public class EnhancedTouchSupport : MonoBehaviour
```
jechter, line 12: So the only reason to have this as a MonoBehaviour is that you can stuff it somewhere in your scene to enable it. Is that a good use of a component? Should we only have one way to enable it (by calling the static `Enable` function)? Or should we maybe add a setting to enable it?

jechter: Also, should it be `Touch.Enable()` instead? Since all the interaction with the system happens through the `Touch` class? Feels odd to have an `EnhancedTouchSupport` class which has the only purpose of enabling the APIs in the `Touch` class.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/Touch.cs
```diff
@@ -0,0 +1,498 @@
...
+        /// </summary>
+        /// <remarks>
+        /// By default, this is zero meaning that no history information is kept for
+        /// touches. Setting this to <c>Int32.maxValue</c> will cause all history from
+        /// the beginning to the end of a touch being kept.
+        /// </remarks>
+        public static int maxHistoryLengthPerFinger
+        {
+            get => s_HistoryLengthPerFinger;
+            set { throw new NotImplementedException(); }
```
jechter: comment out the setter if you can't use it.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs
```diff
@@ -0,0 +1,437 @@
...
+    {
+        public Touchscreen simulatedTouchscreen { get; private set; }
+
+        public static TouchSimulation instance => s_Instance;
+
+        public static void Enable()
+        {
+            if (instance == null)
+            {
+                ////TODO: find instance
```
jechter, line 46: Guess this TODO should not be lost, I suppose this might break apart when you have both a MonoBehaviour you added to the scene _and_ call the static `Enable` function somewhere. But same question as for EnhancedTouch - do we need both methods to enable this?

Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
```diff
@@ -223,6 +223,67 @@ public int ReadInt(void* statePtr)
             return value;
         }
 
+        public void WriteInt(void* statePtr, int value)
+        {
+            Debug.Assert(sizeInBits != 0);
+
+            var valuePtr = (byte*)statePtr + (int)byteOffset;
+
+            if (format == FormatInt || format == FormatUInt)
```
jechter, line 232: use a switch statement?

Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
```diff
@@ -358,6 +425,168 @@ public void WriteFloat(void* statePtr, float value)
...
+        ////        way to use a type argument here, not sure how to get rid of it.
+
+        public double ReadDouble(void* statePtr)
+        {
+            Debug.Assert(sizeInBits != 0);
+
+            var valuePtr = (byte*)statePtr + (int)byteOffset;
+
+            double value;
+            if (format == FormatFloat)
```
jechter, line 450: use a switch?

Assets/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: Think we should axe all of those.

Assets/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: I'll try.

This refactor I made because CoreTests_Actions.cs reached the size limit on code analysis in Rider and I couldn't figure out how to increase the limit. So I split the file into three. Didn't anticipate that it would result in such a clusterfuck of a diff.

Assets/Tests/InputSystem/CoreTests_Actions.cs
jechter: +1

Assets/Tests/InputSystem/CoreTests_Actions.cs
jechter: I guess the way to do it then is to PR a separate branch where you move the tests, and merge that back into your branch. (But, yes, it is a hassle)

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
Rene-Damm: Yup, no change in agreement on press points and no change in behavior. Merely removed this particular form of getter from the public API. But looking at it, maybe better the other way round.

Packages/com.unity.inputsystem/InputSystem/Controls/TouchControl.cs
Rene-Damm: Could go either way. My intention here was to hide the fact that it uses `TouchPhaseControl` as that control and `TouchButtonControl` are born out of implementation concerns. API-wise they don't really have much of a reason to exist.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
Rene-Damm: Agree on the "shouldn't be per-app kinda thing" and admit to the exposed more out of insecurity part :) Think I'm fine with removing it from the UI, though at the same time, I'm hesitant a bit given ATM it's the only discoverable way for the user to counteract a bad choice on my end.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/EnhancedTouchSupport.cs
Rene-Damm: >Is that a good use of a component?

IMO for prototyping this is the kind of stuff that is super convenient. LeanTouch has a number of MonoBehaviours that are similar in nature and probably wouldn't be deemed "good use of a component". But fact is, they are immensely convenient to use.

Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs
Rene-Damm: `case` clauses have to be compile-time constant so can't use `FormatInt` etc. We could go and use plain ints instead but think that would make the code here quite a lot less readable.

Assets/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: And when you are two weeks late with your promised changes, it's the kind of stuff that tends to be the first to land on the "ooops, sorry" list :D

Rene-Damm: Hmm, looking at the fallout, only now realize how intrusive that splitting up of files actually is. Didn't think much about it when I did it.

Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs
Rene-Damm: Yup makes sense. Done.

Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs
Rene-Damm: Indeed. Fixed.

Packages/com.unity.inputsystem/InputSystem/Controls/TouchControl.cs
jechter: So what's the purpose of TouchPhaseControl?

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
jechter: what do we do with this? Don't care strongly, tbh.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
jechter: But our insecurity about the choices we made should not have to be burdened by users who then have to think about it. I'd rather not have these at all, and say we need to get them right (and treat it as a bug when we don't). I guess for the radius, getting it right may be tricky, though as it depends on the resolution used. So I'm ok with having the APIs for them, but let's hide them in the UI?

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/EnhancedTouchSupport.cs
jechter: I guess we could come up with a ton of "super convenient" MonoBehaviours for all our public APIs, but we don't usually ship any of those, but instead let users make them for themselves (in this case it is literally calling a link of code in your enable function). So I'd prefer not to have this (but will let you ultimately decide).

jechter: Any thoughts on my second comment?

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
Rene-Damm: Heh, me neither :) Fine either way. I'll go revert the change here. Might actually be a useful one.

Packages/com.unity.inputsystem/InputSystem/Controls/TouchControl.cs
Rene-Damm: To not expose a raw `int` but rather a fully typed `TouchPase` value.

Thinking about it, think the fact the setup here is unexpected is probably reason enough not to do it this way. I'll go make it `TouchPhaseControl phase`.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
Rene-Damm: Ok, can do.

Do wonder whether we are doing the right thing. LeanTouch, for example, has all these configuration values as inspector-settable properties: https://carloswilkes.github.io/Documentation/LeanTouch#LeanTouch.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/Touch.cs
Rene-Damm: Done.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/EnhancedTouchSupport.cs
Rene-Damm: Ok, looking at this again, I'm fine not making `EnhancedTouchSupport` a `MonoBehaviour`.

Can see the point RE `Touch` vs `EnhancedTouchSupport` but somewhat on the fence. Just that

```
EnhancedTouchSupport.Enable();
```

so nicely communicates what it does vs

```
Touch.Enable();
```

It's not that touch is disabled, it's that the "EnhancedTouch" subsystem is not enabled by default. But maybe I'm splitting hairs here. Can go and roll it into one.

Should the namespace be `UnityEngine.InputSystem.Touch` or `UnityEngine.InputSystem.EnhancedTouch` BTW? The latter would also get rid of the `Touch.Touch` namespace vs class conflict.

jechter: I'd lean towards `UnityEngine.InputSystem.EnhancedTouch`.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
jechter: Clearly this is an area where I hold strong opinions, and not everyone has to agree :) 

But in general: lot of developer tools have a lot of settings. Like, whenever in doubt, make it a setting. I believe that this is wrong. And I believe that one of Unity's biggest strengths has always been that we make the user experience of the developer not an after-thought. So my approach is always: When in doubt that we need this, _don't_ add a setting. We can easily add a setting if it turns out we were wrong, but not the other way around.

Rene-Damm: Yeah, disagree on that one the same way I disagree every time Apple doesn't allow me to configure something in order to "make my user experience better". And there I am finding myself typing obscure "defaults write -g NSWhatTheShitNot -float 0.03" commands in the shell because someone opted to "spare me the UI setting" :)

Aaaanyway, I'll go remove the UI parts.

PR #658, opened 2019-06-03
Adding "relative mode" to OnScreenStick

eblabs:
Hi,
Currently, the OnScreenStick controller uses your first touch to determine a "start point". In many cases it's actually more helpful to always have the "start point" at the center of the joystick.
Scenario, if I tap right on the controller, currently this would not register any movement because the origin of the movement is reset to this point. Ideally if I tap right, I would expect the controller to register this tap as being on the right side of the controller not at zero.
I think this should be the default behaviour, please let me know your thoughts.
Thanks!!


jechter:
Conceptually, seems right. I guess I don't know what common usage expectation for on screen sticks are, but I think it makes sense to add this option.

Would be good to have a test for this in `Assets/Tests/InputSystem/Plugins/OnScreenTests.cs`, though - and we need to run the code formatter script on this before landing, I don't think the indentation on this would pass CI.

eblabs:
Thanks for your feedback @jechter. 

Would you mind clarifying your comments on the formatting/indentation issues? Id be happy to make any needed changes.

Cheers,

Eric

jechter:
Don't bother with the formatting. We can fix that for you, when we land it. (We have automated formatting tools, and PRs will fail if they are not run, but I don't think this stuff is publicly available atm).

eblabs:
Great, thanks @jechter. I'll leave this as is then...

Cheers,

Eric

jechter:
> Great, thanks @jechter. I'll leave this as is then...

It would still be great to add a test in `Assets/Tests/InputSystem/Plugins/OnScreenTests.cs`, though.

eblabs:
Hey @jechter, I'm not familiar with how your unit testing works. So I'll leave that up to you then.
Thanks!!

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=658) <br/>All committers have signed the CLA.

eblabs:
Hi @jechter,
I just resolved a conflict on my code, but of course please check it out when you have a chance.
Thanks!
Eric

jamesmcgill:
Thank you for the PR.
We have taken your suggestion on board and will add this feature. We have an internal PR with this change here: https://github.com/Unity-Technologies/InputSystem/pull/1613

PR #660, opened 2019-06-05
Regenerate inputactions to have guids, and change json code to not ad…

jechter:
…d guids during serialization

Since we switched to writing the action map assets into the generated wrapper code, our project would suffer from wrapper code always being rewritten. This is because the inputactionassets we had did not contain guids (as they were written in an old version), and the json serializer would then generate guids during serialization which would be different every time.

This PR:
1. Resaves our inputactionassets, so they have guids
2. Removes GUID generation during serialization (we only did this for bindings, now bindings work the same way as actions and maps).


Rene-Damm (approved):

douglas-piconi (approved):
Do I need to do anything with the tanks demo?

jechter:
> Do I need to do anything with the tanks demo?

Probably not.

PR #662, opened 2019-06-05
NEW: Added support for Nintendo Switch Pro Controllers on desktops.

jechter:


yebata:
They look good to me.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -343,16 +344,112 @@ public static NPadDeviceIOCTLOutputCommand Create()
...
 {
+#if UNITY_EDITOR || UNITY_SWITCH
     /// <summary>
     /// An NPad controller for Switch, which can be a Joy-Con.
     /// </summary>
     /// <seealso cref="NPadInputState"/>
-    [InputControlLayout(stateType = typeof(NPadInputState), displayName = "Switch Controller (on Switch)")]
-    public class NPad : Gamepad, INPadRumble
+    [InputControlLayout(stateType = typeof(NPadInputStateSwitch), displayName = "Switch Controller (on Switch)")]
+    public class NPadSwitch : Gamepad, INPadRumble
```
Rene-Damm: I wonder about the naming here. Think it's unclear what `NPadSwitch` and `NPadHID` really are. Is it Joy-Cons? Is it Switch Pro controllers? Is it both?

From what I understand, `NPadSwitch` is actually both, i.e. might be Joy-Cons, might be a Pro controller. It's basically "controller picked up through Switch SDK API". Whereas `NPadHID` specifically is the Switch Pro controller (and thus has no support for orientations and such). Is that correct?

If so, I think it'd be good to make that distinction clearer in the API. It's different from DualShockGamepad, where the PS4 and HID version really are one of the same controller with the same capabilities (minus the fact that we are unable to surface all of them on the HID version).

jechter: Yeah, I was unsure about the naming as well, also because NPad doesn't seem to be a name used in consumer branding (?). I guess we *could* support the Joy-Cons on Desktop as well, but not sure if there is much value in doing so. 

What do you propose? Should I just rename `NPadHID` to `SwitchProControllerHID`, and leave `NPadSwitch` as it was (`NPad`)?

Rene-Damm: Have only done reading so far and no actual investigation but it seemed like chances for making JoyCon work on desktop would be slim. Didn't seem like the device is even a HID IIRC.

>What do you propose? Should I just rename NPadHID to SwitchProControllerHID, and leave NPadSwitch as it was (NPad)?

`SwitchProControllerHID` sounds good to me. Long but clear and unambiguous.

For `NPad`, personally, I'd prefer `SwitchController`. Wasn't too fond of `NPad`.

Sidetrack: I think the PS4 controller stuff suffers from similar problems. I think having `Gamepad` and then `XboxGamepad` and `PlayStationGamepad` and `SwitchGamepad` etc. would be much clearer. What do you think? Should we do a naming pass on those? We could have aliases to keep existing bindings working as is.

jechter: As we just discussed, JoyCon support on Desktop should be possible. Listed it for nice to have >1.0.

Also, as discussed, renamed NPadHID to SwitchProControllerHID, and reverted NPad to be renamed later.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
yebata: FYI: NPad is an SDK internal name so it can be a common name to Switch developers, but not PC develoers. Anyway, SwitchController sounds good to me. Sorry for my sense for names!

douglas-piconi (approved):

PR #668, opened 2019-06-07
NEW: Dualshock3 support on Desktops

jechter:
Adds support for DualShock 3 gamepads on Desktops via HID. Only supports Input, no rumble - the DS3 gamepad I have for testing  is a Sixaxis controller, which has no rumble capabilities.


Assets/Tests/InputSystem/Plugins/DualShockTests.cs
```diff
@@ -78,6 +78,67 @@ public void Devices_SupportsDualShockAsHID()
         // Sensors not (yet?) supported. Needs figuring out how to interpret the HID data.
     }
 
+    [Test]
+    [Category("Devices")]
+    public void Devices_SupportsDualShock3AsHID()
```
Rene-Damm: Test looks identical? Recommend just using `TestCase` in that case.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
```diff
@@ -77,6 +77,68 @@ public FourCC format
         }
     }
 
+    /// <summary>
+    /// Structure of HID input reports for PS3 DualShock 3 controllers.
+    /// </summary>
+    [StructLayout(LayoutKind.Explicit, Size = 32)]
+    public struct DualShock3HIDInputReport : IInputStateTypeInfo
```
Rene-Damm, line 84: How different are the two? Any chance we can just have the same struct and use a layout variant?

Assets/Tests/InputSystem/Plugins/DualShockTests.cs
jechter: Differences are in types, not suitable for testcase. But I was able to get rid of the code duplication by making a shared generic method for the test code.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
jechter: Nope. 100% different in binary layout.

Rene-Damm (approved):

douglas-piconi (approved):

PR #669, opened 2019-06-07
create il2cpp yamato configs

jechter:
This adds yamato test configs for mac and windows standalone builds running in il2cpp.

It also fixes xbox controller support on macOS in il2cpp. It turns out that this was broken, because the il2cpp implementation of `Marshall.OffsetOf` will not correctly handle structs with `LayoutKind.Explicit`. Running tests in il2cpp revealed this. The fix is to fix the field order and to add padding fields, so the field offsets will be correct for sequential layout as well. But this once again raises the question whether our explicit layouts are a bad idea. It is not unlikely that we have more similar bugs on other platforms.


douglas-piconi:
Hey this is cool you got this to work.  I did not know they had a Bokken image with Il2cpp build of unity

jechter:
@douglas-piconi it is not cool, as I did not get it to work (yet?). the green tests here are fake news :)

jechter:
As far as I can tell, utr is actually broken, and cannot correctly run playmode tests in il2cpp

douglas-piconi (approved):
Awesome Jonas!!!!

Rene-Damm (approved):

PR #676, opened 2019-06-12
FIX: Allow using InputSystem package if the XR, VR or Physics modules…

jechter:
… are disabled for smaller builds.

Currently, the InputSystem package has a hard dependency on XR, which again drags in Physics, which means you need to add a lot of code (build size) to get input in a simple 2D game.

This PR allows the InputSystem to build with the XR module disabled, to avoid this.

I wonder if we should make the XR stuff a separate package instead, so people don't have to disable it in the first place? Also, would make sense to test this, but I guess we'd need some sort of integration test framework for that, so we could set up a separate project.


douglas-piconi (approved):
Ok, so if I was a 2D user, all I would have to do is remove the enable XR define in the asmdf.  Are they going to be able to do that if they get this package from the package manager, or would this be something the undefine in their own projects asmdf?

PR #679, opened 2019-06-13
FIX: Xbox gamepad sticks for down and left could not navigate UI.

douglas-piconi:
The xbox gamepad on windows desktop could not navigate the new input system UI left or down movements.


jechter:
Hmm, that looks odd. Don't we still need parameters to clamp it to only the positive/negative side of the axis? Or are those inherited? Did you check that that works?

jechter (approved):

PR #680, opened 2019-06-14
Simplify tracked device handling

jechter:
Since we want to get rid of the handling of multiple touches using individual actions in the UI module, I wanted to take a look at whether we still need this set up for tracked devices in VR. After talking to @StayTalm I concluded that we do need the ability to track multiple devices separately (for left and right hand VR controllers, which can both point to UI elements at the same time, which unlike in multitouch UIs is expected to work). But I think we can still simplify that by using single actions with multiple bindings mapped as pass-through, and by creating separate TrackedDeviceModels for each device. This should also come with default bindings which should work out of the box.

This PR implements that. I could not test it in a real set up, but did write a test to make sure that multiple tracked devices would produce events with different pointer Ids (which I don't think could have actually actually worked before, because the pointer Id was not correctly passed into the UI events).

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -4,11 +4,11 @@ platforms:
   - name: win
     type: Unity::VM
     image: package-ci/win10:latest
-    flavor: m1.large
+    flavor: b1.large
   - name: win_standalone
     type: Unity::VM
     image: package-ci/win10:latest
-    flavor: m1.large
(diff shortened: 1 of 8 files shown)


StayTalm (approved):
LGTM

douglas-piconi (approved):

Rene-Damm (approved):

PR #687, opened 2019-06-20
Work on making sure scripting API is documented

jechter:
This PR adds tests to check if all our APIs is documented.

There is one test to check if every type in our public API has at least a summary, and a the same for methods (but the latter is still disabled).

Some types are currently ignored because:
-The package documentation system cannot correctly handle #ifdefed platform specific APIs today (we need to get them to fix this)
-The XR area is almost completely undocumented. I started adding some docs, but this was too much, and I don't have the knowledge to write them. @StayTalm please schedule time to work on this.
-I did not want to touch touch code to not interfere with the touch refactor.

For all other undocumented public types the test brought up, I did one of the following:
-Write a summary.
-Make the type internal (if I could not see how the user would benefit from access to this)
-Deleted the type (if I did not see it offering any useful functionality)

Of course, any of these decisions can be reversed if we don't agree.

In particular (but not limited to), 
-I made all `IInputStateTypeInfo` implementations internal, as I don't see how the user can use these internal device-specific data structures.
-I made all `IInputDeviceCommandInfo` implementations internal, as I don't see how the user can use these internal device-specific data structures.
-I removed `ReadWriteArray`, which was only used for making `RebindingOperation.scores` editable, which does not seem to offer much value.
-I removed `PrimitiveValueOrArray`, as non of it's functionality over `PrimitiveValue` was implemented.
-I made all `InputProcessor` implementation internal, as I see no API how users can use these classes (access seems to be exposed only through text mode representations).
-I removed `CurveProcessor` as it was not implemented.

Also, I renamed `XInputControllerOSX` to a more descriptive `XboxGamepadMacOS`.


jechter:
Note: The packman docs team has pushed a fix for #ifdefed APIs so we can add support for this in a future PR.

douglas-piconi (approved):

Rene-Damm (approved):

PR #702, opened 2019-07-01
CHANGE: We now use the same hierarchical device picker in the "Add Co…

jechter:
…ntrol Scheme" popup, which is already used in the "Input Settings" window.

This is nicer, and also we don't need another device list implementation any more.

Part of https://fogbugz.unity3d.com/f/cases/1165494/, but that bug also complains about the popup window placing on windows, which this may not fix (I'm on a mac, I don't know).


douglas-piconi (approved):

Rene-Damm (approved):
Yessss!

PR #703, opened 2019-07-01
FIX: Adjusted some UI Elements to fit the Unity 19.3 font.

jechter:
Fixes https://fogbugz.unity3d.com/f/cases/1165769/


douglas-piconi (approved):

Rene-Damm (commented):
What's the impact on 2019.1? Do we need to #if? (nooooooooooooo)

jechter:
> What's the impact on 2019.1? Do we need to #if? (nooooooooooooo)

The impact on  2019.1 is two buttons growing by 5 pixels, which I think is neglectable.

Rene-Damm (approved):

PR #719, opened 2019-07-09
Fix the compile error on linux with the Switch Pro Controller

douglas-piconi:
bad ifdef in the the middle of NPad.cs

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
@@ -1,4 +1,4 @@
-#if UNITY_EDITOR || UNITY_SWITCH || UNITY_STANDALONE || UNITY_WSA
+#if UNITY_EDITOR || UNITY_SWITCH || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_WSA
 using System;
 using System.Runtime.InteropServices;
 using UnityEngine.InputSystem.Controls;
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -346,7 +346,7 @@ public static NPadDeviceIOCTLOutputCommand Create()
     }
 #endif
 
-#if UNITY_EDITOR || UNITY_STANDALONE_OSX || UNITY_STANDALONE_WIN || UNITY_WSA
+#if UNITY_EDITOR || UNITY_STANDALONE || UNITY_WSA
```
Rene-Damm: Think this should be the opposite. Don't think we're support the Switch Pro controller this way on Linux ATM. Probably makes more sense to have the entire Switch stuff read `UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX` rather than `UNITY_STANDALONE`.

douglas-piconi: ok fixed, went to the other version.  Now consistent.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -346,7 +346,7 @@ public static NPadDeviceIOCTLOutputCommand Create()
     }
 #endif
 
-#if UNITY_EDITOR || UNITY_STANDALONE_OSX || UNITY_STANDALONE_WIN || UNITY_WSA
+#if UNITY_EDITOR || UNITY_SWITCH || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_WSA
```
douglas-piconi: just caught this, checked in a fix.  switch should not be in the HID section.

Rene-Damm (approved):

PR #721, opened 2019-07-12
Add State_PressingAndReleasingKeyInSameFrame_DoesNotShowStateChange

todi1856:
Added test for enforcing behavior where doing key down/key up at the same frame doesn't trigger

wasPressedThisFrame
wasReleasedThisFrame


Assets/Tests/InputSystem/CoreTests_State.cs
```diff
@@ -1014,6 +1014,25 @@ public unsafe void State_CanGetMetrics()
...
+
+        InputSystem.QueueStateEvent(keyboard, firstState);
+        InputSystem.QueueStateEvent(keyboard, secondState);
+
+        InputSystem.Update();
+
+        Assert.That(keyboard.aKey.isPressed, Is.False);
+        Assert.That(keyboard.aKey.wasPressedThisFrame, Is.False);
+        Assert.That(keyboard.aKey.wasReleasedThisFrame, Is.False);
+    }
```
Rene-Damm, line 1034: Wondering, does it make sense to have a separate test for this one given we have `State_PressingAndReleasingButtonInSameFrame_DoesNotShowStateChange`? The underlying cause is the same and the behavior applies to all buttons on any device.

todi1856: If you're certain that will always be the case, then I guess yes the test is redundant.

Rene-Damm: Yup, certain it will be. Will/would be fixed for all ButtonControls on all devices if we find a good solution for the problem with the current front/back buffer mechanism.

PR #722, opened 2019-07-12
CHANGE: Simplify action configuration and drop InputAction.continuous.

Rene-Damm:
- Added polling API for actions.
  * `InputAction.triggered`
  * `InputAction.ReadValue<T>()`
- Removed `InputAction.continuous`.
- Replaced `InputAction.passThrough` and `InputAction.initialStateCheck` along with the roster of toggles in the action properties view. Each action now has a general behavior type (`Value`, `Button`, or `PassThrough`).

image


douglas-piconi (approved):

DataGreed:
@Rene-Damm could you please tell why continuos was removed?
What is the best practice to do something while key is pressed down? E.g. fire continuously while right trigger is pressed?

JulienAtWork:
I'd like to know that too

Rene-Damm:
@DataGreed @JulienAtWork We deemed it not worth the internal code footprint and added complexity it brought to the action code given how straightforward it is to achieve the same effect through the polling API. However, we did fail to see the cost in complexity incurred on the user side when trying to stick to a callback-only approach.

While I do not anticipate that InputAction.continuous as such will necessarily make a comeback, I think it's clear we're missing something here. I'd like to revisit that after we're back to feature work once 1.0 is safe on the road to becoming a verified package for 2020.1.

PR #729, opened 2019-07-17
FIX: GC heap garbage from event processing and when replugging native device.

Rene-Damm:
Fixes
- GC heap garbage on every processing of a `StateEvent` or `DeltaStateEvent`.
- GC heap garbage from `InputManager` when unplugging a native device and then plugging it back in.

Note that the latter fix is only a half win. The native backend will send us a `String` containing a device descriptor. Allocating that string still allocates garbage.

Also has some script ref doc changes in the PR. Doesn't seem worth bothering to put them into separate PRs.


Rene-Damm:
Ok, need to do some proper unbreaking before this one can go further...

douglas-piconi (approved):

PR #734, opened 2019-07-23
CHANGE: Refactor device building to make InputDeviceBuilder internal.

Rene-Damm:
* Makes `InputDeviceBuilder` internal. Rationale:
  - No good reason to expose it and it had an odd API.
  - It's removed from `FinishSetup` now. The API for retrieving child controls has long ago become generally available so no good reason anymore to have a version of it sit on `InputDeviceBuilder`. And it was the only reason the `builder` argument was passed in.
  - Not having it will help with the next stage where we have generated code that bypasses `InputDeviceBuilder` entirely and whips up devices with no use of reflection.
* Added a new API `InputDevice.Build<TDevice>(...)` that preserves the ability to create devices that are not immediately added to the system.
* Added an internal `InputDeviceBuilder.Ref()` API that allows setting up scopes that automatically reuse the same device builder multiple times. Used after domain reloads, for example. All devices will be set up from the same device builder and share cached layout information. Means we only reflect on `ButtonControl` etc. once, for example.
* Turned `InputDeviceBuilder` into a struct. One less alloc.
* Nuked the ability to reuse existing device instances in `InputDeviceBuilder` along with the `InputSystem.SetLayoutVariant` API.
  - Rationale: Added code and complexity for little gain. Changing the layout variant after the fact didn't really solve the problems it was trying to solve and the ability to reuse instances was predicated on certain conditions which in practice meant that sometimes you get a different instance and sometimes you don't. New rule is simply, if we re-create a device, it's a new instance. Period.
* This PR also has @yebata's changes (slightly refactored) to allow a device to have multiple usages.


douglas-piconi (approved):
Admittedly, some things in here, I didn't get a technical grasp on yet. I didn't see anything wrong with the code though.

PR #745, opened 2019-07-29
Input samples fixes

Chippit:
Corrected observed issues with input samples. Almost everything was to do with layer/physics configuration.


jechter:
Will track and land this here: https://github.com/Unity-Technologies/InputSystem/pull/839
(That PR has the same changes, but in a branch on our repo, which will then be picked up by CI)

PR #747, opened 2019-07-29
NEW: Add the touch samples to the package distribution

douglas-piconi:
Have the touch samples included with our actual package.   They will be downloadable in the package manager on demand by the user.   Had to move the samples into a special folder called samples with a .sample.json file.    During the publishing step of the package, this special folder goes to the package server, however will not automatically be included when the user installs the input package.  In order for the assets to be downloaded, the user will have to click on the install sample button.


jechter:
How does this work specifically? I assume that this is a feature in PackMan? Is it specific to "Samples", or can packages include other optional sub packages? Could potentially be useful for many other things (tests, optional plugins, etc).

Rene-Damm:
@douglas-piconi How do package samples work in combination with project settings required by them? In particular, the touch samples require both the scene list in the player build settings to be set in a specific way and require tags to be configured a certain way. The latter is currently broken on the develop branch. #745 has a fix, though I think it's better to move the tags changes over into `TouchSamplesBuildSettings`.

My hope was that the sample mechanism for packages has some mechanism for dealing with project settings and that we can delete `TouchSamplesBuildSettings` and such.

PR #754, opened 2019-08-01
NEW: Add components for visualizing control and action activity.

Rene-Damm:
It's a start. Still needs work but I'd like to push the various things I have accumulated and what's there with the visualizers is already useful.

image

I also added a sample with some initial visualizer setups.

Ideally, a chunk of this code would *not* live in the input system package. Stuff like various graphing and plotting functions should be generally available. However, I found no public API easily available to us so just rolled some custom code.

There are two only semi-related changes in the PR:

1) `InputSystem.onEvent` now receives two arguments (`(InputEventPtr, InputDevice)` instead of one (`(InputEventPtr)`). This is part of the general API cleanup I'm working on but it's relied on by the visualizer code.
2) `InputAction.ReadValue<TValue>` works differently now and has a `ReadValueAsObject` twin now as the other `ReadValue` APIs do. This change is required for the action visualizer to work correctly.


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1299,7 +1299,64 @@ public static int FindControls<TControl>(string path, ref InputControlList<TCont
...
+        /// Instead, they are consumed in the order they are produced. This means that they
+        /// will also surface on this callback in that order.
+        ///
+        /// Note that the callback will be called for any event picked up from the internal
+        /// event buffer. This may include events for devices that have not been created and thus
+        /// may mean that the <see cref="InputDevice"/> parameter to the callback is null.
+        ///
+        /// <example>
+        /// <code>
+        /// // Treat LMB+RMB as MMB.
```
jechter: nitpick: LMB RMB MMB are not necessarily understood abbreviations

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputActionVisualizer.cs
```diff
@@ -0,0 +1,320 @@
...
+            s_EnabledInstances.Add(this);
+        }
+
+        protected new void OnDisable()
+        {
+            base.OnDisable();
+
+            s_EnabledInstances.Remove(this);
+            if (s_EnabledInstances.Count == 0)
+                InputSystem.onActionChange -= OnActionChange;
```
jechter: Should we remove the action event listeners here?

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputActionVisualizer.cs
```diff
@@ -0,0 +1,320 @@
...
+                        if (s_EnabledInstances[i].m_Action?.actionMap == actionOrMap)
+                            s_EnabledInstances[i].OnActionDisabled();
+                    break;
+            }
+        }
+
+        [SerializeField] private Visualization m_Visualization;
+        [SerializeField] private InputActionReference m_ActionReference;
+        [SerializeField] private string m_ActionName;
+        //[SerializeField] private int m_PlayerIndex;
```
jechter: remove

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputControlVisualizer.cs
```diff
@@ -0,0 +1,361 @@
...
+            base.OnGUI();
+        }
+
+        protected new void OnValidate()
+        {
+            ResolveControl();
+            base.OnValidate();
+        }
+
+        [Tooltip("The type of visualization to perform for the control. TODO")]
```
jechter: Don't have "TODO" in a public tooltip

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputControlVisualizer.cs
```diff
@@ -0,0 +1,361 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Determines which aspect of the control should be visualized.
+        /// </summary>
+        public enum Mode
+        {
+            None = 0,
```
jechter: where do these enum integer values come from?

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputVisualizer.cs
```diff
@@ -0,0 +1,82 @@
+#if DEVELOPMENT_BUILD || UNITY_EDITOR || PACKAGE_DOCS_GENERATION
```
jechter: If I have one of these in my scenes, and I make a non-development build, will Unity silently remove it? Or will there be errors?

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/VisualizationHelpers.cs
```diff
@@ -0,0 +1,623 @@
...
+            }
+
+            public override void AddSample(object value, double time)
+            {
+                base.AddSample(value, time);
+
+                if (value != null)
+                {
+                    var val = (TValue)value;
+                    if (Compare(min, val) > 0)
```
jechter: Do we want to support  the case where min and max can shrink again if values stay within a certain bracket later? I guess depends on what you are visualizing.

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/VisualizationHelpers.cs
```diff
@@ -0,0 +1,623 @@
...
+                        min = val;
+                    if (Compare(max, val) < 0)
+                        max = val;
+                }
+            }
+
+            private static unsafe int Compare(TValue left, TValue right)
+            {
+                var leftPtr = UnsafeUtility.AddressOf(ref left);
```
jechter: why do we need the unsafe logic here? Couldn't we just do something like
```
if (typeof(TValue) == typeof(int))  
   return (left as int).CompareTo(right);
```

?

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputControlVisualizer.cs
Rene-Damm: I was adding and removing entries as I was writing this and after I had a bunch of these components saved in scenes, was too lazy to go and update them after each change so ended up hardcoding the enum values :)

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputVisualizer.cs
Rene-Damm: That is a good question. I vaguely remember that we purge all MBs with missing scripts when we do player builds. But I haven't actually gone and made sure. Another reason to put this in the samples, I guess :)

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/VisualizationHelpers.cs
Rene-Damm: I think we do. Overall, the plotting functionality here can be improved by leaps and bounds. The current level of "smarts" it shows is pretty poor.

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/VisualizationHelpers.cs
Rene-Damm: I wish. C# won't allow that. `as` can only be used on reference and nullable types. With the current level of type parameter constraints in C#, I found no other way to do this kind of stuff more easily and without garbage.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: Fixed

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputActionVisualizer.cs
Rene-Damm: Good catch. Fixed.

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputActionVisualizer.cs
Rene-Damm: Fixed.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
jechter: where? :)

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputControlVisualizer.cs
jechter: So I guess it would make sense to remove them now?

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputVisualizer.cs
jechter: So now that we did move them, should we remove this whole `#if`?

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/VisualizationHelpers.cs
jechter: meh, ok.

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputVisualizer.cs
jechter: (Or even if we still think `DEVELOPMENT_BUILD || UNITY_EDITOR` make sense, `PACKAGE_DOCS_GENERATION` does not in the Samples case)

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputControlVisualizer.cs
Rene-Damm: Does it really matter? Sure I can remove them and update the prefabs I created. Is it important?

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputVisualizer.cs
Rene-Damm: Ah yup, those indeed don't make sense anymore. Fixed.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: Crap, turns out you actually have to push your changes for them to be visible :) Fixed for reals now.

Packages/com.unity.inputsystem/InputSystem/Plugins/Visualizers/InputControlVisualizer.cs
Rene-Damm: Fixed

jechter (approved):

PR #758, opened 2019-08-05
Composite rebinding ui

jechter:
Proposal for how to do rebinding of composites (as demonstrated in the Tanks rebinding screen).

Implemented a `RemapUI` class, where you can pass the buttons for an action and it's composite replacements, and it will do the logic of detecting if a bound control matches the action directly or should go to the composite (and show the corresponding UI).

Added two required changes to the Input System code itself to get here.
-Changed the m_OnGeneratePath handling in the rebinding code (I think that was a bug)
-Exposed `effectivePath` and friends (needed to show correct labels in a rebinding UI).
(need to add API docs for this if we agree to land it!)

Let me know if this makes sense, and if we should land it or wait  for @douglas-piconi 's changes first to avoid conflicts.

Video: https://oc.unity3d.com/index.php/s/9pmVH5dS7n4p8kk


Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
```diff
@@ -26,53 +36,141 @@ public class ButtonRemapScreenController : MonoBehaviour
...
-    private void RemapButtonClicked(string name, Button uiButton, int mapIndex)
-    {
-        uiButton.enabled = false;
-        okButtonMessageText.text = "Press button/stick for " + name;
+        void RemapButtonClicked(string name, int bindingIndex = 0)
+        {
+            m_Button.enabled = false;
+            m_Text.text = "Press button/stick for " + name;
+            rebindOperation = m_Action.PerformInteractiveRebinding()
+                .WithControlsExcluding("Mouse")
```
Rene-Damm: Probably mean `"<Mouse>"`. In the current form, just literally matches "Mouse". Is the intention to kill off all mouse rebinds?

Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
```diff
@@ -26,53 +36,141 @@ public class ButtonRemapScreenController : MonoBehaviour
...
+            // composite.
+            if (expectedControl != null && !expectedControl.IsInstanceOfType(control))
+                return 0.1f;
 
-    private void OkButtonClicked()
-    {
-        playerActionMap.Enable();
-        SceneManager.LoadScene("NewInput");
-    }
+            return 1;
```
Rene-Damm: In general, probably makes sense to take magnitude into account here. The default score computation takes actuation into account and thus can, for example, find out that you are actuating X more than Y and probably want X then. Not sure this is relevant in the context of this demo, though.

Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
```diff
@@ -26,53 +36,141 @@ public class ButtonRemapScreenController : MonoBehaviour
 
     private InputActionMap playerActionMap;
     private InputActionRebindingExtensions.RebindingOperation rebindOperation;
+    private RemapUI gasUI, brakeUI, fireUI, turrretUI;
```
Rene-Damm: Typo in `turrretUI`.

Assets/TanksDemo/Assets/Scripts/UI/ButtonRemapScreenController.cs
```diff
@@ -26,53 +36,141 @@ public class ButtonRemapScreenController : MonoBehaviour
...
+                            m_IsUsingComposite = true;
+                        else
+                            m_IsUsingComposite = false;
+                        return null;
+                    })
+                    .OnApplyBinding((x, path) =>
+                    {
+                        if (m_IsUsingComposite)
+                        {
+                            m_Action.ApplyBindingOverride(m_DefaultBindingIndex, "unbound");
```
Rene-Damm: Alternatively, can just use an empty string. Overriding with "" effectively causes a binding to be ignored.

douglas-piconi (approved):

douglas-piconi:
I think this one makes sense @jechter  This can land first and I will merge into my changes.

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
```diff
@@ -251,9 +251,29 @@ public static InputBinding MaskByGroups(params string[] groups)
...
+        /// </remarks>
+        public string effectiveInteractions => overrideInteractions ?? interactions;
+
+        /// <summary>
+        /// This is the processor config which is effectively being used.
+        /// </summary>
+        /// <remarks>
+        /// This is either <see cref="overrideProcessors"/> if that is set, or <see cref="processors"/> otherwise.
+        /// </remarks>
+        public string effectiveProcessors => overrideProcessors ?? processors;
```
Rene-Damm, line 293: As I have only just recently discovered with much sadness, properties also need a `<value>...</value>` documentation as otherwise DocFX generates a useless empty box in the docs.

Rene-Damm (approved):
Nice!

Any chance we can pull some of the rebinding UI stuff out and put it in a sample separately in some reusable form?

PR #760, opened 2019-08-06
FIX: Compile issue on linux standalone

douglas-piconi:
Can't build linux standalone player with Switch module in InputSystem.cs.  

NOTE: The format tool did that funky indenting.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
@@ -2202,7 +2202,9 @@ namespace UnityEngine.InputSystem
             #endif
 
             #if UNITY_EDITOR || UNITY_SWITCH || UNITY_STANDALONE || UNITY_WSA
+                #if !UNITY_STANDALONE_LINUX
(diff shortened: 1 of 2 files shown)


jechter (approved):

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -2202,7 +2202,9 @@ private static void PerformDefaultPluginInitialization()
             #endif
 
             #if UNITY_EDITOR || UNITY_SWITCH || UNITY_STANDALONE || UNITY_WSA
+                #if !UNITY_STANDALONE_LINUX
```
Rene-Damm: Please fold into outer #if. Probably easiest to just replace UNITY_STANDALONE with UNITY_STANDALONE_OSX || UNITY_STANDALONE_WIN.

douglas-piconi: @Rene-Damm  fixed.

Rene-Damm (approved):

PR #762, opened 2019-08-07
CHANGE: Refactor/clean up some APIs and add a batch of scripting API docs.

Rene-Damm:
This is still *very* incomplete. Given the cleanup refactors plus the API doc comments tend to create huge PRs, would like to land this stage and keep going from there.

Notable API changes:
* `PlayerInput` has been hoisted out of its own namespace and moved to `UnityEngine.InputSystem`.
* `MouseState`, `GamepadState` and `KeyboardState` are public again.
* `RegisterControlProcessor` is now simply `RegisterProcessor`.
* `PerformedAndGoBackToWaiting` is now simply `Performed`.
* `InputUpdateType` is now in `LowLevel` namespace.
* `InputSystem.onBeforeUpdate` and `InputSystem.onAfterUpdate` no longer receive `InputUpdateType` argument. `InputState.currentUpdateType` makes it redundant and the majority of users should never have to care about `InputUpdateType`.
* Some other miscellaneous stuff (like `IMECompositionString` and `InputMetrics`) is now in `LowLevel` in an effort to weed out the toplevel namespace.
* `InputControlListExtensions` is gone.

I also renamed `VisualizerSamples` to just `Visualizers` and moved the sample for on-screen controls into `Samples`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ however, it has to be formatted properly to pass verification tests.
   * Fixed actions bound to the dpad control performing correctly.
   * Fixed non-present touchpad button control being triggered incorrectly.
 - Fixed compile issues with switch classes on standalone Linux.
+- Leak of unmanaged memory in `InputControlList`.
```
jechter, line 22: "fixed"?

Packages/com.unity.inputsystem/InputSystem/Actions/IInputInteraction.cs
```diff
@@ -19,17 +19,237 @@ namespace UnityEngine.InputSystem
...
+    /// <example>
+    /// <code>
+    /// // Create a fire action with two interactions:
+    /// // 1. Hold. Triggers charged firing. Has to come first as otherwise "Press" will immediately perform the action.
+    /// // 2. Press. Triggers instant firing.
+    /// // NOTE: An alternative is to use "Tap;Hold", i.e. a "Tap" first and then a "Hold". The difference
+    /// //       is relatively minor. In this setup, the "Tap" turns into a "Hold" if the button is held for
+    /// //       longer than the tap time whereas in the setup below, the "Hold" turns into a "Press" if the
+    /// //       button is released before the hold time has been reached.
+    /// var fireAction = new InputAction(type: InputActionType.Button, interactions: "Hold;Press");
```
jechter, line 58: Is there a difference between "Hold;Press" and "Hold,Press"?

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -311,37 +388,64 @@ public enum Key
...
         /// <summary>
         /// Event that is fired for every single character entered on the keyboard.
         /// </summary>
+        /// <remarks>
+        /// <example>
+        /// <code>
+        /// Keyboard.current.onTextInput +=
+        ///     ch =>
+        ///     {
+        ///         TODO
```
jechter, line 433: No TODO in public docs, please - but i guess in this case it is not so much for us to do, but for the reader of the docs to fill? Then add a comment saying so.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -429,45 +537,98 @@ public string keyboardLayout
...
+        /// to the right in the bottom-most row of alphabetic characters.
+        /// </summary>
+        /// <value>Control representing the b key.</value>
         public KeyControl bKey => this[Key.B];
+
+        /// <summary>
+        /// The 'c' key. The key in-between the <see cref="xKey"/> to the left and the <see cref="vKey"/>
+        /// to the right in the bottom-most row of alphabetic characters.
+        /// </summary>
+        /// <value>Control representing the c key.</value>
```
jechter, line 629: So we document all letters up to C? :) (Would probably use some grep to put some standard phrase for all of them)

jechter (approved):
💖

Packages/com.unity.inputsystem/InputSystem/Actions/IInputInteraction.cs
Rene-Damm: Heh, didn't even know. Had to check. It's the same. For historical reasons, both the semicolon and comma are allowed in this case.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
Rene-Damm: Heh, no this is for me. The current script ref docs are still a first pass. Have to admit my approach is not the most structured one. I just go all over the place and repeatedly bang my head against the API doc comments until the result is starting to look decent in the output. So, some stuff here still needs a second pass.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
Rene-Damm: Haha, no. Also incomplete. Locally have it about 75% complete and going. For reviewing, my approach probably sucks. Sorry.

Packages/com.unity.inputsystem/InputSystem/Actions/IInputInteraction.cs
jechter: You didn't know, but used both syntaxes in the same PR :) 

Do those historical reasons still apply? Should we allow only one for less confusion?

Rene-Damm: >You didn't know, but used both syntaxes in the same PR :)

I did? Help me see it :)

>Do those historical reasons still apply? Should we allow only one for less confusion?

Dropping `;` is an option, I think. Initially, I used `,` everywhere as separators but for reasons that escape me now (think there was some data used as elements where `,` was already used) started switching over `;` with the intention of making that the element separator in all the JSON serialized data. However, `NameAndParameters` never really ended up fully making the switch and the editor is still writing out stuff like `Tap,Hold(duration=1,pressPoint=0.5)` instead of `Tap;Hold(duration=1,pressPoint=0.5)` (as it was intended to be).

So I guess we could drop `;` here.

jechter: >Help me see it :)

CoreTests_Actions_Interactions.cs:17

I think dropping one of the syntaxes would be better. I wonder if it would be better to drop `,`, because `;` has the benefit of being distinguished from the `,` between parameters. But maybe that's not important?

Rene-Damm: >CoreTests_Actions_Interactions.cs:17

Hah, indeed :)

> think dropping one of the syntaxes would be better. I wonder if it would be better to drop ,, because ; has the benefit of being distinguished from the , between parameters. But maybe that's not important?

I'd prefer that, too. Was one of the original motivations to change to the `,`. But problem is that change is actually a more breaking one. Then, I think, we'd have to somehow upgrade the .inputactions files that people have as those actually do use the `,` syntax instead of the `;` syntax. Which I'm thinking isn't worth the bother. Overall, allowing both `;` and `,` here seems like a relatively minor wart to me. Think I'm leaning towards "let's just live with it" side.

jechter: Fair enough. Might try to be consistent in only using one in our docs then, though.

douglas-piconi (approved):

PR #765, opened 2019-08-09
FIX: Samples not being included in package.

Rene-Damm:
Samples were included in the packages we tested but not in the package we test but not in the package we push. Was considering removing them entirely from the "Build and Test" step but not sure whether there's something in package validation that makes it good to also have it in that step. Guess the package we test should be identical to the package we push.

### DIFF
--- a/.yamato/promotion.yml
+++ b/.yamato/promotion.yml
@@ -17,6 +17,8 @@ promotion_test_{{ platform.name }}_{{ editor.version }}:
   variables:
     UPMCI_PROMOTION: 1
   commands:
+    - mv ./Assets/Samples ./Packages/com.unity.inputsystem
+    - mv ./Assets/Samples.meta ./Packages/com.unity.inputsystem
     - npm install upm-ci-utils@stable -g --registry https://api.bintray.com/npm/unity/unity-npm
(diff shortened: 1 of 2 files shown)


jechter (approved):

douglas-piconi (commented):
for the record I am really against this PR, this is going to be error prone and difficult to maintain especially with the churn of the package and yamato systems.   If we really need to go this route.  these mv lines need to be added to promotion.yml as well.    For the record, if we can't have samples permanently live under samples in the package root folder, we are doing something wrong.

douglas-piconi (approved):

Rene-Damm:
>if we can't have samples permanently live under samples in the package root folder, we are doing something wrong.

I wish that were the case. I think the `Samples` mechanism still needs work to give us what we need. I agree that what we're doing here isn't what we should be doing but ATM it seems the lesser of two evils. Putting things directly in `Packages/com.unity.inputsystem/Samples` will mean that our samples will never be compiled and on top lead to long rows of warnings when working locally in the editor.

One alternate thing I think we could explore next is to package our samples individually into `.unitypackage` files from the directories in `Assets/Samples` instead of moving them. But it'd still have to be a step in the CI and would probably end up being much more complicated.

PR #769, opened 2019-08-09
CHANGE: Clean up SimpleDemo and move into Samples.

Rene-Damm:
- Added a version using `PlayerInput`. SimpleDemo now goes through all relevant permutations of how to work with the input system
- Removed the version using events. There's no good reason to do it this way, so no point in having the sample.
- Removed jumping code. Just adds complexity with no added demonstration value. Reverted the demo to having a no-physics character (i.e. no RigidBody on there, not even kinematic) and fire, move, look as the only actions.
- Switched move and look actions to generally use polling instead of callbacks. Turns out that SimpleDemo with its charged fire action makes a very good illustration of when the callback-based approach is better and when polling is better.
- Moved the whole thing over into Assets/Samples and added .sample.json file.
- Added a [README.md](https://github.com/Unity-Technologies/InputSystem/blob/simple-demo-sample/Assets/Samples/SimpleDemo/README.md) file that goes through the samples one by one.

Also contains a minor tweak to the input debugger to show interactions present on a binding.


Rene-Damm:
@stefanunity This PR probably also affects the QA run that Jia devised slightly as there is no `SimpleDemo` scene anymore. Instead there are multiple such scenes now. Probably makes sense to have a quick go at all of them. This will also give a bit more of a comprehensive test than just the one variation we had before.

douglas-piconi (approved):

PR #774, opened 2019-08-10
NEW: Add two composites for defining buttons with modifiers.

Rene-Damm:
The question/request for a way to set up things such as "CTRL+1" has come up so often and is so easy to add on our side as two composites that I thought what the heck, we may as well just add that.

Also contains a slight tweak to the action editor to filter out composites that aren't applicable to a given action. Cleans up the "Add Binding" menu to hide options that don't make sense anyway.

PS: First PR where I add both full scripting API docs *and* coverage in the manual. So proud of myself.


jechter (approved):
Nice!

douglas-piconi (approved):

PR #775, opened 2019-08-12
Multi touch ui

jechter:


Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/DefaultInputActions.inputactions
```diff
@@ -250,17 +250,17 @@
...
                     "processors": "",
                     "interactions": ""
                 },
                 {
                     "name": "Click",
-                    "type": "Button",
+                    "type": "PassThrough",
                     "id": "3c7022bf-7922-4f7c-a998-c437916075ad",
-                    "expectedControlType": "Button",
+                    "expectedControlType": "",
```
Rene-Damm: What's the reason for setting this to Any rather than Vector2 above and Button here?

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/DefaultInputActions.inputactions
jechter: The reason is that the actions editor resets it when changing the type (I guess this is a minor UX bug). Reverting.

jechter: Will fix the UX bug as well

douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionPropertiesView.cs
```diff
@@ -47,10 +47,6 @@ protected override void DrawGeneralProperties()
 
             if (EditorGUI.EndChangeCheck())
             {
-                // When changing the action type, clear out the expected control type.
-                if ((InputActionType)m_SelectedActionType != (InputActionType)m_ActionTypeProperty.intValue)
-                    m_SelectedControlType = 0;
```
Rene-Damm, line 52: I think we need to replace this with something else rather than remove it. Problem is that when you change from, say, Value of Vector2 to Button, then the Vector2 constraint should indeed be gotten rid of I think. Probably makes sense to restrict this to that particular direction (anything -> Button), though.

jechter: I think that is still guaranteed by the line below:

if ((InputActionType)m_SelectedActionType == InputActionType.Button)
                    m_ExpectedControlTypeProperty.stringValue = "Button";

Rene-Damm: Oh, you're right. Good catch.

PR #779, opened 2019-08-12
CHANGE: Upgrade CI to 2019.2.

douglas-piconi:
with 2019.1 being a dead release, we should probably run our CI in 2019.2.

### DIFF
--- a/.yamato/promotion.yml
+++ b/.yamato/promotion.yml
@@ -1,5 +1,5 @@
 test_editors:
-  - version: 2019.1
+  - version: 2019.2
 test_platforms:
   - name: win
     type: Unity::VM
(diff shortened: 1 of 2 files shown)


jechter:
I don't think we should do this. At least not exclusively. 2019.1 is the lowest version we support. So we should test against that. In general, Unity makes quite an effort to be backwards compatible, but very little effort to be forward compatible. So if we want to support 19.1 and 19.2, testing on 19.1 is more likely to protect us from breaking 19.2 than vice versa.

Now, it may make sense to add more configs to test different versions. That is, in the end, a question of coverage vs resource utilization. So far I'm not convinced we need it yet, as I've not seen any issues which would have been caught if we tested on later versions (but as the range of versions we support gets wider, that will probably happen).

Rene-Damm:
Agree to all of Jonas's points.

For testing an additional version, seems that adding 2019.3 in the not too distant future probably makes sense. 2019.2 seems to not really have caused much trouble but 2019.3 so far seems like a different story. Could be it's just early-alpha glitches, though, and will straighten itself out just fine.

PR #781, opened 2019-08-13
FIX: Device pairing ignoring controls that don't support magnitudes (#764).

### LINKED ISSUE #764: Difficulty with control scheme switching
I split Mouse and Keyboard into separate control schemes (mainly so I can hide the cursor and the pointer-type selection when navigating menus with the keyboard, and so I can disable arrow navigation of menus when the mouse is being used). For the most part this works, except moving the mouse doesn't cause the control scheme to switch to Mouse, I have to click somewhere on the screen. I've attached the inputactions file, in case that's helpful.
[ControlsMap.zip](https://github.com/Unity-Technologies/InputSystem/files/3484281/ControlsMap.zip)

Rene-Damm:
Fixes #764.


jechter (approved):

Rene-Damm:
Turns out this change makes it too eager to switch now and while it fixes the original problem, it now causes the opposite kind of problem. Tweaking.

Rene-Damm:
Ok should be fixed properly now. Can run use the actions from #764 and get completely robust switching between mouse and keyboard.

Also found a bug in GetStatePtrFromEvent where it wasn't checking bounds properly.

PR #784, opened 2019-08-13
FIX: Non-functional 'view' and 'menu' buttons on XInputController.

Rene-Damm:
This fixes https://fogbugz.unity3d.com/f/cases/1170454/.

`XInputController` added two buttons called "view" and "menu" that shouldn't be there. Reason was that since `XInputController` does not use a `stateType`, the system would look on the type itself for controls to add. This would pick up

```
    public class XInputController : Gamepad
    {
        public ButtonControl menu { get; private set; }
        public ButtonControl view { get; private set; }
```

The PR changes the layout code such that properties require an explicit `[InputControl]` attribute to inject new controls into the layout -- which in some cases we want (see the control properties in `DualShockGamepad`).


Rene-Damm:
Turns out quite a few devices needed to have `[InputControl]` attributes added to them. Still think it's the right thing to do and be explicit about it. Adds a bit of noise to the various device layouts but prevents layouts accidentally getting messed up just by someone adding a convenience getter for a control to some layout.

jechter (approved):

PR #788, opened 2019-08-15
Docs fixes3

jechter:
- Cursor docs
- Remove some controls from cursor types which were not implemented.


douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
```diff
@@ -57,20 +56,16 @@ internal struct PointerState : IInputStateTypeInfo
         [InputControl(layout = "Analog", usage = "Pressure")]
         public float pressure;
 
-        [InputControl(layout = "Axis", usage = "Twist")]
-        public float twist;
-
-        [InputControl(layout = "Vector2", usage = "Tilt")]
-        public Vector2 tilt;
+        float twist;
+        Vector2 tilt;
```
Rene-Damm: These can be removed.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
```diff
@@ -57,20 +56,16 @@ internal struct PointerState : IInputStateTypeInfo
...
 
         [InputControl(layout = "Vector2", usage = "Radius")]
         public Vector2 radius;
 
         [InputControl(name = "press", layout = "Button", format = "BIT", bit = 0)]
         public ushort buttons;
 
-        [InputControl(layout = "Digital")]
-        public ushort displayIndex;
+        ushort displayIndex;
```
Rene-Damm: This can be removed.

Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs
```diff
@@ -78,9 +78,8 @@ public struct TouchState : IInputStateTypeInfo
         [FieldOffset(33)]
         public byte tapCount;
 
-        [InputControl(layout = "Digital")]
         [FieldOffset(34)]
-        public byte displayIndex;
+        byte displayIndex;
```
Rene-Damm, line 84: Can be removed.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
Rene-Damm: On `PenState` and `MouseState`, the `[InputControl...]` annotations for `displayIndex` need to be removed. Fields can be made internal (I guess removing them breaks il2cpp). ATM both `PenState` and `MouseState` will separately re-introduce `displayIndex`.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
jechter: Nope, that will break il2cpp struct layout.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
jechter: Nope, removing this will break il2cpp struct layout. (Brought it back after I removed it first and stuff stopped working on my phone).

Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs
jechter: Nope.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
jechter: But will remove the [InputControl...] annotations  i missed

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
Rene-Damm: That's fine. `PointerState` is not used by native, i.e. doesn't have to adhere to any particular layout.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
Rene-Damm: Would prefer a comment on these fields why they are there. Probably also makes sense to just rename all the leftover fields we are keeping for the sake of il2cpp to something like `_paddingXXX` or so.

Rene-Damm (approved):

PR #789, opened 2019-08-15
Move 02_Input test screen to Samples

jechter:


Rene-Damm:
It'll need a `.sample.json` file added to the sample.

RE the name, I'd favor "InputTest" over "InputDeviceTester".

jechter:
Will add the json file. As for the name: I think "InputTest" is too generic - that could be _anything_. InputDeviceTester is maybe not great, but it describes a bit better what this does (offer a visual representation of supported device types and their input state).

douglas-piconi (approved):

Rene-Damm (approved):

PR #796, opened 2019-08-20
FIX: PlayerInput.Instantiate not handling control schemes correctly.

Rene-Damm:
User report: https://forum.unity.com/threads/playerinput-instantiate-ignores-controller-schemes-for-multiplayer-with-players-on-same-device.730868/

The following used to result in two players with both having all keyboard bindings from all control schemes active rather than one using "KeyboardWASD" only and the other using "KeyboardArrows" only. Was a problem in `InputUser` where the order in which actions were set on a user WRT when its control scheme was set ended up mattering when it shouldn't.

```
PlayerInput.Instantiate(player, controlScheme: "Keyboard WASD", pairWithDevice: Keyboard.current);
PlayerInput.Instantiate(player, controlScheme: "Keyboard Arrows", pairWithDevice: Keyboard.current);
```

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
@@ -535,7 +535,7 @@ internal class PlayerInputTests : InputTestFixture
     // schemes.
     [Test]
     [Category("PlayerInput")]
-    public void PlayerInput_CanSetUpSlitKeyboardPlay()
+    public void PlayerInput_CanSetUpSplitKeyboardPlay()
     {
         var keyboard = InputSystem.AddDevice<Keyboard>();
 
--- a/Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
@@ -11,6 +11,20 @@ using UnityEngine.InputSystem.Processors;
 using Object = UnityEngine.Object;
 using Gyroscope = UnityEngine.InputSystem.Gyroscope;
 
+//what if we want to join by button but NOT directly process the input? is the previous behavior actually better?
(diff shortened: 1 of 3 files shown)


Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -11,6 +11,20 @@
...
+//why is InputSystem.Update() triggering an initial state check on the "fire" button action?
+//should the unpaired device activity check in InputUser be moved to *before* state is incorporated into the device
+//   (the current setup doesn't seem to make it possible to immediately react to the state change via actions)
+
+//TODO:
+// - fix playerinput hotplugging
+// - add message for control scheme switches
+// - add message for device changes
+// - add ability to intercept playerinputmanager player instantiation
+
```
Rene-Damm: Darn. This should not be here. Committed by accident. Reverting.

douglas-piconi (approved):

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -531,6 +531,91 @@ public void PlayerInput_CanAutoSwitchControlSchemesInSinglePlayer()
         Assert.That(listener.messages, Is.EquivalentTo(new[] {new Message("OnFire", 1f)}));
     }
 
+    // Test setup where two players both use the keyboard but with two different control
+    // schemes.
+    [Test]
+    [Category("PlayerInput")]
+    public void PlayerInput_CanSetUpSlitKeyboardPlay()
```
jechter: "Split"

jechter (approved):

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
Rene-Damm: Fixed

PR #805, opened 2019-08-23
FIX: Don't pass events for null devices (for devices which have not b…

jechter:
…een created) to `InputSystem.onEvent` callbacks.

Fixes https://fogbugz.unity3d.com/f/cases/1175878/.

If we think we _want_ onEvent callbacks for null devices, then we need to make sure all handlers correctly handle that case. But I think that's also likely going to make it harder for users, and I don't see why such events would be interesting, so I went with this solution.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2396,6 +2377,22 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+                        m_EventListeners[i](new InputEventPtr(currentEventReadPtr), device);
+
+                    // If a listener marks the event as handled, we don't process it further.
+                    if (currentEventReadPtr->handled)
+                    {
+                        eventBuffer.AdvanceToNextEvent(ref currentEventReadPtr, ref currentEventWritePtr,
+                            ref numEventsRetainedInBuffer, ref remainingEventCount, leaveEventInBuffer: false);
+                        continue;
+                    }
+                }
```
Rene-Damm, line 2402: Hmm, not sure about this one. I deliberated moving this here with myself before and ended up not doing it. While it did seem intuitively right to me at first to not invoke the callback, it means that we just silently swallow input that is indeed coming in. And the user has no way anymore of getting to it.

Whether this actually matters is debatable. Think I'm just leaning more towards the side of caution and simply saying we surface all input as is and that's it. And the `InputDevice` parameter may consequently be null.

If we do end up changing this, the documentation on `InputSystem.onEvent` needs to be updated.

douglas-piconi (approved):

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jechter: I expected you to raise this concern ;) 

I think the problem is that null device callbacks are rare, and thus people will always write code which does not expect null, which then breaks later (including ourselves, which the bug is about - our own InputUser code is throwing on this). This is why I deliberately chose this approach over fixing our code to check for null.

I guess if we think we need the events even for devices we can't handle, it might be better to add an onRawEvent, which does not pass a device? But I don't know if we'd actually need this.

jechter: Just came along https://forum.unity.com/threads/errors-during-rebinding.733082/#post-4888859

So you fixed another instance of this bug, making this two bugs caused by this  _in our own code_, found within one week. If we cannot use this API correctly, how can we expect users to? IMO, this is a bad API, and must be changed to not pass null. If we think we need this (not sure about that myself), let's offer raw events for null devices in a separate handler.

jechter:
Updated the docs and tests to match the behavior

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: Hmm, torn on this and not liking the change so much but ok, let's merge it. Reason I'm reluctant is that "surface *all* platform data we get into user land" has been one of the core design tenets of the system. Adding an extra callback for it doesn't seem like the right move.

At the end of the day `InputSystem.onEvent` is super low-level and I'm fine with it not being the most straightforward to use. In many ways, it is intended as the `onRawEvent` thing. Events are raw. They are binary dumps. `onEvent` is basically dumping a pointer to unmanaged memory in your lap. Checking whether `InputDevice` is null should be the smallest of your worries.

These days I'm wondering whether `onEvent` is actually sitting in the right place and whether it would better be moved to some LowLevel API. Its location in `InputSystem` seems to suggest it has a much wider applicability than it really has.

PR #806, opened 2019-08-26
FIX: A number of problems with HID joysticks.

Rene-Damm:
Looking into problems with the G25 Racing Wheel, I found quite a number of issues around joysticks so unfortunately, this branch ended up accumulating quite a few changes.

Most important changes:
- `HID` can now deal with sticks that have their X and Y control sitting in unrelated places in the HID input report and being of different geometry. On the G25 Racing Wheel in particular, we have a 14bit X axis with a 2-bit offset (!) and several controls later an 8-bit Y axis.
- There was a problem in `InputDeviceBuilder` with settings on controls like up/down/left/right not coming through because of an unfinished refactor of mine. Turned out to be very simple to solve and removed a bunch of code.
- The Linux joystick code had a mismatching field and thus lost data. `featureSize` was called `size` in C# and thus always 0.


douglas-piconi:
can't approve just yet, lot of red on these tests.  Other than that code looks fine

Rene-Damm:
> can't approve just yet, lot of red on these tests. Other than that code looks fine

Yeah, hit "add reviewers" a tad bit too early :) I'll go green it first.

douglas-piconi (approved):

jechter (approved):

PR #810, opened 2019-08-28
FIX: Fixed pointer coordinates in editor windows for non-mouse pointi…

jechter:
…ng devices.

Fixes https://fogbugz.unity3d.com/f/cases/1156873/


Packages/com.unity.inputsystem/InputSystem/Controls/Processors/EditorWindowSpaceProcessor.cs
```diff
@@ -32,9 +32,10 @@ public override Vector2 Process(Vector2 value, InputControl<Vector2> control)
                 return value;
 
             var command = QueryEditorWindowCoordinatesCommand.Create(value);
-            ////TODO: don't issue this on the device itself but rather on the system mouse; this way
-            ////      it's not necessary for all pointer devices to implement the IOCTL separately
-            if (control.device.ExecuteCommand(ref command) > 0)
+            // Not all pointer devices implement the editor window position IOCTL,
+            // so we try the global mouse device if available.
+            var device = Mouse.current ?? control.device;
```
Rene-Damm: I think `Mouse.current` isn't robust enough in this case. It'll just grab any mouse whereas ideally, this here would always go to the system mouse.

If we look up and cache the system mouse device ID once, we could even go and just issue the IOCTL directly on `IInputRuntime` with just the mouse ID here. So no matter what happens with the mouse InputDevice, it would do the right thing and get to the native code it needs to run here.

jechter: So, how do we define "the system mouse"?

Rene-Damm: For now it's probably good enough to just remember the first Mouse device that is native. We could probably do this in `Mouse.OnAdded`. Something like

```
internal static int s_PlatformMouseDeviceId = InputDevice.InvalidDeviceId;
protected void OnAdded()
{
    base.OnAdded();
    
    if (native && s_PlatformMouseDeviceId == InputDevice.InvalidDeviceId)
        s_PlatformMouseDeviceId = id;
}
```

Rene-Damm: Just to add that, I think the use of an IOCTL here wasn't well thought through by me. In retrospect this is one of the cases where it probably would have been better to just add a method to UnityEditor.dll that does the conversion. Maybe something we can fix >1.0.

douglas-piconi (approved):

Rene-Damm (approved):

PR #811, opened 2019-08-28
Fix: Fixed using the input system in il2cpp when managed stripping le…

jechter:
…vel is set higher then "Low".

In the higher stripping levels, we would strip control type constructors, as these types were only created from layouts through reflection. This would cause the input system to fail when set to higher stripping levels (and is a pretty serious bug). This PR explicitly marks all control types to "Preserve" and verifies that in a test.

Will try switching our test project to run in high stripping setting to see if there are more issues in a separate PR.


Rene-Damm (approved):
Awesome!

douglas-piconi (approved):

douglas-piconi:
so would this explain the weird behavior we see in tanks on IL2CPP platforms?

PR #814, opened 2019-08-28
Enable stripping in test project

jechter:
More fixes to get the Input System to work with stripping. Now will run all our tests in stripped players, so we should see when stuff like this breaks.


douglas-piconi (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ however, it has to be formatted properly to pass verification tests.
 
 - Don't pass events for null devices (for devices which have not been created) to `InputSystem.onEvent` callbacks.
 - Will close debugger input state windows, when the state is no longer valid instead of throwing exceptions.
+- Fixed using the input system in il2cpp when managed stripping level is set higher then "Low".
```
Rene-Damm: "than"

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XInputControllerWindows.cs
```diff
@@ -93,6 +93,7 @@ namespace UnityEngine.InputSystem.XInput
     /// An <see cref="XInputController"/> compatible game controller connected to a Windows desktop machine.
     /// </summary>
     [InputControlLayout(stateType = typeof(XInputControllerWindowsState), hideInUI = true)]
+    [Scripting.Preserve]
```
Rene-Damm: Makes me wonder, do you think it would make sense to derive `InputControlLayoutAttribute` from `PreserveAttribute`? Seems a little hackish and not sure it'd even work. If we did it, probably would make sense for `InputControlAttribute`, too. For the latter we'd have to find another solution to it being co-opted as a `PropertyAttribute`.

On the upside, seems like this would implicitly solve the problem for users, too.

jechter: That was my first thought as well. But it is not supported.

Rene-Damm: Ugh, sad :(

Rene-Damm (approved):

Issue #714, opened 2019-07-06
InputInteractionContext setTimeout seems to be working weird ?


gotenxds:
Hello, I'm experimenting with creating custom interactions and I'm having issues with setTimeout
This may be by lack of documentation about this but I think it is acting weird

take this code for example: 

`

        public void Process(ref InputInteractionContext context)
        {
            Debug.Log("Proccessing");

            context.SetTimeout(1);
        }
`

What I would expect to happen is that every 1 second I would get the output `Proccessing` in the console
What I get instead is just 2, the intial invocation of the action and the timeout

![image](https://user-images.githubusercontent.com/3519520/60760556-dd700280-a03f-11e9-8930-d3a9ee3430c0.png)

So I tought, maybe setting the timeout every time causing some issue and actually resets the timer ?

`

        private bool timeoutInitialized;

        public void Process(ref InputInteractionContext context)
        {
            Debug.Log("Proccessing");

            if (timeoutInitialized)
            {
                context.SetTimeout(1);
                timeoutInitialized = true;
            }
        }
`

But again, I get the same result, just two calls.

sttz:
I think I hit the same bug, it's currently not possible to set a new timeout when another timeout has just been triggered.

The issue is in `InputManager.ProcessStateChangeMonitorTimeouts()`. When a timeout expires, it calls `IInputStateChangeMonitor.NotifyTimerExpired()` inline, which in turn immediately calls `IInputInteraction.Process()`. If you try to set a new timeout during that `Process` call, it will get added to `m_StateChangeMonitorTimeouts` while `ProcessStateChangeMonitorTimeouts()` is still iterating over them.

The issue is that `ProcessStateChangeMonitorTimeouts()` caches the length of the array, meaning it will ignore newly added timeouts and its compaction strategy then truncates the array, removing them.

To test this, I removed the length caching and the timeouts started to work as expected:

```cs
        private void ProcessStateChangeMonitorTimeouts()
        {
            if (m_StateChangeMonitorTimeouts.length == 0)
                return;

            // Go through the list and both trigger expired timers and remove any irrelevant
            // ones by compacting the array.
            // NOTE: We do not actually release any memory we may have allocated.
            var currentTime = m_Runtime.currentTime - InputRuntime.s_CurrentTimeOffsetToRealtimeSinceStartup;
            var remainingTimeoutCount = 0;
            for (var i = 0; i < m_StateChangeMonitorTimeouts.length; ++i)
            {
[...]

PR #816, opened 2019-08-29
FIX: Not being able to add timeouts while processing a timeout (#714).

Rene-Damm:
Incorporated fix supplied by @sttz and added a test.


jechter (approved):

Rene-Damm:
Found some more problems with the add/remove logic of timeouts.

douglas-piconi (approved):

PR #820, opened 2019-08-30
Move samples

jechter:
This PR:

1. Moves the TanksDemo and TouchSamples out of the main project in the repo into separate project folders under `ExternalSampleProjects`.
2. Adds a new yamato script to run on every PR to import these projects to make sure they compile.
3. Changes the "promotion" yamato script to take these projects, package them as .unitypackages, create a new release on github, and upload the unitypackages as build artifacts.
4. Adds new fake "Samples" to the project to add as a proxy for these external samples, which will offer to open a link to download the unitypackage in the editor.

Notes:
- Currently uploading to github is done using a yamato secret using a github OAuth key which links to my personal github account. We should probably set up a virtual user account for github to use for this instead. I'll look into that next.

- I did not actually test the changed promotion script, as I did not want to upload a new package. But I did test a separate "publish samples" script with the same code, which only uploads the samples and will not publish a new package, and that works. So I hope the same code also works in  the promotion script, but it is possible we will need to fix it when we want to push 0.9.6-preview.


ExternalSampleProjects/ExternalSamplesUtility.cs
```diff
@@ -0,0 +1,135 @@
...
+        p.Start();
+        p.WaitForExit();
+	}
+	
+	static string[] GetAssetsToPublish()
+	{
+		return 
+			Directory.GetFileSystemEntries("Assets").Where(x => !x.Contains("ExternalSamplesUtility"))
+			.Concat(new []{"ProjectSettings/ProjectSettings.asset"})
+			.ToArray();
```
Rene-Damm, line 86: We'll probably TagManager from ProjectSettings, too (used by TouchSamples). Wondering if we should just grab everything from ProjectSettings?

Also, shouldn't this grab `Packages/manifest.json` too?

ExternalSampleProjects/publish.sh
```diff
@@ -0,0 +1,17 @@
+for Sample in ExternalSampleProjects/*/
+do
+	echo Running ${Method} on ${Sample}
+	mkdir -p ${Sample}Assets/ExternalSamplesUtility/Editor
+	ln -s ../../Packages ${Sample}Packages
```
Rene-Damm, line 5: I assume this won't work correctly. Our manifest from the project doesn't mention the input system as it is an "embedded" (or whatever it is called) package in `Packages/`.

jechter: That's why we don't copy the manifest file, but link the whole directory.

Rene-Damm: Ah right, so we can load it without depending on the package. Didn't realize that. Makes me wonder, how do we get manifest.json including the input system reference into the .unitypackage then?

ExternalSampleProjects/ExternalSamplesUtility.cs
jechter: We can add more, but I don't want to unnecessarily wipe any user settings, so I just wanted to grab the stuff we actually need. So, if TouchSamples needs TagManager, we can get that  - but then where do I find the TagManager set up for TouchSamples? Because the TagManager currently in our input system project just has the default settings in it.

As for `Packages/manifest.json` - can't grab that - it's not an asset.

ExternalSampleProjects/publish.sh
jechter: We can't get the input system reference into the unitypackage. But: If you get the unitypackage into a project using the samples mechanism, then you already have the Input System package.

ExternalSampleProjects/ExternalSamplesUtility.cs
Rene-Damm: >where do I find the TagManager set up for TouchSamples? Because the TagManager currently in our input system project just has the default settings in it.

It's still pending on PR #745. Didn't make it further yet. Now what we have a proper way to deal with per-sample project settings, would like to land that after your PR.

>As for Packages/manifest.json - can't grab that - it's not an asset.

Doh, I thought we register that as a constant GUID :/

ExternalSampleProjects/publish.sh
Rene-Damm: The user could install that .unitypackage anywhere, though, including a fresh project. But guess we're kinda out of options there. Not surprised .unitypackage and package manager don't mix. Hope that eventually `Samples` matures into a mechanism that can handle all this for us.

jechter: Yes, indeed. OTOH, the discovery mechanism for these packages makes it more likely for the user to put them into a project which has the Input System package.

ExternalSampleProjects/ExternalSamplesUtility.cs
jechter: Ok, so let's land it, and then add the TagManager

Rene-Damm (approved):

.yamato/promotion.yml
```diff
@@ -1,30 +1,24 @@
...
-    image: package-ci/win10:stable
-    flavor: b1.large
----
 promote:
   name: Promote to Production
   agent:
-    type: Unity::VM
-    image: package-ci/win10:stable
-    flavor: b1.large
+    type: Unity::VM::osx
```
douglas-piconi, line 4: @jechter What is the reason for this change?    This script pushes the package to production, really need to make sure this script works 100%.   The mac agents tend to be very unreliable.

jechter: Because the publish.sh script packing and uploading the samples is a mac shell script. I guess we'd have to turn it into a windows PowerShell script or something. Got any experience with that?

douglas-piconi (approved):

PR #826, opened 2019-09-02
Supported hid usages

jechter:
Replaced `HIDSupport.shouldCreateHID` event with a new `HIDSupport.supportedHIDUsages` property, which takes an array of supported usages.


Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HIDSupport.cs
```diff
@@ -28,29 +28,64 @@ namespace UnityEngine.InputSystem.HID
...
+        /// By default, this includes only <see cref="HID.GenericDesktop.Joystick"/>,
+        /// <see cref="HID.GenericDesktop.Gamepad"/> and <see cref="HID.GenericDesktop.MultiAxisController"/>,
+        /// but you can set this property to include any other HID usages.
+        ///
+        /// Note that currently on macOS, the only HID usages which can be enabled are
+        /// <see cref="HID.GenericDesktop.Joystick"/>, <see cref="HID.GenericDesktop.Gamepad"/>,
+        /// <see cref="HID.GenericDesktop.MultiAxisController"/>, <see cref="HID.GenericDesktop.TabletPCControls"/>,
+        /// and <see cref="HID.GenericDesktop.AssistiveControl"/>.
+        /// </remarks>
+        public static ReadOnlyArray<HIDPageUsage> supportedHIDUsages { get; set; }
```
Rene-Damm: I think the setter should make a private copy of the data. Otherwise, the array is basically readwrite for the caller.

jechter: @Rene-Damm I'm not sure I understand? Maybe I don't get how ReadOnlyArray works, but how would the caller write to the array?

Rene-Damm:
```
var usages = new HIDPageUsage[10];
HIDSupport.supportedHIDUsages = usages; // implicit conversion to ReadOnlyArray
usages[0] = ...;
```

Overall, `ReadOnlyArray` pretends to be more than it really is. Was recently trying to find a .NET standard type that could replace the thing now that we're finally in .NET 4.5 land but unfortunately, nothing seems to quite do exactly what we want :/

In the case here, with the assigning of ReadOnlyArray, the writability IMO isn't that huge a deal. From my POV, more about the API behaving in a predictable way and not allowing a scenario where the data can suddenly change behind your back.

On the other hand, ReadOnlyArray itself can lead to some surprising effects. These days I'm thinking ReadOnlyArray should really be a ref struct. Anyway, digressing.

jechter: Oh, I see. I didn't realize that when constructing a ReadOnlyArray from an array, it would keep the source array as backing storage. Yeah, makes complete sense now, I will change it.

jechter: Fixed.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HIDSupport.cs
jechter: Oops, backing array should have been private, fixing.

douglas-piconi (approved):

PR #828, opened 2019-09-02
FIX: Fixed `CallbackContext.ReadValue` throwing when invoked during d…

jechter:
…evice removal

Fixes https://fogbugz.unity3d.com/f/cases/1170449/


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1928,7 +1928,9 @@ internal TValue ReadValue<TValue>(int bindingIndex, int controlIndex, bool ignor
                     throw new InvalidOperationException(
                         $"Cannot read value of type '{TypeHelpers.GetNiceTypeName(typeof(TValue))}' from control '{control.path}' bound to action '{GetActionOrNull(bindingIndex)}' (control is a '{control.GetType().Name}' with value type '{TypeHelpers.GetNiceTypeName(control.valueType)}')");
 
-                value = controlOfType.ReadValue();
+                // Avoid reading from controls of devices which are being destroyed.
+                if (control.device.m_DeviceIndex != InputDevice.kInvalidDeviceIndex)
```
Rene-Damm: On the fence about this one. Strictly speaking, this fixes a symptom. Conceptually, what should happen is that all actions referring to the device and that are enabled, should get disabled. Then the device should be removed. Then all actions that were disabled, should get reenabled. In-between these, actions that simply reference controls on the device but aren't enabled should simply have their states wiped.

The way the code does it instead ATM is running it all in one step when the setup is already invalid. Strictly speaking, `InputAction.controls` should never contain controls that aren't added to the system.

My guess is that making it behave the way it should isn't a trivial fix, though, given the way `LazyResolveBindings` operates. Gut feeling is that, given how organically some of that stuff has grown over the many iterations of the action system, there's likely a much simpler way to do all this which cuts a lot of crap from the current logic.

jechter: @Rene-Damm hmm, yeah. 

So how about this approach (tested locally, works)?

1. Add an `InputDevice.removing` flag
2. Making bindings resolving ignore devices with `removing` flag
3. In RemoveDevice, after calling `RemoveStateChangeMonitors`, set `device.removing = true;`, and zero the device state.
4. When done removing set `device.removing = false;`, in case the device gets reused.

Rene-Damm: Internal `InputDevice.removing`, sounds reasonable to me.

We'd also be moving the `InputActionState.OnDeviceState` call almost all the way up in `InputManager.RemoveDevice`, right?

jechter: Pushed the change.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1544,6 +1551,7 @@ internal void InstallRuntime(IInputRuntime runtime)
             m_Runtime.onUpdate = OnUpdate;
             m_Runtime.onDeviceDiscovered = OnNativeDeviceDiscovered;
             m_Runtime.onPlayerFocusChanged = OnFocusChanged;
+            Application.focusChanged += b => Debug.Log($"Focus changed {b}");
```
Rene-Damm: Probably unintentional commit.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1122,6 +1124,15 @@ public void RemoveDevice(InputDevice device, bool keepOnListOfAvailableDevices =
             // Remove state monitors while device index is still valid.
             RemoveStateChangeMonitors(device);
 
+            device.removing = true;
+            if (!device.remote)
```
Rene-Damm: Why exclude remote devices?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1122,6 +1124,15 @@ public void RemoveDevice(InputDevice device, bool keepOnListOfAvailableDevices =
...
+                UnsafeUtility.MemClear(device.currentStatePtr, device.valueSizeInBytes);
```
Rene-Damm: To reset to defaults will need to copy from `defaultStatePtr`. For context, unfortunately, default state don't need to be zeroes. E.g. an 8-bit stick axis on a HID may have its 0 point at 127.

However, generally state shouldn't be mutated this way. We're basically changing inputs here in a way that bypasses the logic responsible for handling state changes.

I assume the goal is to get the action cancellations to return default values. I'm thinking that part is probably better handled in the action code instead of here as it affects cancellations in general and not just on device removal.

The fact that 'canceled' may not see a default value returned has turned out to be a problem in general. I'm guessing best thing is probably we just change that and make it always return `default(TValue)`.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2213,6 +2221,7 @@ private bool ShouldRunUpdate(InputUpdateType updateType)
             var mask = m_UpdateMask;
 #if UNITY_EDITOR
             // Ignore editor updates when the game is playing and has focus. All input goes to player.
+            // Debug.Log($"gameIsPlayingAndHasFocus {gameIsPlayingAndHasFocus} {Application.isFocused}");
```
Rene-Damm: Probably same here.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jechter: indeed. Will fix!

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jechter: Should remote devices have a valid device.currentStatePtr? At least in one of our tests, they did not.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jechter: Hmm, you mean to have context.ReadValue always return default(TValue) for canceled actions? Yes, we can do that - that  would then be much closer to my original fix (which would return default if the device was removed), and would make all the RemoveDevice changes in this change unneeded.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: >At least in one of our tests, they did not.

That is curious. Sounds like a potential bug. From the perspective of the input system, a remote device looks (or should look) identical to a local one except that one flag.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: >you mean to have context.ReadValue always return default(TValue) for canceled actions?

Yeah, I'm starting to think that should be the case.

Not sure whether it'd obsolete the `removing` change outright. One thing that's great about it is that it now runs the action cleanup while the system setup for the device is still intact. So there won't be a case where, for example, a callback may find `InputAction.controls` to suddenly have controls that shouldn't be there.

But maybe I'm too concerned there about a detail that in practice won't really matter.

jechter:
So, some back and forth on this. I have now changed this to the last fix suggested by @Rene-Damm, to just always return the default value on canceled. This is the simplest of all fixes suggested so far.

I chose to _not_ keep the logic for InputDevice.removing. We can bring it back if we find that we need it, but I'm not sure that we do (there are valid arguments for either behavior here, IMO), so I decided that the better solution is the less complex one.

Note that always returning default on canceled actions changes behavior (as can be seen in the one change I did to PlayerInput tests). I think in that case the change is better. But it might be less obviously the "right thing" in other cases, such as returning 0 for the interaction value when a Tap interaction times out because the button was held for too long. The button is still held down, but we return 0. I think that can be considered correct in that case, because we are reading the _interaction context_, not the _control_, but it's still worth pointing this out as a possible drawback of this change.

douglas-piconi (approved):

PR #837, opened 2019-09-04
XR Layout Removal

StayTalm:
XR Device Layout removal (mvoed to separate packages), as well as additional testing scenes and test utilities.

Also added Eyes and Bone feature types, for use with XR.


Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -2409,9 +2409,9 @@ private static void PerformDefaultPluginInitialization()
             Switch.SwitchSupport.Initialize();
             #endif
 
-            #if UNITY_EDITOR || UNITY_STANDALONE || UNITY_ANDROID || UNITY_IOS || UNITY_WSA
+            //#if UNITY_EDITOR || UNITY_STANDALONE || UNITY_ANDROID || UNITY_IOS || UNITY_WSA || UNITY_LUMIN
```
jechter: Why is this commented out?

StayTalm: Reverted it back, was just some WIP changes

Packages/com.unity.inputsystem/InputSystem/Editor/EditorInputControlLayoutCache.cs
```diff
@@ -135,7 +135,7 @@ public static IEnumerable<OptionalControl> GetOptionalControlsForLayout(string l
         public static Texture2D GetIconForLayout(string layoutName)
         {
             if (string.IsNullOrEmpty(layoutName))
-                throw new ArgumentNullException(nameof(layoutName));
+                return null;
```
Rene-Damm: Why?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -94,13 +104,14 @@ internal static string OnFindLayoutForDevice(int deviceId, ref InputDeviceDescri
...
-#else
+#else //UNITY_2019_3_OR_NEWER
                 if (deviceDescriptor.deviceRole == InputDeviceRole.LeftHanded || deviceDescriptor.deviceRole == InputDeviceRole.RightHanded)
                     matchedLayout = "XRController";
                 else if (deviceDescriptor.deviceRole == InputDeviceRole.Generic)
                     matchedLayout = "XRHMD";
-#endif
+#endif //UNITY_2019_3_OR_NEWER
             }
+            Debug.Log($"[TOMB] Matched Layout: {matchedLayout}");
```
Rene-Damm: Debug leftover.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -4,7 +4,9 @@
 using UnityEngine.InputSystem.Utilities;
 using System.Text;
 using UnityEngine.InputSystem.Layouts;
+#if UNITY_INPUT_SYSTEM_ENABLE_XR
```
Rene-Damm: Wouldn't it make more sense to have this #if around all the code in Plugins/XR?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
```diff
@@ -117,6 +128,9 @@ internal static string OnFindLayoutForDevice(int deviceId, ref InputDeviceDescri
             InputSystem.RegisterLayoutBuilder(() => layout.Build(), layoutName, matchedLayout);
 
             return layoutName;
+#else //UNITY_INPUT_SYSTEM_ENABLE_XR
+            return null;
```
Rene-Damm: Shouldn't hook itself in if it's not doing something.

Packages/com.unity.inputsystem/InputSystem/Editor/EditorInputControlLayoutCache.cs
StayTalm: Woops, that was to bypass an error temporarily.
Will investigate the error before comitting.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
StayTalm: That makes sense, will adjust

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
StayTalm: I'm thinking of just wrapping it around the functionality (function calls and event registration).

I don't like the idea of wrapping it around types, as it can break serialization if you temporarily lose or disable the XR module.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRTemplateBuilder.cs
StayTalm: Removed

Rene-Damm:
>I don't like the idea of wrapping it around types, as it can break serialization if you temporarily lose or disable the XR module.

Excluding `TrackedPoseDriver` is fine. For the other types, having them disappear IMO is the right thing. Say there was an #if to disable gamepad support. I'd expect it to cut `Gamepad`, not to leave it around as a non-functional API.

jechter (approved):

Rene-Damm (approved):

PR #840, opened 2019-09-05
Removed all references to NDA consoles to prepare moving console devi…

martinh-unity:
…ces to platform packages.


Packages/com.unity.inputsystem/InputSystem/Devices/Commands/InputDeviceCommand.cs
```diff
@@ -28,6 +28,8 @@ namespace UnityEngine.InputSystem.LowLevel
     public struct InputDeviceCommand : IInputDeviceCommandInfo
     {
         internal const int kBaseCommandSize = 8;
+        public const int BaseCommandSize = 8;
```
Rene-Damm, line 31: For these kind of changes, please leave comment RE intention here. Something like

```
////TODO: Remove kBaseCommandSize
```

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -264,31 +264,31 @@ public Action onProjectChange
...
+
         #endif // UNITY_EDITOR
-
-        #if UNITY_ANALYTICS || UNITY_EDITOR
-
+
         public void RegisterAnalyticsEvent(string name, int maxPerHour, int maxPropertiesPerEvent)
-        {
+        {
+            #if UNITY_ANALYTICS
```
Rene-Damm, line 273: Wondering if the above #if should just be turned into `#if UNITY_ANALYTICS` (dropping the editor case) instead.

IIRC my intention here was to send analytics from the editor even if the project has it disabled thinking that whether the user has analytics disabled for his/her own game shouldn't matter in the editor. But I'm probably just mixing up stuff here and that's not how it works.

If `UNITY_ANALYTICS` not being defined should cull all analytics code, we should probably just drop the `UNITY_EDITOR` part from the `#if UNITY_ANALYTICS || UNITY_EDITOR` defines everywhere.

jechter:
@martinh-unity could you please push your changes to a branch on Unity-Technologies/InputSystem (you have write access) and re-create this PR to use that as a source? That way our CI will pick it up and verify the PR.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
jechter: Yeah, AFAIK, this will not compile, even in the editor, when targeting a platform without analytics, hence the change.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -1,776 +0,0 @@
...
-        }
-    }
-#endif
-
-#if UNITY_EDITOR || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_WSA
-    /// <summary>
-    /// Structure of HID input reports for Switch Pro controllers.
-    /// </summary>
-    [StructLayout(LayoutKind.Explicit, Size = 20)]
-    internal struct SwitchProControllerHIDInputState : IInputStateTypeInfo
```
jechter, line 354: We want to keep this, and move it to a separate file - this is the  Switch controller HID implementation used on Desktops!

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/NPad.cs
```diff
@@ -1,776 +0,0 @@
...
-    }
-#endif
-
-#if UNITY_EDITOR || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_WSA || PACKAGE_DOCS_GENERATION
-    /// <summary>
-    /// A Nintendo Switch Pro controller connected to a desktop mac/windows PC using the HID interface.
-    /// </summary>
-    [InputControlLayout(stateType = typeof(SwitchProControllerHIDInputState), displayName = "Switch Controller (on HID)")]
-    [Scripting.Preserve]
-    public class SwitchProControllerHID : Gamepad
```
jechter, line 771: Same here.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchSupport.cs
```diff
@@ -1,35 +0,0 @@
...
-        {
-        #if UNITY_EDITOR || UNITY_SWITCH
-            InputSystem.RegisterLayout<NPad>(
-                matches: new InputDeviceMatcher()
-                    .WithInterface("Switch")
-                    .WithManufacturer("Nintendo")
-                    .WithProduct("Wireless Controller"));
-        #endif
-        #if UNITY_EDITOR || UNITY_STANDALONE || UNITY_WSA
-            InputSystem.RegisterLayout<SwitchProControllerHID>(
```
jechter, line 26: Keep this too!

jechter (changes requested):

martinh-unity:
OK, I'll create a new branch on here and PR my changes into that branch. I can then update it with the changes mentioned in this PR.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
martinh-unity: Just to clarify is the change OK? I took it from what @sambickley recommended as he found the problem when testing on PS4.

jechter:
Replaced by https://github.com/Unity-Technologies/InputSystem/pull/842

Issue #829, opened 2019-09-03
Input rebinding does not clear cached per action bindings


sttz:
I'm implementing input rebinding in my game and observed that after rebinding, the `overridePath` on the rebound action was still `null`, but the rebinding did work properly in the game.

I compared my code to the Tanks demo and noticed it was working properly there. After some digging, I realized that in the Tanks demo, the bindings are already sorted by action and the action map can therefore directly use the bindings array for the actions. However, in my case it was not properly sorted. The rebinding would update the action map's bindings but the `m_BindingsForEachAction` array did not get invalidated, causing the actions to show outdated bindings.

I'm not sure what the proper fix would be but I added `actionMap.ClearPerActionCachedBindingData();` to every `actionMap.LazyResolveBindings();` in `InputActionRebindingExtensions.cs` to invalidate the cache after rebinding has occurred and now it's working properly.

As an aside: I've set up my actions and bindings using the editor, so I'm not sure why they aren't sorted properly. Ideally, the editor would make sure they bindings are sorted and the fast path ca be used?

Rene-Damm:
Indeed, looking at it, `InputActionMap.ResolveBindings` seems to be missing a call to `ClearPerActionCachedBindingData`. I'll have a closer look.

>As an aside: I've set up my actions and bindings using the editor, so I'm not sure why they aren't sorted properly. Ideally, the editor would make sure they bindings are sorted and the fast path ca be used?

Overall, when possible, would recommend not relying on the ordering though with the current level of editor functionality it can end up being a bit tricky not to. For rebinding, if a specific binding is to be targeted, setting a mask/group is recommended.

Still, agree that the editor should make the ordering on disk reflect that in the editor.

Rene-Damm:
Should be fixed by #843.

Turns out that there was a single missing line in `Resolve` where it only cleared the control cache but not the binding cache. Refactored the code slightly to make this more robust.

PR #821, opened 2019-08-30
NEW: More API docs (and more API cleanup).

Rene-Damm:
The Big Bertha of PRs. Straight from hell.

Extremely sorry for inflicting this PR on you guys. If I hadn't simply run out of time, I would have split all this up into a sequence of PRs. Now it's all coming in as one big unreviewable package.

To at least somewhat help, I'll be leaving some inline comments for the more noteworthy changes. Summary here.

Breaking changes:
- Made `IInputRuntime` and `InputTestRuntime` internal.
  * This necessitated an API change to `InputSystem.onFindLayoutForDevice` which now receives a delegate that encapsulates the runtime dependency.
- Made `InputEditorUserSettings` internal.
- Made the unfinished `InputUserSettings` internal.
- Removed "Lefty" variant of gamepad. Turned out to not be the right way to solve the problem and I doubt anyone ever used it.
- Dropped `TValue` constraint of `InputControl` parameter for `InputProcessor<TValue>.Process` as it's simply wrong. If a processor is applied to a binding composite, the control is not guaranteed to have (and actually likely not to) the same value type as the composite. Overall, minor API change.

Changes:
- Removed `Demo`.
- Added a `CustomComposite` and `CustomDevice` sample to help with implementing custom composites and devices.
- `InputDevice.id` is now `InputDevice.deviceId`. Two-letter identifiers tend to suck.
- Started consolidating all the TryGet/Get/Find methods to cut down on all the variations.
- Dropped a few minor things from the public API here and there.
- Moved a couple methods from `InputActionAsset` to `InputActionSetupExtensions`. The idea with the latter class is that it should contain *all* methods to modify the setup of assets, maps, and actions after they have already been created (i.e. stuff you don't usually call at runtime; unlike `InputActionRebindingExtensions`).

PR #843, opened 2019-09-05
FIX: InputAction.bindings not getting refreshed after ApplyBindingOverride (#829).

Rene-Damm:
Actually ended up finding a more fundamental mistake in how `SetUpPerActionCachedBindingData` works. Left a `////FIXME` comment and made it work such that that it at least fixes the original problem.

Something to revisit after 1.0.


Rene-Damm:
Argh, managed to totally break a bunch of stuff with this change. Looking.

douglas-piconi (approved):

jechter (approved):

Rene-Damm:
Putting this one on hold for until I get to take another look after Unite. Looking at the changes again the other day, I don't think these were the right ones. As for the problem #829 I believe that is fixed by some changes that happened as part of #821.

sttz:
Can confirm that bindings on actions now correctly update after rebinding with the latest `develop`.

PR #844, opened 2019-09-05
NEW: Add sample for custom composites and custom devices.

Rene-Damm:


Rene-Damm:
Routing this in via #821.

PR #861, opened 2019-09-13
FIX: Get tanks demo to compile.

douglas-piconi:


jechter:
I believe that `FindActionMap` was added by @Rene-Damm latest PR, where he renamed the API. Did you test with the latest InputSystem branch? 

If the package didn't compile like this, it shouldn't pass the Test Sample Project config.

PR #865, opened 2019-09-17
NEW:  Allow tank movement and firing to support keyboard and mouse.

douglas-piconi:


jechter:
1. Why did you add these bindings using Axis Composites, if you have only one binding to add to the composite?
2. This breaks the rebinding screen.
3. The rebinding screen already kind of solves this problem (you can just bind to the keyboard). Why add hardcoded binding for the keyboard then? 

I guess if we want to make the keyboard experience nicer, we could consider adding a popup allowing you to choose different presets for the rebinding screen. That would actually be pretty nice.

PR #883, opened 2019-09-26
FIX: Fixed a bug where multiple composite bindings for the same contr…

jechter:
…ols but on different action maps would throw exceptions.

Fixes https://fogbugz.unity3d.com/f/cases/1183314/


Rene-Damm (approved):
Great!

douglas-piconi (approved):

PR #886, opened 2019-09-27
FIX: Multiple minor UI glitches.

Rene-Damm:
- Fixes "Listen" button not being wide enough for its text on 2019.3.
- Fixes multiple items in the "Layouts" branch of the input debugger receiving the same ID.
- Fixes composites leading to duplicate mentions of controls in the input debugger.


Rene-Damm:
Damn... forgot to change to `develop`... out of practice :)

PR #887, opened 2019-09-27
FIX: Multiple minor UI glitches.

Rene-Damm:
- Fixes "Listen" button not being wide enough for its text on 2019.3.
- Fixes multiple items in the "Layouts" branch of the input debugger receiving the same ID.
- Fixes composites leading to duplicate mentions of controls in the input debugger.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -7,11 +7,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
+## [1.0.0-preview2] - 2019-?-??
+
+### Fixed
+
+- Selecting a layout in the input debugger no longer selects its first child item, too.
+
+#### Action
+
+- The text on the "Listen" button is no longer clipped off on 2019.3.
+- Controls bound to actions through composites no longer as duplicates in the input debugger.
+
 ## [1.0.0-preview] - 2019-9-20
 
 ### Fixed
 
--Will now close Input Action Asset Editor windows from previous sessions when the corresponding action was deleted.
+- Will now close Input Action Asset Editor windows from previous sessions when the corresponding action was deleted.
 - Fixed an issue where Stick Controls could not be created in Players built with medium or high code stripping level enabled.
 - Fixed incorrect default state for axes on some controllers.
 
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -7,11 +7,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
+## [1.0.0-preview2] - 2019-?-??
```
jechter: Is `preview` followed by `preview2` or `preview1`? (I have been using `preview1` in my currently open PRs). Also, you will need to update your package.json to get tests green likely.

jechter: Hmm, if the `com.unity.package-manager-doctools` package we are using uses the scheme correctly, then we are both doing it wrong, and it should be: `1.0.0-preview.1` (with a dot)

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -7,11 +7,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+
+### Fixed
+
+- Selecting a layout in the input debugger no longer selects its first child item, too.
+
+#### Action
+
+- The text on the "Listen" button is no longer clipped off on 2019.3.
+- Controls bound to actions through composites no longer as duplicates in the input debugger.
```
jechter: This sentence a verb.

jechter (approved):
Some comments, but code changes look fine.

Packages/com.unity.inputsystem/CHANGELOG.md
jechter: I will fix it correctly (using -preview.1) in https://github.com/Unity-Technologies/InputSystem/pull/885, then we can land that and merge it to the other PRs

Rene-Damm: >I will fix it correctly (using -preview.1) in #885, then we can land that and merge it to the other PRs

Awesome.

>Hmm, if the com.unity.package-manager-doctools package we are using uses the scheme correctly, then we are both doing it wrong, and it should be: 1.0.0-preview.1 (with a dot)

Looking at the semver spec, looks like `preview.1` is indeed the correct way.

>Is preview followed by preview2 or preview1?

Good question. Given that everything else in the version is zero-based, should probably be 1 and not 2.

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: :)

douglas-piconi (approved):

PR #888, opened 2019-09-30
FIX: Layout overrides corrupting layout state.

Rene-Damm:
I got several questions at Unite around distinguishing devices based on how they are used (e.g. orientations) and added a sample that uses a layout override to distinguish a "Player1" gamepad from a "Player2" gamepad. However, doing so exposed multiple bugs with layout overrides.

So, here we go. New sample plus a number of fixes to layout overrides.


Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPickerDropdown.cs
```diff
@@ -197,6 +197,10 @@ private void AddDeviceTreeItemRecursive(InputControlLayout layout, AdvancedDropd
 
                     AddCharacterKeyBindingsTo(byCharacterGroup, keyboard);
                     AddPhysicalKeyBindingsTo(byLocationGroup, keyboard, searchable);
+
+                    // AnyKey won't appear in either group. Add it explicitly.
+                    AddControlItem(deviceItem, null,
+                        layout.FindControl(new InternedString("anyKey")).Value, layout.name, null, searchable);
```
Rene-Damm, line 203: Minor, unrelated fix. When I switched the control picker to a special path for keyboards, I didn't realize that now `anyKey` got culled from the list. This change adds the control back to the list.

jechter (approved):

douglas-piconi (approved):

PR #890, opened 2019-10-01
Ios gamepads support

jechter:
Now that  iOS 13 supports xbox one and ps4 controllers, this PR makes them be detected as such (and not as generic gamepads), so you can get correct button names, etc.

Not sure if we should land this now, or post 1.0?


douglas-piconi (approved):

Rene-Damm:
>Not sure if we should land this now, or post 1.0?

IMO it's fine landing it at this point.

Rene-Damm (approved):

PR #910, opened 2019-10-10
CHANGE: Move XR layouts (except Magic Leap) back into input system.

Rene-Damm:
### Why?
Moving them out at this point was a mistake. We operated under the assumption that updated releases of the various packages would immediately follow suit but in practice, we underestimated the gap that would result. To bridge this over, we're putting the layouts back into the main package such that they can be overridden and ultimately replaced individually by packages as they become available.

### Changes

- Add layouts for Oculus, Daydream, OpenVR, and WMR devices back to the repo
- Move several controls up into `XRHMD` and `XRController`
- Introduce concept of `TrackedDevice`
- Adding missing noise flags to respective controls on XR devices (`noisy = true`)


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/GoogleVR.cs
```diff
@@ -0,0 +1,77 @@
+#if ENABLE_VR && !DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR
```
Rene-Damm: @StayTalm 
So the `#if ENABLE_VR` part I'm not 100% sure about. My thinking here was that with the package dependency broken/inverted for now, `#if UNITY_INPUT_SYSTEM_ENABLE_XR` doesn't seem like the right check. Seemed to me that if someone flipped the VR checkbox in player settings, the XR stuff here should come live in the input system. Does that make sense?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -1,35 +1,66 @@
...
+        public Vector3Control rightEyePosition { get; private set; }
+        [InputControl]
+        [Preserve]
+        public QuaternionControl rightEyeRotation { get; private set; }
+        [InputControl]
+        [Preserve]
+        public Vector3Control centerEyePosition { get; private set; }
+        [InputControl]
+        [Preserve]
+        public QuaternionControl centerEyeRotation { get; private set; }
```
Rene-Damm, line 34: @StayTalm
Users have been asking why no controls are available at the HMD level. I moved the tracking-related controls (`isTracked`, etc) all the way up to `TrackedDevice` (same for `XRController`) and pulled the left/right/center eye controls (position&rotation) up to the HMD level. Make sense?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/GoogleVR.cs
Rene-Damm: The individual `DISABLE_BUILTIN_INPUT_SYSTEM_XXX` would be for the packages to get rid of the built-in classes.

Rene-Damm:
Broke some XR tests. I'll fix those up as part of finalizing the PR.

StayTalm (approved):
LGTM, added defines that will remove the built-in types with the correct versions.

PR #915, opened 2019-10-13
Fix Keyboard indexer being unable to index the last key

uwx:
The m_Keys array is shifted left by 1 in comparison to enum, and the indexer incorrectly checks for a number one too small for the range. 
This fix makes it possible to do keyboard[Key.OEM5] without having to use Keyboard.allKeys. Also, the documentation was incorrectly stating how to index the allKeys array.


Rene-Damm:
Thank you!

PR #918, opened 2019-10-16
FIX: Compile errors when using C# reserved identifiers in assets.

Rene-Damm:
Fixes https://fogbugz.unity3d.com/f/cases/1189861/ https://issuetracker.unity3d.com/product/unity/issues/guid/1189861/.


douglas-piconi (approved):

jechter (approved):
Makes sense. What if I name the class itself like a keyword, though? Anything preventing that atm?

Rene-Damm:
> Makes sense. What if I name the class itself like a keyword, though? Anything preventing that atm?

Agree it should protect against that. Improved a bit to at least handle it for class names. Will leave making it more robust for later. For certain names, the generator should probably raise errors and not generate code.

PR #920, opened 2019-10-17
FIX: Deactivated input bindings leading to initialization errors (case 1187377, 1189867).

Rene-Damm:
Fixes 
- https://fogbugz.unity3d.com/f/cases/1187377/, https://issuetracker.unity3d.com/issues/input-system-composite-bindings-are-not-responsive-when-one-of-their-path-is-set-to-an-empty-string-or-null
- https://fogbugz.unity3d.com/f/cases/1189867/, https://issuetracker.unity3d.com/issues/input-system-action-with-no-bindings-in-one-scheme-remains-disabled-in-other-schemes

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
@@ -73,7 +73,7 @@ namespace UnityEngine.InputSystem
         /// </remarks>
         public void StartWithArraysFrom(InputActionState state)
         {
-            Debug.Assert(state != null);
+            Debug.Assert(state != null, "Received null state");
 
             maps = state.maps;
             interactions = state.interactions;
@@ -113,7 +113,7 @@ namespace UnityEngine.InputSystem
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1809:AvoidExcessiveLocals", Justification = "TODO: Refactor later.")]
         public unsafe void AddActionMap(InputActionMap map)
         {
-            Debug.Assert(map != null);
+            Debug.Assert(map != null, "Received null map");
 
             var actionsInThisMap = map.m_Actions;
             var bindingsInThisMap = map.m_Bindings;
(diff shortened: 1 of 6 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
```diff
@@ -474,7 +474,7 @@ void AddEntry(InputAction action, PlayerInput.ActionEvent actionEvent)
             var selectedDefaultActionMap = !string.IsNullOrEmpty(playerInput.defaultActionMap)
                 ? asset.FindActionMap(playerInput.defaultActionMap)
                 : null;
-            m_SelectedDefaultActionMap = 0;
+            m_SelectedDefaultActionMap = asset.actionMaps.Count > 0 ? 1 : 0;
```
Rene-Damm, line 477: Minor tweak to PlayerInputEditor so that if you select an .inputactions asset, it will default to the first action map in the asset (except if you already had an action map selected and there's one with the same name in the new asset).

Obsoletes the need to not forget to set the default action map. In practice, the majority of users will want an action map to be enabled by PlayerInput.

Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
```diff
@@ -227,132 +218,140 @@ public unsafe void AddActionMap(InputActionMap map)
...
+                            else
+                            {
+                                // Search globally.
+                                numControls = InputSystem.FindControls(path, ref resolvedControls);
                             }
-                        }
-                        else
-                        {
-                            // Search globally.
-                            numControls = InputSystem.FindControls(path, ref resolvedControls);
```
Rene-Damm, line 349: The diff above looks scary. But the actual change is pretty straightforward and just rearranges the existing code a bit.

Before:

```
1. Default-initialize *part* of the binding state.
2. Check whether we should skip the binding. If so, early out.
3. Resolve action. If action says we should skip the binding, early out.
4. Actually initialize the binding.
```

This left bindings that were skipped with a partially initialized state. In the revised version, bindings are never just skipped.

Now:

```
1. Set up default values for everything that we will write into the binding state.
2. Resolve the binding's action (if any).
3. Now check whether we need to resolve the binding. If yes, then update all the default values above with initialized values.
4. Write the final values into the binding.
```

This initializes the binding state fully and also makes sure that *every* binding, regardless of whether it is used or not, ends up with a correct reference to the action it is associated with. This fixes a number of issues with binding data getting lost through various operations.

douglas-piconi (approved):

jechter (approved):

PR #928, opened 2019-10-22
FIX: UI input module finding unrelated actions when switching assets (case 1193287).

Rene-Damm:
Fixes https://fogbugz.unity3d.com/f/cases/1193287/

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.cs
@@ -754,7 +754,8 @@ internal class UITests : InputTestFixture
                     {
                         ""name"" : ""Gameplay"",
                         ""actions"" : [
-                            { ""name"" : ""Point"" }
+                            { ""name"" : ""Point"" },
+                            { ""name"" : ""Navigate"" }
                         ]
                     },
                     {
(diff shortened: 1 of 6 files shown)


douglas-piconi (approved):

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -745,13 +740,59 @@ public IEnumerator JoystickActions_CanDriveUI()
...
+        go.AddComponent<EventSystem>();
+        var uiModule = go.AddComponent<InputSystemUIInputModule>();
+
+        const string kActions = @"
+            {
+                ""maps"" : [
+                    {
+                        ""name"" : ""Gameplay"",
+                        ""actions"" : [
+                            { ""name"" : ""Point"" }
```
jechter: Maybe add a Gameplay/Navigate here, so we also test to make sure that reassigning will pick the right one?

jechter (approved):

Assets/Tests/InputSystem/Plugins/UITests.cs
Rene-Damm: Done

PR #931, opened 2019-10-24
FIX: Custom devices lost on domain reload (case 1192379).

Rene-Damm:
First piece of `[InitializeOnLoad]` code accessing the input system would cause devices that couldn't be instantly recreated to be lost even if a needed `RegisterLayout` call would come in later during initialization after domain reload.


douglas-piconi (approved):

jechter (approved):

Rene-Damm:
@jechter I found my changes broke some tests and were actually partially incorrect. Added some more changes to attempt to properly fix the problem. Also added some docs to add some more details about how devices get (re)created.

PR #932, opened 2019-10-24
FIX: InputAction.triggered missing from manual.

### LINKED ISSUE #923: Polling API(s) are undocumented
I believe for my use the polling API makes more sense than the message callbacks. (Or at least I prefer "pulling" data where possible rather than having it pushed.)

However using the polling API is not exactly documented which making it a bit tricky, and relatedly programatically accessing the generate C# code for an input system is also seemingly undocumented.

Did I just miss this? I don't think so?

Rene-Damm:
Fixes #923.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/Actions.md
+++ b/Packages/com.unity.inputsystem/Documentation~/Actions.md
@@ -283,7 +283,7 @@ InputSystem.onActionChange +=
 
 #### Polling Actions
 
-Instead of using callbacks, it might be simpler sometimes to poll the value of an Action where you need it in your code. You can poll the value of an Action using[`InputAction.ReadValue<>()`](../api/UnityEngine.InputSystem.InputAction.html#UnityEngine_InputSystem_InputAction_ReadValue__1):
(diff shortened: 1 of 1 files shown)


douglas-piconi (approved):

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -295,13 +295,39 @@ Instead of using callbacks, it might be simpler sometimes to poll the value of a
...
 
-    void OnUpdate()
+    void Update()
     {
         var moveDirection = moveAction.ReadValue<Vector2>();
         position += moveDirection * moveSpeed * Time.deltaTime;
     }
 ```
 
+For button-type actions, you can also use [`InputAction.triggered`](../api/UnityEngine.InputSystem.InputAction.html#UnityEngine_InputSystem_InputAction_triggered) which will true if at any time in the current frame, the action was performed.
```
jechter: "which will true"? which will _be_ true?

jechter: if at any time in the current frame, the action was performed. -> 
if the action was performed at any time in the current frame.

Rene-Damm: Fixed

PR #938, opened 2019-10-31
FIX: EditorWindow space not appearing to work (case 1191859).

Rene-Damm:
Fixes https://fogbugz.unity3d.com/f/cases/1191859/

Also added a simple demo for how to use the input system in EditorWindow code.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -2274,9 +2274,15 @@ namespace UnityEngine.InputSystem
                 var layout = TryFindMatchingControlLayout(ref m_AvailableDevices[i].description, id);
                 if (IsDeviceLayoutMarkedAsSupportedInSettings(layout))
                 {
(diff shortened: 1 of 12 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2274,9 +2274,15 @@ private void AddAvailableDevicesThatAreNowRecognized()
...
-                        deviceFlags: m_AvailableDevices[i].isNative ? InputDevice.DeviceFlags.Native : 0);
+                    try
+                    {
+                        AddDevice(m_AvailableDevices[i].description, false,
+                            deviceId: id,
+                            deviceFlags: m_AvailableDevices[i].isNative ? InputDevice.DeviceFlags.Native : 0);
+                    }
+                    catch (Exception)
+                    {
+                    }
```
Rene-Damm, line 2285: This is an unrelated tweak to a change I just landed. Some HID devices led to exception spam in the console. The code here should just go and ignore those devices.

douglas-piconi (approved):

PR #939, opened 2019-10-31
FIX: Touch control scheme associations in default actions.

Rene-Damm:
Continuing to tweak the default actions. Some touch bindings were not associated with the touch control scheme. Started adding an XR scheme but there's some controls that with the currently available built-in tools can't be satisfied properly.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XInputController.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XInputController.cs
@@ -65,8 +65,8 @@ namespace UnityEngine.InputSystem.XInput
         [InputControl(name = "leftTrigger", shortDisplayName = "LT")]
         [InputControl(name = "rightTrigger", shortDisplayName = "RT")]
         // This follows Xbox One conventions; on Xbox 360, this is start=start and select=back.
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XInputController.cs
```diff
@@ -65,8 +65,8 @@ public class XInputController : Gamepad
         [InputControl(name = "leftTrigger", shortDisplayName = "LT")]
         [InputControl(name = "rightTrigger", shortDisplayName = "RT")]
         // This follows Xbox One conventions; on Xbox 360, this is start=start and select=back.
-        [InputControl(name = "start", displayName = "Menu")]
-        [InputControl(name = "select", displayName = "View")]
+        [InputControl(name = "start", displayName = "Menu", alias = "menu")]
+        [InputControl(name = "select", displayName = "View", alias = "view")]
```
Rene-Damm, line 69: Minor, unrelated fix. There's an open bug about Xbox One controls not resolving when using the "menu" and "view" names. These are already exposed as getters under those names but the controls were missing aliases to also allow lookup of control by those names.

douglas-piconi (approved):

PR #943, opened 2019-11-03
FIX: Typo in Doc/Components.md

SiunCyclone:
excplusively -> exclusively


Rene-Damm:
Hey @SiunCyclone, thank you for the fix!

I've folded your change into #946. Hope that's okay. Reason is that all PRs need to go into the `develop` branch -- which, however, we haven't documented anywhere.

PR #945, opened 2019-11-06
FIX: InlinedArray.RemoveAt sometimes removing the wrong element (case 1196143).

Rene-Damm:
[FogBugz](https://fogbugz.unity3d.com/f/cases/1196143/) | [IssueTracker](https://issuetracker.unity3d.com/issues/input-system-last-event-handler-is-removed-when-removing-a-middle-event-handler-in-script)


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/WindowsMR.cs
```diff
@@ -13,7 +13,7 @@ namespace UnityEngine.XR.WindowsMR.Input
     public class WMRHMD : XRHMD
     {
         [Preserve]
-        [InputControl(noisy = true)]
+        [InputControl]
```
Rene-Damm, line 16: Unrelated fix for which I didn't bother opening a PR (esp. since this XR stuff has no associated tests). `userPresence` should not be classified as noisy. Was erroneously added by me.

PR #946, opened 2019-11-06
FIX: LayoutNotFoundExceptions from ToHumanReadableString.

Rene-Damm:
Ended up suppressing the exception in a number of places.


Packages/com.unity.inputsystem/package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "com.unity.inputsystem",
 	"displayName": "Input System",
-	"version": "1.0.0-preview.2",
+	"version": "1.0.0",
```
jechter: 💖

jechter (approved):

PR #947, opened 2019-11-08
Tech writer review

keiramoon:
I've gone over all the .md files. Most of them needed additional changes, a few didn't. One question needs follow-up - in Migration.md, for  UnityEngine.Input.GetAxis, there's a missing word:

> There is no global setup corresponding exactly to virtual axis setups in the old Input Manager settings. Instead, you can create sets of Input Actions as independent Assets, or put the, directly on your C# components.

There should be something after the "or put the," in the last sentence, I'm guessing.

Either way, the content is good to go!


Rene-Damm:
> There should be something after the "or put the," in the last sentence, I'm guessing.

Think that was meant to read "or put the Input Actions directly on your C# components".

Rene-Damm:
Routed in via #955.

Issue #902, opened 2019-10-07
Could not create a device for "vJoy"


Rallix:
Every time I open a Unity project with the Input System, the following error appears:

```
Could not create a device for 'Shaul Eizikovich vJoy - Virtual Joystick (HID)' (exception: System.NotImplementedException: The method or operation is not implemented.
  at UnityEngine.InputSystem.Utilities.PrimitiveValue.FromString (System.String value) [0x0013a] in F:\Unity\Pyromance\Library\PackageCache\com.unity.inputsystem@1.0.0-preview\InputSystem\Utilities\PrimitiveValue.cs:407 
  at UnityEngine.InputSystem.Utilities.NamedValue.ParseParameter (System.String parameterString, System.Int32& index) [0x0015d] in F:\Unity\Pyromance\Library\PackageCache\com.unity.inputsystem@1.0.0-preview\InputSystem\Utilities\NamedValue.cs:157 
  at UnityEngine.InputSystem.Utilities.NamedValue.ParseMultiple (System.String parameterString) [0x00050] in F:\Unity\Pyromance\Library\PackageCache\com.unity.inputsystem@1.0.0-preview\InputSystem\Utilities\NamedValue.cs:98 
  at UnityEngine.InputSystem.Layouts.InputControlLayout+Builder+ControlBuilder.WithParameters (System.String parameters) [0x00014] in F:\Unity\Pyromance\Library\PackageCache\com.unity.inputsystem@1.0.0-preview\InputSystem\Controls\InputControlLayout.cs:666 
  at UnityEngine.InputSystem.HID.HID+HIDLayoutBuilder.Build () [0x001ac] in F:\Unity\Pyromance\Library\PackageCache\com.unity.inputsystem@1.0.0-preview\InputSystem\Plugins\HID\HID.cs:375 
  at UnityEngine.InputSystem.HID.HID+<>c__DisplayClass12_0.<OnFindLayoutForDevice>b__0 () [0x00000] in F:\Unity\Pyromance\Library\PackageCache\com.unity.inputsystem@1.0.0-preview\InputSystem\Plugins\HID\HID.cs:185 
[...]

Issue #871, opened 2019-09-21
NotImplementedException when clicking on the binding path dropdown


sodacolor:
When I click on this dropdown (see below) I get a console error saying `NotImplementedException: The method or operation is not implemented.` and nothing else happens.
![image](https://user-images.githubusercontent.com/48622577/65376187-fd08e580-dc94-11e9-8938-0840b21e79b2.png)

Rallix:
For me, there's a `NotImplementedException` when I try to add a device in _Project Settings > Input System Package > Supported Devices > (+) Choose to add to list_:

```
NotImplementedException: The method or operation is not implemented.
UnityEngine.InputSystem.Utilities.PrimitiveValue.FromString (System.String value) (at Library/PackageCache/com.unity.inputsystem@1.0.0-preview/InputSystem/Utilities/PrimitiveValue.cs:407)
UnityEngine.InputSystem.Utilities.NamedValue.ParseParameter (System.String parameterString, System.Int32& index) (at Library/PackageCache/com.unity.inputsystem@1.0.0-preview/InputSystem/Utilities/NamedValue.cs:157)
UnityEngine.InputSystem.Utilities.NamedValue.ParseMultiple (System.String parameterString) (at Library/PackageCache/com.unity.inputsystem@1.0.0-preview/InputSystem/Utilities/NamedValue.cs:98)
UnityEngine.InputSystem.Layouts.InputControlLayout+Builder+ControlBuilder.WithParameters (System.String parameters) (at Library/PackageCache/com.unity.inputsystem@1.0.0-preview/InputSystem/Controls/InputControlLayout.cs:666)
UnityEngine.InputSystem.HID.HID+HIDLayoutBuilder.Build () (at Library/PackageCache/com.unity.inputsystem@1.0.0-preview/InputSystem/Plugins/HID/HID.cs:375)
UnityEngine.InputSystem.HID.HID+<>c__DisplayClass12_0.<OnFindLayoutForDevice>b__0 () (at Library/PackageCache/com.unity.inputsystem@1.0.0-preview/InputSystem/Plugins/HID/HID.cs:185)
UnityEngine.InputSystem.Layouts.InputControlLayout+Collection.TryLoadLayoutInternal (UnityEngine.InputSystem.Utilities.InternedString name) (at Library/PackageCache/com.unity.inputsystem@1.0.0-preview/InputSystem/Controls/InputControlLayout.cs:1811)
[...]

PR #949, opened 2019-11-09
FIX: HID devices with 32 bit reportSize failing to load (#871, #902)

MellowArpeggiation:
Tested with the Official Wii U Gamecube Controller Adapter. I also profiled memory usage in editor, since whilst working on this I did cause the editor to crash a few times due to introduced memory leaks.


jimon:
Hi, thanks a lot for taking your time to fix them, appreciate greatly! As I'm currently making sure that whole stack supports 32 fields better, I think it makes sense to incorporate this PR as part of #1263, so I'm closing this one for now.

PR #951, opened 2019-11-11
CHANGE: Move tests back into package.

Rene-Damm:
Rationale:
- Both upm-ci and tests on trunk now expect tests to be present in packages.
- `createSeparatePackages` does not work for us as our tests have package dependencies that our normal package does not have (see branch `separate-tests-package`).
- Trunk tests have been refactored such that excluding the test requirement is now reserved for just the NUnit package (making adding an exclusion for us very ugly and likely to be rejected with good reason).
- Without being listed under `testables` our C# test code (which sits in a separate DLL) *will not get compiled*. The .cs files *will* get downloaded and imported (which sucks) but it's 38 files so guess we can ignore the problem for now and wait for the testing mechanisms for packages to mature.

Or in short, working around the requirement at this point seemed like a worse option than just complying with it as is.


Rene-Damm:
Abandoning this one too. Same problems as with `createSeparatePackages`. Going to create a dummy test to satisfy the damn tooling (that mandates tests yet seems to have no ability to actually deal with our tests) and call it a day.

PR #952, opened 2019-11-11
NEW: Add integration test that uses native runtime.

Rene-Damm:
This both satisfies the requirement to have tests inside a verified package *and* actually adds a useful sanity check on trunk where we run through a simple input update loop and make sure that we can push events into the native runtime and have them come back out and get processed.


jechter (approved):

Packages/com.unity.inputsystem/Tests/IntegrationTests/IntegrationTests.cs
```diff
@@ -0,0 +1,24 @@
+using NUnit.Framework;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.LowLevel;
+
```
jechter, line 4: maybe it makes sense to leave a comment here (or maybe better a readme.txt file in this folder?) pointing out to critical eyes that there are many more tests outside of the package?

Rene-Damm: Doh shit, that's what I planned to do but looks like I just plain forgot. Thanks for catching it. Adding.

Rene-Damm: Fixed

PR #962, opened 2019-11-21
FIX: Correct the maxValue when reading a multi-bit axis

jamre:
When reading a multi-bit axis, the returned float (or double) value will never return 1.0.

For example, an axis with 7 bits and a raw value range of [0, 127], will return [0, 0.9921875].

This change fixes the maxValue calculation so that it will return [0, 1].

The maxValue for writing was already correct.


Rene-Damm:
Routed to `develop` via #1219.

Thank you for the fix @jamre!

PR #966, opened 2019-11-27
NEW: Add controls for combined modifiers and method to look up keys by display name.

Rene-Damm:
Two small additions to address shortcomings that have become evident in the `Keyboard` API.

* Add `shiftKey`, `ctrlKey`, and `altKey` so that it is not necessary to always bind both left and right versions if no distinction is desired.
* Add `FindKeyOnCurrentKeyboardLayout` as a simple helper to make it obvious how to, for example, go from "q" to the key that generates "q" on the current keyboard layout.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/Keyboard.md
+++ b/Packages/com.unity.inputsystem/Documentation~/Keyboard.md
@@ -11,7 +11,7 @@ You can look up keys based on the character they produce using [Control paths](C
 >__Note__
 >* Keyboards usually have hardware limitations on both the number of simultaneous keypresses they report, and the combinations of keys they support. This means that certain simultaneous keypresses may not register correctly. For example, a given keyboard may report a simultaneous press of the "QWERT" keys correctly but may not report "QWERA" correctly.
 >* At the moment, the new Input System doesn't support on-screen keyboards. For now, please Unity's existing API in `UnityEngine.TouchScreenKeyboard`.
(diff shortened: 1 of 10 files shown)


Packages/com.unity.inputsystem/Documentation~/Keyboard.md
```diff
@@ -11,6 +11,7 @@ You can look up keys based on the character they produce using [Control paths](C
 >__Note__
 >* Keyboards usually have hardware limitations on both the number of simultaneous keypresses they report, and the combinations of keys they support. This means that certain simultaneous keypresses may not register correctly. For example, a given keyboard may report a simultaneous press of the "QWERT" keys correctly but may not report "QWERA" correctly.
 >* At the moment, the new Input System doesn't support on-screen keyboards. For now, please Unity's existing API in `UnityEngine.TouchScreenKeyboard`.
+>* At the moment, Unity platform backends generally do not support distinguishing between multiple keyboards. While the Input System supports having arbitray many [`Keyboard`](../api/UnityEngine.InputSystem.Keyboard.html) devices at any point, platform backends will generally only report a single keyboard and route input from all attached keyboards to the one keyboard device. Supporting multiple keyboards at the platform level is on the TODO list.
```
jechter: I would avoid talking about future plans in the docs. That often only leads to embarrassment when someone reads it in a few years.

jechter (approved):

Packages/com.unity.inputsystem/Documentation~/Keyboard.md
Rene-Damm: Fixed.

PR #975, opened 2019-12-11
NEW: Add sample for virtual mouse cursors.

Rene-Damm:
Adds a `Gamepad Mouse Cursor` sample with a reusable `VirtualMouseInput` component.

image
image

I've opted to keep this as just a sample for now. While one can argue that this functionality should come with the input system itself (and I'd agree), I'm not sure the mechanism I chose to implement this is the best one. So, was thinking, let's just leave it as a sample for now and see how it fares.

Contains a semi-related change that adds a `mode` parameter to `Vector2Composite` which allows switching to analog mode. The `normalize` parameter on the composite has been obsoleted.

image


Rene-Damm:
Still looking for ways to simplify this. Ideally, we'd do all this entirely in bindings and then make the problem of "display image at current cursor position" an entirely separate problem. We'd not need a separate component for this at all in that case.

Rene-Damm:
> Still looking for ways to simplify this. Ideally, we'd do all this entirely in bindings and then make the problem of "display image at current cursor position" an entirely separate problem. We'd not need a separate component for this at all in that case.

Decided to just "demote" it to being a sample only and keep the current approach for now. Let's see how this fares and whether we can do this better at a later point.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector2Composite.cs
```diff
@@ -67,17 +67,37 @@ public class Vector2Composite : InputBindingComposite<Vector2>
...
 
+        /// <summary>
+        /// If true (default), the up/down/left/right inputs will be treated as analog meaning that
+        /// their actual values will be used. If this is false, the values will be read as buttons
+        /// meaning that instead if an input value is below the button press threshold, the value
+        /// that is used is 0 and otherwise the value that is used is 1.
+        /// </summary>
+        /// <value>Whether to treat part bindings as analog controls.</value>
+        public bool analog = true;
+
```
jechter, line 113: 1. Are we ok with changing the default behavior here?
2. Does it make sense to have a normalize and an analog flag? If normalize is on, that kind of cancels out the effect of analog. Is it a desirable option to have analog and normalized enabled? Or should we have an enum with three modes (analog, digital, digital-normalized)?

Rene-Damm: >Are we ok with changing the default behavior here?

Was deliberating with myself, too. Leaning towards yes. Reasoning:

1) The current default behavior has caught users by surprise.
2) The vast majority of existing bindings are things like WASD and they are unaffected.
3) Restoring the old behavior is easy.

>Does it make sense to have a normalize and an analog flag? If normalize is on, that kind of cancels out the effect of analog. Is it a desirable option to have analog and normalized enabled? Or should we have an enum with three modes (analog, digital, digital-normalized)?

Hmm, good question. The enum mode toggle sounds like a good idea. Agree that going with analog makes the value of the normalize option (especially as it's on by default) questionable.

Rene-Damm:
Postponing to after 1.0.0-preview.4.

Changes to make:

1) Switch `Vector2Composite` config to an enum (may have to perserve "normalize" flag for now to keep API compatibility).
2) Make it more obvious in the GamepadMouseCursor sample that it's a software cursor (add TODO that we may want to give an option to use the system mouse device instead and use WarpMouseCursor instead) and the implications of that (you may not want to use it if there's a system mouse, etc).

jechter (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector2Composite.cs
```diff
@@ -1,8 +1,14 @@
+using System;
+using UnityEditor;
```
sbeca, line 2: @Rene-Damm: At a quick skim, this `using UnityEditor;` looks like it would cause a compile error in builds so should be moved into the `#if UNITY_EDITOR` block below.

Rene-Damm: Good catch. No idea why this didn't break on the farm. Fixed in https://github.com/Unity-Technologies/InputSystem/commit/06d8b42a81d6b83e9926fd83bd05d15456bccea7.

PR #979, opened 2019-12-11
CHANGE: Make Vector2Composite treat controls as analog by default.

Rene-Damm:
Unlike `AxisComposite`, which already uses the actual values from controls, `Vector2Composite` ATM treats every part control as a button and thus uses 1 and 0 values only. This is usually undesirable.

If made the default behavior be for source values to come through and added an `analog` parameter to allow reverting to the old behavior. For classic WASD, it makes no difference as keys themselves are digital controls.


Rene-Damm:
Folded into #975.

PR #981, opened 2019-12-16
Support for XRSDK Subcontrol Modifications & Public Serialization Structures

StayTalm:
- Added a Name for the TPD controls for better functionality with the InputDebugger
- Added support in the XRLayoutBuilder so that subcontrols can be modified and inserted by XRSDK
- Made Serialization structures public for externally implemented types that need access to device descriptors.


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -102,6 +103,7 @@ void BindRotation()
         {
             if (!m_RotationBound && m_RotationAction != null)
             {
+                m_RotationAction.Rename($"{gameObject.name} - TPD - Rotation");
```
Rene-Damm, line 106: Alternatively, do the renaming on the serialized data? e.g. in `TrackedPoseDriverEditor` or `OnValidate`.

Rene-Damm (approved):

PR #993, opened 2020-01-07
NEW: Significantly extend and refine rebinding UI support.

Rene-Damm:
- Extends and overhauls API for rebinds and displaying binding strings
- Adds example for rebinding UI including a ready-made prefab that can be reused in user projects

image

image

- Adds another example demonstrating how to show in-game UI hints

image

- `PlayerInput` now has a new `Controls Changed` event that is triggered for any changes to bound controls of the player.

image

- Contains one unrelated improvement (ended up hard to separate): control scheme picking now goes for the best match rather than the first match. Means you can have, for example, a PS4-specific control scheme and a generic gamepad control scheme and the system will reliably pick the appropriate one regardless of how the control schemes are ordered in the asset.


Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -693,7 +779,18 @@ public void ActivateInput()
                 m_CurrentActionMap?.Enable();
         }
 
-        public void PassivateInput()
```
jechter, line 696: While this change is the right thing to do, it will break user's projects. Leave an obsolet `PassivateInput` method which calls the new one.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -219,13 +219,20 @@ public class PlayerInput : MonoBehaviour
...
+
         /// <summary>
         /// Whether input is on the player is active.
         /// </summary>
         /// <value>If true, the player is receiving input.</value>
         /// <seealso cref="ActivateInput"/>
-        /// <seealso cref="PassivateInput"/>
-        public bool active => m_InputActive;
+        /// <seealso cref="DeactivateInput"/>
+        public bool inputIsActive => m_InputActive;
```
jechter, line 235: Same here, when renaming public API, keep the old one as obsolete (at least until you make a breaking change, ie increase major version number).

jechter:
Overall, looks great. I don't have the time to do a thorough review of all the code (if you want that, please do smaller PRs, instead of piling up so many changes in one). But I did test the new UI, so I can leave a few comments:

-How does the UI know which devices to allow binding to? Can I make a single UI button which allows binding to both keyboard and gamepad buttons for example (not saying that's better, but it might be desirable in some cases)?

-In the InGameHints sample scene, I couldn't see any text until I made my Game View very big. Should probably fix the text placement here to work for everyone.

-Button names are reported all wrong for the Switch controller I used to test this. Should we really default to Xbox naming scheme for "Anything not PS4"? I get it that we cannot correctly name all controllers (though we should definitely support Switch, as that one is kinda common), but maybe we should then use generic names for those instead?

-In the Sample, rebinding the mouse delta for Look does not work.

-Please never just rename public APIs like that without leaving the old ones alive. I'm not a strong proponent of SemVer as being an absolute thing (see current email thread on devs for my thoughts on that), but we should at least _try_ to not break projects or depending packages :) Along that line, we should consider adding versioned generated files with all the public API (like the .api files we have on trunk), to make API changes easy to spot in a review.

-This is probably not an issue, but I wonder if we should check with the console guys what the rules are of distributing PS4/Xbox controller button icons in games. It is definitely desirable and common to have them, but they are trademarked symbols, so there might be rules to follow as to how they should appear. If we build assets we expect people to use in games, it would be nice if those are actually known to be legally unproblematic.

Rene-Damm:
>-How does the UI know which devices to allow binding to?

If present, it goes by what's in the control scheme. If there's none, it accepts everything that's compatible with the action.

>Can I make a single UI button which allows binding to both keyboard and gamepad buttons for example (not saying that's better, but it might be desirable in some cases)?

Yup, either by not having control schemes or by having a control scheme that e.g. works with any InputDevice (something that ATM cannot yet be configured from the UI, though).

>-In the InGameHints sample scene, I couldn't see any text until I made my Game View very big. Should probably fix the text placement here to work for everyone.

Good catch. Will fix.

>-Button names are reported all wrong for the Switch controller I used to test this.

That sounds like a bug. If the controller is connected, it should pick up the actual display names from the device and the `Gamepad`-dictated display names shouldn't matter. My guess would be that the display names are not set up correctly for the controller?

>Should we really default to Xbox naming scheme for "Anything not PS4"?

My logic sorta went like this:

- On desktop, Xbox naming seems to be the de facto standard for referring to face buttons.
- On mobile, gamepads still seem to be a niche so the same de facto kinda logic doesn't seem to apply there but going with Xbox doesn't seem unreasonable.
- On consoles, there's no ambiguity; we just go with what the platform universally dictates.
- South/north/west/east, while IMO great for internal names as they obsolete any ambiguity, don't seem great for use in player-facing UIs as pretty much no one does that.

So, Xbox seemed like the most sensible fallback when showing things to players. But wherever it can, the system will try not to rely on the fallback.

>-In the Sample, rebinding the mouse delta for Look does not work.

Yeah, that control is explicitly filtered out. Should be possible to bind it to the scroll wheel instead.

Not sure whether it'd be better to delete that rebind. The thing is that, as mentioned in one of the text blobs on the screen, in reality you'd pretty much never have generic rebinds for those controls. At least, I know of no game that has rebinds for generic mouse controls such as mouse look. So maybe I should really just nuke that from the UI. Was more for completeness sake.

>I wonder if we should check with the console guys what the rules are of distributing PS4/Xbox controller button icons in games

Given the sheer amount of precedent of people doing exactly this, I'd say we're safe. Even if these were trademarked (I doubt they are), the trademark would surely be lost to them due to lack of enforcement.

jechter:
> > -Button names are reported all wrong for the Switch controller I used to test this.
> 
> That sounds like a bug. If the controller is connected, it should pick up the actual display names from the device and the `Gamepad`-dictated display names shouldn't matter. My guess would be that the display names are not set up correctly for the controller?


Hmm, the code in SwitchProControllerHID.cs looks correct, though:

```
        [InputControl(name = "buttonNorth", displayName = "X", bit = (uint)Button.North)]
        [InputControl(name = "buttonSouth", displayName = "B", bit = (uint)Button.South, usage = "Back")]
        [InputControl(name = "buttonWest", displayName = "Y", bit = (uint)Button.West, usage = "SecondaryAction")]
        [InputControl(name = "buttonEast", displayName = "A", bit = (uint)Button.East, usage = "PrimaryAction")]
```
 
Button names are also correct in the Input Debugger.

> > Should we really default to Xbox naming scheme for "Anything not PS4"?
> 
> My logic sorta went like this:

Ok, makes sense, but then let's make sure we at least get Switch right.

> > -In the Sample, rebinding the mouse delta for Look does not work.
> 
> Yeah, that control is explicitly filtered out. Should be possible to bind it to the scroll wheel instead.
> 
> Not sure whether it'd be better to delete that rebind. The thing is that, as mentioned in one of the text blobs on the screen, in reality you'd pretty much never have generic rebinds for those controls. At least, I know of no game that has rebinds for generic mouse controls such as mouse look. So maybe I should really just nuke that from the UI. Was more for completeness sake.

Yeah, might make sense to remove it. But, either way, at least, let's not have a rebind control with a default value you cannot actually bind to. 

> > I wonder if we should check with the console guys what the rules are of distributing PS4/Xbox controller button icons in games
> 
> Given the sheer amount of precedent of people doing exactly this, I'd say we're safe. Even if these were trademarked (I doubt they are), the trademark would surely be lost to them due to lack of enforcement.

The PS4 controller icons are most definitely trademarked. And I'm sure they have some design document specifying exactly how to use them (just as we have for our logos). I'm sure not every game follows that to the letter, I'm just thinking that if we deliver these assets which would potentially be used by a huge number of games, it can't hurt to ask what the rules are.

jechter:
Another minor issue: In the rebinding screen, when I bind to WASD, it is  printed as "w/A/s/d". Ie, all key names are lower case except for "A". Same when binding keys to "Interact".

So, this, and Switch Controller names being wrong tells me something still amiss with the control display names.

Rene-Damm:
>In the rebinding screen, when I bind to WASD, it is printed as "w/A/s/d". Ie, all key names are lower case except for "A". Same when binding keys to "Interact".

Not seeing it on Windows but seeing the same on Mac. However, looking at the device, this actually comes *from* the device. I.e. the display name of "a" is "A" and that of "w" is "w". This suggests that the key lookup for the "a" key is failing on OSX for some reason. I've opened a ticket for it: https://fogbugz.unity3d.com/f/cases/1210302/

Rene-Damm:
>Hmm, the code in SwitchProControllerHID.cs looks correct, though:

Oh wait, this is with the icons, right? The icon example is pretty simplistic and yeah, it has no Switch case and, unlike the core system, indeed will simply replace any `buttonSouth` with an `X` icon if it's not PS4. I'll add a Switch case that reuses the Xbox icons but reorders the face buttons.

////EDIT: It's probably not very obvious that the icons are indeed not part of the reusable rebinding prefab but rather just an example added on top of it.

jechter:
>Oh wait, this is with the icons, right? The icon example is pretty simplistic and yeah, it has no Switch case and, unlike the core system, indeed will simply replace any buttonSouth with an X icon if it's not PS4. I'll add a Switch case that reuses the Xbox icons but reorders the face buttons.

Not only there, I also see it being wrong in the text labels in the InGameHints sample.

Rene-Damm:
>Not only there, I also see it being wrong in the text labels in the InGameHints sample.

Ok, that definitely sounds buggy. I'll go write a test for it and see what I get.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
Rene-Damm: Fixed

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
Rene-Damm: Fixed

Rene-Damm:
- Removed "Look" action from Keyboard&Mouse rebind section.
- Fixed text not being visible in InGameHints sample at smaller resolutions.
- Added back `[Obsolete]` versions of `PlayerInput.active` and `PlayerInput.PassivateInput`.
- Fixed Switch Pro controller not showing correct display names.
- Checked with PS4 guys and got go-ahead for icons.

jechter (approved):

PR #998, opened 2020-01-14
NEW: Can save and load event traces.

Rene-Damm:
This PR significantly builds out the event recording and replay functionality of the system.

__Rationale for doing this now:__

One problem I'm repeatedly having with bugs is that there's always the trickier bug where you need to be on some specific platform, have some specific hardware, run some specific project, and perform some specific input for the problem to show. For example, we have a bug where key presses are getting lost. In my debugging so far, everything points to events but with the tools in the input system right now, it requires laborious and tedious debugging to dig into problems like this.

So, instead I want to be able to repro a problem once and to simply capture the repro such that I can replay it over and over and dissect it at will. Even without being on the specific platform, having the specific hardware, running a specific project, and definitely without trying to get the repro sequence just right (which makes debugging no fun).

A side benefit is that it also addresses the question of "how do I record and replay input?" which has repeatedly come up. So, two birds, one stone kinda deal.

__What can this do:__

You record your input problem, you send/upload/post the resulting trace file, and I can debug your problem even without having anything else from you. You have a BuzzFit Twitchomatic 4000 controller with the ChamferEdge Rainbow LED add-on running on Linux and I don't? No problem, I can still repro your input.

__What can this not do:__

Trace back problems to *before* event generation, i.e. to where the event sequence is already buggy. But it can still at least clearly point in the direction.

__What this adds:__

A new "InputRecording" sample with a new `InputRecorder` MonoBehaviour in it.

image

Some additional controls to event traces in the input debugger.

image

A number of additions to the `InputEventTrace` API.
* Buffers can be made growable now.
* Can now capture frame markers to capture distribution of input over frames.
* Can write traces to streams/files and load them from such.
* Traces now capture information about the devices they capture events for (including writing out entire device layouts as JSON for devices with auto-generated layouts; this allows reproducing problems on HIDs without the original HID data being available).
* Traces now have built-in replay functionality including the ability to replay the original frame distribution or to simulate original event timing.


Assets/Samples/InputRecorder/InputRecorderInspector.cs
```diff
@@ -0,0 +1,234 @@
...
+                if (oldIsRecording != newIsRecording)
+                {
+                    if (newIsRecording)
+                        recorder.StartCapture();
+                    else
+                        recorder.StopCapture();
+                }
+                EditorGUI.EndDisabledGroup();
+
+                ////TODO: finish
```
jechter: Fix TODO here.

Packages/com.unity.inputsystem/InputSystem/Events/InputEventTrace.cs
```diff
@@ -1,85 +1,648 @@
...
+                if (!m_EventListeners.ContainsReference(value))
+                    m_EventListeners.Append(value);
+            }
             remove => m_EventListeners.Remove(value);
         }
 
-        // Create a disabled event trace that does not perform any allocation
-        // yet. An event trace only starts consuming resources the first time
-        // it is enabled.
-        public InputEventTrace(int bufferSize = kDefaultBufferSize)
```
jechter, line 44: FYI: Removing this constructor breaks assembly compatibility. You still have source compatibility, as there is the constructor which takes buffer size and more parameters, which all have default values, which the compiler can then map to. But that's a compile-time mapping, not runtime. So any dll referencing the old constructor would fail. I don't particularly care because:
1. Package still has the preview label
2. Forward/Backwards compatibility of assemblies referencing packages is grey area in Unity anyways. It is not something we really want to commit to supporting, because it is a mess, but people do it all the time anyways.

jechter:
Changes lgtm (except for the #ifdefed out code) - though I'm wary of the amount of changes landing in 1.0.0-preview.x at this point. I would suggest a branch-off for any new work soon. Please have someone do some manual QA on this before landing, though!

Assets/Samples/InputRecorder/InputRecorderInspector.cs
Rene-Damm: Fix how? Ultimately I'd like to be able to inspect the contents of any trace with something better than that simple tree view in the input debugger and I started on that here but then decided it makes more sense to defer this I disabled the code that was already there.

Packages/com.unity.inputsystem/InputSystem/Events/InputEventTrace.cs
Rene-Damm: > Changes lgtm (except for the #ifdefed out code) - though I'm wary of the amount of changes landing in 1.0.0-preview.x at this point. I would suggest a branch-off for any new work soon. Please have someone do some manual QA on this before landing, though!

Asked Stefan for QA validation.

Rene-Damm: (wtf... no idea why GitHub put the response here...)

Assets/Samples/InputRecorder/InputRecorderInspector.cs
jechter: I'm fine with //TODO comments in code. But in my experience, ifdefed out or commented out code never ages well and by the time you get to revisit it you have to rewrite it anyways. So I would either do it now, or remove the code, but not ship disabled code like this.

Rene-Damm: Removed.

Packages/com.unity.inputsystem/InputSystem/Events/InputEventTrace.cs
Rene-Damm: Stefan uncovered a couple cosmetic issues. Fixed. Should be ready to go.

Rene-Damm:
Looking into the Windows standalone failures.

jechter (approved):

PR #1000, opened 2020-01-14
FIX: UI component references broken on 2019.1.

Rene-Damm:
Made the same stupid mistake again of working on 2019.3 and committing UI stuff. Fixed the `RebindingUI` and `InGameHints` sample to work with 2019.1.

Will fix the `InputRecording` scene in #998.


jechter (approved):

Rene-Damm:
Added one extra change to fix the UI getting garbled up at different resolutions which Stefan discovered. Spelled "UI" as "ui" in the branch name and git got confused all kinds of ways. Routed the changes to develop in #1000.

PR #1007, opened 2020-01-15
CHANGE: Overhaul focus logic to fix problems such as stuck keys on alt-tabbing (case 1206199).

Rene-Damm:
This PR intends to address bugs arising from the current focus handling, like e.g. a stuck alt key when alt-tabbing on Windows (case 1206199).

It essentially replaces the previous focus handling with a much more nuanced mechanism.

* When focus is lost in the player, devices are reset.
* A reset consists of sending a reset *request* to the backend, like before, but regardless of how the backend responds to that, the input system will now perform an immediate reset of a device's state back to its default state.
* Noisy controls are not touched and will remain at their current value. This preserves sensor readings at their last value instead of snapping them back to default.
* When `Application.runInBackground` is enabled, devices that return `true` from `InputDevice.canRunInBackground` will be exempted from resets and will be left alone.

Working on this, I also found a couple bugs with noise masks and associated memory checks.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Utilities/MemoryHelpers.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Utilities/MemoryHelpers.cs
@@ -150,18 +150,16 @@ namespace UnityEngine.InputSystem.Utilities
                 // to mask out parts of the bits we're reading.
                 var byteMask = 0xFF << (int)bitOffset;
                 if (bitCount + bitOffset < 8)
-                {
                     byteMask &= 0xFF >> (int)(8 - (bitCount + bitOffset));
-                }
 
                 if (maskPtr != null)
                 {
-                    byteMask &= *maskPtr;
+                    byteMask &= ~*maskPtr;
                     ++maskPtr;
                 }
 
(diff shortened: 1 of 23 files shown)


Packages/com.unity.inputsystem/InputSystem/Utilities/MemoryHelpers.cs
```diff
@@ -228,6 +226,93 @@ public static bool MemCmpBitRegion(void* ptr1, void* ptr2, uint bitOffset, uint
             return true;
         }
 
+        public static void MemSet(void* destination, int numBytes, byte value)
```
jechter, line 229: Wow, surprised this does not exist (well, it does, Unsafe.InitBlock, but only in .NET Core). Guess there is the Initblk opcode, but no C# code to generate it. Weird.

jechter (approved):
Seems reasonable. Have @stefanunity or someone do some QA on it, though!

Rene-Damm:
> Seems reasonable. Have @stefanunity or someone do some QA on it, though!

Yup, QA pass is scheduled to happen shortly.

PR #1018, opened 2020-01-23
FIX: Several issues uncovered in QA pass.

Rene-Damm:
Just a couple random tweaks and fixes, mainly to the input debugger UI code.

One more serious issue that's fixed is that the code that now allows InputEventTraces to survive domain reloads caused their data to be written into EditorWindow layouts. On startup, this led to crashes. Fixed by making sure the data is serialized only when the debug flag is included in serialization.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
+++ b/Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
@@ -14,6 +14,7 @@ The following is a list of known limitations that the Input System currently has
 * (Windows) Pen input will not work with Wacom devices if "Windows Ink" support is turned off.
 * Joy-Cons are only supported on Switch.
 * Sensors in the PS4 controller are currently only supported on PS4.
+  >NOTE: Support for NDA platforms is distributed as separate packages due to licensing restrictions. The packages, at this point, are made available separately to licensees for download and installation.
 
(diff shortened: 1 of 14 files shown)


Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
```diff
@@ -0,0 +1,20 @@
...
+
+* `PlayerInput` split-screen support does not work with Cinemachine virtual cameras.
+* The Input System cannot generate input for IMGUI or UIElements.
+* The Input System does not yet support the new 2019.3 mode where domain reloads are disabled when entering play mode.
+
+## Device support
+
+* (Windows) Mouse input from remote desktop connections does not come through properly.
+* (Windows) Pen input will not work with Wacom devices if "Windows Ink" support is turned off.
+* Joy-Cons are only supported on Switch.
```
jechter, line 15: Should the comments about Switch and PS4 clarify that support for this is in separate packages?

jechter (approved):

Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
Rene-Damm: Done

PR #1019, opened 2020-01-23
FIX: Allow overriding composite binding parameters

jamre:
Currently, the ```InputBindingResolver``` uses the binding's path to instantiate a composite binding.
Any parameters are stored in the path as well.
For example, ```"MyCustomComposite(someParameter=true)"```

If any overrides to the path were made at runtime (e.g. to change the parameter to false) they are ignored when the binding is resolved.

This PR changes the ```InputBindingResolver``` to use ```effectivePath``` instead of ```path``` so that parameters of the composite can be changed at runtime.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
@@ -289,7 +289,7 @@ namespace UnityEngine.InputSystem
                                     : InputActionState.kInvalidIndex;
 
                                 // Instantiate. For composites, the path is the name of the composite.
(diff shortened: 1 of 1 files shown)


Rene-Damm:

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1019) <br/>All committers have signed the CLA.

ekcoh:
Thanks for your contribution @jamre. Following-up on this PR we have noticed that nowadays `effectivePath ` is used, hence this PR can be closed since what you wanted to achieve can now be done.

PR #1020, opened 2020-01-24
FIX: mousestates from different devices don't share the point position

itsmeted:
When the m_PointAction device is different from the m_ScrollWheelAction, m_LeftClickAction, m_RightClickAction, or m_MiddleClickAction, the position for the mouseStates for the different actions are set to the default value of (0,0).

For example, if I set the binding for the m_LeftClickAction to a keyboard press, and leave the m_PointAction as the mouse position, two different mouseStates are created in InputSystemUIInputModule.cs since they are two different devices. The subsequent ProcessMouseButton() function call will process the mouseState set by the m_LeftClickAction separately from the mouseState set by the m_PointAction. Thus, ProcessMouseButton() does not get the position set in the mouseState set by the m_PointAction and always has the default value of (0,0).

This change stores the position retrieved from the m_PointAction, and stores it on the mouseStates set on the subsequent m_ScrollWheelAction, m_LeftClickAction, m_RightClickAction, or m_MiddleClickAction. This removes the dependency that the device action mapping for m_PointAction must match that of the other actions.  Also, this allows us to define virtual cursors that can be used to determine mouse position.


Rene-Damm:

Rene-Damm:
Thank you for the bringing this to our attention. I've taken a look and opted to solve it in a different way as part of a larger rewrite of `InputSystemUIInputModule`. Should be fixed by #1082.

PR #1024, opened 2020-01-27
FIX: Don't set group for pasted composite binding

jamre:
When pasting a copied composite binding in the asset editor, if the editor is filtering to show only one scheme, the group will get set to that scheme on the composite binding when it should be left empty.

This PR skips setting the group if the binding is a composite.


Rene-Damm:

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1024) <br/>All committers have signed the CLA.

jimon:
Hi thanks for the PR, we will merge it as part of #1604

Issue #1021, opened 2020-01-26
Specific device names no longer labelled correctly as of preview.4


BHSPitMonkey:
After updating from preview.3 to preview.4, "More Specific" device models now all appear under their more generic name. See below:

![image](https://user-images.githubusercontent.com/33672/73130435-2c19c580-3fad-11ea-99ba-e711ac26891c.png)

As a result, it's no longer possible to tell which bindings are which without toggling to the raw text path display for each entry.

Downgrading back to preview.3 restores the correct titles.

Rene-Damm:
Argh, that one is embarrassing. Unfortunate side effect of a cosmetic change where the `XRController` now sets a display name so it reads "XR Controller". But display names are inherited and none of the derived layouts is overriding it. Fixing.

Rene-Damm:
Thank you for reporting :)

Rene-Damm:
Should be fixed as of #1035.

PR #1028, opened 2020-01-29
FIX: Wrong display names for devices in control schemes.

Rene-Damm:
Minor cosmetic issue. You'd pick "PS4 Controller" (display name) from the list and then it'd say "DualShockGamepad" (layout name) in the list instead.


BHSPitMonkey:
Possibly fixes #1021?

jechter (approved):

PR #1035, opened 2020-02-03
FIX: Display names in UI being wrong for XR controllers and HMDs (#1021).

Rene-Damm:
Was regressed by the fact that `XRController` and `XRHMD` now assign a `displayName`. Since that propery is inherited and none of the XR layouts was setting a display name, this broke display names on all the XR layouts.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/OpenVR.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/OpenVR.cs
@@ -7,7 +7,7 @@ using UnityEngine.Scripting;
 
 namespace Unity.XR.OpenVR
 {
-    [InputControlLayout]
+    [InputControlLayout(displayName = "OpenVR Headset")]
     [Preserve]
     public class OpenVRHMD : XRHMD
     {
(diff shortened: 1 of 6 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/OpenVR.cs
```diff
@@ -51,7 +51,7 @@ protected override void FinishSetup()
         }
     }
 
-    [InputControlLayout(commonUsages = new[] { "LeftHand", "RightHand" })]
+    [InputControlLayout(displayName = "OpenVR Controller (Windows MR)", commonUsages = new[] { "LeftHand", "RightHand" })]
```
StayTalm: This one should follow the same pattern as Oculus' 'Oculus Touch Controller (OpenVR)'
Maybe
Windows Spatial Controller (OpenVR)
?

StayTalm (approved):
Small suggestion for consistency.  Approved either way, I see what I need to make sure for future controls.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/OpenVR.cs
Rene-Damm: Fixed. Dropped the "Spatial" for brevity. Hope that's okay.

PR #1049, opened 2020-02-10
Handle parent directory renaming whilst asset is being edited.

karljj1:
*Purpose*:
When an asset was being edited and the parent directory was renamed, the asset path would become incorrect and an exception would occur during SaveChangesToAsset.
I removed all caching of the asset path and we now use the guid when the path is needed. From what I can see, the path is not used very often so performance should not be an issue.

*Release Notes*
Fixed exception when saving changes to an Input Action asset and the parent directory has been renamed. (case 1207527)

*Functional Testing status*:
Includes test

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Editor.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Editor.cs
@@ -545,6 +545,31 @@ partial class CoreTests
         Assert.That(map.bindings[0].groups, Is.EqualTo(""));
     }
 
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputActionAssetManager_SaveChangesToAsset_DoesNotThrow_WhenParentDirectoryWasRenamed()
+    {
+        const string kAssetPath = "Assets/DirectoryBeforeRename/InputAsset." + InputActionAsset.Extension;
+
+        AssetDatabase.CreateFolder("Assets", "DirectoryBeforeRename");
+        File.WriteAllText(kAssetPath, "{}");
+        AssetDatabase.ImportAsset(kAssetPath);
+
+        var asset = AssetDatabase.LoadAssetAtPath<InputActionAsset>(kAssetPath);
+        Assert.NotNull(asset, "Could not load asset: " + kAssetPath);
+
+        var inputActionAssetManager = new InputActionAssetManager(asset);
+        inputActionAssetManager.Initialize();
+        inputActionAssetManager.onDirtyChanged = (bool dirty) => { };
+
+        FileUtil.MoveFileOrDirectory("Assets/DirectoryBeforeRename", "Assets/DirectoryAfterRename");
+        AssetDatabase.Refresh();
+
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -140,19 +142,20 @@ public void ApplyChanges()
 
         internal void SaveChangesToAsset()
         {
-            Debug.Assert(!string.IsNullOrEmpty(m_AssetPath));
+            Debug.Assert(importedAsset != null);
```
karljj1, line 145: We could probably just remove this

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
```diff
@@ -827,7 +827,7 @@ public static AssetDeleteResult OnWillDeleteAsset(string path, RemoveAssetOption
 
             // Handle .inputactions asset being moved.
             // ReSharper disable once UnusedMember.Local
-            public static AssetMoveResult OnWillMoveAsset(string sourcePath, string destinationPath)
+            public static AssetMoveResult OnWillMoveAsset(string sourcePath, string _)
```
Rene-Damm: Need to suppress the warning here for the analyzer to be happy.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
Rene-Damm: +1

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
Rene-Damm: (the breakage is from users now seeing a compiler warning pop up in the console when they import the input system package)

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -545,6 +545,31 @@ public void Editor_InputAsset_CanReplaceBindingGroupThroughSerialization()
         Assert.That(map.bindings[0].groups, Is.EqualTo(""));
     }
 
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputActionAssetManager_SaveChangesToAsset_DoesNotThrow_WhenParentDirectoryWasRenamed()
```
Rene-Damm: `Editor_InputActionAssetManager_CanMoveAssetOnDisk`? The throwing seems more incidental to the previous bug than instrumental to the behavior of interest here.

Rene-Damm:
For the release notes, please add them to Packages/com.unity.inputsystem/CHANGELOG.md.

(Sorry that our PR process is so badly documented ATM...)

Rene-Damm (approved):

PR #1053, opened 2020-02-12
CHANGE: Enforce started/performed/canceled cycle on button and value actions.

Rene-Damm:
Fixes [1196050](https://issuetracker.unity3d.com/issues/input-system-ui-becomes-unresponsive-after-the-first-ui-button-press) ([internal](https://fogbugz.unity3d.com/f/cases/1196050/)).

This is an attempt to make started/performed/canceled more predictable and consistent by forcing a transition to `performed` to go through `started` and forcing a transition back to waiting to go through `canceled`.

Includes a do-over of `PressInteraction` to solve the many surprises the interaction has led to.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -854,7 +854,7 @@ namespace UnityEngine.InputSystem
                 interactionIndex = kInvalidIndex,
                 time = time,
                 startTime = time,
-                passThrough = actionIndex != kInvalidIndex && actionStates[actionIndex].passThrough,
+                isPassThrough = actionIndex != kInvalidIndex && actionStates[actionIndex].isPassThrough,
             };
 
             // If we have pending initial state checks that will run in the next update,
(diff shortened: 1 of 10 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -2631,7 +2688,7 @@ public uint lastTriggeredInUpdate
             /// Whether the action associated with the trigger state is marked as pass-through.
             /// </summary>
             /// <seealso cref="InputActionType.PassThrough"/>
-            public bool passThrough
+            public bool isPassThrough
```
jechter, line 2691: Make an obsolete property with the old name.

jechter (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
Rene-Damm: This is an internal API.

jechter: Oh, nvm then

PR #1061, opened 2020-02-14
FIX: Enabling single actions leading to errors when enabling their map.

Rene-Damm:
Fixes [thread](https://forum.unity.com/threads/error-while-switching-between-action-maps.825204/).

Turns out that somewhere along the way of `InputActionState` emerging as a thing, the code no longer kept track of which controls it had already installed state monitors for and which it hadn't. So if you enabled a single action and then enabled the entire map that contained it, the code would go and install the state monitor for the already enabled action again. So now it would trigger twice. And if later the map was disabled, one of the state monitors would stick around then fire for an action that is now disabled.

Added code to track the enabled/disabled state for each individual control in an `InputActionState`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -3590,6 +3590,69 @@ partial class CoreTests
 
     [Test]
     [Category("Actions")]
+    public void Actions_CanMixEnablingSingleActionsAndEntireActionMaps()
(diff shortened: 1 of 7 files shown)


Rene-Damm:
Breaks firing with the gamepad in `SimpleDemo_UsingPlayerInput`. Needs more work. (And looks like the test rig has also caught on to that fact...)

jechter (approved):

PR #1063, opened 2020-02-17
FIX: Runtime errors when action for binding does not exist (case 1213085).

Rene-Damm:
Fixes [1213085](https://issuetracker.unity3d.com/issues/input-system-input-actions-cause-exceptions-and-should-not-get-here-errors-to-appear-after-deleting-an-action-map) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1213085)).

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
@@ -218,23 +218,38 @@ namespace UnityEngine.InputSystem
                             action = currentCompositeAction;
                         }
 
+                        // If it's a composite, start a chain.
+                        if (isComposite)
+                        {
+                            currentCompositeBindingIndex = bindingIndex;
+                            currentCompositeAction = action;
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
```diff
@@ -218,23 +218,38 @@ public unsafe void AddActionMap(InputActionMap map)
                             action = currentCompositeAction;
                         }
 
+                        // If it's a composite, start a chain.
```
Rene-Damm, line 221: The changes starting here look pretty invasive but essentially only one simple thing changed: before, the binding state for a composite was only set up if the binding was not disabled. Also, binding states for composites were written out from a dedicated path which then earlied out of the current loop iteration.

Now, the binding state is *always* written at the end. For composites, for part bindings, and for "normal" bindings. Which now allows to safely disable any kind of binding.

This is now used if the action for a "normal" or composite binding cannot be found. If that is the case, the binding and all its part bindings are disabled and thus will not get their controls resolved.

jechter (approved):

Issue #1045, opened 2020-02-10
Functionality to simulate touch with mouse in player


eliphatfs:
There's such an option in Input Debug window. Why not also in player? It is nice to have universal input handling logic between platforms.

Rene-Damm:
Touch simulation is implemented by the `TouchSimulation` MonoBehavior. You can add that to your game or enable touch simulation through `TouchSimulation.Enable()` to get the same behavior in the player that you get in the editor with the switch in the input debugger.

eliphatfs:
Thanks. Maybe this should be mentioned briefly in the documentation, hard to find :D

Rene-Damm:
Indeed :) Fixed in #1067.

PR #1067, opened 2020-02-19
FIX: useStateFrom not working with injected controls.

Rene-Damm:
Edge case that's likely not really impacting anyone ATM. But simple, one-line fix (also moved the respective code to a separate function, though).

Also snuck in a doc change to fix missing mention of `TouchSimulation` as pointed out in #1045.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Layouts.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Layouts.cs
@@ -1371,6 +1371,60 @@ partial class CoreTests
 
     [Test]
     [Category("Layouts")]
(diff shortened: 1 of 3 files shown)


jechter (approved):
Might want to add a release note?

Rene-Damm:
> Might want to add a release note?

Was thinking about it but after trying to word this in an understandable way, I wondered if this is just such an obscure thing that the outcome would only be a confusing blob of garbage in the relnotes. But maybe I just didn't put enough effort into it :)

jechter:
Fair enough

PR #1078, opened 2020-02-28
Software Cursor update using Position rather than Anchored Position

ad-walker:
Using Position rather than Anchored Position allows use of Canvas Scaler settings other than Constant Pixel Size. Software cursor image no longer needs to be anchored to bottom left.


Rene-Damm:
Thank you for the PR. Integrated into #1119.

PR #1080, opened 2020-03-03
CHANGE: Added Windows Input tests to ensure Mouse movement works usin…

karljj1:
Added a test to go with the fix in https://ono.unity3d.com/unity/unity/pull-request/102187/_/se/input/rdp-mouse-position-fix

CHANGE: Added Windows Input tests to ensure Mouse movement works using RDP.


karljj1:
The tests are expected to fail until the fix lands

Assets/Tests/InputSystem/WindowsTests.cs
```diff
@@ -0,0 +1,54 @@
...
+using UnityEngine.InputSystem;
+using UnityEngine.TestTools;
+
+partial class WindowsTests
+{
+    [UnityTest]
+    [Category("WindowsInput")]
+    public IEnumerator WindowsInput_RemoteDesktopMouseMovements_AreDetected()
+    {
+        var mouse = InputSystem.GetDevice<Mouse>();
```
Rene-Damm: `Mouse.current`

Rene-Damm (commented):
Since this goes end-to-end, would prefer to have this as part of `IntegrationTests`.

Also, I assume this can't land until we've landed the fix on the native side?

Assets/Tests/InputSystem/WindowsTests.cs
karljj1: You want it moved into the IntegrationTests folder?

karljj1:
Wont be able to backport this fix further than 2019.3 so these tests wont pass. Ill have a think about a solution, maybe move the test into trunk.

PR #1082, opened 2020-03-04
FIX: Redo much of InputSystemUIInputModule.

Rene-Damm:
This PR is a larger scale rewrite of `InputSystemUIInputModule`.

## Why?

I found so many issues that after I initially just fixed individual issues as I went, I ultimately decided to take a knife to the current code and cut away a great deal of what I perceived as unnecessary complexity and systematically build out tests (which were greatly lacking) and functionality.

Many of the issues were deviations from `StandaloneInputModule` behavior which were regressions but some of the issues also arose from peculiarities of how `InputSystemUIInputModule` sources input and thus needs different handling than `StandaloneInputModule`. For example, `StandaloneInputModule` simply used `Input.mousePosition` which by nature is an aggregate whereas `InputSystemUIInputModule` sources position values from individual devices and may thus receive multiple conflicting pointer inputs whereas the old input had one already unified input.

These changes come in very late for 1.0 but IMO the current `InputSytemUIInputModule` is simply too broken to be 1.0 and UI is too critical to not fix this.

## Changes

- Added lots of tests. Our UI tests were greatly lacking. Situation is still not stallar but probably already a great deal more extensive than the stuff we ever had for `StandaloneInputModule` :)
- Renamed `MouseModel` (internal) to `PointerModel` and simplified the code.
- Collapsed `TrackedDeviceModel` (internal) into `PointerModel` as tracked devices are ultimately just a special type of pointers (like touch).
- In the same vein, removed `trackedDeviceSelect`. Tracked input is pointer input so this is simply equivalent to `leftClick` (same as for touch). As part of this change, it is now also possible to right and middle click from tracked devices by simply binding `rightClick` and/or `middleClick` for them.
- Renamed `JoystickModel` (internal) to `NavigationModel` and simplified it.
- Added a public `ExtendedPointerEventData` which surfaces input system-specific information in UI events. Also replaces the internal TrackedPointerEventData class. Tracked device input and pointer input now sit on the same single codepath.
- Exposed max raycast distance for `TrackedDevice` input as a `maxDistance` setting on `TrackedDeviceRaycaster`.
- Added `TrackedDeviceRaycaster` to the component menu (like `GraphicRaycaster`) and added instructions for how to set it up to the docs.
- Added documentation about how input is sourced from actions by `InputSystemUIInputModule`.
- Fixed pointer states in `InputSystemUIInputModule` growing infinitely with `Touchscreen` if touch IDs were not reused.
- Fixed `lastPress` value on PointerEventData getting lost.
- Fixed button press&release happening in same frame resulting in no UI input.
- Fixed trigger clicks with non-pointer devices leading to pointer events with (0,0) positions. Now triggers clicks on the last pointer and thus uses its position.
- Fixed huge screen deltas on pointer events from tracked devices.
- Fixed `TrackedDeviceRaycaster` not setting `screenPosition` in `RaycastResult`.
- Fixed GC heap garbage from mouse UI input.
- Fixed touch input not sending pointer exit events.
- Fixed `InputSystem.onActionChange` getting invoked too many times when actions got enabled/disabled/changed.
- Renamed `repeatDelay` and `repeatRate` in `InputSystemUIInputModule` to `moveRepeatDelay` and `moveRepeatRate`.
- Added `InputSystemUIInputModule.pointerBehavior` to give control over how the input module resolves concurrent input from multiple pointers. Default behavior is to unify all pointer input that isn't from touch or tracked devices. This preserves multi-point ability for touch and tracked devices while collapsing input from other pointers such as mice and pens into a single unified pointer.

Pretty sure I missed a couple issues I found and fixed.

## Known Issues

- UI input module does not react to pointer positions until pointers are moved.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,8 +9,23 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased]
 
+### Changed
+
+* `InputSystemUIInputModule.trackedDeviceSelect` has been removed. Simply use `InputSystemUIInputModule.leftClick` instead.
```
keiramoon: Suggest removing "Simply" from the second sentence.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,8 +9,23 @@ however, it has to be formatted properly to pass verification tests.
...
+* `InputSystemUIInputModule.repeatDelay` has been renamed to `moveRepeatDelay` and `repeatRate` has been renamed to `moveRepeatRate`.
+
 ### Fixed
+
 - Fixed CS0109 warning being generated during player build due to use of `new` with the `PlayerInput.camera property`. (case 1174688)
+- Fixed a number of issues in `InputSystemUIInputModule`.
+  * Fixed GC heap garbage when click-dragging.
+  * Fixed number of pointer states growing indefinitely if OS did not reuse touch IDs.
+  * Fixed `lastPress` on `PointerEventData` getting lost.
+  * Fixed button press&release happening in same frame resulting in no UI input.
```
keiramoon: Suggest changing press&release to press-and-release (easier to read).

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -20,6 +35,18 @@ however, it has to be formatted properly to pass verification tests.
 - `PressInteraction` no longer misses the next button press if it gets reset from within the `performed` callback ([case 1205285](https://issuetracker.unity3d.com/issues/inputsystem-problem-with-button-state-after-deactivating-and-reactivating-an-action-map)).
 - `InputBinding.DisplayStringOptions.DontIncludeInteractions` is now properly respected.
 - Reading the value of a composite binding no longer causes processors from the last active part binding to be applied rather than the processors of the composite itself, if any ([case 1207082](https://issuetracker.unity3d.com/issues/input-system-invert-processors-have-no-effect-on-the-inputaction-dot-callbackcontext-value)).
+- Fixed `InputSystem.onActionChange` getting invoked too many on binding changes.
```
keiramoon: Is there a missing word here - should it be "too many times"?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -20,6 +35,18 @@ however, it has to be formatted properly to pass verification tests.
...
+- Pasting composite bindings while having a specific control scheme active no longer incorrectly applies the control scheme's binding group to the entire composite (fix contributed by [jamre](https://github.com/jamre) in #1024).
+- Can now have spaces in action names when using `PlayerInput` with the `SendMessages` or `BroadcastMessages` behavior. Previously an incorrect method name was generated (fix contributed by [BHSPitMonkey](https://github.com/BHSPitMonkey) in #1022).
```
keiramoon: Add a comma after "Previously" in the second sentence.

Packages/com.unity.inputsystem/Documentation~/Installation.md
```diff
@@ -26,3 +26,5 @@ To fully switch from the old Input Manager to the new Input System for a Project
 ![Switch Active Input Handling](Images/ActiveInputHandling.png)
 
 >__Note__: You must restart the Unity Editor before this setting takes effect.
+
+When the new input backends are enabled, the `ENABLE_INPUT_SYSTEM=1` C# `#define` is added to builds. Similarly, when the old input backends are enabled, the `ENABLE_LEGACY_INPUT_MANAGER=1` C# `#define` is added. As both can be enabled at the same time, it is possible for __both__ defines to be 1 at the same time.
```
keiramoon: In the last sentence, change "As" to "Because".

Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
```diff
@@ -8,6 +8,10 @@ The following is a list of known limitations that the Input System currently has
 * The Input System cannot generate input for IMGUI or UIElements.
 * The Input System does not yet support the new 2019.3 mode where domain reloads are disabled when entering play mode.
 
+### uGUI
+
+* After enabling, the UI will not react to a pointer's position until the position is changed.
```
keiramoon, line 13: Can we specify what we're enabling here, just for the sake of clarity?
Also, please change the last verb from passive to active ("is changed" --> "changes").

Packages/com.unity.inputsystem/Documentation~/Migration.md
```diff
@@ -4,6 +4,21 @@ This guide provides a list of APIs in `UnityEngine.Input` (and other related API
 
 >__Note__: All of the new APIs are in the `UnityEngine.InputSystem` namespace. The namespace is omitted here for brevity. `UnityEngine.InputSystem` is referenced in full for easy disambiguation.
 
+Note that it is possible to have code for the old and the new input system at the same time using conditional compilation. When the new input system is enabled in the player preferences (see [here](./Installation.md#enabling-the-new-input-backends)), the `ENABLE_INPUT_SYSTEM` preprocessor directive is available.
```
keiramoon, line 7: Should this paragraph be highlighted as a **Note** as well? If not, I'd skip "Note that" and just dive in: "It is possible to have code from the old and new..."

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+|[`Tracked Device Orientation`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_trackedDeviceOrientation)|An Action that delivers a `Quaternion` representing the rotation of one or multiple spatial tracking devices, such as XR hand controllers. In combination with [`Tracked Device Position`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_trackedDevicePosition), this allows XR-style UI interactions by pointing at UI [selectables](https://docs.unity3d.com/Manual/script-Selectable.html) in space. See [tracked-type input](#tracked-type-input).|
+
+### How the bindings work
+
+The UI input module can deal with three different types of input:
+
+1) pointer-type input,
```
keiramoon, line 44: Let's update this list to match the Microsoft style guidelines [here](https://docs.microsoft.com/en-us/style-guide/scannable-content/lists):

1. Pointer-type input
2. Navigation-type input
3. Tracked-type input

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+
+>NOTE: The UI input module does not have an association between pointers and cursors. In general, the UI is oblivious to whether a cursor exists for a particular pointer. However, for mouse and pen input, the UI input module will respect [`Cusor.lockState`](https://docs.unity3d.com/ScriptReference/Cursor-lockState.html) and pin the pointer position at `(-1,-1)` whenever the cursor is locked.
```
keiramoon, line 54: Can we change the second sentence to:
In general, the UI isn't aware whether a cursor exists for a particular pointer.

Bit clearer, especially for ESL folks.

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+
+Multiple pointer devices may feed input into a single UI input module. Also, in the case of [`Touchscreen`](../api/UnityEngine.InputSystem.Touchscreen.html), a single device can have the ability to have multiple concurrent pointers (each finger contact is one pointer).
```
keiramoon, line 56: Please change "may" to either "can" or "might", whichever's appropriate (Microsoft style guidelines forbid "may" because it "might be interpreted as providing permission").

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+
+You can influence how the input module deals with concurrent input from multiple pointers using the [`Pointer Behavior`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_pointerBehavior) setting.
```
keiramoon, line 60: Change this row to:

Use the [`Pointer Behavior`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_pointerBehavior) setting to influence how the input module deals with concurrent input from multiple pointers.

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
...IPointerBehavior.html#UnityEngine_InputSystem_UI_UIPointerBehavior_SingleMouseOrPenButMultiTouchAndTrack)|Behaves like [`Single Unified Pointer`](../api/UnityEngine.InputSystem.UI.UIPointerBehavior.html#UnityEngine_InputSystem_UI_UIPointerBehavior_SingleUnifiedPointer) for all input that is not classified as touch or tracked input, and behaves like [`All Pointers As Is`](../api/UnityEngine.InputSystem.UI.UIPointerBehavior.html#UnityEngine_InputSystem_UI_UIPointerBehavior_AllPointersAsIs) for tracked and touch input.<br><br>If concurrent input is received on a [`Mouse`](../api/UnityEngine.InputSystem.Mouse.html) and [`Pen`](../api/UnityEngine.InputSystem.Pen.html), for example, the input of both is fed into the same UI pointer instance. The position input of one will overwrite the position of the other.<br><br>Note that when input is received from touch or tracked devices, the single unified pointer for mice and pens is __removed__ including [`IPointerExit`](https://docs.unity3d.com/Packages/com.unity.ugui@1.0/api/UnityEngine.EventSystems.IPointerExitHandler.html) events being sent in case the mouse/pen cursor is currently hovering over objects.<br><br>This is the default behavior.|
```
keiramoon, line 64: Let's break the Note into two sentences to make it easier to digest. Also, please ensure the copy below is still accurate re: what actually happens - I've rephrased it slightly:

**Note:** When input is received from touch or tracked devices, the single unified pointer for mice and pens is removed. This includes [`IPointerExit`](https://docs.unity3d.com/Packages/com.unity.ugui@1.0/api/UnityEngine.EventSystems.IPointerExitHandler.html) events being sent if the mouse or pen cursor is currently hovering over objects.

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+|[`Single Unified Pointer`](../api/UnityEngine.InputSystem.UI.UIPointerBehavior.html#UnityEngine_InputSystem_UI_UIPointerBehavior_SingleUnifiedPointer)|All pointer input is unified such that there is only ever a single pointer. This includes touch and tracked input. This means, for example, that regardless how many devices feed input into [`Point`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_point), only the last such input in a frame will take effect and become the current UI pointer's position.|
```
keiramoon, line 65: For the first sentence, is it accurate to rephrase it as "All pointer input is unified into a single pointer in all scenarios." or something along those lines? That would feel easier to read.

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+|[`All Pointers As Is`](../api/UnityEngine.InputSystem.UI.UIPointerBehavior.html#UnityEngine_InputSystem_UI_UIPointerBehavior_AllPointersAsIs)|The UI input module will not unify any pointer input. Any device, including touch and tracked, devices feeding input pointer-type actions will be its own pointer (for touch, multiple pointers).<br><br>Note that this means that there may be arbitrary many pointers in the UI and several objects may be pointed at concurrently.|
```
keiramoon: Let's rephrase the second sentence to:

Any device, including touch and tracked devices that feed input pointer-type actions, will be its own pointer (or multiple pointers for touch input).

And the Note to:

**Note:** This might mean that there will be an arbitrary number of pointers in the UI, and several objects might be pointed at concurrently.

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+
+#### Navigation-type input
+
+Navigation-type input controls the current selection based on motion read from the [`move`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_move) action. Additionally, input from
+[`submit`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_submit) will trigger `ISubmitHandler` on the currently selected object and
+[`Cancel`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_cancel) will trigger `ICancelHandler` on it.
+
+Inputs received on these actions is independent of each other. Thus, an arbitrary set of devices can be bound to these actions and feed input in arbitrary order.
```
keiramoon: Change "is independent" to "are independent" (the verb relates to "Inputs", right?)

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+
+Navigation input is non-positional, i.e. unlike with pointer-type input, there is no screen position associcated with these actions. Rather, navigation actions always operate on the current selection.
```
keiramoon: Change "i.e." to "that is," (per Unity's [style guidelines](https://unity-docs.gitbook.io/style-guide/terminology/terminology-list#i), we don't use i.e.)

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+
+For this raycasting to work, you need to add [`TrackedDeviceRaycaster`](../api/UnityEngine.InputSystem.UI.TrackedDeviceRaycaster.html) to the `GameObject` that has the UI's `Canvas` component. This `GameObject` will usually have a `GraphicRaycaster` component which, however, works for 2D screen-space raycasting only. You can put [`TrackedDeviceRaycaster`](../api/UnityEngine.InputSystem.UI.TrackedDeviceRaycaster.html) alongside `GraphicRaycaster` and both can be enabled at the same time without advserse effect.
```
keiramoon: Change "which, however, works for 2D screen-space raycasting only." "which, however, only works for 2D screen-space raycasting."

Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -16,26 +16,80 @@ You can use the following properties to configure [`InputSystemUIInputModule`](.
...
+
+![TrackedDeviceRayster Add Component](Images/TrackedDeviceRaycasterComponentMenu.png)
+
+![TrackedDeviceRayster Properties](Images/TrackedDeviceRaycaster.png)
+
+Note that clicks on tracked devices do not differ from other [pointer-type input](#pointer-type-input) meaning that actions such as [`Left Click`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_leftClick) work for tracked devices just like they work for other pointers.
```
keiramoon: Split this sentence in two:

Clicks on tracked devices do not differ from other [pointer-type input](#pointer-type-input). Therefore, actions such as [`Left Click`](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html#UnityEngine_InputSystem_UI_InputSystemUIInputModule_leftClick) work for tracked devices just like they work for other pointers.

keiramoon (changes requested):
Added some style suggestions.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -739,6 +739,42 @@ public void Actions_CanReadValueFromAction_InCallback()
...
+        action.performed +=
+            ctx =>
+        {
+            Assert.That(receivedValue, Is.Null);
+            receivedValue = ctx.ReadValueAsButton();
+        };
+
+        Set(gamepad.leftTrigger, 0.25f);
+
+        Assert.That(receivedValue, Is.False);
```
karljj1, line 763: Be good to have a message in the assert for when it fails to explain why we expected it to be true.

e.g

"Expected received value to be false because its value was below the threshold however it was true"

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -739,6 +739,42 @@ public void Actions_CanReadValueFromAction_InCallback()
...
+
+        receivedValue = null;
+
+        Set(gamepad.leftTrigger, 0.75f);
```
karljj1, line 767: Similar thing here

"Expected received value to be true because its value was above the threshold however it was false."

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -995,341 +2344,39 @@ public IEnumerator TODO_UI_CanStartNavigationWhenNothingIsSelected()
...
-        }
-    }
-
-#endif
+    ";
 
     private enum EventType
     {
-        Click,
-        Down,
```
karljj1, line 1326: Should we mark these as [Obsolete] and provide an upgrade rule?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1429,6 +1429,26 @@ public TValue ReadValue<TValue>()
...
+            /// global <see cref="InputSettings.defaultButtonPressPoint"/> will be used.
+            /// </remarks>
+            /// <seealso cref="InputSettings.defaultButtonPressPoint"/>
+            /// <seealso cref="ButtonControl.pressPoint"/>
+            public bool ReadValueAsButton()
+            {
+                var value = false;
+                if (m_State != null && phase != InputActionPhase.Canceled)
+                    value = m_State.ReadValueAsButton(bindingIndex, controlIndex);
+                return value;
```
karljj1, line 1449: Do we need value? We could just do:

```
if (m_State != null && phase != InputActionPhase.Canceled)
	return m_State.ReadValueAsButton(bindingIndex, controlIndex);
return false;
```

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -405,98 +507,98 @@ internal void ProcessJoystick(ref JoystickModel joystickState)
...
+        [Obsolete("Use moveRepeatRate")]
         public float repeatRate
         {
-            get => m_RepeatRate;
-            set => m_RepeatRate = value;
+            get => moveRepeatRate;
+            set => moveRepeatRate = value;
         }
 
+        [Obsolete("Use moveRepeatDelay")]
```
karljj1: Should be able to use the api updater:

`[Obsolete("'repeatDelay' has been obsoleted; use 'moveRepeatDelay' instead. (UnityUpgradable) -> moveRepeatDelay", true)]`

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -405,98 +507,98 @@ internal void ProcessJoystick(ref JoystickModel joystickState)
...
+        /// <see cref="move"/>
+        public float moveRepeatRate
         {
-            get => m_RepeatDelay;
-            set => m_RepeatDelay = value;
+            get => m_MoveRepeatRate;
+            set => m_MoveRepeatRate = value;
         }
 
+        [Obsolete("Use moveRepeatRate")]
```
karljj1: Support  api updater:
[Obsolete("'repeatRate' has been obsoleted; use 'moveRepeatRate' instead. (UnityUpgradable) -> moveRepeatRate", true)]

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -608,33 +713,26 @@ public InputActionReference submit
...
+            set => SwapAction(ref m_TrackedDeviceOrientationAction, value, m_ActionsHooked, m_OnTrackedDeviceOrientationDelegate);
         }
 
         public InputActionReference trackedDevicePosition
         {
             get => m_TrackedDevicePositionAction;
-            set => SwapAction(ref m_TrackedDevicePositionAction, value, m_ActionsHooked, OnAction);
-        }
-
-        public InputActionReference trackedDeviceSelect
```
karljj1, line 626: Mark as obsolete?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/TrackedDeviceRaycaster.cs
```diff
@@ -226,6 +240,39 @@ private bool RayIntersectsRectTransform(RectTransform transform, Ray ray, out Ve
...
+
+        [FormerlySerializedAs("checkFor2DOcclusion")]
+        [SerializeField]
+        private bool m_CheckFor2DOcclusion;
+
+        [FormerlySerializedAs("checkFor3DOcclusion")]
+        [SerializeField]
+        private bool m_CheckFor3DOcclusion;
+
+        [Tooltip("Maximum distance (in 3D world space) that rays are traced to find a hit.")]
```
karljj1, line 256: Are we localizing Input System? We may want to localize these tooltips.

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Done

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Done

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Indeed. Fixed.

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: The last two entries actually made it in here by accident. They are related to changes not in the PR. Removed the two entries.

Packages/com.unity.inputsystem/Documentation~/UISupport.md
Rene-Damm: Done

Packages/com.unity.inputsystem/Documentation~/UISupport.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/UISupport.md
Rene-Damm: Done

Packages/com.unity.inputsystem/Documentation~/UISupport.md
Rene-Damm: Done

Packages/com.unity.inputsystem/Documentation~/UISupport.md
Rene-Damm: Done

Packages/com.unity.inputsystem/Documentation~/Installation.md
Rene-Damm: Done

Assets/Tests/InputSystem/Plugins/UITests.cs
Rene-Damm: This is internal stuff used only in the tests.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
Rene-Damm: Done

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
Rene-Damm: Done

Assets/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: Not a fan of test assert messages TBH (unlike for asserts in normal code). It adds a *shit ton* of clutter to the tests and makes them extremely wordy while IMO adding not that much in value. If an assert fails, I'll have to go and usually debug the test anyway to find what I've broken. Having the message wouldn't speed up the process. If the test is well-written, I will very quickly see why the expectation is the way it is.

But... maybe it's all lame excuses and I'm just lazy. Not saying this might not be the case :)

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
Rene-Damm: Done.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/TrackedDeviceRaycaster.cs
Rene-Damm: Do we support localization for TooltipAttributes?

In general, while there's some EditorGUIUtility.TrTextContent calls are in the UI frontend, no serious effort has yet been put into localization. I'd like to see this tackled after 1.0. For player stuff (like rebind screens) as well as for in-editor UIs.

Assets/Tests/InputSystem/CoreTests_Actions.cs
karljj1: Haha no problem

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/TrackedDeviceRaycaster.cs
karljj1: We do have support for localizing packages. We can look into it after we leave preview.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/TrackedDeviceRaycaster.cs
```diff
@@ -1,12 +1,23 @@
...
+    /// <summary>
+    /// Racyasting implementation for use with <see cref="TrackedDevice"/>s.
+    /// </summary>
+    /// <remarks>
+    /// This component needs to be added alongside the <c>Canvas</c> component. Usually, raycasting is
+    /// performed by the <c>GraphicRaycaster</c> component found there but for 3D raycasting necessary for
+    /// tracked devices, this component is required.
+    /// </remarks>
+    [AddComponentMenu("Event/Tracked Device Raycaster")]
+    [RequireComponent(typeof(Canvas))]
```
pcosgrave, line 20: Why do we require the canvas? I think it would be nice if this just worked for the whole EventSystem instead of just with UI.

Rene-Damm: It seems the TrackedDeviceRaycaster needs a canvas to work. I'm not very familiar with the internals of this thing. @StayTalm wrote this raycaster. With the attributes here I mainly followed what `GraphicRaycaster` is doing. It also does the `RequirementComponent(typeof(Canvas))` thing.

pcosgrave: Yeah would have been interesting to extend it past just the UI but it does 1 get for the canvas elements so i know why its necessary. Maybe future improvement if we'd want.

Rene-Damm: Would great if we can do this as part of the UI Elements work. My guess is that when it comes to implementing that, there's several things in the current uGUI support code of the input system that make sense to reuse. Hopefully we can generalize these at that point. Ideally, I'd love for users to be able to rather easily hook up their own UI systems, too. Maybe a pipe dream :)

keiramoon (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
Rene-Damm: Flip-flopping on this one. I'm certainly neither an advocate of strict one-return-per-method nor of the opposite.

For methods that do nothing but compute and return one value, I find it nice to have a single point in the method where I can set a breakpoint. If you return the value directly, that becomes harder.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/TrackedDeviceRaycaster.cs
Rene-Damm: Let's do it :)

PR #1084, opened 2020-03-09
Initialize last update time to MinValue to account for negative times

cbrwn:
Rarely, devices' internal times can be negative (not sure the cause), so the initial value of `m_lastUpdateTimeInternal` being 0 results in no input detected from these devices.

Here I just initialize it to `MinValue` in the constructor which fixed it for me, but a better fix may exist.

Here's a screenshot of my mouse and keyboard events having negative times (which detected no input) next to my Xbox controller events with positive times, which detected input.

image


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1084) <br/>All committers have signed the CLA.

jimon:
Hi! We've tried to reproduce the issue with latest Unity versions, and while device event times becoming negative is expected when entering playmode, any new events after that should contain a positive event time in the debugger.

If the issue still persist please create a bug report instead and attach this PR.

PR #1095, opened 2020-03-19
FIX: VirtualMouseInput SW cursor not working when using HardwareCursorIfAvailable (case 1224591).

Rene-Damm:
Fixes [1224591](https://fogbugz.unity3d.com/f/cases/1224591/).

### DIFF
--- a/Assets/Samples/GamepadMouseCursor/VirtualMouseInput.cs
+++ b/Assets/Samples/GamepadMouseCursor/VirtualMouseInput.cs
@@ -2,6 +2,8 @@ using System;
 using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.UI;
 
+////TODO: respect cursor lock mode
+
 ////TODO: investigate how driving the HW cursor behaves when FPS drops low
 ////      (also, maybe we can add support where we turn the gamepad mouse on and off automatically based on whether the system mouse is used)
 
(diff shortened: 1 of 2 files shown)


Assets/Samples/GamepadMouseCursor/VirtualMouseInput.cs
```diff
@@ -2,6 +2,8 @@
 using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.UI;
 
+////TODO: respect cursor lock mode
```
karljj1, line 5: Should we be capturing this stuff somewhere else? Maybe a favro board?

karljj1 (approved):

Assets/Samples/GamepadMouseCursor/VirtualMouseInput.cs
Rene-Damm: Most of the bigger tickets are captured in our [JIRA backlog](https://unity3d.atlassian.net/secure/RapidBoard.jspa?rapidView=87&view=planning.nodetail). However, in general I find that keeping things *with* the source is indeed the most effective way to track stuff. These are the TODOs and REVIEWs that I find myself coming back to over and over whereas the JIRA stuff is out of view and slowly decaying most of the time. Like with docs, I find that the further you go from the source, the more useless and outdated the stuff becomes.

PR #1099, opened 2020-03-24
FIX: Fixed compilation errors when disabling built-in VR and XR packages. (case 1230225)

karljj1:
ENABLE_VR is not the correct define to use. We need to create a define for both the XR and VR module.
Fixes: https://fogbugz.unity3d.com/f/cases/1230225/


Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/GoogleVR.cs
```diff
@@ -1,4 +1,4 @@
-#if ENABLE_VR && !DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR && !PACKAGE_DOCS_GENERATION
+#if UNITY_INPUT_SYSTEM_ENABLE_XR && !DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR && !PACKAGE_DOCS_GENERATION
```
Rene-Damm, line 1: Not directly related to this PR but `&& !PACKAGE_DOCS_GENERATION` just caught my eye. That doesn't look right. Seems like that should be `(...) || PACKAGE_DOCS_GENERATION`.

PR #1100, opened 2020-03-24
FIX: Missing InputActionReferences when renaming actions (case 1129145).

Rene-Damm:
Fixes [1129145](https://fogbugz.unity3d.com/f/cases/1229145/) ([Issue Tracker](https://issuetracker.unity3d.com/issues/inputsystem-inputactionreference-loses-guid-when-its-action-is-moved-or-renamed-in-the-inputaction-asset)).

This one isn't so great. The problem is fixed but a reimport will break all existing InputActionReferences that are referenced elsewhere in the project. I couldn't come up with a way to avoid that.

The problem is that I had used the "<map>/<action>" names for actions as the names passed to `AddObjectToAsset` which meant these names were fed into the name hashing that ultimately generates local file IDs. So... change action name -> different file ID -> missing InputActionReference.

With the change, we now use the action GUID as the name to generate a hash for an InputActionReference. So that's stable. BUT.... it breaks any existing InputActionReference already in the project :(

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -17,6 +17,9 @@ however, it has to be formatted properly to pass verification tests.
 
 #### Actions
 
+- References to `InputActionReference` objects created by the importer for `.inputactions` files are no longer broken when the action referenced by the object is renamed ([case 1229145](https://issuetracker.unity3d.com/issues/inputsystem-inputactionreference-loses-guid-when-its-action-is-moved-or-renamed-in-the-inputaction-asset)).
+  * __THIS IS A BREAKING CHANGE__: Unfortunately, this change impacts Unity internal file IDs of `InputActionReference` objects created by the `.inputactions` file importer. This means that previously created references to `InputActionReference` objects will become missing and need to be recreated.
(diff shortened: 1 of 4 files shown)


karljj1 (approved):
Could we provide some tool to auto fix peoples projects when they upgrade the package?

Rene-Damm:
> Could we provide some tool to auto fix peoples projects when they upgrade the package?

Unfortunately, I don't know of a good way to do this :( We'd have to load basically take all Unity serialized files in the project, check them for InputActionReferences, determine the action they *would* reference in the asset and update them to the action that they now reference.

I'm thinking that the way the current InputActionReferences work is fundamentally flawed. The way the whole thing is set up they are stored *with* the imported asset when really the references should be stored with the data containing the reference.

I hope a better solution will emerge after 1.0.

Rene-Damm:
This PR was a brainfart. Breaks any existing InputSystemUIInputModule instance. Reverted.

Rene-Damm:
> Could we provide some tool to auto fix peoples projects when they upgrade the package?

@karljj1 Doh, I think I just realized there is a super simple solution... just create two InputActionReferences in the importer. One that is hidden from the project browser and has the old file ID and one that is visible and has the new one. This way all existing references keep working and any newly established reference will pick up the new behavior.

PR #1101, opened 2020-03-24
FIX: Messages from PlayerInput not received when actions contain spaces (case 1214519).

Rene-Damm:
Fixes [1214519 ](https://fogbugz.unity3d.com/f/cases/1214519/)([Issue Tracker](https://issuetracker.unity3d.com/issues/player-input-send-messages-wont-trigger-when-input-action-name-contains-spaces)).

This incorporates the fix supplied by @BHSpitMonkey in #1022 and adds a test for it.

Also took the opportunity to add a no-GC-garbage test for PlayerInput events/messages.


Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -826,6 +829,25 @@ public void PlayerInput_ByDefaultChoosesMostSpecificControlSchemeAvailable()
...
+        InputSystem.AddDevice<Gamepad>();
+
+        var go = new GameObject();
+        var playerInput = go.AddComponent<PlayerInput>();
+        playerInput.defaultActionMap = "gameplay";
+        playerInput.notificationBehavior = PlayerNotifications.SendMessages;
+        playerInput.actions = InputActionAsset.FromJson(kActions);
+        var listener = go.AddComponent<MessageListener>();
+
+        Press((ButtonControl)playerInput.actions["Action With Spaces!!"].controls[0]);
```
karljj1, line 846: Maybe we could have a parametric test with a few different variations on special characters and spaces

karljj1 (approved):

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
Rene-Damm: Probably. Opting on the lazy side for this one :)

PR #1106, opened 2020-03-27
FIX: Warnings from package validation suite.

Rene-Damm:


John-Whigham (approved):
Tested locally and now passes NDA validation without any warnings

PR #1110, opened 2020-03-30
FIX: Newly added actions or maps being outside of visible area.

Rene-Damm:
[Forum thread](https://forum.unity.com/threads/input-system-update.508660/page-16#post-5643580)

Looks like staging registry is finally defunct. Embedded doctools package to get CI green.


karljj1 (approved):
152 files ;-)

Rene-Damm:
> 152 files ;-)

Only 3 files modified outside of `Packages/com.unity.package-manager-doctools`, though. Just a huge stack of added files (end megabytes) due to having to embed the package (for now; hopefully we can eventually kill this).

PR #1113, opened 2020-04-01
CHANGE: Moved UI to be an optional dependency.

karljj1:
- UI was a package dependency which forced it to always be included when the Input package was included. Moved it to be an optional dependency. UI code will only be compiled when the UI package is also in the project.
- Added defines to Physics and Physics2d code so that the modules can be disabled.
- Fixed some typos I spotted ;)


Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -40,7 +43,7 @@ namespace UnityEngine.InputSystem
     /// <remarks>
     /// PlayerInput is a high-level wrapper around much of the input system's functionality
     /// which is meant to help getting set up with the new input system quickly. It takes
-    /// care of <see cref="InputAction"/> bookkeeping and has a custom UI to help
+    /// care of <see cref="InputAction"/> bookkeeping and has a custom UI(Requires UI package) to help
```
Rene-Damm: Nitpick... ` (requires UI package)`

Rene-Damm: Actually, on second thought, given the ambiguities of what "UI" may refer to, probably better to explicitly use the name of the package here (`requires the "Unity UI" package`).

Packages/com.unity.inputsystem/package.json
```diff
@@ -18,8 +18,5 @@
 		"touch",
 		"vr",
 		"xr"
-	],
-	"dependencies": {
-		"com.unity.ugui": "1.0.0"
-	}
-}
```
Rene-Damm, line 25: With this gone, do we need to add the ugui dependency to Packages/manifest.json? We do want uGUI to be pulled into the InputSystem project as a whole.

Rene-Damm (commented):
Nice!

Packages/com.unity.inputsystem/package.json
karljj1: > 
> 
> With this gone, do we need to add the ugui dependency to Packages/manifest.json? We do want uGUI to be pulled into the InputSystem project as a whole.

UGUI did not become a package until 2019.2 so we can't really add it as the project is 2019.1 at the moment. It will be automaticvally added when the project upgrades. If it doesn t then the project will fail as it has a lot of ugui code in it.

Rene-Damm: Ah right, forgot about the automatic upgrade for the 2019.1 project. All good then :)

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenButton.cs
```diff
@@ -1,3 +1,4 @@
+#if UNITY_INPUT_SYSTEM_ENABLE_UI
```
Rene-Damm: I realized we need to add `|| PACKAGE_DOCS_GENERATION` to the #ifs that have public APIs in the files.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenButton.cs
karljj1: That would explain my failing test

Rene-Damm (approved):

PR #1114, opened 2020-04-01
FIX: Actions and bindings getting forcibly expanded all the time.

Rene-Damm:
The action editor was forcibly expanding the action tree view all the time. This happened every single time we reloaded the tree, i.e. any time an item (action or binding) was added or removed and on every domain reload.

I've removed the force expansion altogether. Which also means that the tree view now opens in a collapsed state, i.e. the bindings are hidden by default and only the actions visible. This seems in line with what we do elsewhere in Unity (e.g. we don't expand the scene hierarchy by default either).

Added context menu entries to force-expand/collapse the entire tree view.


karljj1 (approved):
Seems fine although I wonder if we could store the expansion state instead so that it looks the same after a reload? I have seen Editor windows store SerializedValues so maybe we could do the same or use a EditorPref?

Rene-Damm:
> Seems fine although I wonder if we could store the expansion state instead so that it looks the same after a reload? I have seen Editor windows store SerializedValues so maybe we could do the same or use a EditorPref?

I think storing the view state in Library the same way we store expandedItems is probably a good idea. Added a TODO to the code.

PR #1128, opened 2020-04-23
Fix AxisControl types without clamps not updating properly

StayTalm:
There was an issue where an AxisControl without a min/max clamp value would only receive the first 'performed' event, and no subsequent events until the action was cancelled.  This was because AxisControl.EvaluateMagnitude early outs with -1 and so underlying action systems think that the action has not changed between frames.  It feels appropriate that the magnitude of an AxisControl is just the absolute value if there are no other parameters, but I'm not sure if there was a specific reason for it.

This fixes internal bug #1183862


jackpr-unity (approved):
Solves https://fogbugz.unity3d.com/f/cases/1183862

StayTalm:
@Rene-Damm 
Updated with test.  Also with additional info:
https://github.com/Unity-Technologies/InputSystem/blob/4dff741829ccaa9e5ff12349ebd5b71f2fb74423/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs#L1009
This function relies on EvaluateMagnitude in order to identify a change in the control.  When updating the control frame-by-frame, it leaves this function here: https://github.com/Unity-Technologies/InputSystem/blob/4dff741829ccaa9e5ff12349ebd5b71f2fb74423/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs#L1236
Because the control is either cancelled or the result of EvaluateMagnitude is -1 (so the trigger and action state magnitudes are the same).

Not sure this is the right fix.  This test uses a second gamepad so that conflict resolution turns on.  This axis won't be normalized, so it will be able within a potentially different range.

Rene-Damm:
I suspect the right fix for this is in the action system's disambiguation code, not in `EvaluateMagnitude`. But let's go and merge it. There's some other stuff related to disambiguation not working as it should that needs to be looked at. I'll try to have a look at this when I get around to that.

PR #1133, opened 2020-04-28
XR Haptics can now use an interface.

StayTalm:
- Added IXRRumble as an interface that can do XR-type rumble events
- XRControllerWithRumble now inherits from IXRRumble
- XR Rumble can now check if rumble is supported, and can send a rumble event to multiple channels.
- Updated Test Scene.


jackpr-unity (approved):

Rene-Damm (changes requested):
Missing API docs.

StayTalm:
@Rene-Damm Added API Docs

PR #1140, opened 2020-05-07
Fixed PrimitiveValue.FromString to handle NaN value

STARasGAMES:
Issue: https://issuetracker.unity3d.com/issues/inputsystem-error-when-vjoy-is-installed
vJoy may produce NaN value that breaks the process of adding a device.


onnimie:
How can you edit the code in PrimitiveValues.cs without Unity somehow blocking the editing? When I edit the code and launch Unity, it seems to restore the code back to the way it was before editing. If I open PrimitiveValues.cs after launching Unity, the edit has been removed and the code restored.

STARasGAMES:
@onnimie yeah that sucks. I needed to just check if joystick is working so for me it was enough.
For you I can only suggest importing the whole input system into the assets folder.

jimon:
Hi, I'm sorry for the delay, we had another look on the root cause of the issue with input system and vJoy, and fixed it in #1263, hence I'm closing this PR. Thanks for taking an effort and care to contribute and fix the issue! 👍

PR #1141, opened 2020-05-07
FIX: CHANGELOG dates not conforming to ISO 8601.

Rene-Damm:
upm-ci has been spilling warnings about dates not conforming to the expected format. Fixed the dates and added a test to make sure that dates added in the future also conform.


karljj1 (approved):
Sorry I thought I had already reviewed this one

Issue #1142, opened 2020-05-08
Could not create a device for 'DUALSHOCK 4 Wireless Controller (iOS)'


nnul:
On tvOS the touchpadButton input control is not setup and causes the following runtime exception:

### Callstack

```sh
Could not create a device for 'DUALSHOCK 4 Wireless Controller (iOS)' (exception: System.ArgumentException: Cannot find input control 'touchpadButton'
Parameter name: path
  at UnityEngine.InputSystem.InputControl.GetChildControl (System.String path) [0x00000] in <00000000000000000000000000000000>:0 
  at UnityEngine.InputSystem.InputControl.GetChildControl[TControl] (System.String path) [0x00000] in <00000000000000000000000000000000>:0 
  at UnityEngine.InputSystem.DualShock.DualShockGamepad.FinishSetup () [0x00000] in <00000000000000000000000000000000>:0 
  at UnityEngine.InputSystem.InputControl.CallFinishSetupRecursive () [0x00000] in <00000000000000000000000000000000>:0 
  at UnityEngine.InputSystem.InputDevice.Build[TDevice] (System.String layoutName, System.String layoutVariants, UnityEngine.InputSystem.Layouts.InputDeviceDescription deviceDescription) [0x00000] in <00000000000000000000000000000000>:0 
  at UnityEngine.InputSystem.InputManager.AddDevice (UnityEngine.InputSystem.Utilities.InternedString layout, System.Int32 deviceId, System.String deviceName, UnityEngine.InputSystem.Layouts.InputDeviceDescription deviceDescription, UnityEngine.InputSystem.InputDevice+DeviceFlags deviceFlags, UnityEngine.InputSystem.Utilities.InternedString variants) [0x00000] in <00000000000000000000000000000000>:0 
  at UnityEngine.InputSystem.InputManager.AddDevice (UnityEngine.InputSystem.Layouts.InputDeviceDescription description, System.Boolean throwIfNoLayoutFound, System.String deviceName, System.Int32 deviceId, UnityEngine.InputSystem.InputDevice+DeviceFlags deviceFlags) [0x00000] in <00000000000000000000000000000000>:0 
  at UnityEngine.InputSystem.InputManager.OnNativeDeviceDiscovered (System.Int32 deviceId, System.String deviceDescriptor) [0x00000] in <00000000000000000000000000000000>:0 
[...]

PR #1143, opened 2020-05-08
FIX: DUALSHOCK 4 controller initialization bug on iOS w/ touchpadButton

nnul:
A fix for issue #1142


nnul:
I'm working on a proper fix

PR #1145, opened 2020-05-09
FIX: Controller input on tvOS

nnul:
+ XInputSupport and DualShockSupport are now initialized for tvOS


nnul:
This fixes #1142

nnul:
https://issuetracker.unity3d.com/issues/tvos-input-system-dualshock-controller-causes-an-could-not-create-a-device-error

karljj1:
Hi thanks for the PR. We have moved this fix over to an internal PR so we can run our CI and QA on it before merging. https://github.com/Unity-Technologies/InputSystem/pull/1155

PR #1150, opened 2020-05-19
FIX: Drag-reordering losing bindings (case 1228000).

Rene-Damm:
Fixes [1228000](https://fogbugz.unity3d.com/f/cases/1228000/) ([Issue Tracker](https://issuetracker.unity3d.com/issues/input-system-bindings-get-cleared-for-other-control-scheme-actions-when-reordering-an-action-in-a-specific-control-scheme)).

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
+## [Unreleased]
+
+### Fixed
+
+#### Actions
+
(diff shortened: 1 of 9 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
+## [Unreleased]
+
+### Fixed
+
+#### Actions
```
karljj1, line 14: Does Fixed need a sub category for Actions? Will we have multiple fixes in this area?

karljj1 (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Not sure it's strictly necessary. There will definitely be multiple entries in this area. My initial thought behind having an explicit subheading for actions was that most of the fixes elsewhere tended to not be that relevant to most users whereas the action-related changes/fixes tend to be the most important ones.

PR #1153, opened 2020-05-29
FIX: Added ENABLE_VR to XR and VR define checks

karljj1:
Checking if the VR or XR module is present is not enough. In the XR and VR module, there are some parts of code that are only compiled if **ENABLE_VR** is defined. This means that the module API can change and even though the module is included it may not actually have the API we expect. For example _XRDevice_:

image

We now check the ENABLE_VR define and the module.
Also cleaned up the asmdef which had a duplicate entry


Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -36,16 +31,16 @@
             "expression": "2.0.3",
             "define": "DISABLE_BUILTIN_INPUT_SYSTEM_WINDOWSMR"
         },
-        {
-            "name": "Select...",
```
karljj1, line 40: Unused entry

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -36,16 +31,16 @@
...
-            "expression": "2.0.3",
-            "define": "DISABLE_BUILTIN_INPUT_SYSTEM_WINDOWSMR"
-        },
         {
             "name": "com.unity.modules.vr",
             "expression": "1.0.0",
             "define": "UNITY_INPUT_SYSTEM_ENABLE_VR"
         },
+        {
```
karljj1, line 39: I moved it so its next to vr

Rene-Damm (approved):

PR #1155, opened 2020-06-01
FIX: Fixed DualShock controller causing a "Could not create a device" erro…

karljj1:
Copy of this PR: https://github.com/Unity-Technologies/InputSystem/pull/1145
Fixes https://fogbugz.unity3d.com/f/cases/1221223/

Fixes a bug that prevented DualShock controllers from working on tvOS.

I created this PR for @itsneil (neil.skinner) as he does not have permission to push. He will be verifying the fix with QA before we merge.


Rene-Damm:
Seems like the existing test in `iosTests` would have failed on tvOS and caught the error. Wondering... can we set up a test where we simply cycle the project through all platforms and run the tests in the editor on each? Not the same as running on the device but it should already be able to catch a number of errors that otherwise go unnoticed.

Rene-Damm (approved):

PR #1156, opened 2020-06-02
NEW: Add concept of 'precompiled layouts'.

Rene-Damm:
## Why?

Creating devices is *slow* and creates tons of GC heap garbage (some of which is VM reflection data so will stick around for the entire rest of the game).

## What does it do?

GeneratePrecompiledLayout

This change preserves the layout system with all its flexibility yet gives us (and users) the ability to just "bake" specific layouts down. It's done in a way where specific baked versions are automatically removed/ignored if the user alters layouts such that they are affected but as long as that is not the case, whenever there is a precompiled version, the input system will *NOT* load the respective `InputControlLayout` at all and simply invoke a factory method that directly `new`s the respective device (containing the auto-generated code).

```CSharp
InputSystem.RegisterPrecompiledLayout<FastKeyboard>(FastKeyboard.metadata);
```

I have some performance tests which, along with another batch of performance tests, I'm adding in a separate PR. On my own machine, I've measured a >20x speedup for creating a `Touchscreen`.

For an example of what the generated code looks like, see [FastKeyboard](https://github.com/Unity-Technologies/InputSystem/blob/691e55419c1c1053a8f2194888a6b28f8cb42e1c/Packages/com.unity.inputsystem/InputSystem/Devices/Precompiled/FastKeyboard.cs).

I've added an API test to make sure that the precompiled versions we have are in fact up-to-date.

## What's the next step?

In the precompiled devices, we can make certain assumptions that we otherwise can't. For example, the memory layout of the device is entirely fixed. For `Mouse`, I want to exploit that in the future and get rid of the expensive reads and writes it performs to accumulate and reset delta controls. For `FastMouse`, we can simply hardcode much of that logic and get rid of a whole bunch of virtual method calls. With mouse events being rather high-frequency, this actually matters.


Assets/Samples/SimpleDemo/SimpleControls.cs
```diff
@@ -50,7 +59,18 @@ public @SimpleControls()
                     ""path"": ""*/{PrimaryAction}"",
                     ""interactions"": """",
                     ""processors"": """",
-                    ""groups"": """",
+                    ""groups"": ""Gamepad;KeyboardMouse"",
```
Rene-Damm: Ooppps... any of the changes from here to the end of the file are accidental commits. Will revert.

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,84 @@ const string json = @"
...
+InputSystem.RegisterLayout<ExtendedButtonControl>("Button");
+
+// PrecompiledMyDevice has implicitly been removed now as the ButtonControl it uses
+// has now been replaced with ExtendedButtonControl.
+```
+
+Note that you can add `#if` checks to the generated code, if needed. If you add them like in the following code snippet, the checks will be preserved even if the code is regenerated. Similarly, you can change the generated class from `public` to `internal` and the modifier will be preserved when regenerating the class.
+
+```CSharp
+//!!DEFINES
```
Rene-Damm: @karljj1 Looking at this now, I think this is stupid. I want the user to be able to add #ifs to the generated file but required an explicit markup with `//!!DEFINES`. I think the code should simply scan for an #if and if so, preserve it.

Packages/com.unity.inputsystem/InputSystem/Controls/Processors/AxisDeadzoneProcessor.cs
```diff
@@ -6,15 +6,46 @@
...
+    /// </code>
+    /// // Bind to right trigger on gamepad such that the value is clamped and normalized between
+    /// // 0.3 and 0.7.
+    /// new InputAction(binding: "&lt;Gamepad&gt;/rightTrigger", processors: "axisDeadzone(min=0.3,max=0.7)");
+    /// </example>
+    /// </remarks>
+    /// <seealso cref="StickDeadzoneProcessor"/>
     [Preserve]
-    internal class AxisDeadzoneProcessor : InputProcessor<float>
+    public class AxisDeadzoneProcessor : InputProcessor<float>
```
Rene-Damm, line 28: Most of the processors were internal. Precompiled layouts need to be able to just `new` them so I made them public. Also, don't really see a good reason why they should not be.

Packages/com.unity.inputsystem/InputSystem/Devices/Joystick.cs
```diff
@@ -61,7 +61,7 @@ public class Joystick : InputDevice
         /// This is the <see cref="ButtonControl"/> type control on the joystick
         /// that has the <see cref="CommonUsages.PrimaryTrigger"/> usage.
         /// </remarks>
-        public ButtonControl trigger { get; private set; }
+        public ButtonControl trigger { get; protected set; }
```
Rene-Damm, line 64: These changes to various devices are required so that the generated C# class can set these properties directly. I could have simply invoked `FinishSetup` like `InputDeviceBuilder` does, but I really wanted to go all the way with this and get rid of all the path lookups done in `FinishSetup` methods through `GetChildControl`. So, `InputControlLayoutGenerator` actually goes and looks at which controls these kind of properties lead to and then just hardwires that directly into the generated class.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -808,5 +809,337 @@ internal static string BuildPath(this InputControl control, string deviceLayout,
 
             return builder.ToString();
         }
+
+        // Undocumented APIs. Meant to be used only by auto-generated, precompiled layouts.
+        // These APIs exist solely to keep access to the various properties/fields internal
+        // and only allow their contents to be modified in a controlled manner.
+        #region Undocumented
```
Rene-Damm, line 816: This one's a compromise. The generated classes need a way to mutate data that is otherwise considered immutable. So far, this was not a problem as `InputDeviceBuilder` was the only place that required access. But the code generated for precompiled layouts may not be part of the input system and thus may not have `internal`-level access (the test I added in `CoreTests_Editor` sets up such a scenario).

However, I also did not want to simply expose all the properties/fields for arbitrary mutation. So I chose to have a "builder" kind of API that has time-limited write access which is turned off as soon the setup of a device is completed. But... I also didn't want this to be an API the user has to deal with so I chose to keep it out of thie documentation.

I added `AggressiveInlining` annotations to attempt to get rid of the compiler just generating tons and tons of function call overhead for what really is simple field writes.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -17,6 +24,10 @@ however, it has to be formatted properly to pass verification tests.
 
 - Fixed drag&drop reordering actions while having one control scheme selected causing bindings from other control schemes to be lost ([case 122800](https://issuetracker.unity3d.com/issues/input-system-bindings-get-cleared-for-other-control-scheme-actions-when-reordering-an-action-in-a-specific-control-scheme)).
 
+### Added
+
+- Device layouts can now be "precompiled" for speed. `Keyboard`, `Mouse`, and `Touchscreen` are now included as precompiled layouts greatly reducing instantiation time and GC heap cost for these devices. for `Touchscreen`, this results in a >20x speedup for `InputSystem.AddDevice<Touchscreen>()`.
```
keiramoon:
```suggestion
- Device layouts can now be "precompiled" for speed. `Keyboard`, `Mouse`, and `Touchscreen` are now included as precompiled layouts, greatly reducing instantiation time and GC heap cost for these devices. For `Touchscreen`, this results in a >20x speed-up for `InputSystem.AddDevice<Touchscreen>()`.
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,84 @@ const string json = @"
 
 InputSystem.RegisterLayoutOverride(json);
 ```
+
+## Precompiled layouts
+
+Building a device at runtime from an [`InputControlLayout`](../api/UnityEngine.InputSystem.Layouts.InputControlLayout.html) is a slow process. The layout instance itself has to be built (which may involve reflection) and then interpreted in order to put the final [`InputDevice`](../api/UnityEngine.InputSystem.InputDevice.html) instance together. This process usually involves the loading of multiple [`InputControlLayout`](../api/UnityEngine.InputSystem.Layouts.InputControlLayout.html) instances each of which may be the result of merging multiple layouts together (if the layout involves [inheritance](#layout-inheritance) or [overrides](#layout-overrides)).
```
keiramoon:
```suggestion
Building a device at runtime from an [`InputControlLayout`](../api/UnityEngine.InputSystem.Layouts.InputControlLayout.html) is a slow process. The layout instance itself has to be built (which might involve reflection) and then interpreted in order to put the final [`InputDevice`](../api/UnityEngine.InputSystem.InputDevice.html) instance together. This process usually involves the loading of multiple [`InputControlLayout`](../api/UnityEngine.InputSystem.Layouts.InputControlLayout.html) instances, each of which might be the result of merging multiple layouts together (if the layout involves [inheritance](#layout-inheritance) or [overrides](#layout-overrides)).
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,84 @@ const string json = @"
...
+
+You can speed this process up dramatically by "baking" the final form of a layout into a "precompiled layout". This is simply generated C# code that when run, will build the corresponding device without relying on loading and interpreting an [`InputControlLayout`](../api/UnityEngine.InputSystem.Layouts.InputControlLayout.html). Aside from running faster, this will also create far less garbage and will not involve C# reflection (which generally causes runtime overhead by inflating the number of objects internally kept by the C# runtime).
```
keiramoon:
```suggestion
You can speed up this process up by "baking" the final form of a layout into a "precompiled layout". A precompiled layout is generated C# code that, when run, will build the corresponding device without relying on loading and interpreting an [`InputControlLayout`](../api/UnityEngine.InputSystem.Layouts.InputControlLayout.html). Aside from running faster, this will also create far less garbage and will not involve C# reflection (which generally causes runtime overhead by inflating the number of objects internally kept by the C# runtime).
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,84 @@ const string json = @"
...
+
+>__NOTE__: Precompiled layouts must be device layouts. It is not possible to precompile the layout for an [`InputControl`](../api/UnityEngine.InputSystem.InputControl.html).
+
+### Creating a precompiled layout
+
+The first step in setting up a precompiled layout is to generate it. To do so, open the [Input Debugger](./Debugging.md), navigate to the layout you want to precompile within the "Layouts" branch, right-click it and select "Generate Precompiled Layout".
```
keiramoon:
```suggestion
The first step in setting up a precompiled layout is to generate it. To do so, open the [Input Debugger](./Debugging.md), navigate to the layout you want to precompile within the **Layouts** branch, right-click it, and select **Generate Precompiled Layout**.
```

keiramoon: Style Guide: We format UI elements with bold (more info [here](https://unity-docs.gitbook.io/style-guide/format/bold-italics-and-underline)).

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,84 @@ const string json = @"
...
+
+![Generate Precompiled Layout](./Images/GeneratePrecompiledLayout.png)
+
+ You will be prompted for where to store the generated code. Pick a directory in your project and file name and click "Save".
```
keiramoon:
```suggestion
Unity will ask you where to store the generated code. Pick a directory in your project, enter a file name, and click **Save**.
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,84 @@ const string json = @"
...
+
+ Once generated, you can register the precompiled layout with the Input System using [`InputSystem.RegisterPrecompiledLayout`](../api/UnityEngine.InputSystem.InputSystem.html#UnityEngine_InputSystem_InputSystem_RegisterPrecompiledLayout__1_System_String_). The method expects a string argument containing metadata for the precompiled layout. This string is automatically emitted as a `const` inside the generated class.
+
+ ```CSharp
+ InputSystem.RegisterPrecompiledLayout<MyPrecompiledDevice>(MyPrecompiledDevice.metadata);
+ ```
+
+>__IMPORTANT__: It is very important that this method is called with all relevant layout registrations being in the same state as at the time the layout was precompiled. There is no internal check whether the precompiled layout will still generate an identical result to the non-precompiled version.
```
keiramoon, line 331:
```suggestion
>__IMPORTANT__: When you call this method, all relevant layout registrations must be in the same state as the state they were in when the layout was precompiled. There is no internal check whether the precompiled layout still generates an identical result to the non-precompiled version.
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,84 @@ const string json = @"
...
+
+Once registered, a precompiled layout is automatically used whenever the layout that the precompiled layout is based on is instantiated.
```
keiramoon, line 333:
```suggestion
Once registered, a precompiled layout is automatically used whenever the layout it is based on is instantiated.
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,97 @@ const string json = @"
...
+InputSystem.RegisterLayout<MyDevice>();
+
+// And you register a precompiled version of it then as well.
+InputSystem.RegisterPrecompiledLayout<PrecompiledMyDevice>(PrecompiledMyDevice.metadata);
+
+// Then the following will implicitly use the precompiled version.
+InputSystem.AddDevice<MyDevice>();
+```
+
+A precompiled layout will automatically be unregistered in the following cases:
```
keiramoon, line 364:
```suggestion
A precompiled layout is automatically unregistered in the following cases:
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,97 @@ const string json = @"
...
+
+* A [layout override](#layout-overrides) is applied to one of the layouts used by the precompiled Device. This also extends to [controls](Controls.md) used by the Device.
+* A layout with the same name as one of the layouts used by the precompiled Device is registered (which replaces the layout already registered under the name).
+* A [processor](Processors.md) is registered that replaces a processor used by the precompiled Device.
+
+This causes the Input System to fall back to the non-precompiled version of the layout. Note also that a precompiled layout will not be used for layouts [derived](#layout-inheritance) from the layout the precompiled version is based on. In the example above, if someone derives a new layout from `MyDevice`, the precompiled version is unaffected (it will not be unregistered) but is also not used for the newly created type of device.
+
+```CSharp
+// Let's constinue from the example above and assume that sometime
```
keiramoon, line 373:
```suggestion
// Let's continue from the example above and assume that sometime
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,97 @@ const string json = @"
...
+// later, someone replaces the builtin button with an extended version.
```
keiramoon:
```suggestion
// later, someone replaces the built-in button with an extended version.
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,97 @@ const string json = @"
...
+InputSystem.RegisterLayout<ExtendedButtonControl>("Button");
+
+// PrecompiledMyDevice has implicitly been removed now as the ButtonControl it uses
```
keiramoon:
```suggestion
// PrecompiledMyDevice has implicitly been removed now, because the ButtonControl it uses
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,97 @@ const string json = @"
...
+// has now been replaced with ExtendedButtonControl.
+```
+
+Note that you can add `#if` checks to the generated code, if needed. The code generator will scan the start of an existing file for a line starting with `#if` and, if found, preserve it in newly generated code and generate a corresponding `#endif` at the end of the file. Similarly, you can change the generated class from `public` to `internal` and the modifier will be preserved when regenerating the class. Finally, you can change the namespace
```
keiramoon:
```suggestion
If needed, you can add `#if` checks to the generated code, if needed. The code generator will scan the start of an existing file for a line starting with `#if` and, if found, preserve it in newly generated code and generate a corresponding `#endif` at the end of the file. Similarly, you can change the generated class from `public` to `internal` and the modifier will be preserved when regenerating the class. Finally, you can change the namespace
```

keiramoon: The last sentence looks incomplete. Was there meant to be more there?

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,97 @@ const string json = @"
...
+
+Also note that the generated class is marked as `partial`. This means you can add additional overloads and other code by simply having a parallel, `partial` class definition.
```
keiramoon:
```suggestion
The generated class is marked as `partial`, which means you can add additional overloads and other code by  having a parallel, `partial` class definition.
```

Packages/com.unity.inputsystem/Documentation~/Layouts.md
```diff
@@ -295,3 +305,97 @@ const string json = @"
...
+```CSharp
+// The next line will be preserved when regenerating the precompiled layout. A
+// corresponding #endif will be emitted at the end of the file.
+#if UNITY_EDITOR || UNITY_STANDALONE
+
+// If you change the namespace to a different one, the name of the namespace will be
+// preserved when you regenerate the precompiled layout.
+namepace MyNamespace
+{
+    // If you changing `public` to `internal`, the change will be preserved
```
keiramoon:
```suggestion
    // If you change `public` to `internal`, the change will be preserved
```

keiramoon (changes requested):
Tech writer review (documentation only). Left a few suggestions and fixed some typos.

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed and completed last sentence. Thanks for noticing :)

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Removed the need for an explicit `//!!DEFINES`. Just grabs an `#if` expression now if it finds one.

Packages/com.unity.inputsystem/Documentation~/Layouts.md
Rene-Damm: Fixed

keiramoon (approved):

Assets/Tests/InputSystem/CoreTests_Layouts.cs
```diff
@@ -691,6 +691,154 @@ public void TODO_Layouts_RegisteringMatcherForLayout_OverridesExistingMatchers()
         #endif
     }
 
+    [Preserve]
```
karljj1, line 694: It should be possible to just do _[assembly: Preserve]_ so that no stripping is done to the tests assembly.
Probabaly easier in the long run ;) 
https://docs.unity3d.com/Manual/ManagedCodeStripping.html

Rene-Damm: This one is actually mainly there to make `API_TypesCreatedByReflectionHavePreserveAttribute` happy. Which, however, makes me wonder... why does that one even look at the test assemblies...

Packages/com.unity.inputsystem/Documentation~/Processors.md
```diff
@@ -100,18 +100,9 @@ If you [create a layout from JSON](Layouts.md#layout-from-json), you can specify
...
-|---|---|
-|__Operand Type__|`float`|
-|__Parameters__|`float min`<br>`float max`|
-
-An axis deadzone Processor scales the values of a Control so that any value with an absolute value smaller than `min` is 0, and any value with an absolute value larger than `max` is 1 or -1. Many Controls don't have a precise resting point (that is, they don't always report exactly 0 when the Control is in the center). Using the `min` value on a deadzone Processor avoids unintentional input from such Controls. Also, some Controls don't consistently report their maximum values when moving the axis all the way. Using the `max` value on a deadzone Processor ensures that you always get the maximum value in such cases.
-
 ### Clamp
 
-|__Name__|`Clamp`|
+|__Name__|[Clamp](../api/UnityEngine.InputSystem.Processors.ClampProcessor.html)|
```
karljj1: You removed the quotes `` here but added them in the later tables. Should this one also have quotes or should the others also have them removed? As its a url I think they probably need removing.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -3,6 +3,8 @@
 using System.Collections.Generic;
 using UnityEngine.InputSystem.Utilities;
 
+////TODO: make the FindAction logic available on any IEnumerable<InputAction> and IInputActionCollection via extension methods
```
karljj1, line 6: Should we be moving the TODO code into somewhere where we can capture and track the tasks? Maybe favro/jira cards?

Packages/com.unity.inputsystem/InputSystem/Controls/DiscreteButtonControl.cs
```diff
@@ -81,7 +81,7 @@ public override unsafe float ReadUnprocessedValueFromState(void* statePtr)
 
         public override unsafe void WriteValueIntoState(float value, void* statePtr)
         {
-            throw new NotImplementedException();
+            throw new NotSupportedException();
```
karljj1: Why is this not supported? Can we put a reason into the exception?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
Rene-Damm: From my POV, moving them anywhere other than the code is not an improvement. I found that the only TODOs/FIXMEs/etc. that I consistently see and am reminded of are the ones I constantly stumble over in the code. Whenever I have moved them to separate lists in the past, it was the first step towards having them become obsolete and outdated and having a list full of random no-longer-relevants that I would soon stop looking at :)

Also, other than these being in *my* face, they are there for *everyone* to see on GitHub instead of on a private board inside of Unity (which 2 years from now will for sure get deleted as part of us moving to yet another ticket tracking service).

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -30,6 +30,31 @@
 ////REVIEW: useStateFrom seems like a half-measure; it solves the problem of setting up state blocks but they often also
 ////        require a specific set of processors
 
+////REVIEW: Can we allow aliases to be paths rather than just plain names? This would allow changing the hierarchy around while
+////        keeping backwards-compatibility.
+
+// Q: Why the heck is there this layout system instead of just new'ing everything up in hand-written C# code?
```
karljj1: Can we say heck in the docs? ;)

keiramoon: Technically, it's a .cs file (not docs) so... I didn't see anything :)

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -566,7 +595,19 @@ public class Builder
             /// </summary>
             /// <value>Name of base layout.</value>
             /// <seealso cref="InputControlLayout.baseLayouts"/>
-            public string extendsLayout { get; set; }
+            public string extendsLayout
+            {
+                get => m_ExtendsLayout;
+                set
+                {
+                    if (!string.IsNullOrEmpty(value))
```
karljj1, line 603: What if the user enters a load of whitespace? Is that ok or should we use IsNullOrWhiteSpace?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
Rene-Damm: Why the heck not? :D

Seriously, though, as a non-native speaker, I find it sometimes hard to judge what's offensive/explicit/getting-you-sideways-glances-at-parties and what isn't. I'll go remove it :)

Packages/com.unity.inputsystem/InputSystem/Controls/Processors/EditorWindowSpaceProcessor.cs
```diff
@@ -19,13 +19,21 @@ namespace UnityEngine.InputSystem.Processors
...
+        /// <param name="value">GameView screen space coordinate.</param>
+        /// <param name="control">Ignored.</param>
+        /// <returns>The given coordinate transformed into <c>EditorWindow</c> space.</returns>
+        /// <remarks>
+        /// This method will only succeed if the editor is currently in an <c>EditorWindow</c> callback such
+        /// as <c>OnGUI</c>.
+        /// </remarks>
         public override Vector2 Process(Vector2 value, InputControl control)
         {
-            if (control == null)
```
karljj1, line 26: Do we not mind if it is null anymore?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
```diff
@@ -86,7 +92,7 @@ public static string GenerateWrapperCode(InputActionAsset asset, Options options
             }
 
             // Begin class.
-            writer.WriteLine($"public class @{options.className} : IInputActionCollection, IDisposable");
+            writer.WriteLine($"public partial class @{options.className} : IInputActionCollection, IDisposable");
```
karljj1, line 95: Could we use templates for generated code, like how MonoBehaviours and other Unity assets work?

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
```diff
@@ -370,6 +370,7 @@ private static class Contents
             public static readonly GUIContent copyDeviceDescription = new GUIContent("Copy Device Description");
             public static readonly GUIContent copyLayoutAsJSON = new GUIContent("Copy Layout as JSON");
             public static readonly GUIContent createDeviceFromLayout = new GUIContent("Create Device from Layout");
+            public static readonly GUIContent generateCodeFromLayout = new GUIContent("Generate Precompiled Layout");
```
karljj1, line 373: We should look at localizing these in the future.

Packages/com.unity.inputsystem/InputSystem/Editor/InputLayoutCodeGenerator.cs
```diff
@@ -0,0 +1,335 @@
...
+            if (defines != null)
+            {
+                writer.WriteLine($"#if {defines}");
+                writer.WriteLine();
+            }
+
+            if (@namespace == null)
+                @namespace = baseTypeNamespace;
+
+            writer.WriteLine("using UnityEngine.InputSystem;");
```
karljj1, line 103: Feels like we could do with a template file for this stuff.

Packages/com.unity.inputsystem/InputSystem/Controls/DiscreteButtonControl.cs
Rene-Damm: >Why is this not supported?

It's technically not impossible but a bit unclear how exactly this would operate. The control maps multiple input states to a single input value so when you write an input value back, it's unclear what state that should correspond to.

>Can we put a reason into the exception?

Yup, I'll go add a blurb.

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
```diff
@@ -754,5 +754,14 @@ public static void SwapElements<TValue>(this NativeArray<TValue> array, int inde
             array[index1] = array[index2];
             array[index2] = temp;
         }
+
+        public static T[] ToArray<T>(this Array array)
```
karljj1: Can you not just cast the array?
`public static T[] ToArray<T>(this Array array) => (T[])array;`

Packages/com.unity.inputsystem/InputSystem/Controls/Processors/EditorWindowSpaceProcessor.cs
Rene-Damm: Yup, correct. It used to rely on the control to get to the device on which the position request was issued and then would send the coordinate mapping command to that device. That turned out to be nonsense as native backends then needed to implement the same command for every single type of pointer device they supported. So now we just send the command to the system mouse and ignore the control we get.

karljj1 (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionCodeGenerator.cs
Rene-Damm: Maybe. For now, seems a bit overkill and probably makes for an additional upgrade headache.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
Rene-Damm: Yup, indeed.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
karljj1: Haha. Its not really offensive(I think), just a bit too casual perhaps ;)

Packages/com.unity.inputsystem/InputSystem/Editor/InputLayoutCodeGenerator.cs
Rene-Damm: Maybe. Bit reluctant as it will blow up what is now one single self-contained, very obvious and easy to change file into a something that needs to be modified in multiple places and is far less straightforward to digest and modify.

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
Rene-Damm: Turns out we can in this case. Thought I was using this for some System.Arrays returned from the reflection API but nope, not the case.

Packages/com.unity.inputsystem/Documentation~/Processors.md
Rene-Damm: Think having the `` makes sense. Fixed.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
Rene-Damm: Nuked :)

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
Rene-Damm: I'd classify that under *what* you put in those strings is up to you as long as you put *something* in there :) The check is here mainly to filter out empty strings and make them come out as null. Should probably be `value != ""` instead of `IsNullOrEmpty`.

PR #1157, opened 2020-06-02
FIX: Missing HelpURLs.

Rene-Damm:
Add missing `[HelpURL]` attributes and add tests to make sure public MonoBehaviours in the input system have it.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenButton.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenButton.cs
@@ -11,7 +11,7 @@ namespace UnityEngine.InputSystem.OnScreen
     /// input.
     /// </summary>
     [AddComponentMenu("Input/On-Screen Button")]
-    [HelpURL("https://docs.unity3d.com/Packages/com.unity.inputsystem@latest?subfolder=/manual/OnScreen.html%23on-screen-buttons")]
+    [HelpURL(InputSystem.kDocUrl + "/manual/OnScreen.html#on-screen-buttons")]
     public class OnScreenButton : OnScreenControl, IPointerDownHandler, IPointerUpHandler
     {
         public void OnPointerUp(PointerEventData eventData)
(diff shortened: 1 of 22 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenButton.cs
```diff
@@ -11,6 +11,7 @@ namespace UnityEngine.InputSystem.OnScreen
     /// input.
     /// </summary>
     [AddComponentMenu("Input/On-Screen Button")]
+    [HelpURL("https://docs.unity3d.com/Packages/com.unity.inputsystem@latest?subfolder=/manual/OnScreen.html%23on-screen-buttons")]
```
karljj1: Should we use @latest or can we somehow link it to the actual package version that is in the project? This may take them to a page that is incorrect for the version they are using

karljj1: Maybe do something like this: Define the version in C# somewhere as a const:
```
class Version
{
    public const string kVersion = "1.0";
}

...
[HelpURL("https://docs.unity3d.com/Packages/com.unity.inputsystem@" + Version.kVersion + "?subfolder=/manual/OnScreen.html%23on-screen-buttons")]

```

karljj1: https://gist.github.com/karljj1/0c9b8fa8d737dc69f810f6ec90ea83ab

Rene-Damm: Good point. Will change. Also, occurs to me I should probably move the entire first part of that string all the way up to the actual doc path into a string const. I'll put it in AssemblyInfo.cs so that the stuff including the version is all in one place.

Rene-Damm: Ok, went one step further and also locked this in with tests.

karljj1 (approved):

PR #1159, opened 2020-06-03
Update CI Tests

karljj1:
- Fixed python error when using unity-downloader
- Added support for 2019.3, 2020.1 and 2020.2.
- Fixed failing XR tests on 2019.3+

- Disabled Windows Mouse tests as they wont work in batchmode. We can fix this later


Packages/manifest.json
```diff
@@ -1,13 +1,7 @@
 {
   "dependencies": {
-    "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "3.3.2",
-    "com.unity.cinemachine": "2.3.4",
-    "com.unity.multiplayer-hlapi": "1.0.2",
     "com.unity.package-manager-doctools": "1.1.1-preview.5",
     "com.unity.package-manager-ui": "2.1.2",
-    "com.unity.probuilder": "4.0.5",
```
Rene-Damm, line 9: This one's needed for TouchSamples

Packages/manifest.json
```diff
@@ -1,13 +1,7 @@
 {
   "dependencies": {
-    "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "3.3.2",
-    "com.unity.cinemachine": "2.3.4",
```
Rene-Damm: Not sure about this one... is it used by TouchSamples?

Packages/manifest.json
```diff
@@ -1,13 +1,7 @@
 {
   "dependencies": {
-    "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "3.3.2",
```
Rene-Damm: Don't we need that for UNITY_ANALYTICS?

Packages/manifest.json
karljj1: Ah. It causes compilation issues in 2019.3+
Maybe It just needs upgrading to a new version.

ProjectSettings/EditorSettings.asset
```diff
@@ -20,3 +20,4 @@ EditorSettings:
   m_CollabEditorSettings:
     inProgressEnabled: 1
   m_EnableTextureStreamingInPlayMode: 1
+  m_AssetPipelineMode: 1
```
karljj1, line 23: Suppresses the popup when running outside of batchmode

Rene-Damm (approved):

PR #1160, opened 2020-06-03
FIX: Gravity, LinearAcceleration, and Attitude not initialized on iOS (case 1251382).

Rene-Damm:
Fixes [1251382](https://issuetracker.unity3d.com/product/unity/issues/guid/1251382/) ([internal](https://fogbugz.unity3d.com/f/cases/1251382/)).


todi1856 (approved):
Looks good, but QA needs to verify

Rene-Damm:
> Looks good, but QA needs to verify

Yup, expecting that to happen as part of verification for the closed ticket when the next package comes out.

todi1856:
> > Looks good, but QA needs to verify
> 
> Yup, expecting that to happen as part of verification for the closed ticket when the next package comes out.

Can't QA clone this branch and test? I thought we test first, then release, not the opposite :)

aleksandrasdzikia (approved):
All three now are initialised on iOS, issue seems to be fixed.

karljj1 (approved):

PR #1165, opened 2020-06-09
NEW: Add performance tests.

Rene-Damm:
This PR switches `CorePerformanceTests` to the performance testing framework and adds a couple new tests.

The tests are restricted to the editor for now. The performance testing framework triggered a `NullReferenceException` in players. Probably already fixed in a newer version. We should be able to upgrade once we bump everything to 2019.4 (prepping a PR).


Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -17,39 +22,176 @@ internal class CorePerformanceTests : InputTestFixture
...
+    public void Performance_Update10Gamepads()
     {
         const int kNumGamepads = 10;
 
         var gamepads = new Gamepad[kNumGamepads];
         for (var i = 0; i < kNumGamepads; ++i)
             gamepads[i] = InputSystem.AddDevice<Gamepad>();
 
-        // Perform initial update to get any first-update-only stuff out of the way.
+        // Warm up.
```
karljj1: You can tell Measure to run some warmups e.g

```
Measure.Method(() =>
{
...
}.WarmupCount(100)
```

karljj1 (approved):

Assets/Tests/InputSystem/CorePerformanceTests.cs
Rene-Damm: Not in this version of the package.

karljj1: Are you using `com.unity.test-framework.performance`?

PR #1167, opened 2020-06-10
Introduction of iOS/Android CI

todi1856:
This PR introduces Android/iOS CI to Input System.

Notice, Android/iOS build and run configs are split, the reason is Android/iOS devices are valuable resources, thus if you're not running on it, you shouldn't be allocating it.

There were about 34 failing tests, disabled them and created bug reports for some of them, the parent case is https://fogbugz.unity3d.com/f/cases/1254558/

Disabled Assets/Tests/InputSystem/Plugins/HIDTests.cs  without bug reports on Android since there are no HID devices on Android, not sure if this is correct.

Note: This PR cannot land until Input System bumps the Unity version to 2019.3 or higher, since there are package dependencies which require that version, specifically these one:
  "com.unity.test-framework": "1.1.13",
    "com.unity.test-framework.build": "0.0.1-preview.12",
    "com.unity.test-framework.utp-reporter": "0.1.3-preview.18",


AndroidBuildTests.cmd
```diff
@@ -0,0 +1,17 @@
...
+if "%EDITOR_LOCATION%" == "Empty" (
+ echo Invalid editor location, exiting
+ pause
+ GOTO :eof
+)
+
+echo Editor location is %EDITOR_LOCATION%
+curl -s https://artifactory.internal.unity3d.com/core-automation/tools/utr-standalone/utr.bat --output utr.bat
+call ./utr.bat --suite=playmode --platform=Android --editor-location="%EDITOR_LOCATION%" --testproject=F:/Projects/InputSystem --player-save-path=build/players --artifacts_path=build/logs --scripting-backend=il2cpp --build-only
+pause
```
Rene-Damm: If possible, please use PowerShell. Also will get rid of the dependence on `curl`.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -4493,6 +4496,7 @@ public void Actions_WhenFindingControlSchemeUsingGivenDevice_MostSpecificControl
             Is.EqualTo(ps4GamepadScheme));
         Assert.That(InputControlScheme.FindControlSchemeForDevice(xboxController, new[] { genericGamepadScheme, ps4GamepadScheme, xboxGamepadScheme, mouseScheme }),
             Is.EqualTo(xboxGamepadScheme));
+#endif
```
Rene-Damm, line 4505: The two tests here aren't really platform-specific. Would prefer an `[Ignore]` and case so I can fix this along with the other tests.

todi1856: Will do

todi1856: Done

Rene-Damm: Awesome, thank you :)

Rene-Damm (approved):

PR #1168, opened 2020-06-10
CHANGE: Move Unity min version up to 2019.4 LTS.

Rene-Damm:
Changes:

* Bump package version to 1.1
* Bump CI from 2019.1 to 2019.4 and make 2019.4 our min supported version
* Add `trunk` to CI runs
  - HOWEVER: I've excluded `trunk` runs from being required to be green for merging. Right now, the asset DB crashes with trunk, for example, on `Test Sample Projects`. Overall, I think `trunk` runs should be considered informative and "early warnings" but will likely be too unstable for them to be a hard requirement.
* Replace Yamato branch triggers with PR triggers
  - We no longer run CI on every push but rather only on PRs targeting `develop`
  - I've enabled GitHub branch rules for `develop` that require the respective Yamato configs to pass
* Disable API Validation checks performed by Package Validation Suite due to false positive
  - A PR fixing the false positive in PVS is [in the queue](https://github.cds.internal.unity3d.com/unity/com.unity.package-validation-suite/pull/159)
* Put custom API validation check in place while we're waiting for PVS to get fixed
* Undo breaking API change to `InputTestFixture`


Assets/Tests/InputSystem/APIVerificationTests.cs
```diff
@@ -575,6 +576,187 @@ public void API_MonoBehavioursHaveHelpUrls()
         Assert.That(brokenHelpUrls, Is.Empty);
     }
 
+    private const string kAPIDirectory = "Tools/API";
+
+    ////FIXME: The .api-based checks are temporary and don't account for platform-specific APIs. Nuke these tests as soon
+    ////       as we can switch back to API validation performed by the Package Validation Suite (as soon as Adriano's fix
+    ////       for the access modifier false positive has landed).
```
Rene-Damm, line 583: Talked this over with Adriano. He'll go and fix the false positive and PR the change into Package Validation Suite. When that change is out, we can probably switch back to using API Validation from PVS.

However, with the exclusion mechanism that PVS currently has, we have no ability to exclude individual breaking changes. So could be, that we'll run into another problem that we can't exclude.

I opted to just run with this test for now, as imperfect as it is. At least it gives us back control and so far, the detections it comes up with all seem sensible.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased]
 
+>__The minimum version requirement for the Input System package has been moved up to 2019.4 LTS.__
```
karljj1, line 12: Whats the reason for moving the min version to LTS?
Is it just because we want people to move to LTS?

Rene-Damm: Pretty much. We haven't been able to backport native fixes to our min required version for quite some time now. With 2019.4 out, we can in good conscience move on to a newer version of Unity without forcing users onto 2020.x and at the same time give us a Unity version that we can keep backporting native fixes to and thus fully support. Plus we'll focus on the version that for now will likely have the majority of users. Last but not least, it'll give us a main target that will stay immovable for a good while.

Ah and... XR was already having a quasi-2019.3+ requirement so... time to move on :)

Tools/CodeAnalyzerTestProject/Assets/Tests/CleanConsoleTest.cs
```diff
@@ -18,7 +18,11 @@ public class CleanConsoleTest
         public void TestCleanConsole()
         {
             string logs = Utilities.GetLogs();
-			Assert.That(logs, Is.Empty, string.Format("Found logs in the console on project startup.\n\nLogs: {0}",logs));
+            // This test is to pick up warnings generated by the Microsoft Roslyn analyzers. However, these may
```
jechter, line 21: Which other packages are those? Maybe it would make more sense to make sure no other packages are in the project?

jechter (approved):

Tools/CodeAnalyzerTestProject/Assets/Tests/CleanConsoleTest.cs
Rene-Damm: Unfortunately, those warnings all came from the test framework -- which we can't get rid off..

jechter: Ok, makes sense!

Assets/Tests/InputSystem/APIVerificationTests.cs
Rene-Damm: I realize now that we'll probably have to live with this API test instead of relying on API Validation even after @adrianoc-unity3d's fix has landed. Problem is, we will need to slap the preview label back on our package instead of going directly to 1.1. Adding the label back will bypass API Validation checks from what I know. So seems that even then, we'll still need our own thing :(

PR #1173, opened 2020-06-17
Added a Pose Control and support for injecting it into the XRLayout Builder

StayTalm:
Added detection for 'SubControls' to the XRLayoutBuilder so we can declare a deeper hierarchy, and added hte first control, the PoseControl, and a way to detect and implement it.

Included test that the PoseControl gets generated.


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        public bool isTracked { get; set; }
+        public TrackingState trackingState { get; set; }
+
+        public Vector3 position { get; set; }
+        public Quaternion rotation { get; set; }
+
+        public Vector3 velocity { get; set; }
+        public Vector3 angularVelocity { get; set; }
+    }
```
Rene-Damm, line 111: If possible, would recommend making this a `PoseState` struct for `PoseControl` and locking the memory layout of the control down to this struct the same way `TouchControl` does. The reason is that as is, `PoseControl.ReadUnprocessedValueFromState` is *extremely* costly. ATM every `ReadValue` involves several virtual method calls and walking all over the place. In addition, you have several controls that are in turn made up of subcontrols. Reading a single `Vector3`, for example, will involve a `ReadValue` overhead for the vector itself and then a `ReadValue` overhead for three subcontrols.

Some of the `ReadValue` overhead will eventually disappear but even then, I would strongly recommend locking the memory formats of bigger controls down.

Admittedly, that does lead to inconsistencies. For controls that do this, you can do stuff like putting processors on individual child controls but then, when you read a value through the parent control, the process on the child control won't take effect. Though, with how the `PoseControl.ReadUnprocessedValueFromState` looks like ATM you get the same effect (of ignoring processors on child controls).

In the DOTS version, we'll likely have none of these issues as all the data will probably be mass-converted once but for better or worse, in the current implementation, `ReadValue` is not cheap :(

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRLayoutBuilder.cs
```diff
@@ -157,11 +200,13 @@ private InputControlLayout Build()
                 ? InputSystem.LoadLayout(parentLayout)
                 : null;
 
+            var parentControls = new List<string>();
             var currentUsages = new List<string>();
 
             uint currentOffset = 0;
-            foreach (var feature in descriptor.inputFeatures)
+            for(int i = 0; i < descriptor.inputFeatures.Count; i++)
```
Rene-Damm: `var`

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRLayoutBuilder.cs
```diff
@@ -144,6 +144,49 @@ private static string ConvertPotentialAliasToName(InputControlLayout layout, str
...
+            FeatureType.Axis3D,
+            FeatureType.Rotation,
+            FeatureType.Axis3D,
+            FeatureType.Axis3D
+        };
+
+        // A PoseControl consists of 6 subcontrols with specific names and types
+        private bool IsPoseControl(List<XRFeatureDescriptor> features, int startIndex)
+        {
+            for(int i = 0; i < 6; i++)
```
Rene-Damm: var

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
StayTalm: Swapped to use a full state block.

StayTalm:
@Rene-Damm swapped to state black, ready for re-review.
Added changelog as well.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
```diff
@@ -0,0 +1,101 @@
...
+            m_IsTracked = isTracked;
+            m_TrackingState = trackingState;
+            m_Position = position;
+            m_Rotation = rotation;
+            m_Velocity = velocity;
+            m_AngularVelocity = angularVelocity;
+        }
+
+        [FieldOffset(0), InputControl(name = "isTracked", displayName = "Is Tracked", layout = "Button")]
+        public bool m_IsTracked;
```
Rene-Damm: No `m_` for struct fields. Just `isTracked` and so on.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        [FieldOffset(48), InputControl(name = "angularVelocity", displayName = "Angular Velocity", noisy = true)]
+        public Vector3 m_AngularVelocity;
+
+        public bool isTracked => m_IsTracked;
+        public TrackingState trackingState => m_TrackingState;
+        public Vector3 position => m_Position;
+        public Quaternion rotation => m_Rotation;
+        public Vector3 velocity => m_Velocity;
+        public Vector3 angularVelocity => m_AngularVelocity;
```
Rene-Damm: Okay to here just go with the fields only.

PR #1175, opened 2020-07-07
FIX: TouchControl.tapCount resetting to 0 with ScriptDebugging enable…

vaidas-unity:
Fix InputState.currentTime calculation where s_CurrentTimeOffsetToRealtimeSinceStartup is added to currentTime instead of being subtracted. This exposes a bug that resets tapCount to 0 when "Script Debugging" is enabled because the value of s_CurrentTimeOffsetToRealtimeSinceStartup is greater due to delayed startup while waiting for debugger to attach.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_State.cs
+++ b/Assets/Tests/InputSystem/CoreTests_State.cs
@@ -760,7 +760,7 @@ partial class CoreTests
 
     [Test]
     [Category("State")]
-    public void State_CurrentTimeCalculatedCorrectly()
+    public void State_CurrentTimeTakesOffsetToRealtimeSinceStartupIntoAccount()
     {
         runtime.currentTime += 2;
         runtime.currentTimeOffsetToRealtimeSinceStartup = 1;
--- a/Assets/Tests/InputSystem/CoreTests_State.cs
+++ b/Assets/Tests/InputSystem/CoreTests_State.cs
@@ -758,6 +758,17 @@ partial class CoreTests
         Assert.That(positionMonitorFired);
     }
 
+    [Test]
+    [Category("State")]
+    public void State_CurrentTimeCalculatedCorrectly()
+    {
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
```diff
@@ -30,7 +30,7 @@ public static class InputState
         /// </summary>
         public static uint updateCount => InputUpdate.s_UpdateStepCount;
 
-        public static double currentTime => InputRuntime.s_Instance.currentTime + InputRuntime.s_CurrentTimeOffsetToRealtimeSinceStartup;
+        public static double currentTime => InputRuntime.s_Instance.currentTime - InputRuntime.s_CurrentTimeOffsetToRealtimeSinceStartup;
```
Rene-Damm, line 33: Ouch... that's an embarrassing one... first reaction was "no way" :)

But... needs test.

Assets/Tests/InputSystem/CoreTests_State.cs
```diff
@@ -758,6 +758,17 @@ public void State_StateChangeMonitorsStayIntactWhenOtherDevicesAreRemoved()
         Assert.That(positionMonitorFired);
     }
 
+    [Test]
+    [Category("State")]
+    public void State_CurrentTimeCalculatedCorrectly()
```
Rene-Damm: Plz avoid things like `xxxCorrectly` and `xxxWorks` kind of tests. From these, it's never quite evident what "correct" implies.

For this test, for example, something like `State_CurrentTimeTakesOffsetToRealtimeSinceStartupIntoAccount` would advertise more clearly what the behavior is that the test wants to establish.

vaidas-unity: Fixed the test name, thanks

PR #1177, opened 2020-07-14
Yamato test (not an actual PR)

benoit-alain:
Will delete immediately, I just want to see how the tests look when they pass and if something's wrong with my PR creation process.


benoit-alain:
This shows that Test Sample Projects trunk is very likely unstable.

PR #1181, opened 2020-07-23
Adding a Min & Max value for all XR AxisControl and Vector2Control subcontrols

StayTalm:
Not adding a min and max value for axes causes EvaluateMagnitude to not function which breaks press & hold interactions.  This forces a range of [-1:1], which is true in all current circumstances.  In future verisons of Unity 2021.x, we will be looking to add additional context to controls so it doesn't need to be set in bulk.

Also added test.


Rene-Damm (approved):

Rene-Damm (changes requested):
Ah just realized... please add CHANGELOG entry (and, if applicable, link to bug report to test and changelog entry).

PR #1184, opened 2020-07-30
FIX: Memory corruption in InputEventTrace (case 1262496).

Rene-Damm:
Fixes [1262496](https://issuetracker.unity3d.com/issues/input-system-crash-with-various-stack-traces-when-using-inputactiontrace-dot-subscribetoall) ([internal](https://fogbugz.unity3d.com/f/cases/1262496/)).

Was a stupid thing where I changed the meaning of `capacity` to be in line with how it is used elsewhere in .NET but then didn't correctly update all the code that depended on the previous meaning.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Utilities/NumberHelpers.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Utilities/NumberHelpers.cs
@@ -13,6 +13,15 @@ namespace UnityEngine.InputSystem.Utilities
             return number + alignment - remainder;
         }
 
+        public static long AlignToMultipleOf(this long number, long alignment)
+        {
+            var remainder = number % alignment;
+            if (remainder == 0)
+                return number;
+
+            return number + alignment - remainder;
+        }
+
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Utilities/NumberHelpers.cs
```diff
@@ -13,6 +13,15 @@ public static int AlignToMultipleOf(this int number, int alignment)
             return number + alignment - remainder;
         }
 
+        public static long AlignToMultipleOf(this long number, long alignment)
```
karljj1, line 16: Should we add a test just for this function?

karljj1 (approved):

Packages/com.unity.inputsystem/InputSystem/Utilities/NumberHelpers.cs
Rene-Damm: I'll go add something in the next PRs.

PR #1191, opened 2020-08-18
FIX: RMB menu in action editor not filtering composite types.

Rene-Damm:
Fix action editor showing this

image

instead of this

image


karljj1 (approved):

karljj1:
Looks fine. Is it possible to test the contents of the menu? Maybe we could abstract the menu creation and use faking to see what goes into it?

TJHeuvel:
I dont understand this fix, what do i do if i want to add a 2d vector composite action now? I dont understand any of those new options (button with x modifiers one modifier composite), where will they be documented?

Rene-Damm:
> I dont understand this fix, what do i do if i want to add a 2d vector composite action now?

The change only filters out composites that aren't applicable to the action as per its chosen type. E.g. if the action is set to "Value" with a type of "Vector2", then it filters out the "Axis" composite as that doesn't apply. But the "Vector2" composite will be visible.

PR #1194, opened 2020-08-21
Remoting fixes to send layouts used by devices to remotes

chris-massie:
- Updated the remoting system to send a JSON representation of the layout used by a device over the wire to allow the receiver to have the layout the device depends on. This fixes an issue where XR Controllers and HMDs could not be seen in the Input Debugger window since those layouts may not be present in that Editor.
- Added remote tests for devices that use custom layouts
- Fixed bug where a layout override would cause a `NullReferenceException` when converting to JSON representation due to a null type
- Fixed bug where layouts sent to remotes would not use the name of the layout, since it may be different from the name field in JSON

Tested change by successfully using the Input Debugger window without error to observe device values for the HMD and controllers with the Oculus Quest (using the Oculus Link and an Android build for untethered Quest). Before this change, the console window would log errors about the layout not being found to create the devices.

[XRE-2692 Investigate issues with New Input System debugger and XR Devices](https://jira.unity3d.com/browse/XRE-2692)


Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
```diff
@@ -499,29 +517,48 @@ public struct Data
...
                 };
 
-                var json = JsonUtility.ToJson(data);
-                var bytes = Encoding.UTF8.GetBytes(json);
-
                 return new Message
                 {
                     type = MessageType.NewDevice,
-                    data = bytes
+                    data = SerializeData(data)
```
mattdatwork, line 534: guess we're going to have to get this tested on all platforms? as this could impact mobile et al?

mattdatwork (approved):

Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
```diff
@@ -499,29 +517,48 @@ public struct Data
...
+            public static Message Create(InputRemoting sender, InputDevice device)
             {
                 Debug.Assert(!device.remote, "Device being sent to remotes should be a local device, not a remote one");
 
+                // Try to load the layout. Ignore the layout if it couldn't
+                // be loaded.
+                InputControlLayout layout = null;
+                try
+                {
+                    layout = sender.m_LocalManager.TryLoadControlLayout(new InternedString(device.layout));
```
Rene-Damm: Minor thing... `device.m_Layout` avoids the interning here.

Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
```diff
@@ -499,29 +517,48 @@ public struct Data
...
+                    Debug.Log(string.Format(
+                        "Could not load layout '{0}' associated with device '{1}'; not sending to remote listeners (exception: {2})",
+                        device.layout, device.name, exception));
+                }
+
                 var data = new Data
                 {
                     name = device.name,
                     layout = device.layout,
+                    layoutJson = layout?.ToJson(),
```
Rene-Damm: What's the rationale for sending the layouts with the device?

Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
```diff
@@ -545,12 +582,38 @@ public static void Process(InputRemoting receiver, Message msg)
...
+                            {
+                                layoutAlreadyRegistered = true;
+                                break;
+                            }
+                        }
+                    }
+
+                    if (!layoutAlreadyRegistered)
+                    {
+                        receiver.m_LocalManager.RegisterControlLayout(data.layoutJson, data.layout);
```
Rene-Damm: This has potential to go badly. If a player creates a `Touchscreen` device, then this here causes the editor on the remote end to overwrite the built-in `Touchscreen` layout with the JSON version coming from the player. Thus losing the connection to the `Touchscreen` class.

Remoted layouts should be registered in a separate namespace so as to not clash with layouts local to the editor.

Assets/Tests/InputSystem/CoreTests_Remoting.cs
```diff
@@ -72,6 +72,249 @@ public void Remote_AddingNewControlLayout_WillSendLayoutToRemotes()
...
+            Assert.That(remote.manager.devices, Has.All.With.Property("remote").True);
+
+            var layout = InputSystem.LoadLayout("Gamepad");
+
+            Assert.That(layout.appliedOverrides, Is.EquivalentTo(new[] {new InternedString("MyOverride")}));
+            Assert.That(layout.commonUsages.Count, Is.EqualTo(2));
+            Assert.That(layout.commonUsages, Has.Exactly(1).EqualTo(new InternedString("LeftHand")));
+            Assert.That(layout.commonUsages, Has.Exactly(1).EqualTo(new InternedString("RightHand")));
+        }
+    }
```
Rene-Damm: This test doesn't quite make sense to me. All the layout override-specific parts that are tested are tested on the "local" input system and thus I'd expect them to pass regardless of what the remoting code does or does not do with the override.

Rene-Damm: Overrides are tricky to handle for the remoting code as they are getting merged into layouts and are largely transparent once `LoadLayout` is done. However, ff we restrict remoting of layouts to `RegisterLayoutBuilder` and otherwise require layout information to match between the editor and the player, we shouldn't have to worry about overrides.

Rene-Damm (changes requested):
First off, this PR is amazing.

Couple general comments.

While sending all layouts all the time is the safest route in a way, I found it results in a ton of data being sent over the PlayerConnection, often exceeding the maximum bandwidth it is able to handle.

Overall, I think we can/should restrict this to just generated layouts, i.e. ones coming from layout builders (`InputSystem.RegisterLayoutBuilder`). While the tests use that term, the layouts they involve are not actually dynamically built.

For background, the way the layout stuff is meant to work is that the editor always has the full gamut of layouts available. I.e. all layouts available in the player are expected to be present in the editor as well.

The one big exception to this are layouts built at runtime, i.e. the `InputControlLayout.Builder` stuff. This mainly comprises HIDs and XR devices. Those we should definitely send over the wire to fix the problem you guys are facing with XR layouts missing in the editor.

A side effect of doing this is that the devices in the editor will end up using the "right" layouts. For example, if something is an `Touchscreen` in the player, it is an `Touchscreen` in the player. If every layout is used from a remoted version, then, for example, the  `Touchscreen` in the editor will use a remoted version of the layout instead. In the current code where remoted layouts are registered with the same name as local ones, this will even be destructive to an extent and cause malfunctions on local devices.

Overall, I found that remoting layouts always comes with a lot of problems. IMO by only remoting layouts entering the system via `InputSystem.RegisterLayoutBuilder`, we can avoid many of those problems.

Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
chris-massie: I wanted to ensure that the receiver would have the layout available to be able to add the device. If it wasn't included in the `NewDeviceMsg`, then I would need to create and send a `NewLayoutMsg` first and then send the `NewDeviceMsg`. To me, it seemed better to just combine it into one message.

Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
chris-massie: Looking at `InputManager.RemoveControlLayout`, it has a `namespace` parameter that it uses to prepend some string to the name, however I don't see a similar parameter for `InputManager.RegisterControlLayout`. I can add that on to the register function, and make use of that if that's what you are referring to.

However, there's no guarantee that whatever namespace I construct in the remoting code won't conflict with the name of a layout the user registers if they happen to collide. I could generate a GUID for each connection to use as the namespace to remove the chance, but that would make it not as nice to look at in the Input Debugger window than something else like "Remote{participantId}".

I'll probably do the latter and log an error if there is already a layout with that name registered.

Assets/Tests/InputSystem/CoreTests_Remoting.cs
chris-massie: I should have been using `remote.manager.TryLoadControlLayout` instead of `InputSystem.LoadLayout`. Now with that corrected, some of my tests are now failing since the receiver does not have the base layouts to successfully load the derived layout.

I suppose if I change my solution to send all layouts registered with `RegisterLayoutBuilder` upon connection, and assume that any base layouts registered by other means (e.g. JSON) would have also been registered on the receiver, than I'll have to adjust the way I'm testing anyway.

Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
Rene-Damm: It will send the layout over and over then, though. I think the code should only send dynamically generated layouts and send them as soon as the layout is registered with `RegisterLayoutBuilder`. The majority of layouts would never have to be sent this way.

chris-massie: I'm working on changing the solution to serialize all the `RegisterLayoutBuilder` layouts to JSON and send them all upon connection (right before sending all devices), and then do the same as each new one is registered.

chris-massie:
I believe I've addressed your review comments.  I would appreciate you taking another look.

For remoting changes:
- Changed remoting to only send generated layouts at connection, with the assumption that other layouts should be registered already on the receiver. Removed the JSON representation of the layout from the `AddDevice` message to reduce the amount of data sent.
- Added a namespace parameter to some layout registration functions, which was previously only used within `RemoveControlLayout`, to be used by remoting code
- Update remoting to put received layouts in a namespace so that they don't overwrite layouts present on the receiver
- Implement TODO to remove layouts added by remote when disconnecting
- Add support for sending whether the layout is an override layout
- Updated JSON conversion of layouts to serialize common usages and before render
- Added the layout name within the Input Debugger Window (so that the namespace can be seen on layouts that also have a Display Name)

For testing changes:
- Fixed test setup for remoting by having the `InputControlLayout.s_Layouts` field get set to the `InputManager.m_Layouts` for the receiver. This is achieved by hooking into the subscribers array and setting the globals for the right `InputManager` before the receiver processes the message.

Rene-Damm:
Giving this a closer look, I opted for cutting the layout stuff down to a clear set of rules about how layouts should be handled across platforms and thus in remoting.

1. Any given layout 'X' should always denote the same layout regardless of platform.
2. The same applies to layout overrides (which really are just another form of layouts).
3. The editor is expected to have *all* layouts and overrides relevant to the project.
4. The only exception to this rule are layouts generated on the fly -- which by nature cannot be known to the editor.
5. These are the only layouts we send over the wire.
6. We do not bother removing layouts from the editor which have been transmitted to it from a remote.

With these rules in place, there is no longer a need for namespacing. Any given layout 'X' in the player is expected to universally denote the same layout. HIDs have the potential for causing some chaos there but I don't think it'll be relevant in practice.

No 3 we actually violate with some of our own builtin stuff (I know there's some Win vs Mac stuff we don't register properly). Something to fix.

Rene-Damm:
2019.4 il2cpp tests on Mac still red because of il2cpp bug. Fix seems to have made it out to 2020.1 but not yet to 2019.4.

PR #1197, opened 2020-08-24
Fix: Hold action with multiple bindings (case 1239551)

karljj1:
Fixed Action with multiple bindings becoming unresponsive after a Hold interaction was performed. ([case 1239551](https://issuetracker.unity3d.com/issues/input-system-hold-interaction-makes-an-input-action-unresponsive-when-2-or-more-binding-are-attached-to-the-same-input-action))

When we copy the state over we need to do so before we copy the flags as we don't want the Have Magnitude flag.

Includes Test which fails before the fix and passes afterwards. I commited the test first so you can run the tests against it to verify if you wish.


Assets/Tests/InputSystem/CoreTests_Actions_Interactions.cs
```diff
@@ -240,30 +241,49 @@ public void Actions_CanPerformHoldInteraction()
...
         action.Enable();
 
+        TestHoldInteraction(action, gamepad.buttonSouth);
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_CanPerformHoldInteraction_WithMultipleBindings()
+    {
+        //var gamepad = InputSystem.AddDevice<Gamepad>();
```
Rene-Damm: Plz remove

Assets/Tests/InputSystem/CoreTests_Actions_Interactions.cs
```diff
@@ -240,30 +241,49 @@ public void Actions_CanPerformHoldInteraction()
...
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+
+        var action = new InputAction(binding: "<Keyboard>/space", interactions: "hold(duration=0.4)");
+        action.AddBinding("<Keyboard>/s");
+        action.Enable();
+
+        TestHoldInteraction(action, keyboard.spaceKey);
+    }
+
+    void TestHoldInteraction(InputAction action, ButtonControl button)
```
Rene-Damm: Please no methods like this. Makes it hard to understand the tests. Would actually recommend simply folding `_WithMultipleBindings` into the other test.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1675,10 +1675,10 @@ private void ChangePhaseOfActionInternal(int actionIndex, TriggerState* actionSt
 
             // Update action state.
             var newState = trigger;
-            newState.flags = actionState->flags; // Preserve flags.
-            newState.phase = newPhase;
             if (!newState.haveMagnitude)
                 newState.magnitude = ComputeMagnitude(trigger.bindingIndex, trigger.controlIndex);
+            newState.flags = actionState->flags; // Preserve flags.
```
Rene-Damm, line 1819: Please add comment about the dependency here.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1675,10 +1675,14 @@ private void ChangePhaseOfActionInternal(int actionIndex, TriggerState* actionSt
 
             // Update action state.
             var newState = trigger;
-            newState.flags = actionState->flags; // Preserve flags.
-            newState.phase = newPhase;
+
+            // We need to update the magnitude before we copy the flags across as the copied flags may contain
+            // the `HaveMagnitude` flag which would give us an incorrect magnitude value. (case 1239551)
             if (!newState.haveMagnitude)
```
karljj1: @Rene-Damm This change is causing several tests to fail.
They all have messages like this
_InvalidOperationException: Cannot query value of control '/VirtualMouse/scroll' before 'VirtualMouse' has been added to system!'_

Should I check if they have been added to the system before calling ComputeMagnitude? How can i do this?

midopooler (approved):
LGTM but can't comment on what is causing several tests to fail.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1715,10 +1715,15 @@ private void ChangePhaseOfActionInternal(int actionIndex, TriggerState* actionSt
 
             // Update action state.
             var newState = trigger;
+
+            // We need to update the magnitude before we copy the flags across as the copied flags may contain
+            // the `HaveMagnitude` flag which would give us an incorrect magnitude value. (case 1239551)
```
Rene-Damm: Now I notice this is actually incorrect this way, too. It computes the magnitude which sets the `HaveMagnitude` flag on `newState`. Which may then get nuked by the copying over of flags.

I'll go fix it and merge the PR.

PR #1199, opened 2020-09-02
ScreenKeyboard implementation.

todi1856:
Introducing Screen Keyboard implementation for iOS and Android.

The important key of this PR - screen keyboard implementation is fully inside a package:
* for iOS, there are corresponding iOSScreenKeyboard.mm, iOSScreenKeyboardBridge.mm, iOSScreenKeyboardBridge.h files
* for Android, AndroidScreenKeyboard.java.

This should help with iteration and bug fixes, because we won't depend on core Unity.

The editor and platforms which don't implement screen keyboard, FakeScreenKeyboard will be used instead, while it doesn't do any visual appearance, it ensures that API behaves in the same manner as platform specific screen keyboards.

This is also backed up by tests Assets\Tests\InputSystem\ScreenKeyboardTests.cs

The screen keyboard code is marked with internal keyword, since the API for it is not finalized. UI toolkit and similar packages shouldn't access it directly anyways, but instead abstraction layer should be used, where input system would implement it.
This will give flexibility to do any major refactoring, until we're fully happy with it, only then the API will be made public.

Differences between this implementation and https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard.html
* Screen keyboard can/will be accessible via InputRuntime.s_Instance.screenKeyboard;
* Screen keyboard can be shown via Show method by providing ScreenKeyboardShowParams or using overload without any params, in that case keyboard with default params will be shown.
* There's no alternative for https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-hideInput.html, you decide if input field should be shown or not during call of Show method.
* No https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-isInPlaceEditingAllowed.html
* For https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-visible.html & https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-active.html see state property.
* No https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-canGetSelection.html, https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-canSetSelection.html
* No https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-characterLimit.html, since this property is very limiting, in this implementation the same can be achieved by subscribing to inputFieldTextChanged and changing the text from within, this provide much more powerful behavior, since beside character limitation, you can do all kinds of text manipulations
* No https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard-targetDisplay.html 
* Added events - stateChanged, inputFieldTextChanged, selectionChanged which gives more control over the keyboard
* All events are called on Unity main thread.

Update 2020-10-12 added keyboard emulation in Editor - https://www.youtube.com/watch?v=ISAv3A9NY2s&ab_channel=UnityTactic

QA since keyboard implementation is internal, you need to have an assembly with name Unity.InputSystem.ScreenKeyboardTests to access keyboard directly. Alternatively checkout Input System project and scene ScreenKeyboardTests.


Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
```diff
@@ -0,0 +1,132 @@
...
+        RangeInt m_Selection;
+        class FakeScreenKeyboardDispatcher : MonoBehaviour
+        {
+        }
+
+        private FakeScreenKeyboardDispatcher Dispatcher
+        {
+            get
+            {
+                var go = GameObject.Find(nameof(FakeScreenKeyboard));
```
aurimasc: IMO GameObjects we create for our own purposes should have "Unity" in their names.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,306 @@
...
+    /// </summary>
+    internal enum ScreenKeyboardType
+    {
+        /// <summary>
+        /// The default keyboard layout of the target platform.
+        /// </summary>
+        Default = 0,
+
+        /// <summary>
+        ///
```
aurimasc: Document me :)

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,306 @@
...
+        protected ScreenKeyboardShowParams m_ShowParams;
+
+        private InlinedArray<Action<ScreenKeyboardState>> m_StatusChangedListeners;
+        private InlinedArray<Action<string>> m_InputFieldTextListeners;
+        private InlinedArray<Action<RangeInt>> m_SelectionChangedListeners;
+
+        /// <summary>
+        /// Subscribe to an event which is fired whenever screen keyboard state changes
+        /// </summary>
+        public event Action<ScreenKeyboardState> stateChanged
```
aurimasc: I'm skeptical about event's on a keyboard, given that we only have one instance of it. As a user, if I have multiple parts of my application using keyboard, I would have to either write my own central place for event handling or connect events from multiple places and then manually handle, which part of my code has currently activated the keyboard and ignore events in other parts.
I believe it would be nicer to pass an implementation of an interface when opening keyboard, where callbacks would be called on that implementation until the keyboard is closed.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,306 @@
...
+    ///   - [WSA] When screen keyboard is shown, the main Unity window will continue receiving input events, when clicking on Unity window, screen keyboard will continue to be shown
+    /// </summary>
+    internal abstract class ScreenKeyboard
+    {
+        protected ScreenKeyboardState m_KeyboardState;
+        protected ScreenKeyboardShowParams m_ShowParams;
+
+        private InlinedArray<Action<ScreenKeyboardState>> m_StatusChangedListeners;
+        private InlinedArray<Action<string>> m_InputFieldTextListeners;
+        private InlinedArray<Action<RangeInt>> m_SelectionChangedListeners;
```
aurimasc: Whenever you have a "changed" event, there is a question of whether the even handler receives new value or both old and new. Quite often even handler are interested not only in new value, but also in the old one.

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            get => m_InputFieldText;
+            set
+            {
+                if (m_InputFieldText.Equals(value))
+                    return;
+                m_InputFieldText = value;
+                // Note: Order is important, the selection is reported first
+                OnSelectionChange(new RangeInt(m_InputFieldText.Length, 0));
+                ReportInputFieldChange(value);
```
aurimasc: This is very important question to answer: are event fired only for on-screen action or for programmatic changes as well?
Currently they are fired for programmatic changes too (assign value to property and event is fired).

With later choice you have a before/after problem: is event fired before the change has occurred or after?

Consider the scenario where changed event does modify the value in some way and then sets the property. In current implementation it would lead to cascading event firing with potential of infinite recursion. Not to mention that with another issue of multiple even listeners such cascade would lead to events being fired twice with some handlers receiving the newest change first and older change later (as it is currently coded).

Suggested options for fixing:
1. Make events non-programmatic
2. Single handler only event implementation and new value being passed as ref to be modifiable in an event
3. Same as (2) but with keeping multiple handlers per event: make it illegal for event handlers to modify the properties

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.cs
```diff
@@ -0,0 +1,151 @@
...
+            AndroidScreenKeyboard m_Parent;
+            private int m_MainThreadId;
+            public ScreenKeyboardCallbacks(AndroidScreenKeyboard parent)
+                : base("com.unity.inputsystem.AndroidScreenKeyboard$IScreenKeyboardCallbacks")
+            {
+                m_MainThreadId = Thread.CurrentThread.ManagedThreadId;
+                m_Parent = parent;
+            }
+
+#if UNITY_ANDROID
```
aurimasc: I don't think we need this guard Preserve attribute, given that entire class is only present in Android where we need this and in Editor where it does nothing. Or are there cases where the attribute is unavailable in Editor?

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.java
```diff
@@ -0,0 +1,418 @@
...
+public class AndroidScreenKeyboard extends Dialog implements OnClickListener, TextWatcher, OnDismissListener, OnShowListener
+{
+    interface IScreenKeyboardCallbacks
+    {
+        void OnTextChanged(String text);
+        void OnStateChanged(int state);
+        void OnSelectionChanged(int start, int length);
+    }
+
+    private enum ScreenKeyboardState
```
aurimasc, line 45: In Java a new Object is created for each enum value. For something as simple as int value this looks a bit overkill.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            if (m_KeyboardObject == null)
+                m_KeyboardObject = new AndroidJavaObject("com.unity.inputsystem.AndroidScreenKeyboard");
+            return m_KeyboardObject;
+        }
+
+        protected override void InternalShow()
+        {
+            GetOrCreateKeyboardObject().Call("show",
+                new ScreenKeyboardCallbacks(this),
```
aurimasc: Why do we create a new callback proxy every time we open the keyboard? With only one instance on C# side backed by one instance on Java side it should be enough to have only one proxy.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidSupport.cs
```diff
@@ -151,6 +154,19 @@ public static void Initialize()
...
+            // AndroidSupport.Initialize can be called multiple times when running tests
+            // Don't recreate keyboard
+            NativeInputRuntime.instance.screenKeyboard = AndroidScreenKeyboard.instance;
+            #endif
+        }
+
+        public static void Shutdown()
+        {
+            #if !UNITY_EDITOR
+            NativeInputRuntime.instance.screenKeyboard = null;
```
aurimasc: We should no only null the instance, but also dispose AndroidJavaObject/Proxy stuff.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSScreenKeyboard.mm
```diff
@@ -0,0 +1,592 @@
...
+        {
+            [self hideUIDelayed];
+        }
+    }
+    m_ShowParams = param;
+
+    if (m_Active)
+        [self hide: StateDone];
+
+    m_InitialText = initialTextCStr ? [[NSString alloc] initWithUTF8String: initialTextCStr] : @"";
```
aurimasc: [NSString stringWithUTF8String: initialTextCStr] is shorter

iOSBuildTests.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+curl -s https://artifactory.internal.unity3d.com/core-automation/tools/utr-standalone/utr --output utr
+chmod +x utr
+EDITOR_LOCATION=
```
aurimasc: What is the purpose of this line?

aurimasc (changes requested):
The callback stuff currently is problematic and error prone.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Hmph, I think it makes sense what you're saying. I'll check

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: hmph, I am not sure, I don't recall seeing much events which would provide old value as well. Also not sure, if it's possible to provide an old value in all of these cases

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
todi1856: yes, invoking property, would invoke the event

>With later choice you have a before/after problem: is event fired before the change has occurred or after?
after of course, then if in callback you query text property it would contain changed value.

>Consider the scenario where changed event does modify the value in some way and then sets the property. In current implementation it would lead to cascading event firing with potential of infinite recursion.
the callback wouldn't be called if you call property but provide same value. Obviously if you code your callback in a wrong way, like modifying text by adding some letter, this would cause callback to be infinetly called, but that's a legit error from user, which I think would be expected.

> Make events non-programmatic
sadly this is not possible, for ex., in Android case native onSelectionChanged callback doesn't differentiate between UI change or programmatic change. The same is with setText. It might be possible on IOS I think...

> Single handler only event implementation and new value being passed as ref to be modifiable in an event
not good, if event are coming as dispatched calls (I think that's the case on WSA), where events are called on another thread, and we'll be forced to dispatch it on main thread, so you can't change the value you receive. This also doesn't sound very safe. Unless I am misunderstanding.



Again, I don't think this is such a big problem, since like I said, the callback is called only in case where VALUE CHANGES.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.java
todi1856: I don't think it's such a big problem, having an enum provides you with properly typed values. This doesn't affect performance that much

iOSBuildTests.sh
todi1856: for testing purposes you can temporary set the editor path, thus removing the need to reenter it every time you run the tests

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.cs
todi1856: you're right

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidSupport.cs
todi1856: currently AndroidScreenKeyboard continues to live, since it's assigned to static instance in AndroidScreenKeyboard itself. But ye, should probably fix it.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.cs
todi1856: you're right

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /// <summary>
+    /// Provides screen keyboard callbacks
+    /// </summary>
+    /// Note: We use delegates here instead of events to disallow a potential bug from users
+    ///       If these would be events, the order in which callbacks would invoked would be undefined.
+    ///       Imagine if in one of the inputFieldTextChanged callbacks we change the input field text.
+    ///       The other inputFieldTextChanged would still receive the original changed text which is wrong and could lead to potential problems.
+    /// Note: We use this approach with class callbacks, since it's more future proof, if we add an additional callback in the future, it won't break existing user projects.
```
aurimasc: Is it OK for these comments to be part of doc-comment (three slashes)?

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,298 @@
...
+    internal class ScreenKeyboardCallbacks
```
aurimasc: Is there a reason for this to be a class, not a struct?

iOSBuildTests.sh
aurimasc: I mean this variable is assigned twice, empty value here and then hardcoded value below.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,298 @@
...
+        /// </summary>
+        public ScreenKeyboardState state
+        {
+            get => m_KeyboardState;
+        }
+
+        /// <summary>
+        /// Cleanup any platform specific resources
+        /// </summary>
+        public virtual void Dispose()
```
aurimasc, line 203: Looks like every single implementation is expected to override this in order to at least null the static instance?
If such expectation is correct, then this better be abstract.

aurimasc: Actually, we do we even have per-class static instance and one more global instance?

aurimasc (commented):
Looks much better now. Some more cleanup is left and awaiting answers to questions.

iOSBuildTests.sh
todi1856: hardcoded value should have been commented :(

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
todi1856: Fixed

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Fixed

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: As discussed, you can now pass callbacks to Show method, and to make it less chance for user to make a mistake, I turned callbacks to be delegates instead of events

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
todi1856: Resolved in dicussion

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidSupport.cs
todi1856: Added disposing both for iOS and Android

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: since they're outside summary, I don't think the go to docs , not sure @Rene-Damm

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: Didn't see a reason for it to be a struct

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: no, that shouldn't be possible, only one static instance of specific keyboard should be ever alive

todi1856: I think static instance might not be even needed anymore, i'll check

todi1856: Removed static instances

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
aurimasc: No object created and less GC if it is struct and I don't see any reason for it to be class and not struct.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
aurimasc: Now that you have removed statics from implementation, it makes sense for base class to have do-nothing Dispose(), instead of each implementation overriding it and only Android one doing something :)

aurimasc (commented):
Fix some leftovers from earlier comments and class vs struct issue is to be answered.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: so it's still makes sense to have virtual in base class, otherwise Android Dispose won't get called

todi1856: iOS has also dispose

todi1856: Fixed now

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,298 @@
...
+        {
+            Show(new ScreenKeyboardShowParams());
+        }
+
+        /// <summary>
+        /// Show the screen keyboard with customized options.
+        /// </summary>
+        /// <param name="showParams"></param>
+        /// <param name="callbacks"></param>
+        public void Show(ScreenKeyboardShowParams showParams, ScreenKeyboardCallbacks callbacks = default)
```
aurimasc: You can do "= default" for both parameters, then there is no reason for Show() with no arguments, since both do the same.

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.cs
aurimasc: Reminder

Packages/com.unity.inputsystem/InputSystem/Plugins/Android/AndroidScreenKeyboard.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// Note: We're not creating AndroidJavaObject in constructor, since AndroidScreenKeyboard is called when global static variables are initialized
+        ///       But it seems you cannot create AndroidJavaObject's at this time yet
+        /// </summary>
+        /// <returns></returns>
+        //private AndroidJavaObject GetOrCreateKeyboardObject()
+        //{
+        //    if (m_KeyboardObject == null)
+
+        //    return m_KeyboardObject;
+        //}
```
aurimasc: leftover?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSScreenKeyboardBridge.mm
```diff
@@ -0,0 +1,172 @@
...
+    UnityRect unityRC = {(float)rc.origin.x, (float)rc.origin.y, (float)rc.size.width, (float)rc.size.height};
+    return unityRC;
+}
+
+extern "C" void _iOSScreenKeyboardSetInputFieldText(const char* text)
+{
+    iOSScreenKeyboardBridge* keyboard = [iOSScreenKeyboardBridge getInstance];
+    if (keyboard == NULL)
+        return;
+    NSString* convertedText = text ? [[NSString alloc] initWithUTF8String: text] : @"";
```
aurimasc: Use [NSString stringWithUTF8String] here.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSScreenKeyboard.mm
```diff
@@ -0,0 +1,651 @@
+#include "iOSScreenKeyboardBridge.h"
+#include "DisplayManager.h"
+#include "UnityAppController.h"
+#include "UnityForwardDecls.h"
+#include <string>
+
+#define KEYBOARD_LOG(...) LoggingIndentation s_IndentationScope; if (s_KeyboardLogging) NSLog(@"ScreenKeyboard -%@%@", LoggingIndentation::GetIndentation(), [NSString stringWithFormat: __VA_ARGS__])
```
aurimasc: I don't like this macro for two issues it has:
1. It's function-like, but you can only call it once per scope due to variable definition. (And why variable is like static?)
2. It has potentially harmful if at the end

Suggestion:
- Make variable unique for each call, something like: LoggingIndentation indentationScope ## __LINE__
- Pass __FUNCTION__ to scope constructor and indent all same function logs at the same level
- Wrap if in a do { } while(0)

iOSBuildTests.sh
aurimasc: Remove this entirely. It will expand to empty string anyway and will open the opportunity to set via env variable.

aurimasc (approved):
Minor fixes suggested, but no reason to block this anymore.

todi1856:
> For additional safety, wrap entire if statement in do { } while (0)

That's no good, since destructor would be called too soon on scope. I didn't really understood, how function macro would help, it might cover some scenarios, but not all

aurimasc:
> > For additional safety, wrap entire if statement in do { } while (0)
> 
> That's no good, since destructor would be called too soon on scope. I didn't really understood, how function macro would help, it might cover some scenarios, but not all

No, leave the scope variable before the "do", just place "if" in it.
What I'm suggesting is the the following code should not compile:

KEYBOARD_LOG("xxxx");
else
    do_something_funny();

currently it would. It may not seem like an issue, until you write:

if (something)
    KEYBOARD_LOG("something");
else
    something_else();

todi1856:
> if (something)
> KEYBOARD_LOG("something");

Hmph, I see. Though code like

`
if (something)
    KEYBOARD_LOG("something");
`

will still compile, but will be executed not in the way you would expect.

aurimasc:
> > if (something)
> > KEYBOARD_LOG("something");
> 
> Hmph, I see. Though code like
> 
> `if (something) KEYBOARD_LOG("something");`
> 
> will still compile, but will be executed not in the way you would expect.

Indeed. It seems that the indentation thingy is more risky than it is valueable. It could be solved by making macro expand into object creation and the logging happening inside the constructor of the object.

todi1856:
> > > if (something)
> > > KEYBOARD_LOG("something");
> > 
> > 
> > Hmph, I see. Though code like
> > `if (something) KEYBOARD_LOG("something");`
> > will still compile, but will be executed not in the way you would expect.
> 
> Indeed. It seems that the indentation thingy is more risky than it is valueable. It could be solved by making macro expand into object creation and the logging happening inside the constructor of the object.

Good idea, did that. It looks fine now. The indentation doesn't behave completely in a way I wanted, since

KEYBOARD_LOG("sdsd");
KEYBOARD_LOG("nextline"); <-- this will be indented.

But I think it's fine for now

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
```diff
@@ -0,0 +1,128 @@
+using System;
+using System.Collections;
+
+namespace UnityEngine.InputSystem
+{
+    /// <summary>
+    /// Mock screen keyboard class, which simulates screen keyboard behavior and is validated by the tests in ScreenKeyboardTests.cs
+    /// </summary>
+    class FakeScreenKeyboard : ScreenKeyboard
```
Rene-Damm: Think this one we should move to the test rig instead of having it in the main assembly.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
Rene-Damm: Needs to be `//` or `////` or there'll be complaints from tooling.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -219,6 +221,17 @@ private void OnFocusChanged(bool focus)
 
         public ScreenOrientation screenOrientation => Screen.orientation;
 
+        public ScreenKeyboard screenKeyboard
+        {
+            get => m_ScreenKeyboard;
```
Rene-Damm: Let's lazy-initialize this one such that for apps that don't care, we never create one.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,290 @@
...
+            m_Callbacks = callbacks;
+            if (m_ShowParams.initialText == null)
+                m_ShowParams.initialText = String.Empty;
+            if (m_ShowParams.placeholderText == null)
+                m_ShowParams.placeholderText = String.Empty;
+            InternalShow();
+        }
+
+        /// <summary>
+        /// Hide screen keyboard
```
Rene-Damm: When going public with the API, the xmldocs will have to be fleshed out (and .md docs will have to be added). Stuff like this here is what IMO makes Unity API docs so poor. No examples, no explanation, no cross-references. Not even a fully formed sentence. Just something that repeats what's already obvious from the method name.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3166,6 +3170,19 @@ private static void PerformDefaultPluginInitialization()
...
+            #endif
+        }
+
+        private static void PerformDefaultPluginShutdown()
+        {
+            if (NativeInputRuntime.instance.screenKeyboard != null)
+            {
+                NativeInputRuntime.instance.screenKeyboard.Dispose();
+                NativeInputRuntime.instance.screenKeyboard = null;
+            }
```
Rene-Damm: Why not just do this as part of the runtime disposal?

TautvydasZilys:
Looks like the design doc wasn't shared in this PR: https://docs.google.com/document/d/1iDsV8WRumTDap23TAyBcVRT7RS3oAslEKtVnBgayivA/edit#

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,290 @@
...
+        /// Returns portion of the screen which is covered by the keyboard.
+        /// </summary>
+        public virtual Rect occludingArea => Rect.zero;
+
+        /// <summary>
+        /// Modify selection of an input field text.
+        /// When input field is hidden, this property does nothing and always returns (text.length, 0)
+        /// If you specify a selection out of bounds of text length, the selection will not be set.
+        /// </summary>
+        public abstract RangeInt selection { set; get; }
```
aurimasc, line 274: While trying to fix a bug in our existing TouchScreenKeyboard I encountered one quite fundamental issue - we should probably consider having ability to atomically change multiple properties.
For example, when one writes code like this:
keyboard.inputFieldText = newText;
keyboard.selection = new RangeInt(0, 5);

one would expect the keyboards text and selection to be set. However, with keyboard running on different thread than this code, user can press a button on a keyboard exactly in between these two calls leading to funny outcome.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
```diff
@@ -0,0 +1,290 @@
...
+
+            var selection = new RangeInt(start, length);
+            m_Callbacks.inputFieldSelectionChanged?.Invoke(selection);
+        }
+
+        /// <summary>
+        /// Modifies text in screen keyboard's input field.
+        /// You can use this propery even if input field is hidden
+        /// </summary>
+        public abstract string inputFieldText { set; get; }
```
aurimasc, line 262: Another side note from what I noticed in current TouchScreenKeyboard: what happens to selection when you set text?

Rene-Damm (commented):
F**k... apparently started commenting on some stuff before I went on vacation but didn't submit... the GitHub "Pending" things get me every time...

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
todi1856: This can have negative impact on performance, if application needs to allocate somekind of resources for the keyboard, the first time the keyboard will show up, you might suffer a stutter. It's better if such stutter would be a part of application initialization rather than happen during gameplay.

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
todi1856: If I do this, there's no way for me to use screen keyboard in editor, sure it isn't visible in Editor, but you can use all the API and it will behave the same as runtime.   If I'll do the move, the user will be forced to write 

```
#if UNITY_EDITOR
...
#endif
```
code, which is not very nice.

Note: when https://docs.unity3d.com/ScriptReference/TouchScreenKeyboard.html is used in editor, it's faking the behavior

Rene-Damm: (Gah... hitting every random button today... :D)

Good point. Makes sense to me. In that case, though, I wonder whether we should provide an *actual* emulation, though. I.e. a screen keyboard implemented in software. Otherwise, seems to not really meet the intention of keeping things going in the editor.

Packages/com.unity.inputsystem/InputSystem/Devices/ScreenKeyboard.cs
todi1856: As per our discussion, the most logical thing to do, is set selection at the end of the text

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
todi1856: Sure, I'll try to cook something quickly, make worth renaming this to EmulatedScreenKeyboard or EditorScreenKeyboard

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
todi1856: Runtime doesn't have a Dispose function ?

Packages/com.unity.inputsystem/InputSystem/Devices/FakeScreenKeyboard.cs
todi1856: Here's my quick draft - https://www.youtube.com/watch?v=ISAv3A9NY2 

As far as I recall though, in one the meetings, someone suggested to make keyboard with UI elements (when it's ready), but for now I think this is better than nothing

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSScreenKeyboard.mm
```diff
@@ -0,0 +1,657 @@
...
+#if PLATFORM_IOS
+- (void)positionInput:(CGRect)kbRect x:(float)x y:(float)y
+{
+    float safeAreaInsetLeft = 0;
+    float safeAreaInsetRight = 0;
+
+#if UNITY_HAS_IOSSDK_11_0
+    if (@available(iOS 11.0, *))
+    {
+        safeAreaInsetLeft = [UnityGetGLView() safeAreaInsets].left;
```
jimon, line 508: Correct me if I'm wrong, but `UnityGetGLView` seems to be an internal unity api. So someone like me can land a change into trunk changing this api's, which at best will results in this code not compiling, at worst working incorrectly. To make a backend for native keyboard robust and exist outside of trunk, IMHO we should make it not use any of internal api's. And we should have some way to test that it works on CI.

PR #1202, opened 2020-09-10
Fix: Clear the dirty flag when building 

karljj1:
Fixes (case-1254502)
The temporary change that we make to PlayerSettings is flushed and causes versioning software to check out the modified file, which we then revert later on. This change clears the dirty flag if we are the only one who modified the asset.

I do have a test but I could not get it to work as it causes an error when running BuildPlayer inside of the player loop which all the Input tests are run from.
I will ask SE QA to verify the fix.


Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsBuildProvider.cs
```diff
@@ -18,22 +18,46 @@ public void OnPreprocessBuild(BuildReport report)
...
             // Revert back to original state
             PlayerSettings.SetPreloadedAssets(m_OriginalPreloadedAssets);
+
+            if (!wasDirty)
+                ClearPlayerSettingsDirtyFlag();
+        }
+
+        static bool IsPlayerSettingsDirty()
+        {
+            var settings = Resources.FindObjectsOfTypeAll<PlayerSettings>();
```
Rene-Damm, line 60: IIRC we have access to internals. Could probably use `PlayerSettings.GetSerializedObject` here.

There's also a `[StaticAccessor("GetPlayerSettings")]` directly on the `PlayerSettings` class. Wonder what that does...

Rene-Damm (commented):
Wondering if we can fix this problem in the same go: https://unity.slack.com/archives/C09Q7LYP9/p1595614625238500

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsBuildProvider.cs
Rene-Damm: Turns out we don't have access after all.

Rene-Damm:
> Wondering if we can fix this problem in the same go: https://unity.slack.com/archives/C09Q7LYP9/p1595614625238500

Added the change to also address this problem. [Ticket](https://fogbugz.unity3d.com/f/cases/1265707/) for reference.

Rapsssito:
@Rene-Damm, it looks like the issue still occurs. Take a look at the last comments in https://forum.unity.com/threads/project-settings-keeps-appending-fileid-0.954147/.

PR #1203, opened 2020-09-11
Docs: Missing and unreferenced images

karljj1:
Bringing over a test I added to the Localization package.
This test checks for unreferenced and missing image files.

I then went through and removed the files that were unreferenced. I managed to use 1 and the rest seem to be using the old editor style.


Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -2593,6 +2594,96 @@ public void Editor_CanRestartEditorThroughReflection()
...
 
+    [Test]
+    public void DocsContainNoMissingOrUnreferencedImages()
+    {
+        const string docsPath = "Packages/com.unity.inputsystem/Documentation~/";
+        const string imagesPath = "Packages/com.unity.inputsystem/Documentation~/images/";
+        var regex = new Regex("\\(.*images\\/(?<filename>[^\\)]*)", RegexOptions.IgnoreCase);
+
+        // Add files here if you want to ignore them being unreferenced.
+        string[] unreferencedIgnoreList = new [] { "InputArchitectureLowLevel.sdxml" };
```
karljj1: I have no idea what this file is but assume its important ;)

Rene-Damm: It's the source file for the architecture diagram. In a rather obscure format, though (SimpleDiagrams).

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -2593,6 +2594,96 @@ public void Editor_CanRestartEditorThroughReflection()
         EditorHelpers.RestartEditorAndRecompileScripts(dryRun: true);
     }
 
+    [Test]
+    public void DocsContainNoMissingOrUnreferencedImages()
```
Rene-Damm: Just to keep them together, would prefer to have them in the `APIVerificationTests` suite alongside `API_DocumentationManualDoesNotHaveMissingInternalLinks` test. Neither is quite about API verification but at least they'll be in the same place.

```
    [Test]
    [Category("API")]
    public void API_DocumentationManualDoesNotContainMissingOrUnreferencedImages()
```

PR #1205, opened 2020-09-14
Update docs for tvOS

todi1856:
tvOS wasn't mentioned in Supported Input Devices, even though we support devices there.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/SupportedDevices.md
+++ b/Packages/com.unity.inputsystem/Documentation~/SupportedDevices.md
@@ -6,28 +6,30 @@ This page lists Input Device types and products that the Input System package su
 
 Support for the following Devices doesn't require specialized support of particular products.
 
-|Device|Windows|Mac|Linux|UWP|Android|iOS|Xbox(3)|PS4(3)|Switch(3)|WebGL|
-|------|-------|---|-----|---|-------|---|----|---|------|-----|
-|[Mouse](Mouse.md)|Yes|Yes|Yes|Yes|Yes|No|Yes|Yes|No|Yes|
-|[Keyboard](Keyboard.md)|Yes|Yes|Yes|Yes|Yes|No|Yes|Yes|No|Yes|
-|[Pen](Pen.md)|Yes|No (1)|No|Yes|Yes|Yes|No|No|No|No|
-|[Touchscreen](Touch.md)|Yes|No|No|Yes|Yes|Yes|No|No|No|No|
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/Documentation~/SupportedDevices.md
```diff
@@ -6,28 +6,30 @@ This page lists Input Device types and products that the Input System package su
...
-|Xbox One|Yes (1)|Yes (3)|Yes (1)|Yes|Yes (1)|Yes (6)|Yes|No|No|Sometimes (2)|
-|PS4|Yes (5)|Yes (5)|Yes (5)|Yes (5)|Yes (5)|Yes (5, 6)|No|Yes|No|Sometimes (2)|
-|Switch|Yes|Yes|Yes|Yes|No|No|No|No|Yes|Sometimes (2)|
+|Device|Windows|Mac|Linux|UWP|Android|iOS(6)|tvOS(6)|Xbox(7)|PS4(7)|Switch(7)|WebGL|
+|------|-------|---|-----|---|-------|---|----|----|---|------|-----|
+|Xbox 360 (4)|Yes|Yes (3)|Yes|Yes|No|No|No|Yes|No|No|Sometimes (2)|
+|Xbox One|Yes (1)|Yes (3)|Yes (1)|Yes|Yes (1)|Yes (6)|Yes (6)|Yes|No|No|Sometimes (2)|
+|PS4|Yes (5)|Yes (5)|Yes (5)|Yes (5)|Yes (5)|Yes (5, 6)|Yes (5, 6)|No|Yes|No|Sometimes (2)|
+|Switch|Yes|Yes|Yes|Yes|No|No|No|No|No|Yes|Sometimes (2)|
+|SteelSeries|No|No|No|No|No|Yes|Yes|No|No|No|No|
```
Rene-Damm: Aren't there more MFi gamepads than the SteelSeries stuff? Wondering if the last row should be something like "MFi (such as SteelSeries)".

Rene-Damm (approved):

Packages/com.unity.inputsystem/Documentation~/SupportedDevices.md
Rene-Damm: (Also wouldn't be surprised if on Linux they are actually supported due to the use of SDL)

todi1856: renamed to your suggestion, as for linux, a quick google shows that you need perform additional steps in OS to do that... so probably it works, but not out of the box, ,leaving No for now, until we can properly check this

PR #1207, opened 2020-09-15
Fix: multiple actions fix (case 1253034)

karljj1:
Fixed https://fogbugz.unity3d.com/f/cases/1253034/

When creating a new `TriggerState` we need to also copy over the `mapIndex`.

Includes a test to reproduce the original error.


Assets/Tests/InputSystem/CoreTests_Actions_Interactions.cs
```diff
@@ -91,6 +91,33 @@ public void Actions_StartedAndCanceledAreEnforcedImplicitly()
         }
     }
 
+    [Test]
```
Rene-Damm, line 94: Having link to ticket here for context probably useful.

Rene-Damm (approved):

Assets/Tests/InputSystem/CoreTests_Actions_Interactions.cs
```diff
@@ -91,6 +91,33 @@ public void Actions_StartedAndCanceledAreEnforcedImplicitly()
...
+    [Category("Actions")]
+    public void Action_WithMultipleInteractions_DoesNotThrowWhenUsingMultipleMaps()
+    {
+        var gamepad = InputSystem.AddDevice<Gamepad>();
+        var asset = ScriptableObject.CreateInstance<InputActionAsset>();
+
+        InputSystem.RegisterInteraction<PressInteraction>();
+        InputSystem.RegisterInteraction<HoldInteraction>();
```
Rene-Damm, line 102: These are always registered.

PR #1212, opened 2020-10-10
Fix potential exception when the reference in InputActionProperty is null

chris-massie:
This PR updates `InputActionProperty` to handle when the reference is `null`.

Tested with a behavior in the XR Interaction Toolkit that serialized an `InputActionProperty` and used the inspector to delete the reference while keeping **Use Reference**=`true`, and accessing the `InputActionProperty.action` property no longer threw.


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionProperty.cs
```diff
@@ -33,12 +33,12 @@ public struct InputActionProperty : IEquatable<InputActionProperty>, IEquatable<
...
         /// <value>The action object contained in the property.</value>
         /// <remarks>
-        /// This property will return <c>null</c> if the property using a <see cref="reference"/> and
+        /// This property will return <c>null</c> if the property is using a <see cref="reference"/> and
         /// the referenced action cannot be found. Also, it will be <c>null</c> if the property
         /// has been manually initialized with a <c>null</c> <see cref="InputAction"/> using
         /// <see cref="InputActionProperty(InputAction)"/>.
         /// </remarks>
-        public InputAction action => m_UseReference ? m_Reference.action : m_Action;
+        public InputAction action => m_UseReference ? m_Reference != null ? m_Reference.action : null : m_Action;
```
Rene-Damm, line 41: `m_UseReference ? m_Reference?.action : m_Action`

Rene-Damm (commented):
Needs test.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionProperty.cs
chris-massie: I would have used `?.` but that wouldn't use the overloaded `operator==` of `UnityEngine.Object` which also checks if the object is valid.

chris-massie:
I added tests including one that would fail if the property was changed to `public InputAction action => m_UseReference ? m_Reference?.action : m_Action;`

I also fixed the `NullReferenceException` in `GetHashCode` for a `null` `InputAction`.

PR #1214, opened 2020-10-12
FIX: Action map and asset name equal leading to compile error (case 1212052).

Rene-Damm:
Fixes [1212052](https://issuetracker.unity3d.com/issues/input-system-user-can-name-inputaction-asset-and-action-map-the-same-creating-compilation-errors-on-generation) ([internal](https://fogbugz.unity3d.com/f/cases/1212052/)).

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
@@ -158,48 +158,61 @@ namespace UnityEngine.InputSystem.Editor
             // Generate wrapper code, if enabled.
             if (m_GenerateWrapperCode)
             {
-                var wrapperFilePath = m_WrapperCodePath;
-                if (string.IsNullOrEmpty(wrapperFilePath))
+                // When using code generation, it is an error for any action map to be named the same as the asset itself.
+                // https://fogbugz.unity3d.com/f/cases/1212052/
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -158,48 +158,61 @@ public override void OnImportAsset(AssetImportContext ctx)
...
+                        className = m_WrapperClassName,
+                    };
+
+                    if (InputActionCodeGenerator.GenerateWrapperCode(wrapperFilePath, asset, options))
+                    {
+                        // When we generate the wrapper code cs file during asset import, we cannot call ImportAsset on that directly because
+                        // script assets have to be imported before all other assets, and are not allowed to be added to the import queue during
+                        // asset import. So instead we register a callback to trigger a delayed asset refresh which should then pick up the
+                        // changed/added script, and trigger a new import.
+                        EditorApplication.delayCall += AssetDatabase.Refresh;
```
karljj1, line 214: Should we not just call `AssetDatabase.ImportAsset` instead of Refresh?

karljj1 (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
Rene-Damm: I think that should be fine. I'll give it a go. Guess there's something to Refresh being more robust in this case (by the time we run the delayed call, the path we have need no longer be valid) but I'm not sure that matters in actual practice.

PR #1216, opened 2020-10-12
FIX: FileNotFoundException (case 1273691).

Rene-Damm:
Fixes internal case [1273691](https://fogbugz.unity3d.com/f/cases/1273691/).

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputEditorUserSettings.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputEditorUserSettings.cs
@@ -105,12 +105,16 @@ namespace UnityEngine.InputSystem.Editor
 
         internal static void Load()
         {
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputEditorUserSettings.cs
```diff
@@ -105,12 +105,16 @@ private static void OnChange()
...
+            if (!File.Exists(kSavePath))
+                return;
+
             try
             {
                 var json = File.ReadAllText(kSavePath);
                 s_Settings = JsonUtility.FromJson<SerializedState>(json);
             }
-            catch (Exception)
+            catch
```
scottbilas, line 117: I don't think this catch block is necessary any more

scottbilas (approved):
Code looks fine, though I'm surprised not to see a test

PR #1224, opened 2020-10-16
FIX: OnScreenButton leaving stuck keys when disabled.

Rene-Damm:
Also contains a minor change to tag virtual devices created by OnScreenControls with "OnScreen" and to show device usage tags in the Input Debugger's device list.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
@@ -657,18 +657,22 @@ namespace UnityEngine.InputSystem.Editor
                 }
             }
 
-            private void AddDevices(TreeViewItem parent, IEnumerable<InputDevice> devices, ref int id, int participantId = InputDevice.kLocalParticipantId)
+            private static void AddDevices(TreeViewItem parent, IEnumerable<InputDevice> devices, ref int id, int participantId = InputDevice.kLocalParticipantId)
             {
                 foreach (var device in devices)
                 {
                     if (device.m_ParticipantId != participantId)
                         continue;
 
+                    var displayName = device.name;
+                    if (device.usages.Count > 0)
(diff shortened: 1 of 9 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
```diff
@@ -657,18 +657,22 @@ private void AddUser(TreeViewItem parent, InputUser user, ref int id)
...
+                    var displayName = device.name;
+                    if (device.usages.Count > 0)
+                        displayName += " (" + string.Join(",", device.usages) + ")";
+
                     var item = new DeviceItem
                     {
                         id = id++,
                         depth = parent.depth + 1,
-                        displayName = device.name,
+                        displayName = displayName,
```
karljj1, line 678: Is this correct? It looks like it is assigning the same value to itself?

karljj1: oh wait they just have the same name.

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenButton.cs
```diff
@@ -39,6 +51,8 @@ public void OnPointerDown(PointerEventData eventData)
         [SerializeField]
         private string m_ControlPath;
 
+        [NonSerialized] bool m_Pressed;
```
karljj1: I dont think you need NonSerialized if the field is already private. Its the default behaviour.

karljj1 (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenButton.cs
Rene-Damm: Ended up reworking the change which removed the field altogether.

PR #1228, opened 2020-10-16
fix for m_LayoutRegistrationVersion not being updated when a layout is removed.

mattdatwork:
fix for m_LayoutRegistrationVersion not being updated when a layout is removed.

we noticed adding / removing a controller layout did not remove the controller layout from the input debugger. debugging through we noticed that m_LayoutRegistrationVersion was only being updated when Layouts were added, not removed. this PR adds that into the correct spot for removal.

Had a look at how to unit test this but everything is internal :(


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -784,6 +784,8 @@ public void RemoveControlLayout(string name)
             // Let listeners know.
             for (var i = 0; i < m_LayoutChangeListeners.length; ++i)
                 m_LayoutChangeListeners[i](name, InputControlLayoutChange.Removed);
+
+            ++m_LayoutRegistrationVersion;
```
Rene-Damm: Move up to before invocation of listeners.

Rene-Damm (changes requested):

Rene-Damm:
> Had a look at how to unit test this but everything is internal :(

In `CoreTests_Editor` 

```
    [Test]
    [Category("Editor")]
    public void Editor_AddingAndRemovingLayoutRefreshesLayoutCache()
    {
        const string json = @"
            {
                ""name"" : ""TestLayout""
            }
        ";
        
        InputSystem.RegisterLayout(json);

        Assert.That(EditorInputControlLayoutCache.allLayouts, Has.Exactly(1).With.Property("name").EqualTo(new InternedString("TestLayout")));
        
        InputSystem.RemoveLayout("TestLayout");

        Assert.That(EditorInputControlLayoutCache.allLayouts, Has.None.With.Property("name").EqualTo(new InternedString("TestLayout")));
    }
```

PR #1231, opened 2020-10-21
Generate precompiled code friendly to IL2CPP (case 1283676)

joshpeterson:
The generated code from IL2CPP can cause problems for some C++
compilers, especially when script debugging is enabled. This happens
with large individual methods. So this change modifies the way that
input control initialization C# code is generated. Now each input
control object will generate to an individual method.

This change also uses that new generated code for keyboard, mouse, and
touchscreen. With these changes, the Switch build with script debugging
enabled takes a few minutes instead of a few hours.


joshpeterson:
I'm unclear why some of these tests failed. I feel like none of these changes should differ for platform or Unity version, but the test results seem to indicate that they do. @Rene-Damm: Do you have any insight?

PR #1232, opened 2020-10-23
FIX: IndexOutOfRangeException when active device is disconnected, and another activated

sandereskens:
I was having issues that when disconnecting the active device, and another device is activated, an IndexOutOfRangeException would be spammed.

It seems that every new event, there would be another attempt at ArrayHelpers.EraseSliceWithCapacity on the s_LostDevices as the lostDeviceCount and lostDeviceStartIndex are not reset. Inspiration was taken from the UnpairDevices-method, where they are reset. After resetting them in UpdateControlSchemeMatch, the exception was gone. I also added it to the RemoveUser method for completeness sake.

These are the first few days I'm exploring the system, so by no means do I claim this is a valid fix.


Rene-Damm:
Hi @sandereskens, good catch. Indeed a bug. Rolled the fix into #1274. Thank you for the PR.

PR #1233, opened 2020-10-28
fixes sizeof calculation in querycanruninbackground IOCTL command.

mattdatwork:
fixes sizeof calculation in querycanruninbackground IOCTL command.


jechter (approved):

Rene-Damm (changes requested):
Test and CHANGELOG or it didn't happen :)

mattdatwork:
> Test and CHANGELOG or it didn't happen :)

not sure how we can add a test for this in any sort of generic way. I can add a test that checks to see if the QCRIB is the right size, but thats limited utility?

Rene-Damm:
Added test and CHANGELOG.

PR #1236, opened 2020-11-02
FIX: Scripting API generation errors.

Rene-Damm:
- Fixes malformed xmldoc comments.
- Adds a test to go red on malformed xmldoc comments.
- Removes code from Samples from docs.
- Updates doctools to latest.


Packages/com.unity.package-manager-doctools/.docgen/docfx_packages/_exported_templates/packages/ManagedReference.html.primary.js
```diff
@@ -22,6 +22,32 @@ exports.transform = function (model) {
   }
   model._disableToc = model._disableToc || !model._tocPath || (model._navPath === model._tocPath);
 
+  //Added to sort enums by value and add value itself to model
```
karljj1, line 25: Should this not be part of the doc tools?

Packages/com.unity.package-manager-doctools/.docgen/docfx_packages/_exported_templates/packages/layout/_master.tmpl
```diff
@@ -36,7 +36,8 @@
           <div class="col-md-12">
           {{/_disableAffix}}
           {{^_disableAffix}}
-          <div class="col-md-10">
+          <div class="col-md-10" id="main_content">
```
karljj1: Did you intend to change the package manager doc tools?

karljj1:
Would it not be better to have the doctools as a git submodule?

Rene-Damm:
> Would it not be better to have the doctools as a git submodule?

Unfortunately, that will break running the tests for anyone outside of Unity (plus you'll have to be on VPN then for checkouts). While we don't have a better solution for that, I opted for just updating our embedded version.

Also, there is actually a modification to the doctools in the changes. There's no support in the current doctools version for getting the build report output through the API. I modified it so it passes that along.

karljj1 (approved):

Assets/Tests/InputSystem/CoreTests_Analytics.cs
```diff
@@ -111,6 +111,9 @@ public void Analytics_SendsStartupEventOnlyOnFirstUpdate()
     #if UNITY_EDITOR
     [Test]
     [Category("Analytics")]
+    #if UNITY_2021_1_OR_NEWER
+    [Ignore("TODO fix me")]
```
Rene-Damm: @jimon Always detail what.

jimon:
Because this branch code a bit overwhelming with git-lfs and rebasing, I've decided to do a second iteration in a new PR #1288

PR #1240, opened 2020-11-12
Bug fixes

todi1856:
* Case 1254567, fix Remote_EventsAreSentToRemotes on Android 32 bit. There was an issue with this code - dataEndPtr.ToInt64() which produced -1230560996 when pointer value is 3064406300, changed code to do pointer comparison instead
* Properly fix ShouldEnableRemoting, UNITY_INCLUDE_TESTS define tells if we're running playmode tests on player. Application.productName  is not good in this case, since at least on Android it doesn't contain UnityTestFramework when running playmode tests


Packages/com.unity.inputsystem/InputSystem/Devices/Remote/InputRemoting.cs
```diff
@@ -624,8 +624,9 @@ public static unsafe void Process(InputRemoting receiver, Message msg)
                     var eventCount = 0;
                     var eventPtr = new InputEventPtr((InputEvent*)dataPtr);
                     var senderIndex = receiver.FindOrCreateSenderRecord(msg.participantId);
-
-                    while ((Int64)eventPtr.data < dataEndPtr.ToInt64())
+                    // Don't use IntPtr.ToInt64() function, on 32 bit systems, the pointer is first converted to Int32 and then casted to Int64
+                    // Thus for big pointer value, you might get a negative value even though the pointer value will be less than Int64.MaxValue
+                    while ((void*)eventPtr.data < dataEndPtr.ToPointer())
```
Rene-Damm, line 629: Ouch. Good find.

Rene-Damm (approved):

PR #1239, opened 2020-11-11
FIX: Improve performance of InputUser.OnEvent and RebindingOperation.OnEvent.

Rene-Damm:
This PR focuses on a number of performance hotspots.

- Performance of `InputUser.OnEvent` has been increased ~9x leading to less overhead when automatic control switching is on.
- Performance of `RebindingOperation.OnEvent` has been increased ~2.5x.
- Performance of `FastMouse.OnStateChange` has been increased ~3x leading to less processing overhead for mouse events.
- Performance of `ReadValue` operations has been tweaked to lead to some small gains across the board.
- Performance of `InputDeviceBuilder` and `AddDevice` has been increased somewhat.

Aside from some reshuffling, the biggest change is the addition of a new `InputEventPtr.EnumerateControls()` extension method which employs a much more efficient algorithm for determining control changes compared to the previous, mostly linear approach based on traversing `InputDevice.allControls`.

Some detailed before/after numbers (all in 2019.4 editor with debug code):

`Performance_ListenForUnpairedDeviceActivity`
- Before: 196 microseconds
- After: 22.7 microseconds

`Performance_Rebinding_OneSuccessfulCycle`
- Before: 226.80 microseconds
- After: 84.60 microseconds

`Performance_AutoSwitchingOfControlSchemesInPlayerInput_UnrelatedDeviceIsFeedingInput`
- Before: 268.90 microseconds
- After: 22.40 microseconds

`Performance_CreateDevice("Keyboard")` (creates a *non-precompiled* version of `Keyboard`)
- Before: 8441 microseconds
- After: 4143.30 microseconds

More work to be done but it's a step forward.

PR #1242, opened 2020-11-16
Update UTP reporter, fixes json import issue on 2020.2b

jimon:


Rene-Damm:
Incorporated this change into #1239 as part of several changes to get CI back to normal.

PR #1243, opened 2020-11-16
Fixing player settings for 2020.2+

jimon:


jimon:
Will make one combined PR

PR #1244, opened 2020-11-17
Fixing line endings consistency in the whole repository and making sure everything is formatted

jimon:
Unity by convention follows LF-everywhere approach, and our tooling, for example the formatter, is inclined to reformat into LF if line endings are not consistent.

This PR attempts to achieve:
- Asks git to store everything in LF.
- Asks git to checkout VS/Windows-specific files as CRLF.
- Provides a .repoconfig for our formatting tool to ignore some folders.
- Renormalizes all stored files, we had tons of files actually stored as CRLF inside the repository.
- Formats all stored files, a few of them had inconsistent tabulation.

Result of this change:
- When running `git add --renormalize .` no files are changed (tested on Windows and Mac).
- When running `perl %pathtotheinstall%/format.pl all` no files are changed (tested on Windows and Mac).


Rene-Damm (approved):
I should have ideally merged this one first. Anyway, I'll go sync it to latest develop and merge.

PR #1248, opened 2020-11-26
Disable upm-ci API validator until we have a proper fix for it on <=2020.1

jimon:


jimon:
the problem was solved internally, so this fix is no longer needed

PR #1250, opened 2020-11-27
Fixing touch updateStepCount invalid state after being disabled/enabled.

jimon:
Potential fix for https://fogbugz.unity3d.com/f/cases/1286865/
Needs to be validated that it is correct.

Analysis:
During debug I found that touch `updateStepCount` was a lot higher than `currentUpdateStepCount`, hence the following check was failing:
image
From that I concluded that `s_PlayerState.updateStepCount` must have been reset somehow, and yes indeed during scene reload it is reset.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
@@ -797,6 +797,41 @@ internal class EnhancedTouchTests : InputTestFixture
             new Tuple<string, Finger>("Up", Touch.fingers[1])
         }));
     }
+    
+    [Test]
+    [Category("EnhancedTouch")]
+    public void EnhancedTouch_TouchWorksAfterDisableEnable()
+    {
+        BeginTouch(1, new Vector2(0.123f, 0.234f), queueEventOnly: true);
+        InputSystem.Update();
+        Assert.That(Touch.activeTouches.Count, Is.EqualTo(1));
+        Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Began));
+
+        MoveTouch(1, new Vector2(0.234f, 0.345f), queueEventOnly: true);
+        InputSystem.Update();
+        Assert.That(Touch.activeTouches.Count, Is.EqualTo(1));
+        Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Moved));
+
+        InputSystem.Update();
+        Assert.That(Touch.activeTouches.Count, Is.EqualTo(1));
+        Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Stationary));
+        
+        EnhancedTouchSupport.Disable();
+        EnhancedTouchSupport.Enable();
+
+        InputSystem.Update();
+        Assert.That(Touch.activeTouches.Count, Is.EqualTo(1));
+        Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Stationary));
+
+        MoveTouch(1, new Vector2(0.123f, 0.234f), queueEventOnly: true);
+        InputSystem.Update();
+        Assert.That(Touch.activeTouches.Count, Is.EqualTo(1));
+        Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Moved));
+
+        InputSystem.Update();
+        Assert.That(Touch.activeTouches.Count, Is.EqualTo(1));
+        Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Stationary));
+    }
 
     [Test]
     [Category("EnhancedTouch")]
--- a/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
@@ -798,9 +798,10 @@ internal class EnhancedTouchTests : InputTestFixture
         }));
     }
 
+    // https://fogbugz.unity3d.com/f/cases/1286865/
     [Test]
     [Category("EnhancedTouch")]
-    public void EnhancedTouch_TouchWorksAfterDisableEnable()
+    public void EnhancedTouch_CanBeDisabledAndReenabled()
     {
         BeginTouch(1, new Vector2(0.123f, 0.234f), queueEventOnly: true);
         InputSystem.Update();
--- a/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
@@ -797,7 +797,7 @@ internal class EnhancedTouchTests : InputTestFixture
             new Tuple<string, Finger>("Up", Touch.fingers[1])
         }));
     }
-    
+
     [Test]
     [Category("EnhancedTouch")]
     public void EnhancedTouch_TouchWorksAfterDisableEnable()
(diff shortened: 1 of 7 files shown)


Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
```diff
@@ -71,27 +71,95 @@ public enum InputUpdateType
...
+        #else
+        public static uint s_UpdateStepCount => s_PlayerUpdateStepCount.value;
+        #endif
         public static uint s_LastUpdateRetainedEventBytes;
         public static uint s_LastUpdateRetainedEventCount;
 
+        [Serializable]
+        public struct UpdateStepCount
+        {
+            public bool wasUpdated;
```
todi1856: maybe make these fields private (with [SerializeField] attribute] and introduce properties with only getters,  since as far as I understood, it's not intended to change wasUpdated/value outside of this struct?

todi1856: actually wasUpdated  is not being used outside, so no need to add property for it

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
```diff
@@ -71,27 +71,95 @@ public enum InputUpdateType
...
+                wasUpdated = true;
+                value++;
+            }
+
+            public void OnUpdate()
+            {
+                // only increment if OnBeforeUpdate was not called
+                if (!wasUpdated)
+                    value++;
+                wasUpdated = false;
```
todi1856: is this correct - wasUpdated being set to false from a function which name is OnUpdate, feels counter intuitive, maybe the name should be different.

jimon: If we call OnBeforeUpdate + OnUpdate the value should only be incremented once as it's treated as one frame, but if we call OnUpdate + OnUpdate then it's two frames and value is incremented twice

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
jimon: Good idea! Will do

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
todi1856: yes, that's fine, what I am talking about is the variable name, it's counter intuitive :) I read it as "was updated", so it's weird that's being set to false from function "where Update is happening"

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
jimon: Fixed

todi1856 (approved):

Packages/com.unity.inputsystem/InputSystem/InputUpdateType.cs
```diff
@@ -71,27 +71,96 @@ public enum InputUpdateType
 
     internal static class InputUpdate
     {
-        public static bool s_HaveUpdatedStepCount;
         public static InputUpdateType s_LastUpdateType;
-        public static uint s_UpdateStepCount;
+        public static UpdateStepCount s_PlayerUpdateStepCount;
+        #if UNITY_EDITOR
+        public static UpdateStepCount s_EditorUpdateStepCount;
+        public static uint s_UpdateStepCount => s_LastUpdateType == InputUpdateType.Editor ? s_EditorUpdateStepCount.value : s_PlayerUpdateStepCount.value;
```
jimon: copy value in-place instead of branching

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/EnhancedTouchSupport.cs
```diff
@@ -124,8 +124,10 @@ internal static void Reset()
         private static void SetUpState()
         {
             Touch.s_PlayerState.updateMask = InputUpdateType.Dynamic | InputUpdateType.Manual;
+            Touch.s_PlayerState.updateStepCount = InputUpdate.s_PlayerUpdateStepCount.value;
```
jimon: remove updateStepCount here, use global static instead

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
```diff
@@ -798,6 +798,41 @@ public void EnhancedTouch_CanReceiveCallbacksOnFingerActivity()
         }));
     }
 
+    [Test]
+    [Category("EnhancedTouch")]
+    public void EnhancedTouch_TouchWorksAfterDisableEnable()
```
Rene-Damm: Preferably something like `EnhancedTouch_CanBeDisabledAndReenabled`.

Also, for tests coming out of tickets, preferably put a comment with a link to the ticket in the code.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ however, it has to be formatted properly to pass verification tests.
   * In our own measurements, `InputUser.OnEvent` is >9 times faster than before and `RebindingOperation.OnEvent` is ~2.5 times faster.
 - Fixed PS4 controller not recognized on Mac when connected over Bluetooth ([case 1286449](https://issuetracker.unity3d.com/issues/input-system-dualshock-4-zct1e-dualshock-2-v1-devices-are-not-fully-recognised-over-bluetooth)).
 - Fixed restart prompt after package installation not appearing on Unity 2020.2+ ([case 1292513](https://issuetracker.unity3d.com/issues/input-system-after-package-install-the-update-slash-switch-and-restart-prompt-does-not-appear)).
+- Fixed incorrect touch phases reported when enhanced is re-enabled ([case 1286865](https://issuetracker.unity3d.com/issues/new-input-system-began-moved-and-ended-touch-phases-are-not-reported-when-a-second-scene-is-loaded)).
```
Rene-Damm:
```
- Fixed `Touch.activeTouches` having incorrect touch phases after calling `EnhancedTouch.Disable()` and then `EnhancedTouch.Enable()` ([case 1286865](https://issuetracker.unity3d.com/issues/new-input-system-began-moved-and-ended-touch-phases-are-not-reported-when-a-second-scene-is-loaded)).
```

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/Finger.cs
```diff
@@ -124,6 +124,10 @@ internal Finger(Touchscreen screen, int index, InputUpdateType updateMask)
                 updateMask = updateMask,
             };
             m_StateHistory.StartRecording();
+
+            // if touch is already in progress, record the current state
```
Rene-Damm: In general, please follow normal English orthography in comments.

Packages/com.unity.inputsystem/CHANGELOG.md
jimon: ok, but this is a real nitpick :P you wrote essentially what I already had 1-to-1

jimon: I'll make this pass, but would like to note that I think the changed version is less readable, as now users need to parse disable+enable statements into one semantical block.

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
jimon: I changed the name, but I think new name doesn't represent what the test is doing, the test only checks that touch phases specifically work, not whole enhanced touch module.

PR #1253, opened 2020-12-11
Add Step Counter sensor for iOS

todi1856:
* Added Step Counter sensor implementation for iOS, as you see both C# code and ObjC code lives in package.
* Made InputDevice.ExecuteCommand virtual this allows handling IOCTL commands on C# side, this also enables users to implement their own devices without waiting for us.
* Added iOS settings for Motion Usage

For testing:
- When you'll produce XCode project, please add Privacy - Motion Usage Description with any message. At this moment I am not sure if Input Package should be adding this value. UPDATE: Added a setting in Input Settings, input system package will add necessary info to Info.plist, see screenshot below
- When you'll launch app and try to Enable Step Counter - a dialog will pop asking for permission to use Motion, accept it. And start walking, after 3-10 seconds, the stepCounter value should start changing
- If you'll answer No to Motion dialog, the device will still be enabled, but no events will be coming. Even if you restart the app, the dialog won't longer show up, you either need to go to App Settings and enable Motion there, or reinstall app. I think though this is a bug in input system, the device should stay disabled if there was an error during Enable operation (like in this case authorization denial), i'll look into that in separate PR.

Outsanding question:
* iOSStepCounter doesn't support sample frequency, so if you'll try to use that property, you'll get NotSupportedException Device 'iOSStepCounter:/iOSStepCounter' does not support querying sampling frequency. Is that user friendly, or I should simply return 0. My personal opinion, I think the exception should stay, as it explicitly tells you not to use samplinFrequency property, but that's my opinion 

ProjectSettingsWindow000


Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,145 @@
...
+                var callbacks = new iOSStepCounterCallbacks();
+                callbacks.onData = OnDataReceived;
+                return _iOSStepCounterEnable(deviceId, ref callbacks, Marshal.SizeOf(callbacks));
+#endif
+            }
+
+            if (t == DisableDeviceCommand.Type)
+            {
+#if UNITY_EDITOR
+                m_Enabled = true;
```
glumazy: should it be false here and on line 108 ?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        [MonoPInvokeCallback(typeof(OnDataReceivedDelegate))]
+        private static void OnDataReceived(int deviceId, int numberOfSteps)
+        {
+            var stepCounter = (iOSStepCounter)InputSystem.GetDeviceById(deviceId);
+            InputSystem.QueueStateEvent(stepCounter, new iOSStepCounterState {stepCounter = numberOfSteps});
+        }
+
+#if UNITY_EDITOR
+        private bool m_Enabled;
```
glumazy: does it need false as default value ?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    s_Wrapper.Enable(deviceId, callbacks);
+
+    return kResultSuccess;
+}
+
+extern "C" int _iOSStepCounterDisable(int deviceId)
+{
+    s_Wrapper.Disable(deviceId);
```
glumazy: is it worth to return kResultFailure in case deviceId is diferent or m_Pedometer is null in s_Wrapper ?

glumazy (approved):
LGTM

Pavel-Unity (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
todi1856: correct, nice find

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
todi1856: Makes sense

renaldasrep (approved):
Step Counter seems to be working fine or at least as expected, my only concern is that if `Privacy - Motion Usage Description` is not set - app crashes on device enable, would be nice to get a warning instead of a hard crash

Another issue Tomas thought of when we were discussing this is that the permission dialog runs on a separate thread - what if we enable multiple devices that require such dialog at once? It might not be an issue right now, but should be thought about and dealt with for the future

Last thing I noticed is that the device has 'samplingFrequency' property that's missing a getter. My test project reflects and prints all properties and it was not happy to find that, I had to add an if/else to bypass this. Might not be a huge issue for average user, but we should probably do something about this. Return a default value and spam a warning in the console 'samplingFrequency is returning 0 and this is expected, but please don't use this' - along those lines maybe?

todi1856:
> Step Counter seems to be working fine or at least as expected, my only concern is that if `Privacy - Motion Usage Description` is not set - app crashes on device enable, would be nice to get a warning instead of a hard crash

Fixed that, it will now print to log instead of crashing

> 
> Another issue Tomas thought of when we were discussing this is that the permission dialog runs on a separate thread - what if we enable multiple devices that require such dialog at once? It might not be an issue right now, but should be thought about and dealt with for the future

That's something for later

> Last thing I noticed is that the device has 'samplingFrequency' property that's missing a getter. My test project reflects and prints all properties and it was not happy to find that, I had to add an if/else to bypass this. Might not be a huge issue for average user, but we should probably do something about this. Return a default value and spam a warning in the console 'samplingFrequency is returning 0 and this is expected, but please don't use this' - along those lines maybe?

Added a setting in Input Settings, where you can enable Motion Usage, and the relevent info will be added to Info.plist

renaldasrep:
The Motion Usage toggle+field is a very welcome addition, now it's way better. And doesn't crash!

renaldasrep (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -126,11 +126,20 @@ public override void OnGUI(string searchContext)
 
                 m_SupportedDevices.DoLayoutList();
 
+                EditorGUILayout.LabelField("iOS", EditorStyles.boldLabel);
```
jimon, line 129: A bit dunno tbh, is it worth putting platform specific detail into input system settings panel versus placing it into player settings where we already have established platform specific settings pattern? If I would be a user I would expect something changing the plist to be in the player settings, not input system.

Not to stop the PR, just want clarify what UX we expect here.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,166 @@
...
+        [DllImport("__Internal")]
+        private static extern int _iOSStepCounterIsAvailable();
+
+        [DllImport("__Internal")]
+        private static extern int _iOSStepCounterGetAuthorizationStatus();
+
+        [MonoPInvokeCallback(typeof(OnDataReceivedDelegate))]
+        private static void OnDataReceived(int deviceId, int numberOfSteps)
+        {
+            var stepCounter = (iOSStepCounter)InputSystem.GetDeviceById(deviceId);
```
jimon, line 101: I guess input system native API is not reachable from Obj-C? Otherwise it would probably be easier to just go directly to it instead of dealing with native->managed->native double transition.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
```diff
@@ -0,0 +1,152 @@
...
+            STEP_COUNTER_LOG(@"Was already enabled?");
+            if (m_Device != deviceId)
+                STEP_COUNTER_LOG(@"Enabling with different device id? Expected %d, was %d. Are you creating more than one iOSStepCounter", m_Device, deviceId);
+        }
+        m_Pedometer = [[CMPedometer alloc]init];
+        m_Callbacks = *callbacks;
+        m_Device = deviceId;
+        [m_Pedometer startPedometerUpdatesFromDate: [NSDate date] withHandler:^(CMPedometerData * _Nullable pedometerData, NSError * _Nullable error)
+        {
+            // Note: We need to call our callback on the same thread the Unity scripting is operating
```
jimon, line 40: If we could get access to input module native api here, then no, it can be queued from any thread

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
```diff
@@ -0,0 +1,152 @@
+#include <CoreMotion/CoreMotion.h>
+
+#define ENABLE_STEP_COUNTER_LOGGING 1
+#if ENABLE_STEP_COUNTER_LOGGING
+#define STEP_COUNTER_LOG(...) NSLog(@"StepCounter - %@", [NSString stringWithFormat: __VA_ARGS__])
```
jimon, line 5: I don't see any code disabling that, meaning we spam logs in release builds?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
```diff
@@ -0,0 +1,152 @@
...
+{
+    return [CMPedometer isStepCountingAvailable] ? 1 : 0;
+}
+
+extern "C" int _iOSStepCounterGetAuthorizationStatus()
+{
+    if (@available(iOS 11.0, *))
+    {
+        return (int)[CMPedometer authorizationStatus];
+    }
```
jimon, line 94: You could also use isAuthorizedForRecording for <11.0 here

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSupport.cs
```diff
@@ -41,6 +43,16 @@ public static void Initialize()
                 new InputDeviceMatcher()
                     .WithInterface("iOS")
                     .WithDeviceClass("LinearAcceleration"));
+
+            InputSystem.RegisterLayout<iOSStepCounter>();
+
+            // Don't add devices for InputTestRuntime
+            // TODO: Maybe there should be a better place for adding device from C#
+            if (InputSystem.s_Manager.m_Runtime is NativeInputRuntime)
```
jimon, line 51: Yea it seems suboptimal to do jumpyloops with checking which runtime we're using ... not quite sure what would be a better way to do it tbh.

jimon (approved):
It's ok PR given the current limitations of input system. A better way IMHO would to avoid native->managed->native double transition on every pedometer event. Also would be great to change input native module API somewhat to accommodate a case of registering "singleton" device better.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
todi1856: We don't put package settings in Player Settings, as far as I am aware.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
todi1856: Yes, it's not reachable, but I don't understand where you see the last entry to native, you mean  InputSystem.GetDeviceById goes into native?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
todi1856: It's only logging in case of errors

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
todi1856: Didn't know that exist, thanks

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
todi1856: I'll try to explore, since dispatching like this - probably delays the input event by quite a lot

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
jimon: - Your Obj-C code invokes manage code
- Your managed code queues event, which invokes native code

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
jimon: Ah ok, missed that

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
todi1856: Ah, I see... Thanks for explanation

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ however, it has to be formatted properly to pass verification tests.
                   Debug.Log($"Button {button} was pressed");
       }
   ```
+  - Added Step Counter sensor for iOS, you'll need to add 'Privacy - Motion Usage Description' to Info.plist before using the sensor.
```
Rene-Damm:
```
- Added support for step counters on iOS.
  ``
  InputSystem.EnableDevice(iOSStepCounter.current);
  ``
  * You need to add `Privacy - Motion Usage Description` to `Info.plist` to be able to use the sensor in your app.
```

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/InputSettingsiOS.cs
```diff
@@ -0,0 +1,34 @@
+#if UNITY_EDITOR || UNITY_IOS || UNITY_TVOS
+using System;
+
+namespace UnityEngine.InputSystem
+{
+    public partial class InputSettings : ScriptableObject
+    {
+        [Serializable]
+        public class iOSSettings
```
Rene-Damm, line 55: Needs API docs.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/InputSettingsiOSProvider.cs
```diff
@@ -0,0 +1,35 @@
+#if UNITY_EDITOR
+using System;
+using UnityEditor;
+
+namespace UnityEngine.InputSystem
+{
+    public class InputSettingsiOSProvider
```
Rene-Damm: Does this need to be public. If so, needs to be documented.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSPostProcessBuild.cs
```diff
@@ -0,0 +1,35 @@
+#if UNITY_EDITOR
+using System.IO;
+using UnityEditor;
+using UnityEditor.Callbacks;
+using UnityEngine;
+using UnityEngine.InputSystem;
+
+public class iOSPostProcessBuild
```
Rene-Damm: Shouldn't be in global namespace. Can this be internal? If it has to be public, needs to be documented.

From the looks of it, seems like this should be internal.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,166 @@
...
+using AOT;
+using Unity.Collections.LowLevel.Unsafe;
+using UnityEngine.InputSystem.Layouts;
+using UnityEngine.InputSystem.LowLevel;
+using UnityEngine.InputSystem.Utilities;
+
+namespace UnityEngine.InputSystem.iOS.LowLevel
+{
+    //See CoreMotion.framework/Headers/CMAuthorization.h
+    public enum MotionAuthorizationStatus : int
```
Rene-Damm: Xmldocs

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -126,11 +126,20 @@ public override void OnGUI(string searchContext)
...
+
                 if (EditorGUI.EndChangeCheck())
                     Apply();
             }
         }
 
+        private static void ShowPlatformSettings()
+        {
+            // Would be nice to get BuildTargetDiscovery.GetBuildTargetInfoList since that contains information about icons etc
+        }
```
Rene-Damm, line 141: Dead code? Add `////TODO` or `////REVIEW` for things that should be changed/considered in the future.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/InputSettingsiOS.cs
```diff
@@ -0,0 +1,34 @@
+#if UNITY_EDITOR || UNITY_IOS || UNITY_TVOS
```
Rene-Damm: Needs `PACKAGE_DOCS_GENERATION` in the #if.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,166 @@
+#if UNITY_EDITOR || UNITY_IOS || UNITY_TVOS
```
Rene-Damm: Probably needs `PACKAGE_DOCS_GENERATION` in the #if.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSupport.cs
```diff
@@ -41,6 +43,16 @@ public static void Initialize()
...
+            {
+                if (iOSStepCounter.IsAvailable())
+                    InputSystem.AddDevice<iOSStepCounter>();
```
Rene-Damm, line 54: How come we're doing this one differently and adding it from managed instead of reporting it from native?

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            Debug.LogWarning($"Unhandled command {command.GetType().Name}");
+            return kCommandFailure;
+        }
+
+        /// <summary>
+        /// Does the phone supports the pedometer?
+        /// </summary>
+        /// <returns></returns>
```
Rene-Damm: No empty stuff.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,166 @@
...
+#endif
+                return kCommandSuccess;
+            }
+
+            Debug.LogWarning($"Unhandled command {command.GetType().Name}");
+            return kCommandFailure;
+        }
+
+        /// <summary>
+        /// Does the phone supports the pedometer?
```
Rene-Damm: "support"

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+#if UNITY_EDITOR
+            return false;
+#else
+            return _iOSStepCounterIsAvailable() != 0;
+#endif
+        }
+
+        /// <summary>
+        /// Query motion authorization status
```
Rene-Damm: Docs like this are so arghhhhhhhhhh. This is the stuff that makes Unity API docs really disheartening to visit.

If the doc comment doesn't doc anything, preferably leave it out altogether.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
```diff
@@ -0,0 +1,166 @@
...
+        public FourCC format => kFormat;
+
+        [InputControl(name = "stepCounter")]
+        public int stepCounter;
+    }
+
+    /// <summary>
+    /// Step Counter (also known as pedometer) sensor for iOS.
+    /// Note: You need to add 'Privacy - Motion Usage Description' to Info.plist to use this sensor.
+    /// </summary>
```
Rene-Damm, line 50: Could use a simple code example that shows that you need to enable the sensor through the API and how to do it.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -301,6 +310,8 @@ private void InitializeWithCurrentSettings()
                     EditorGUI.LabelField(rect, m_Settings.supportedDevices[index]);
                 }
             };
+
+            m_iOSProvider = new InputSettingsiOSProvider(m_SettingsObject);
```
Rene-Damm, line 314: For now, this is probably okay. However, going forward, this won't work across platforms. E.g. with consoles, we can't do this. Preferably, this should be done using the new polymorphic serialization. So it'd just be a list of SerializableObjects. Would also be easy to make the UI extensible this way.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -519,7 +519,7 @@ protected virtual void OnRemoved()
         /// target="_blank">DeviceIoControl</a> on Windows and <a href="https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man2/ioctl.2.html"
         /// target="_blank">ioctl</a> on UNIX-like systems.
         /// </remarks>
-        public unsafe long ExecuteCommand<TCommand>(ref TCommand command)
+        public virtual unsafe long ExecuteCommand<TCommand>(ref TCommand command)
```
Rene-Damm: Hmm, I remember this one led to compilation errors on PS4 back then. Was an il2cpp issue, though, so may have been fixed. Just remember that the particular combination of virtual + parameterized + ref parameter blew up il2cpp. An easy workaround, if we still need it, is to just have a separate, non-parameterized version that is virtual and just takes a direct pointer instead.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.mm
todi1856: Actually there's no such function for CMPedometer

Packages/com.unity.inputsystem/CHANGELOG.md
todi1856: Changed the changelog since, this entry was before I did iOS input settings

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/InputSettingsiOS.cs
todi1856: Done

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/InputSettingsiOSProvider.cs
todi1856: made internal, also caught by tests

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSPostProcessBuild.cs
todi1856: Fixed, also made internal

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
todi1856: fixed

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
todi1856: fixed

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
todi1856: Rephrased

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSStepCounter.cs
todi1856: Fixed

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSupport.cs
todi1856: I believe, we always had the option to have everything in input system package... sadly I went with general flow in the beginning and did like other platforms did - half of the code in Unity core, other half in input system package.

I think the approach where you have the whole implementation in input system package, gives tremendous flexibility (same as screen keyboard impl) - we don't need 2 PRs to fix something, users are able to see implementation, it also shows how users from their project can add their own devices without being blocked by us.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
todi1856: Adjusted the code as per your suggestion

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSupport.cs
Rene-Damm: >I think the approach where you have the whole implementation in input system package, gives tremendous flexibility

Yup, agree.

I think should still report by device description, though. Doing `AddDevice<XXX>`, we bypass the layout matching process and thus don't give an opportunity for someone rewire the layout matching.

Rene-Damm: Indeed needs work on our end. Created a [ticket](https://jira.unity3d.com/browse/ISX-449) for it.

PR #1254, opened 2020-12-16
"Improved" InvokeCallbackSafe method to better guard against concurrent modification

Zeejfps:
Modified all InvokeCallbackSafe methods to Clone the InlineArray of callbacks in order to better guard against removal and addition of new callbacks while still inside the loop.

Currently, when a user registers a new callback, inside another callback, the callback gets executed immediately. I believe this is counter intuitive to what a user might expect. 

For example, when an input action is triggered a game object is enabled. Inside that gameobject OnEnabled is called. Say I want to subscribe to the same input action that triggered this. However, as soon as I subscribe to the input action my method will be executed before even Start is called.

One possible solution could have been to iterate over the callback array backwards, however, this wouldn't account for callbacks being removed and added while still in the loop.

My solution is to use the Clone method of the InlineArray to create a copy of the array we can iterate over to avoid being affected by concurrent modification.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Utilities/DelegateHelpers.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Utilities/DelegateHelpers.cs
@@ -13,10 +13,9 @@ namespace UnityEngine.InputSystem.Utilities
             if (callbacks.length == 0)
                 return;
             Profiler.BeginSample(callbackName);
-            for (var i = 0; i < callbacks.length; ++i)
+            var copy = callbacks.Clone();
+            for (var i = copy.length - 1; i >= 0; --i)
             {
-                var lengthBefore = callbacks.length;
-
                 try
                 {
                     callbacks[i]();
@@ -29,10 +28,6 @@ namespace UnityEngine.InputSystem.Utilities
                         Debug.LogError($"{exception.GetType().Name} while executing '{callbackName}' callbacks");
                     Debug.LogException(exception);
                 }
-
-                ////REVIEW: is this enough?
-                if (callbacks.length == lengthBefore - 1)
-                    --i;
             }
             Profiler.EndSample();
         }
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/Utilities/DelegateHelpers.cs
```diff
@@ -13,10 +13,9 @@ public static void InvokeCallbacksSafe(ref InlinedArray<Action> callbacks, strin
             if (callbacks.length == 0)
                 return;
             Profiler.BeginSample(callbackName);
-            for (var i = 0; i < callbacks.length; ++i)
+            var copy = callbacks.Clone();
```
Rene-Damm, line 16: This will lead to lots of GC heap garbage for callback lists with more than one element.

Zeejfps: I totally didn't realize that ArrayHelpers.Clone actually allocates a new array and clones all the values. Hmm, let me see if I can come up with something better. Assuming this is even something that should be fixed.

Zeejfps: Heck, would just using ToArray work lol?

Rene-Damm: >Assuming this is even something that should be fixed.

How would it create a clone without allocating memory, though?

>Heck, would just using ToArray work lol?

Same deal, will have to create an array to hold the clone.

Zooming back out, what seems like a better solution is to clone only on modification. I.e. have something like a struct that holds the InlinedArray and a bool that indicates whether callbacks are currently being invoked. If so, when the collection gets modified, it clones the array first. So, GC heap hit only on modifying the collection from within callbacks.

Rene-Damm:
Added an internal ticket (ISX-499) to revisit this. Thank you for bringing this to our attention :)

PR #1256, opened 2021-01-08
FIX: Actions not triggered correctly with multiple bindings to same control (case 1293808).

Rene-Damm:
Fixes [1293808](https://issuetracker.unity3d.com/issues/input-system-action-dot-triggered-triggers-only-once-after-remapping-an-action-to-a-non-trigger-button) ([internal](https://fogbugz.unity3d.com/f/cases/1293808/)).


Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
```diff
@@ -252,6 +252,74 @@ public unsafe void AddActionMap(InputActionMap map)
...
+                                    // binding. Each such control we find, we kick out of the list.
+                                    for (var k = 0; k < numControls; ++k)
+                                    {
+                                        var controlOnCurrentBinding = resolvedControls[firstControlIndex + k - memory.controlCount];
+                                        var controlIndexOnOtherBinding = resolvedControls.IndexOf(controlOnCurrentBinding,
+                                            otherBindingState.controlStartIndex - memory.controlCount, otherBindingState.controlCount);
+                                        if (controlIndexOnOtherBinding != -1)
+                                        {
+                                            // Control is bound to a previous binding. Remove it from the current binding.
+                                            resolvedControls.RemoveAt(firstControlIndex + k - memory.controlCount);
```
jimon, line 309: A bit sad that this needs to remove bindings. A more efficient way would be to pass a list of "ignored controls" to `FindControls` imho.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlList.cs
```diff
@@ -348,15 +348,29 @@ public void CopyTo(TControl[] array, int arrayIndex)
...
             var index = ToIndex(item);
             var indices = (ulong*)m_Indices.GetUnsafeReadOnlyPtr();
 
-            for (var i = 0; i < m_Count; ++i)
-                if (indices[i] == index)
-                    return i;
+            if (count < 0)
+                count = Mathf.Max(m_Count - startIndex, 0);
+
+            if (startIndex + count > m_Count)
```
jimon, line 365: Nitpick, can be ignored, but in general it would be great to avoid doing work like `var index = ToIndex(item);` and `var indices = (ulong*)m_Indices.GetUnsafeReadOnlyPtr();` before we're done with all "early outs" like `if (startIndex + count > m_Count)`

jimon (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingResolver.cs
Rene-Damm: Good point. Looked into it and I think adding an argument for a collection of ignored controls would be a straightforward change. Unfortunately, it's only *some* controls in `resolvedControls` that should be ignored and they don't have to be one after the other in the list. That part seems like it'd make moving the check into `FindControls` quite messy.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlList.cs
Rene-Damm: Good catch. Fixed.

PR #1258, opened 2021-01-13
FIX: Touch simulation leading to exceptions and assertions in UI module (case 1190150).

Rene-Damm:
Fixes [1190150](https://issuetracker.unity3d.com/issues/new-input-system-simulated-touch-in-editor-doesnt-work) ([internal](https://fogbugz.unity3d.com/f/cases/1190150/)).

Principal change is that `TouchSimulation` now __disables__ `Pointer` devices (that aren't already Touchscreens) and reroutes their input into a simulated `Touchscreen`.

Turns out this also fixes [1271942](https://issuetracker.unity3d.com/issues/android-onenddrag-not-being-called-when-there-are-at-least-2-touches-on-the-screen).


Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1708,6 +1708,11 @@ public static void EnableDevice(InputDevice device)
         /// Disable the given device, i.e. "mute" it.
         /// </summary>
         /// <param name="device">Device to disable. If already disabled, the method will do nothing.</param>
+        /// <param name="keepSendingEvents">If true, no <see cref="LowLevel.DisableDeviceCommand"/> will be sent
```
jimon, line 1711: What it means is that now there is a possibility for native and managed device states to be different. This is most likely to hit us hard in the future with some random bugs happen because "of course I double checked, device was disabled, oh wait it was only disabled on managed side".

Instead a better way would be to add an explicit device mode/state, for example: 1) active / listening to events 2) disabled / not-listening to events 3) incorporated-in-a-virtual-device

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -873,6 +874,18 @@ internal bool isSetupFinished
             }
         }
 
+        internal bool isButton
+        {
+            get => (m_ControlFlags & ControlFlags.IsButton) == ControlFlags.IsButton;
+            set
+            {
```
jimon, line 881: Nitpick: Setter on question-as-property feels weird, `control.isButton = true`  is hard to understand what it is gonna to. Either we need a better prop, like `controlType` which accepts an enum, and then that is mapped to flags internally. Or remove setter completely, because there is only one place in this PR using the setter

jimon: Maybe it's ok for now because button is a property of control, not type of control. But it's even more confusing.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1490,25 +1567,48 @@ private void OnPoint(InputAction.CallbackContext context)
...
+                return;
+
             ref var state = ref GetPointerStateFor(ref context);
             state.middleButton.isPressed = context.ReadValueAsButton();
             state.changedThisFrame = true;
         }
 
+        private bool CheckForRemovedDevice(ref InputAction.CallbackContext context)
+        {
+            // When a device is removed, we want to simply cancels ongoing pointer
```
jimon: Nitpick: Would it be possible for the native-backend/runtime/input-manager-event-consumption-logic to guarantee if device is disconnected then we always have an event to "depress"/"cancel"/"release" touch-mouse-etc controls? I'm concerned that spreading this logic all over the code would create challenges in the future, for example would composite input actions correctly transition through the phases if I randomly disconnect mouse-like devices when they are in performed state?

jimon (changes requested):
I think we shouldn't create desynchronization between managed and native states unnecessarily, if it can be avoided let's fix it in this PR before merging. I'm ok with ignoring nitpicks for now.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: >What it means is that now there is a possibility for native and managed device states to be different.

That's already the case. Backends are not guaranteed to heed `DisableDeviceCommand`. ATM the majority of devices do not. And IMO that's fine. Sending events for a device and processing events for a device IMO are two separate issues and having this not tied together between the backend and frontend seems an advantage to me.

>Instead a better way would be to add an explicit device mode/state, for example: 1) active / listening to events 2) disabled / not-listening to events 3) incorporated-in-a-virtual-device

You mean something like

```
public class InputDevice : ...
{
    // ...

    public enum InputDeviceMode mode { ... }
}

public enum InputDeviceMode
{
    EnabledAndListeningToEvents,
    DisabledAndNotSendingEvents,
    DisabledButSendingEvents,
}
```

?

jimon: Yes an explicit enum to surface the implicit state machine would be great.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
Rene-Damm: >Would it be possible for the native-backend/runtime/input-manager-event-consumption-logic to guarantee if device is disconnected then we always have an event to "depress"/"cancel"/"release" touch-mouse-etc controls?

In general, this is already happening. When a device is removed, its actions get cancelled (see `Actions_WhenDeviceIsRemoved_OngoingActionsAreCancelled`). So, usually you don't have to do this thing here.

The reason it's done in this case is due to the `ReadValue` on the touch ID control we're doing in `GetPointerStateIndexFor`.

Lemme walk through this all the way from the beginning... (warning: going to get messy....). May be useful for arriving at a better setup.

So, the UI module has a couple of actions sourcing input. However, while these are all separate actions, they are not really independent. `point`, `leftClick`, `scrollWheel`, etc. all go together. As does `navigate`, `submit`, and `cancel`. In other words, these form compound inputs.

Unfortunately, we can't use composites (at least for the pointer-type stuff) as we are sourcing arbitrary many pointers through these actions.

So... when we get input on the `point` action, we somehow need to associate that with a specific pointer in the UI that should receive the position value coming in.

That's where `touchId` is entering the picture. Normally, we simply say "one device, one pointer". So we can just use the device ID as a pointer  ID. But for touches, there may be several pointers coming in from the same device. So we need something more complicated there. And ATM that's handled by `GetPointerStateIndexFor` which reads touch IDs explicitly, if necessary.

However.... that means we have input that we read in the UI module which is *not* coming in through actions. Which is... bad.

And... which leads us back to the beginning. The reason that the UI module, unlike other parts of the code, explicitly is checking for device removals is to prevent itself from reading touch IDs when it shouldn't.

Rene-Damm:
> I think we shouldn't create desynchronization between managed and native states unnecessarily, if it can be avoided let's fix it in this PR before merging. I'm ok with ignoring nitpicks for now.

I'm not clear on what you mean by fixing, though. Having the backend keep sending events with the frontend no longer processing them IMO is something we want. Communicating this fact to the backend just seems like it would duplicate data for no benefit -- the backend wouldn't act any different. But that's probably not what you mean.

Turning the "flat" `disabled` flag on `InputDevice` into a mode that more finely discerns what specifically do with sending and what we do with receiving, that part sounds like a good avenue to explore to me, though. Is that the part you mean by fixing?

What if we have two flags? (something like "Sending events is disabled" and "Processing events is disabled")

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jimon: Got it.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
jimon: Let's refactor this later on.

PR #1261, opened 2021-01-19
Remove stale data from package.json

jimon:
This part of the package.json is always filled by upm-ci, so no need for us to keep track of it.

### DIFF
--- a/Packages/com.unity.inputsystem/package.json
+++ b/Packages/com.unity.inputsystem/package.json
@@ -3,12 +3,7 @@
 	"displayName": "Input System",
 	"version": "1.1.0-preview.2",
 	"unity": "2019.4",
-	"repository": {
-		"type": "git",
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/package.json
```diff
@@ -3,12 +3,7 @@
 	"displayName": "Input System",
 	"version": "1.1.0-preview.2",
 	"unity": "2019.4",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/Unity-Technologies/InputSystem.git",
-		"revision": "9ceb1563f669e33fe7b8562e36375caa59bb6f1b"
```
Rene-Damm, line 9: It's the entire section that is added by release tooling and not just the revision?

jimon: Yes, whole section

PR #1262, opened 2021-01-20
Moving CI to 2020.2

jimon:
No more 2020.1 unity releases are planned, so at this is not possible to fix anything in 2020.1 backend. We can re-add a sparse 2020.1 set (1-2 platforms) later on just to make sure it doesn't regress.

PS. I guess if this going in, we need to change which jobs are required

### DIFF
--- a/.yamato/config.metadata
+++ b/.yamato/config.metadata
@@ -1,6 +1,6 @@
 editors:
   - version: 2019.4
-  - version: 2020.1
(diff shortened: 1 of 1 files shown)


andrew-oc (approved):

jimon:
I'm force merging this and will change set of required jobs to point to 2020.2

PR #1264, opened 2021-01-25
Adding Device Simulator plugin for Input System support

RomanO-Unity:
This is a Device Simulator plugin which takes touchscreen inputs from the Device Simulator and inserts them into the Input System.

There are two issues that I am aware of:

1) Input System throws "A Native Collection has not been disposed, resulting in a memory leak. Enable Full StackTraces to get more details." This happens on domain reload sometimes. Enabling full stack trace does not improve the message.

I thought maybe there is some mismatch between touch begin and end/cancel events, but no everything matches and still the error persists. **Any ideas on this would be appreciated.**

2) IMGUI controls do not work, I've investigated why this happens and the fix needs to happen in the core not in this plugin.


RomanO-Unity:
@Rene-Damm I think the rest of the failures in tests are not caused by my change. What's the next step for this PR?

Rene-Damm:
> @Rene-Damm I think the rest of the failures in tests are not caused by my change.

Yup, only the configs marked "Required" need to be green.

> What's the next step for this PR?

I'll have a look at it.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
```diff
@@ -0,0 +1,92 @@
...
+
+using System;
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEditor.DeviceSimulation;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.LowLevel;
+
+public class InputSystemPlugin : DeviceSimulatorPlugin
```
Rene-Damm: Should be in namespace.

Also, can this be internal? If it has to be public, needs to be documented.

Rene-Damm (commented):
Couple comments.

- Ideally, this should have some level of automated testing in the package.
- Can we add docs to the manual for this?
- Needs a CHANGELOG entry.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
```diff
@@ -0,0 +1,92 @@
...
+    private List<InputDevice> m_DisabledDevices;
+
+    public override string title => "Input System";
+
+    public override void OnCreate()
+    {
+        var playerSettings = Resources.FindObjectsOfTypeAll<PlayerSettings>()[0];
+        SerializedObject playerSettingsSerialized = new SerializedObject(playerSettings);
+
+        var activeInputHandler = playerSettingsSerialized.FindProperty("activeInputHandler").intValue;
```
Rene-Damm: Recommend using `EditorPlayerSettingsHelper`.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
```diff
@@ -0,0 +1,92 @@
...
+        m_InputSystemEnabled = activeInputHandler == 1 || activeInputHandler == 2;
+        if (m_InputSystemEnabled)
+        {
+            deviceSimulator.touchScreenInput += OnTouchEvent;
+
+            // UGUI elements like a button don't get pressed when multiple pointers for example mouse and touchscreen are sending data at the same time
+            m_DisabledDevices = new List<InputDevice>();
+            foreach (var device in InputSystem.devices)
+            {
+                if (device.native && device is Mouse && device.enabled)
```
Rene-Damm: What about `Pen` input?

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        var activeInputHandler = playerSettingsSerialized.FindProperty("activeInputHandler").intValue;
+        m_InputSystemEnabled = activeInputHandler == 1 || activeInputHandler == 2;
+        if (m_InputSystemEnabled)
+        {
+            deviceSimulator.touchScreenInput += OnTouchEvent;
+
+            // UGUI elements like a button don't get pressed when multiple pointers for example mouse and touchscreen are sending data at the same time
+            m_DisabledDevices = new List<InputDevice>();
+            foreach (var device in InputSystem.devices)
```
Rene-Damm: This will not work correctly with devices that are added and removed on the fly.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
```diff
@@ -0,0 +1,92 @@
...
+    public override void OnDestroy()
+    {
+        if (m_InputSystemEnabled)
+        {
+            deviceSimulator.touchScreenInput -= OnTouchEvent;
+            if (m_SimulatorTouchscreen != null)
+                InputSystem.RemoveDevice(m_SimulatorTouchscreen);
+            foreach (var device in m_DisabledDevices)
+            {
+                InputSystem.EnableDevice(device);
```
Rene-Damm: Device may have disappeared by now.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
RomanO-Unity: Indeed it was misbehaving. Fixed it.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
RomanO-Unity: Added code to handle it during InputSystem.onDeviceChange. Was not actually able to make it fire, all devices in the editor seem to be pre-added and disconnecting reconnecting does not seem to do anything...

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
RomanO-Unity: Added check to ignore devices that are not **added**.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
RomanO-Unity: Done

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
RomanO-Unity: Yup, it can be internal, fixed it.

RomanO-Unity:
@Rene-Damm I've addressed the issues that you've raised, it's ready for a second look.

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
```diff
@@ -0,0 +1,110 @@
+#if UNITY_EDITOR && UNITY_2021_1_OR_NEWER
```
andrew-oc: Typo in file name: InputSy**s**temPlugin

Assets/Tests/InputSystem/Plugins/DeviceSimulatorTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        plugin.OnTouchEvent(CreateTouch(0, new Vector2(5, 5), UnityEditor.DeviceSimulation.TouchPhase.Ended));
+        yield return null;
+        Assert.AreEqual(Touch.activeTouches[0].screenPosition, new Vector2(5, 5));
+        Assert.AreEqual(Touch.activeTouches[0].phase, TouchPhase.Ended);
+
+        yield return null;
+        Assert.AreEqual(Touch.activeTouches.Count, 0);
+
+        plugin.OnDestroy();
+    }
```
andrew-oc, line 52: To fix the native array not disposed error, you need to disable the EnhancedTouchSupport at the end of this test. This will dispose the native array that was allocated to hold touch state.

Assets/Tests/InputSystem/Plugins/DeviceSimulatorTests.cs
```diff
@@ -0,0 +1,83 @@
...
+        plugin.OnDestroy();
+        Assert.IsFalse(touchscreen.added);
+    }
+
+    private TouchEvent CreateTouch(int touchId, Vector2 position, UnityEditor.DeviceSimulation.TouchPhase phase)
+    {
+        var touch = new TouchEvent();
+        var type = typeof(TouchEvent);
+        object touchObject = touch;
+        var touchIdProperty = type.GetField("<touchId>k__BackingField", BindingFlags.Instance | BindingFlags.NonPublic);
```
andrew-oc: Could we do something a little more robust here? Maybe use nameof() to get the property name and check that the backing fields actually exists?

andrew-oc (changes requested):

Packages/com.unity.inputsystem/InputSystem/Editor/DeviceSimulator/InputSytemPlugin.cs
RomanO-Unity: Big Oof... Fixed now.

Assets/Tests/InputSystem/Plugins/DeviceSimulatorTests.cs
RomanO-Unity: Done

Assets/Tests/InputSystem/Plugins/DeviceSimulatorTests.cs
RomanO-Unity: Done what you suggested. Also added a few asserts, so hopefully if this code ever breaks due to changes to TouchEvent, errors will be clearer.

andrew-oc (approved):

PR #1266, opened 2021-02-09
Add exception to AddAction when the map is part of an asset

jimon:
Show a better exception what trying to add an action to the map which is part of an input asset which other maps being enabled. Not sure if we need a changelog entry here.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Editor.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Editor.cs
@@ -678,41 +678,6 @@ partial class CoreTests
 
     [Test]
     [Category("Editor")]
-    public void Editor_InputAsset_AddingActionToAssetWithMultipleMaps()
-    {
-        var map1 = new InputActionMap("map1");
-        map1.AddAction(name: "action1", binding: "<Gamepad>/leftStick");
-
-        var map2 = new InputActionMap("map2");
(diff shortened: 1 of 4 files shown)


Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -676,6 +676,41 @@ public void Editor_InputAsset_CanReplaceBindingGroupThroughSerialization()
         Assert.That(map.bindings[0].groups, Is.EqualTo(""));
     }
 
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputAsset_AddingActionToAssetWithMultipleMaps()
```
Rene-Damm: This behavior is not specific to the editor. Should be in `CoreTests_Actions`.

Also, the test name is just a name and doesn't describe behavior. Something `Actions_AddingActionToAssetWithEnabledActions_ThrowsException`.

Rene-Damm (changes requested):
Needs CHANGELOG entry. Also needs updating of the xmldoc comment.

Assets/Tests/InputSystem/CoreTests_Editor.cs
jimon: Ah sorry, missed the fact that CoreTests_Actions had InputActionAsset related tests. Fixed.

jimon:
Added CHANGELOG entry and updated xmldoc comment

PR #1268, opened 2021-02-12
Add partial CI for tvOS + fix tvOS

todi1856:
* iOSStepCounter broke tvOS, which got unnoticed.
* This PR adds CI for tvOS (we only build for tvOS, but don't run it)
* Excluded iOSStepCounter from tvOS


jimon (approved):
lgtm

PR #1269, opened 2021-02-17
DOCS: Details about Switch Pro support on desktop (case 1190216).

Rene-Damm:
Switch Pro support via USB is more involved due to how the controller works. From all we've found, the HID descriptor and default state of the device is useless with the device expecting custom HID output commands to set it up properly. Added as an internal feature request (ISX-451).

Added some text to the docs to this end.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/Gamepad.md
+++ b/Packages/com.unity.inputsystem/Documentation~/Gamepad.md
@@ -135,3 +135,7 @@ On other platforms Unity, uses derived classes to represent Xbox controllers:
 ## Switch
 
 The Input System support Switch Pro controllers on desktop computers via the [`SwitchProControllerHID`](../api/UnityEngine.InputSystem.Switch.SwitchProControllerHID.html) class, which implements basic gamepad functionality.
+
+>__Note__: This support does not currently work for Switch Pro controllers connected via wired USB. Instead, the Switch Pro controller *must* be connected via Bluetooth. This is due to the controller using a prioprietary communication protocol on top of HID which does not allow treating the controller like any other HID.
+
+>__Note__: Switch Joy-Cons are not currently supported on desktop.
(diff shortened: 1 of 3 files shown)


jimon (approved):

mle-ii:
Is there a way to detect and disable input from a switch pro controller that's connected via USB until this issue is fixed?  I know of at least one game where due to the issue of the pro controller not being recognized it sends random inputs and in the case of this game ends up deleting player content randomly due to it going into the UI that handles game/player saves and entering those random button presses there.  :(

Also, is there a spot where devs/users can report Unity issues?  And is there a way for devs to track the issue mentioned here?

I'm just now learning Unity so I don't know enough yet on how to detect this and offer up a potential way to block these controllers in code and off a PR, assuming you take public PRs.

jimon:
@mle-ii 
> Is there a way to detect and disable input from a switch pro controller that's connected via USB until this issue is fixed?

Unlikely, [check how we detect the controller](https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchSupportHID.cs), it seems they share same USB vendor id/device id over wired and BT connection, so telling them apart is difficult on this level.

One potential hot fix for the issue could be to run Steam in background. It happens to send a custom configuration package to the controller, and after that it works as expected.

> Also, is there a spot where devs/users can report Unity issues? And is there a way for devs to track the issue mentioned here?

Please report all issues via "Help->Report a bug" ( [more info here](https://unity3d.com/unity/qa/bug-reporting) ). Ideally please also add a concrete reproduction project (or reproduction steps) and clear clarification what is expected and what is observed. Issues then are available in [the tracker](https://issuetracker.unity3d.com/) and can be voted/discussed/followed upon/etc.

mle-ii:
Ah thank you, and one thing I just realized they're using the old input system as well, and this repo seems to be for the new input system.  So even if you did have a fix here it wouldn't help them unless/until they upgraded to this new system for input.

And thank you for the bug reporting and tracking info.

PR #1271, opened 2021-02-19
Ensuring InputManager checks current state of focus on bootstrap

jimon:
This fixes `EnhancedTouch_SupportsEditorUpdates` sometimes failing on 2021+

Had hard time writing a test to it, because `m_HasFocus` is a private property that is set on bootstrap.
Ideally instead of writing a overspecifying test, I would prefer to remove `m_HasFocus` all together.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -12,7 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 
 - Delete key not working in the input actions editor ([case 1282090](https://issuetracker.unity3d.com/issues/input-system-delete-key-doesnt-work-in-the-input-actions-window)).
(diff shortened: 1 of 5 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 
 - Delete key not working in the input actions editor ([case 1282090](https://issuetracker.unity3d.com/issues/input-system-delete-key-doesnt-work-in-the-input-actions-window)).
+- Inputs in game view sometimes not working if running under editor, as we were not checking game view focus state after domain reload.
```
Rene-Damm: As per Unity documentation guidelines, no "we" in our docs. Took me some time, too, to adjust to that.

PR #1274, opened 2021-02-23
FIX: IndexOutOfRangeExceptions and null elements in InputUser.lostDevices (case 1275148).

Rene-Damm:
Fixes [1275148](https://issuetracker.unity3d.com/issues/disconnecting-and-reconnecting-input-device-causes-exception-in-inputuser) ([internal](https://fogbugz.unity3d.com/f/cases/1275148/)).

There were a couple things I stumbled over around device loss/regain logic along the way that weren't well-defined so made a couple changes beyond just the bugfix. Also added a listing of lost (and not just paired) devices for users in the input debugger.


Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
```diff
@@ -666,43 +672,40 @@ public void UnpairDevices()
...
-            for (var i = 0; i < deviceCount; ++i)
-                Notify(userIndex, InputUserChange.DeviceUnpaired, s_AllPairedDevices[deviceStartIndex + i]);
-
-            // Remove.
-            ArrayHelpers.EraseSliceWithCapacity(ref s_AllPairedDevices, ref s_AllPairedDeviceCount, deviceStartIndex, deviceCount);
-            if (s_AllUserData[userIndex].lostDeviceCount > 0)
+        private static void RemoveLostDevicesForUser(int userIndex)
+        {
+            var lostDeviceCount = s_AllUserData[userIndex].lostDeviceCount;
+            if (lostDeviceCount > 0)
```
andrew-oc, line 703: How do you feel about inverting conditionals like this to reduce nesting? I generally do it as a habit now, where possible.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -924,8 +924,8 @@ private void OnDestroy()
         [NonSerialized] internal InputActionState m_SharedStateForAllMaps;
         [NonSerialized] internal InputBinding? m_BindingMask;
 
-        [NonSerialized] private int m_DevicesCount = -1;
-        [NonSerialized] private InputDevice[] m_DevicesArray;
+        [NonSerialized] internal int m_DevicesCount = -1;
+        [NonSerialized] internal InputDevice[] m_DevicesArray;
```
andrew-oc: Is making these public really a good idea? Could we not add a method to this class and InputActionMap called TryRemoveDevice to support the logic you added to InputActionState without exposing these?

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
Rene-Damm: >to reduce nesting

Wait, that *increases* nesting, right?

Personally, I prefer the opposite, i.e. code that isn't heavily indented. I.e. instead of

```
class Foo
{
    void Bar()
    {
        if (x)
        {
             Flub();
             if (y)
             {
                 Blub();
```

get

```
class Foo
{
    void Bar()
    {
        if (!x)
            return;

         Flub();
         if (!y)
             return;

         Blub();
```

Aside from avoiding brace-craziness, every time you see a `return`, you know it's the exit right there. No need to jump all the way to the end of the `if` to see whether there's something else coming after the block.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
Rene-Damm: Not sure. IMO sharing internals between a very small set of closely related classes is not an issue. The coupled is straightforwardly visible from the references and refactoring in the future, should the need arise, is straightforward. Downside to encapsulating the data is you pay for the encapsulation with more methods and more code.

But yeah, would definitely agree there's strict limits to this. Once it's not just one or two tightly coupled classes but rather something seeping out all over the place, it's getting stupid.

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
Rene-Damm: LOL... so first I respond to the *opposite* of what you're saying and then I accidentally hit the "Resolve Conversation" button. Guess it's Friday...

So.... let me give this another try :)

Yup, fully agree with you there. `if` should be inverted here. Ignore my rambling :)

andrew-oc: All good :)

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
andrew-oc: It's only shared between two closely related classes **now**, but by making it internal you're saying to the next person that writes code relating to it that it's ok to use it directly, which it's really not. Also encapsulating it will result in less code, not more. And if we're at the point of worrying about the cost of a method call, it's time to pack up and write everything in assembly :) So basically, in my only slightly dogmatic book of coding, this is the wrong thing to do, and I would only approve it if you're adamant you don't want to change it.

Rene-Damm: >if we're at the point of worrying about the cost of a method call

Nah, not the cost of the call. Meant just the code.

I'll go wrap it.

Rene-Damm: Ok, finally fixed this. Indeed much better now. And also found some bugs in the original code in the process.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -739,13 +693,84 @@ IEnumerator IEnumerable.GetEnumerator()
...
 
         internal static int s_DeferBindingResolution;
 
+        internal struct DeviceArray
+        {
+            private bool m_HaveValue;
+            private int m_DeviceCount;
+            private InputDevice[] m_DeviceArray; // May have extra capacity; we won't let go once allocated.
+
+            public int IndexOf(InputDevice device)
```
andrew-oc, line 708: Can this method be simplified to just 
`return m_DeviceArray.IndexOfReference(device, m_DeviceCount); `?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -739,13 +693,84 @@ IEnumerator IEnumerable.GetEnumerator()
...
+            }
+
+            public ReadOnlyArray<InputDevice>? Get()
+            {
+                if (!m_HaveValue)
+                    return null;
+                return new ReadOnlyArray<InputDevice>(m_DeviceArray, 0, m_DeviceCount);
+            }
+
+            public bool Set(ReadOnlyArray<InputDevice>? devices)
```
andrew-oc, line 729: Feels like this method is doing too much. Might be worth splitting this into another method called ContainsAll that can do the check for whether all devices exist in m_DeviceArray, and not having Set return a bool. Also, this is currently more of a Sequence Equals than a contains, so you only skip rebinding if the devices exist and are in the same order. That intentional?

andrew-oc (changes requested):

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
Rene-Damm: Indeed. Done.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
Rene-Damm: (fuck... buttons aren't my friends...)

Moved the check into a method.

>Also, this is currently more of a Sequence Equals than a contains, so you only skip rebinding if the devices exist and are in the same order. That intentional?

Sort of. Somewhat gray territory. Technically, it should be a set rather than a sequence. So order should be ignored.

However, when it comes to binding resolution, the devices will be tried in the order they occur. So in the end, order does end up having a subtle influence (which in practice rarely ever matters).

So from that POV, not sure. Ended up leaving it alone.

andrew-oc: I'm a fan of command/query separation, and I'd like if the Set method didn't return anything to fit with that. It's not at all clear what that bool will represent without reading the code...but this is something for another day. PR looks good.

andrew-oc (approved):

PR #1277, opened 2021-02-25
Fix: Bindings added at runtime modifying input action asset

andrew-oc:
This was only an issue in the editor and this commit fixes it by keeping track of "dirty" input action assets, i.e. assets that have had bindings added to them at runtime, and force re-importing those from disk when exiting play mode.


Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3072,6 +3072,27 @@ internal static void OnPlayModeChange(PlayModeStateChange change)
...
+
+                    foreach (var assetGuid in s_TrackedDirtyAssets)
+                    {
+                        var assetPath = AssetDatabase.GUIDToAssetPath(assetGuid);
+
+                        if (string.IsNullOrEmpty(assetPath))
+                        {
+                            Debug.LogWarning($"Input action asset change tracking was tracking asset with GUID '{assetGuid}' but it wasn't found in the asset " +
+                                             $"database on exiting play mode. If you have added bindings to this asset during this play mode session, you may need " +
+                                             $"to reimport the asset from disk to remove the added bindings from memory.");
```
Rene-Damm: Probably better to just silently ignore. Asset may have been deleted in the meantime so it no longer being found would be expected and perfectly fine. If the asset popped up somewhere else (say, someone did a `git checkout` that happened to change the GUID in the .meta file), we're also good as the importer has brought in a clean state.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
```diff
@@ -502,6 +497,19 @@ private static int AddBindingInternal(InputActionMap map, InputBinding binding,
             return bindingIndex;
         }
 
+        private static void InvalidateActionMap(InputActionMap map)
```
Rene-Damm: Name seems a bit odd as it doesn't really invalidate the action map itself. Maybe `MarkActionMapSetupDirty` or something?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
```diff
@@ -487,12 +487,7 @@ private static int AddBindingInternal(InputActionMap map, InputBinding binding,
             else
                 ArrayHelpers.InsertAt(ref map.m_Bindings, bindingIndex, binding);
 
-            // Invalidate per-action binding sets so that this gets refreshed if
-            // anyone queries it.
-            map.ClearPerActionCachedBindingData();
-
-            // Make sure bindings get re-resolved.
-            map.LazyResolveBindings();
+            InvalidateActionMap(map);
```
Rene-Damm: This will have to get called from several more places. Probably each one in `InputActionSetupExtensions` and `InputActionRebindingExtensions` which does the `ClearPerActionCachedBindingData` and `LazyResolveBindings` call pattern here.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ however, it has to be formatted properly to pass verification tests.
 
 - Delete key not working in the input actions editor ([case 1282090](https://issuetracker.unity3d.com/issues/input-system-delete-key-doesnt-work-in-the-input-actions-window)).
 
+- Runtime rebinds are leaking into .inputactions asset ([case 1190502](https://issuetracker.unity3d.com/issues/inputsystem-runtime-rebinds-are-leaking-into-inputactions-asset))
+
```
Rene-Damm, line 57: No space after last entry. Preferably describe change rather than bug. E.g.

`- Fixed runtime rebinds and binding/action setup changes made to .inputactions assets leaking out of play mode (...).`

Assets/Tests/InputSystem/Editor/CoreEditorTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.IO;
+using NUnit.Framework;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.LowLevel;
+using UnityEngine.TestTools;
+
+internal class CoreEditorTests
+{
```
Rene-Damm: Preferably add link to ticket here as comment.

Assets/Tests/InputSystem/Editor/CoreEditorTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    [UnityTest]
+    [Category("Actions")]
+    public IEnumerator Actions_InteractiveRebinding_NewBindingsAreRemovedFromAssetOnExitingPlayMode()
```
Rene-Damm: This will have to cover more than just newly added bindings. Binding overrides and all other action/binding setup changes should also be nuked. If, for example, at runtime, code adds a new control scheme and bindings for it, all the data should also disappear.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
andrew-oc: Hmm, now I think I shouldn't have combined these calls into this method. We probably don't want to mark an asset as dirty in all cases where we use call ClearPerActionCachedBindingData and LazyResolveBindings. For example, ApplyBindingOverride does nothing to the asset that would require it to be reloaded on exit. I've undone this change and we can add a refactoring task later if you still want to combine the other two method calls.

Assets/Tests/InputSystem/Editor/CoreEditorTests.cs
andrew-oc: Overrides are already ignored since the override strings aren't serialized, but I added code to deal with new control schemes

Rene-Damm: >Overrides are already ignored since the override strings

I think the tests should explicitly check for that, though. The current implementation does remove them as a side-effect but if the implementation changes and no longer does that, would be great to have a test immediately go red.

Would prefer a single test that checks all mutations in one go.

* Added/deleted/changed bindings (both overrides and non-overrides)
* Added/deleted/changed actions
* Added/deleted/changed action maps
* Added/deleted/changed control schemes

IMO the easiest way would actually be to just clone the original asset, then apply a couple mutations, and then exit out of play mode and compare the JSON of the clone to the JSON of the original asset and expect them to match.

andrew-oc: Ok, added tests for all asset changes I could find. Let me know if I missed any. I also moved the tests back to play mode because they probably make more sense there now.

Rene-Damm (approved):
LGTM.

One thing that just occurred to me is that it's probably something we want to mention in the docs (on the APIs and in the manual).

Packages/com.unity.inputsystem/CHANGELOG.md
jo-unity: Usually it's something like:

Fixed an issue with runtime rebinds that are added as new bindings, so that they do not leak into .inputactions when Play mode is exited. 

"leak" is a bit of a jargony word though - something like "affect" might be more appropriate

jo-unity (approved):
Made some changes to the formatting and wording. LGTM

PR #1280, opened 2021-02-27
NEW: Added 'IPlayerInputEventSystem' 

Yoraiz0r:
## Changes
- Added a new `IPlayerInputEventSystem`, that allows event systems to restrict which transforms can a `InputSystemUIInputModule`s pointer interact with.
- Changed `InputSystemUIInputModule` to rely on `IPlayerInputEventSystem` instead of `MultiplayerEventSystem`.
- Changed `MultiplayerEventSystem` so it implements `IPlayerInputEventSystem`, preserving its functionality.

## Why?
The `MultiplayerEventSystem` is limited to only having one root object, meaning you cannot use several scenes for UI, unless you avoid using `playerRoot`, which means you can no longer restrict which player can access what. This PR fixes the problem, by letting you use `IPlayerInputEventSystem`, and define your own rules for what each player may target.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1280) <br/>All committers have signed the CLA.

ekcoh:
Thanks for this submission and effort being put into this @Yoraiz0r. We discussed this contribution in the team and based on previous changes, see https://github.com/Unity-Technologies/InputSystem/commit/6752b5d53f69788602fa98b9894b1758629eaefd and https://docs.unity3d.com/Packages/com.unity.inputsystem@1.4/api/UnityEngine.InputSystem.UI.MultiplayerEventSystem.html#UnityEngine_InputSystem_UI_MultiplayerEventSystem_playerRoot it seems that the desired behavior is already covered without introducing an interface. This PR will hence be closed.

PR #1281, opened 2021-03-01
CHANGE: Move VirtualMouseInput into input system DLL.

Rene-Damm:
# What

Moves the `VirtualMouseInput` component (which supports driving cursors from gamepads and joysticks) from the `Gamepad Mouse Cursor` sample to the input system DLL itself. Adds a manual section for the component.

# Why

Being shipped as a sample meant the component will not appear in our API docs and won't have a proper place in our manual. Given what the component does should really be considered stock functionality, it makes sense to just move it.

This is related to [1311114](https://issuetracker.unity3d.com/issues/virtual-mouse-cursor-calls-onclick-event-twice-when-doing-a-submit-action-with-a-gamepad) where a peculiarity of the component surfaced that needed to be documented properly.


jo-unity (approved):
Sorted out the formatting on one of the pages and reworded a few bits, but looks good to me now.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ however, it has to be formatted properly to pass verification tests.
 
 - Adding an action to a `InputActionMap` that is part of an `InputActionAsset` now requires all actions in the asset to be disabled ([case 1288335](https://issuetracker.unity3d.com/issues/adding-actions-at-runtime-to-existing-map-from-asset-triggers-assertion-error)).
   * This used to trigger an `Assert` at runtime but now properly throws an `InvalidOperationException`.
+- The `VirtualMouseInput` component is now part of the Input System DLL. It was previously packaged with the `Gamepad Mouse Cursor` sample.
```
jimon: nitpick: "Input System Assembly" instead of "Input System DLL"

jimon (approved):
Haven't went into too much detail, but seems good to me.

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Changed

PR #1286, opened 2021-03-04
Fixing deleting multiple bindings

jimon:
### Description

When deleting more than one binding, `BindingTreeItem.property` of existing ones becomes invalid until the tree is recreated. But we made one time copy of them inside `InputActionTreeView.DeleteDataOfSelectedItems` which leads to copies going stale as soon as first item is deleted. When enough items is deleted in one go, `property.GetParentProperty()` starts pointing into completely wrong place, which leads to native errors in `SerializedProperty` code.

### Changes made

When deleting, dynamically discover what action the current binding is bound to.


Rene-Damm:
Nice!

andrew-oc (approved):

PR #1287, opened 2021-03-06
GetLastRaycastResult & XRTrackingOrigin

StayTalm:
Adds two new elements to the InputSystemUIInputModule.
1: xrTrackingOrigin.  This is used as a real world to Unity world reference point, and can be set to the 'xr rig' to let the user move the rig and maintain UI interactions.  Serialized, or set as a property, it will trigger tracked pointers to always update, so they will be correctly reporting their positions even if the input state doesn't change.
This fixes: https://fogbugz.unity3d.com/f/cases/1308480/

2: Added a GetLastRaycastResult.  This helps in drawing 3D UI pointers for XR (e.g. using a line renderer to draw out the ray, or putting a small cursor on the hit UI using the raycast's hit position and normal). I copied the IsPointerOverGameObject pattern and used touch or device Ids, in case it could be useful for other users.

Added a test for each one, and put the two in the 'added' section of the changelog.

Changesets are a little funny since I branched off of a merged branch, but changes are correct.


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -162,6 +162,46 @@ public override bool IsPointerOverGameObject(int pointerOrTouchId)
...
+        /// to this are touches as a <see cref="Touchscreen"/> may have multiple pointers (one for each active
+        /// finger). For touch, you can use the <see cref="TouchControl.touchId"/> of the touch.
+        ///
+        /// Negative values will return an invalid <see cref="RaycastResult"/>.</param>
+        /// <returns>The most recent raycast information.</returns>
+        /// <remarks>
+        /// This method is for the most recent raycast, but depending on when it's called is not guaranteed to be for the current frame.
+        /// This method can be used to determine raycast distances and hit information for visualization.
+        ///
+        /// Use <see cref="RaycastResult.isValid"/> to determine if pointer hit anything.
```
Rene-Damm, line 171: Does the `<see` to `RaycastResult` here and above actually work? References to stuff outside the package are painful ATM.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
StayTalm: My Rider doesn't seem to support <See> links anywhere, but just checked in in VS 2019, and the RaycastResult and isValid both sent me to the right location.

Rene-Damm: Unfortunately, the IDE side will paint an incorrect picture there. On the Unity side, it's the output from the  `doctools` package which will have to get the link right. We have a new test on `develop` that checks for XMLDoc errors in the generated, which hopefully, if there's anything wrong it catches that.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -51,6 +51,8 @@ however, it has to be formatted properly to pass verification tests.
 - Added ability to force XR Support in a project by defining `UNITY_INPUT_FORCE_XR_PLUGIN`.
 - Added support for [UI Toolkit](https://docs.unity3d.com/Manual/UIElements.html) with Unity 2021.1+.
   * UITK is now supported as a UI solution in players. Input support for both [Unity UI](https://docs.unity3d.com/Manual/com.unity.ugui.html) and [UI Toolkit](https://docs.unity3d.com/Manual/UIElements.html) is based on the same `InputSystemUIInputModule` code path. More details in the manual.
+- InputSystemUIInputModule now has an xrTrackingOrigin property. When assigned, this will transform all tracked device positions and rotations from it's local space into Unity's world space.
+- Added InputSystemUIInputModule.GetLastRaycastResult. This returns the most recent raycast result and can be used to draw ray visualizations or get information on the most recent UI object hit.
```
chris-massie: Surround code with `backticks` and fix it's -> its.
Add a link to the public issue tracker [1308480](https://issuetracker.unity3d.com/product/unity/issues/guid/1308480/)

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -162,6 +162,46 @@ public override bool IsPointerOverGameObject(int pointerOrTouchId)
...
 
+        /// <summary>
+        /// Returns the most recent raycast information for a given pointer or touch.
+        /// </summary>
+        /// <param name="pointerOrTouchId">ID of the pointer or touch. Meaning this should correspond to either
+        /// <c>PointerEventData.pointerId</c> or <see cref="ExtendedPointerEventData.touchId"/>. The pointer ID
+        /// generally corresponds to the <see cref="InputDevice.deviceId"/> of the pointer device. An exception
+        /// to this are touches as a <see cref="Touchscreen"/> may have multiple pointers (one for each active
+        /// finger). For touch, you can use the <see cref="TouchControl.touchId"/> of the touch.
+        ///
```
chris-massie, line 164: You have to add `<br />` if you want to add a newline like this.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -584,6 +633,10 @@ internal void ProcessNavigation(ref NavigationModel navigationState)
         // Hide this while we still have to figure out what to do with this.
         private float m_TrackedDeviceDragThresholdMultiplier = 2.0f;
 
+        [Tooltip("Transform representing the real world origin for tracking devices.")]
```
chris-massie: It may be worth adding to this tooltip and xml-doc of the property to say this should typically be set to an object representing an "XR Rig". We talk about a rig in https://docs.unity3d.com/Manual/configuring-project-for-xr.html and nudging the user to set this to the expected Transform would probably help.

chris-massie (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -211,8 +252,16 @@ private void ProcessPointer(ref PointerModel state)
...
             else if (pointerType == UIPointerType.Tracked)
             {
-                eventData.trackedDeviceOrientation = state.worldOrientation;
-                eventData.trackedDevicePosition = state.worldPosition;
+                var position = state.worldPosition;
+                var rotation = state.worldOrientation;
+                if (m_XRTrackingOrigin != null)
+                {
+                    position = m_XRTrackingOrigin.TransformPoint(position);
+                    rotation = m_XRTrackingOrigin.rotation * rotation;
```
jimon, line 239: this is probably wrong, you probably need `TransformDirection` instead, otherwise if parent of `m_XRTrackingOrigin` has rotation, the result will be incorrect

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -162,6 +162,46 @@ public override bool IsPointerOverGameObject(int pointerOrTouchId)
...
+        /// </remarks>
+        /// <seealso cref="ExtendedPointerEventData.touchId"/>
+        /// <seealso cref="InputDevice.deviceId"/>
+        public RaycastResult GetLastRaycastResult(int pointerOrTouchId)
+        {
+            var stateIndex = GetPointerStateIndexFor(pointerOrTouchId);
+
+            if (stateIndex == -1)
+            {
+                for (var i = 0; i < m_PointerStates.length; ++i)
```
jimon: nitpick, can be ignored, this is leaking implementation details of `GetPointerStateIndexFor`, a better way would be to add something like `bool getLastKnownState = false` argument to `GetPointerStateIndexFor` and only left `GetLastRaycastResult` to return `pointerCurrentRaycast`

jimon (changes requested):
as per usual, all my nitpicks can be ignored and don't influence merge decision

but it seems transform rotation application is wrong, please double check and fix if necessary

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
StayTalm: I ended up going the other way on this (I wasn't sure what you meant by GetLastKnownState=false).  And I ended up making it so GetPointerStateIndexFor() would also work for Touch & Device Ids.  Both callers had the same secondary check, and it's a private function so I felt it was better to make the getter cover that case too.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
StayTalm: Can you elaborate?  This should transform the InputAction rotation into the tracking origin space.
I made a quick script test:
```
using System.Collections;
using UnityEngine;

public class TransformTest : MonoBehaviour
{
    // Base XR Stage
    [SerializeField]
    Transform m_XrTransform;

    // Transform used as a real world reference (should be parented to m_XrTransform)
    [SerializeField]
    Transform m_ReferenceLocalTransform;

    // World space Transform that should be the same global position & rotation as ReferenceLocalTransform
    [SerializeField]
    Transform m_TargetTransform;
    

    // Update is called once per frame
    void LateUpdate()
    {
        var localPosition = m_ReferenceLocalTransform.localPosition;
        var localRotation = m_ReferenceLocalTransform.localRotation;

        m_TargetTransform.position = m_XrTransform.TransformPoint(localPosition);
        m_TargetTransform.rotation = m_XrTransform.rotation * localRotation;
    }
}

```

And when I move and rotate the XRTransform or ReferenceLocalTransform game objects, the target tracks it consistently.
I feel like using TransformDirection would be more steps for the same results.

jimon: `m_XrTransform` is a hierarchy of transforms, e.g. it has `m_XrTransform.parent`
Does `m_XrTransform.transform` include parent rotation as well?

StayTalm: Ah right! Deeper Hierarchies.
Yeah, I want to include all parent transform rotations as well. I'm using `.rotation` instead of `.localRotation`, so it should be accounting for the whole hierarchy.
Tested my little script with an extra layer of GameObject on top of XrTransform and it still places the object where I'd expect it to be.

jimon:
lgtm

PR #1288, opened 2021-03-12
FIX: Scripting API generation errors, take 2

jimon:
### Description

Updated to latest `com.unity.package-manager-doctools` and fixed all documentation fixes. That should make us more inline with package validator tooling.

### Changes made

Authored by @Rene-Damm , I'm just a guy who made it work on CI :)

- Removed old `com.unity.package-manager-doctools`.
- Cloning `com.unity.package-manager-doctools` on the go instead.
- Fixed all cross reference and xml issues.

### Notes

This PR is a second iteration of #1236, I've decided to make a new one because previous one got a bit stale, and given that we attempted to first check-in a repo, the branch got very messy with git-lfs state. This branch is a squashed rebase split into a few meaningful commits.


Tools/doctools.patch
```diff
@@ -0,0 +1,43 @@
+diff --git a/Editor/Sources/Services/Doc/Documentation.cs b/Editor/Sources/Services/Doc/Documentation.cs
```
Rene-Damm, line 1: Wondering if this mechanism I put in place will turn out to be too brittle. Just thinking of them changing something upstream that breaks the patch and then it will hit a random PR in this repo. Is embedding a copy of the package the better solution after all? What do you think?

jimon: My plan is to next go and land this patch into upstream. Doing that will not make our repo break (as patch will be a no-op), but allows us to land this first.

jimon: I don't expect doctools to break much if at all. If this situation arises we can always pin git checkout to a specific hash. But I would prefer to always be up-to-date until it fails.

PR #1289, opened 2021-03-12
UI scrolling speed kPixelPerLine from 20 to 120

jimon:
# DONT MERGE YET

### Description

We have a zoo on incompatible scroll scales at Unity:

- `Input.mouseScrollDelta` takes value -1 or 1 based if native mouse scroll delta is `<= -1` or `>= 1`, usually mouse scroll delta is equal to 120 based on historical [WHEEL_DELTA](https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-mousehwheel) constant.

- The `Input.GetAxis("Mouse ScrollWheel")` has 0.1f scaling, so it will return -12 / 12 for usual mouse scroll delta of 120

- `EventType.ScrollWheel` `Event.delta` divides by 120 and multiplies by 3, so the resulting delta is -3 / 3 for usual scroll delta of 120.

Now UGUI scroll is driven by `StandaloneInputModule` which is using `Input.mouseScrollDelta`, so it's -1/0/1 per scroll event, regardless of native value.

Our `InputSystemUIInputModule` replacement currently uses scaling of 20, so it's -6/6 per usual scroll delta event of 120. Meaning using our code scroll lists are scrolling 6 times faster.

### Changes made

Adjusting kPixelPerLine from 20 to 120, this will not lead to feature parity due to `Input.mouseScrollDelta` to be utterly broken, but should give a more consistent velocity feel.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -1615,7 +1615,10 @@ namespace UnityEngine.InputSystem.UI
             return false;
         }
 
-        internal const float kPixelPerLine = 20;
+        // Due to different input systems at Unity all piping input into UI,
+        // we need to have a consistent scroll speed across the board.
+        // Hence this constant should be equal to Windows WHEEL_DELTA constant.
+        internal const float kPixelPerLine = 120.0f;
 
         private void OnScroll(InputAction.CallbackContext context)
         {
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -15,6 +15,7 @@ however, it has to be formatted properly to pass verification tests.
   * This used to trigger an `Assert` at runtime but now properly throws an `InvalidOperationException`.
 - The `VirtualMouseInput` component is now part of the Input System assembly. It was previously packaged with the `Gamepad Mouse Cursor` sample.
   * The component has a different GUID from before, so existing setups that use the component from the sample are not broken. To use the built-in component you must explicitly switch over.
+- Changed UI scroll speed from 20 to 120 pixels per line to make UI scrolling speeds more consistent with `Input.mouseScrollDelta`.
 
 ### Fixed
 


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1615,7 +1615,10 @@ private bool CheckForRemovedDevice(ref InputAction.CallbackContext context)
             return false;
         }
 
-        internal const float kPixelPerLine = 20;
+        // Due to different input systems at Unity all piping input into UI,
+        // we need to have a consistent scroll speed across the board.
+        // Hence this constant should be equal to Windows WHEEL_DELTA constant.
+        internal const float kPixelPerLine = 120.0f;
```
Rene-Damm, line 1621: This sounds Windows-specific. What's the effect on Mac? Will it scroll at super high speed there now? Should we #if this?

jimon: Just tried on my macbook 2017 model, printing `Mouse.current.scroll` values:
- Touch pad provides values in range 0-100 on average, starting from 5-10 and going to 50-100 when I scroll hard.
- Mouse wheel starts with 1-2 for scrolling single wheel notches, but then goes to 50-110 when I scroll faster.

So macOS implements acceleration curve there ... meaning oh well .. that `kPixelPerLine` is completely broken as well?
Maybe whole concept of "scrolling in lines" doesn't make sense, and it should be "scrolling in pixels", meaning `StandaloneInputModule` should use pixel based scrolling and `kPixelPerLine` be removed?

Seems like this is a much bigger change than just our package

Rene-Damm: Agree that "scrolling in lines" isn't a good concept. Think `InputSystemUIInputModule` is the only place where we should attempt the thing. Basically, there the guiding thing is that swapping out `StandaloneInputModule` should (more or less) not result in any visible change in behavior. So whatever the old thing ends up doing for scrolling, we'll need to try to emulate.

Given that the old thing does the scroll amount adjustment already at the platform level, probably means we'll have to actually handle the `kPixelPerLine` thing in some platform-dependent way here instead of just one simple constant.

IMO we don't need to be 100%. If the scrolling behavior is reasonably close, IMO that's good enough.

jimon: @Rene-Damm we can do `clamp((int)scroll.y, -1, 1)` and that would be almost 100% corresponding to `Input.mouseScrollDelta`. But it's insanely broken, it means that your scroll speed is FPS + mouse event rate dependent.

chris-massie: Also on Windows, when I scroll my mouse wheel really fast, I can sometimes get line values larger than 1 from `Input` (the `Input.mouseScrollDelta` value can return (0, -2) for instance).

jimon: Hm, I assume that is because it's clamping on event .. so if there are two events in the frame .. you get 2, oh well why it's so hard 🤣

jimon:
This needs more work, we will put it back into our backlog and approach and a later time

PR #1292, opened 2021-03-16
Fixed UnhookActionChange setting s_OnActionChangeHooked to the wrong …

SvenRH:
Fix for UnhookFromActionChange not setting s_OnActionChangeHooked correctly.

Setting it to true is an obvious error that prevents to add a new hook. This in turn breaks the PlayerInput ControlsChangedEvent if previously, in another scene, anther PlayerInput had added a listener to ControlsChangedEvent. See 
https://forum.unity.com/threads/controls-changed-event-not-invoking-on-scene-change.885592 for more information about the issue.


Rene-Damm:
Thank you for the fix! Added test and CHANGELOG and routing the change in via #1296.

PR #1293, opened 2021-03-17
FIX: Editing collections of input actions in the inspector duplicates bindings (case 1258578)

andrew-oc:
Fixes [1258578](https://issuetracker.unity3d.com/issues/adding-a-binding-to-one-inputaction-element-in-a-list-adds-the-same-binding-to-all-the-other-elements-in-the-list) ([Fogbugz](https://fogBugz.unity3d.com/f/cases/1258578/))

## Bug
The property drawer for InputAction was implemented in a way that would only work if one property drawer was created for each InputAction, but Unity only creates one property drawer instance and reuses it to draw any number of InputActions. The property drawer code would ensure that the tree view used to display the InputAction bindings was only created once, so subsequent InputActions ended up using the same tree view.

## Fix
Modified InputActionDrawerBase to maintain a dictionary of tree views and other state, keyed on the serialized property path, so for every new input action created, one new tree view is also be created.


Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
```diff
@@ -38,21 +50,56 @@ private void InitTreeIfNeeded(SerializedProperty property)
...
-                    // With the tree in the inspector, the foldouts are drawn too far to the left. I don't
-                    // really know where this is coming from. This works around it by adding an arbitrary offset...
-                    foldoutOffset = 14,
-                    drawActionPropertiesButton = true
-                };
-                m_TreeView.Reload();
+                property.FindPropertyRelative(nameof(InputAction.m_Id)).stringValue = "";
+                property.FindPropertyRelative(nameof(InputAction.m_Name)).stringValue = "Input Action";
+                property.FindPropertyRelative(nameof(InputAction.m_SingletonActionBindings)).ClearArray();
+                property.serializedObject.ApplyModifiedProperties();
```
Rene-Damm: Probably better to use `ApplyModifiedPropertiesWithoutUndo` here.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
```diff
@@ -144,6 +209,12 @@ private void OnGUI()
 
             private PropertiesViewBase m_PropertyView;
         }
+
+        private class InputActionDrawerViewData
```
Rene-Damm, line 182: Nitpick: struct seems to do just as well.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
```diff
@@ -22,12 +26,20 @@ public override bool CanCacheInspectorGUI(SerializedProperty property)
 
         public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
         {
+            // don't use the custom property drawer when multi-editing for now. Just show the default
```
Rene-Damm: Ideally should contain why we do that.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
```diff
@@ -38,21 +50,56 @@ private void InitTreeIfNeeded(SerializedProperty property)
...
+            // clear out some properties of the InputAction (id, name, and singleton action bindings) and
+            // recreate the tree view.
+
+            var array = property.GetArrayPropertyFromElement();
+            var index = property.GetIndexOfArrayElement();
+
+            for (var i = 0; i < index; i++)
+            {
+                if (property.FindPropertyRelative(nameof(InputAction.m_Id)).stringValue == array
+                    .GetArrayElementAtIndex(i).FindPropertyRelative(nameof(InputAction.m_Id)).stringValue)
```
Rene-Damm: Now occurs to me for the first time... did you check this with an `InputActionMap` property? `InputActionDrawerBase` is used as the basis both for `InputAction` and `InputActionMap` properties. Do the changes work for either?

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
andrew-oc: :+1:

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
andrew-oc: If I change it to struct, I'd have to change GetOrCreateViewData to have a ref argument because the instance it returns is modified in various places. I can do that, but I think it's cleaner as is. Thoughts?

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
andrew-oc: Actually multi-editing seems to be working now, so I've taken this out.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
andrew-oc: I did not...and they did not. But they do now.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
Rene-Damm: Totally up to you. In my head, paying 36 bytes (IIRC) for an object header plus 8 bytes for a pointer on a payload of 16 bytes just brings up an immediate "hmmm....". But in the context here, even classifying it as a "nitpick" is inflating this *massively*. Matter of fact, I most likely wasted two persons' mental cycles on absolutely nothing here :D

andrew-oc: Haha, it's good to be critical. Off the hot path, I think maintainability and readability are more important than squeezing every byte, so if you can still sleep at night with this in the code, I'll leave it as is.

PR #1298, opened 2021-03-23
NEW: API to parse control paths.

Rene-Damm:
# What?

Adds a new API that wraps around our internal parser for control paths.

```CSharp
foreach (var part in InputControlPath.Parse("<XRController>{LeftHand}/trigger")
{
    Debug.Log(part.layout);
    Debug.Log(part.usages.First());
    Debug.Log(part.name);
}
```

# Why?

XR team needs to programmatically process the bindings in an `.inputactions` asset for use with OpenXR. Rather than them copypasting our internal code to work with paths, makes sense to expose the parsing so that everyone can process our path language where needed.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Controls.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Controls.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using NUnit.Framework;
 using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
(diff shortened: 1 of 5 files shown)


andrew-oc (approved):
The PR looks fine but I'm a little concerned that this API shouldn't really be made public until it's more complete. Calling it a parser while it's internal is fine but for it to be public, I feel it should really be able to report parsing errors.

jo-unity:
Removing myself from review because I don't have time, but @duckets can help review the docs for this one 😊

Rene-Damm:
@andrew-oc 
>should really be able to report parsing errors.

Indeed. Think this goes for the parser in general, even when used internally. Added a [ticket](https://jira.unity3d.com/browse/ISX-530) to our tech debt epic.

bryandube (approved):
Looks good.  I merged this into a local branch and updated the OpenXR v2 input code to use it and all looks good.

duckets (approved):
Approved with minor docs amendments

PR #1304, opened 2021-03-31
Making keyboard display names consistent

jimon:
### Description

Platforms don't have any consistency to keyboard key names, it can be all upper case, all lower case, title case. And can even change between keyboard layouts on the same platform. For example Swedish Windows have key names in all upper case.

### Changes made

Enforced title case (even on all upper case) names.

### Checklist

Before review:

- [X] Changelog entry added.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/KeyControl.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/KeyControl.cs
@@ -69,7 +69,7 @@ namespace UnityEngine.InputSystem.Controls
                     displayName = rawKeyName;
                     return;
                 }
-                
+
                 displayName = textInfo.ToTitleCase(keyNameLowerCase);
             }
         }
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Controls/KeyControl.cs
```diff
@@ -50,7 +51,13 @@ protected override void RefreshConfiguration()
             if (device.ExecuteCommand(ref command) > 0)
             {
                 m_ScanCode = command.scanOrKeyCode;
-                displayName = command.ReadKeyName();
+
+                var rawKeyName = command.ReadKeyName();
+                var textInfo = CultureInfo.CurrentCulture.TextInfo;
```
andrew-oc: Should we use InvariantCulture instead? From the docs, it doesn't really make a difference now but it might in the future. The docs also have one of my favourite lines of documentation: "We reserve the right to make this API slower in the future."

jimon: Fixed

andrew-oc (approved):

PR #1306, opened 2021-03-31
FIX: Can't rebind to a key that was pressed when the rebinding operation started (case 1317225)

andrew-oc:
### Description

Fixes an issue where it is impossible to bind a control to an action if that control is pressed when a rebind operation begins. This was caused by the rebinding logic using the flag IgnoreControlsInCurrentState when iterating over controls which would prevent any actuated control from ever being considered during rebinding.

### Changes made

* Removed the IgnoreControlsInCurrentState flag when iterating over controls in the RebindingOperation.OnEvent method. This affects performance because all controls in the current device will now be checked, so changed the starting actuation code to use a dictionary rather than arrays and linear search, and optimized path part matching by changing from InvariantCulture to Ordinal string comparisons.
* Made a small change to starting actuation logic so once the pressed key is released, the next time it is pressed it will be chosen as the binding, rather than having to be released, pressed and released, then pressed again.
* This change broke the skip noisy controls path in InputEventControlCollection, which only works when also ignoring controls in current state, so added some extra code there to compensate.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions_Rebinding.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions_Rebinding.cs
@@ -625,6 +625,35 @@ internal partial class CoreTests
         }
     }
 
+    // Tests that controls with discrete actuations successfully rebind when the control is already
+    // actuated when rebinding begins.
+    // https://fogbugz.unity3d.com/f/cases/1317225/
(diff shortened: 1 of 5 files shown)


jimon (approved):
Haven't went into full details, but the tests are green and conceptually seems reasonable. LGTM

PR #1309, opened 2021-04-07
Properly selecting dropdown items when going up through hierarchy

jimon:
### Description

In input action editor when selecting binding path via dropdown window, when clicking on items in the window, closing window by clicking away/pressing escape, and then opening window again, it remembers the state where it was last in. That is helpful so you don't need to traverse the hierarchy again. But when clicking up through hierarchy (going to parent), the state was not updated, so closing/opening the window was resulting in the wrong section.

### Changes made

Correctly selecting right items when going up through hierarchy.

### Notes

Doesn't seem like we have any tests in this region, so this fix was manually tested.

### Checklist

Before review:

- [X] Changelog entry added.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/Internal/AdvancedDropdown/AdvancedDropdownWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/Internal/AdvancedDropdown/AdvancedDropdownWindow.cs
@@ -518,7 +518,17 @@ namespace UnityEngine.InputSystem.Editor
             else
                 m_NewAnimTarget = -1;
             m_AnimationTree = m_CurrentlyRenderedTree;
-            m_CurrentlyRenderedTree = m_ViewsStack.Pop();
+            var parentItem = m_ViewsStack.Pop();
+
+            m_State.ClearSelectionOnItem(m_CurrentlyRenderedTree);
+
+            if (parentItem != null)
+            {
+                var suggestedIndex = parentItem.GetIndexOfChild(m_CurrentlyRenderedTree);
+                m_State.SetSelectionOnItem(parentItem, suggestedIndex);
+            }
+
+            m_CurrentlyRenderedTree = parentItem;
         }
 
         private void GoToChild()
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/Internal/AdvancedDropdown/AdvancedDropdownWindow.cs
```diff
@@ -518,7 +518,14 @@ protected void GoToParent()
             else
                 m_NewAnimTarget = -1;
             m_AnimationTree = m_CurrentlyRenderedTree;
-            m_CurrentlyRenderedTree = m_ViewsStack.Pop();
+            var parentItem = m_ViewsStack.Pop();
+
+            m_State.ClearSelectionOnItem(m_CurrentlyRenderedTree);
+
+            var suggestedIndex = parentItem.children.IndexOf(m_CurrentlyRenderedTree);
```
andrew-oc: Can parentItem be null here?

andrew-oc (commented):
The fix looks fine, but I'm wondering about changing the 'children' collection to an IList. That collection shouldn't be exposed at all IMO. It's used in too many places right now to simply change it (would be a bigger refactoring) but for this bug fix, maybe leaving it as an IEnumerable and adding a method to AdvanceDropDownItem called GetIndexOfChild or something similar would be better. It would be a tiny step towards having some actual encapsulation in this system, but I'd like if we could start moving things in that direction.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/AdvancedDropdown/AdvancedDropdownWindow.cs
jimon: I don't think so, as it always has a root and we can't go higher than a root item. But I can add a check just to be extra sure :)

jimon: Fixed

andrew-oc (approved):

PR #1310, opened 2021-04-07
Fixing Keyboard.current becoming null after OnScreenButton is disabled/destroyed

jimon:
### Description

When `OnScreenButton` is disabled, it removes it's own virtual device, which triggers method `Keyboard.OnRemoved` which sets `Keyboard.current` static variable to `null`. [We have a bug](https://issuetracker.unity3d.com/issues/inputsystem-keyboard-dot-current-becomes-null-after-onscreenbutton-is-destroyed) complaining that it's suboptimal. Given keyboard is of a more fundamental device, it makes sense to set current to already known device without waiting for it's input.

### Changes made

If any Keyboard-derived device is removed, set first known Keyboard-derived device as current.

### Notes

This change fixes the bug by introducing technical debt, a more proper fix would be to redesign whole `.current` concept, but due lack of time/priorities it's not possible right now.

### Checklist

Before review:

- [X] Changelog entry added.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -4196,4 +4196,20 @@ partial class CoreTests
         InputSystem.RemoveDevice(keyboard2);
         Assert.That(Keyboard.current, Is.EqualTo(keyboard1));
     }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_RemovingDevice_MakesNextDeviceOfTypeCurrent()
+    {
+        var mouse = InputSystem.AddDevice<Mouse>();
+        Press(mouse.leftButton);
+        Assert.That(Pointer.current, Is.EqualTo(mouse));
+
+        var pointer = InputSystem.AddDevice<Pointer>();
+        Move(pointer.position, Vector2.right);
+        Assert.That(Pointer.current, Is.EqualTo(pointer));
+
+        InputSystem.RemoveDevice(pointer);
+        Assert.That(Pointer.current, Is.EqualTo(mouse));
+    }
 }
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -4180,4 +4180,20 @@ partial class CoreTests
         keyboard.SetIMECursorPosition(Vector2.one);
         Assert.That(commandWasSent, Is.True);
     }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_RemovingKeyboardMakesNextKeyboardCurrent()
+    {
+        var keyboard1 = InputSystem.AddDevice<Keyboard>();
+        Press(keyboard1.spaceKey);
+        Assert.That(Keyboard.current, Is.EqualTo(keyboard1));
+
+        var keyboard2 = InputSystem.AddDevice<Keyboard>();
+        Press(keyboard2.spaceKey);
+        Assert.That(Keyboard.current, Is.EqualTo(keyboard2));
+
+        InputSystem.RemoveDevice(keyboard2);
+        Assert.That(Keyboard.current, Is.EqualTo(keyboard1));
+    }
 }
--- a/Assets/Tests/InputSystem/Plugins/OnScreenTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/OnScreenTests.cs
@@ -263,6 +263,27 @@ internal class OnScreenTests : CoreTestsFixture
         Assert.That(Gamepad.all[0].buttonSouth.isPressed, Is.False);
     }
 
+    // https://fogbugz.unity3d.com/f/cases/1305016/
+    [Test]
+    [Category("Devices")]
+    public void Devices_CanUseKeyboardCurrentAfterDisablingOnScreenButton()
+    {
+        var systemKeyboard = InputSystem.AddDevice<Keyboard>();
+
+        Assert.That(Keyboard.current, Is.EqualTo(systemKeyboard));
+
+        var gameObject = new GameObject();
+        var button = gameObject.AddComponent<OnScreenButton>();
+        button.controlPath = "<Keyboard>/a";
+
+        Assert.That(Keyboard.current, Is.Not.EqualTo(systemKeyboard));
+        Assert.That(Keyboard.current, Is.Not.Null);
+
+        gameObject.SetActive(false);
+
+        Assert.That(Keyboard.current, Is.EqualTo(systemKeyboard));
+    }
+
     private class TestEventSystem : EventSystem
     {
         public new void OnApplicationFocus(bool hasFocus)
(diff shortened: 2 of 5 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1303,6 +1303,20 @@ public void RemoveDevice(InputDevice device, bool keepOnListOfAvailableDevices =
...
+            {
+                var candidate = m_Devices[i];
+                if (candidate == null || !candidate.enabled)
+                    continue;
+
+                if (!(device is Keyboard) || !(candidate is Keyboard))
+                    continue;
+                candidate.MakeCurrent();
+                break;
+            }
```
Rene-Damm: Still thinking about how we could solve this on a generic basis.

At least as far as `Keyboard` is concerned, though, this can more elegantly be solved by changing the code in `Keyboard.OnRemove`.

```
        /// <summary>
        /// Called when the keyboard is removed from the system.
        /// </summary>
        protected override void OnRemoved()
        {
            base.OnRemoved();
            if (current == this)
                current = null; // <------- Look for other keyboard here instead.
        }
```

Assets/Tests/InputSystem/Plugins/OnScreenTests.cs
```diff
@@ -263,6 +263,27 @@ public IEnumerator Devices_CanHaveOnScreenJoystickControls()
         Assert.That(Gamepad.all[0].buttonSouth.isPressed, Is.False);
     }
 
+    // https://fogbugz.unity3d.com/f/cases/1305016/
+    [Test]
+    [Category("Devices")]
+    public void Devices_CanUseKeyboardCurrentAfterDisablingOnScreenButton()
```
Rene-Damm, line 269: Recommend not making this specific to on-screen controls but rather just have a generic test `Devices_RemovingKeyboardMakesNextKeyboardCurrent` along with the other device tests.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: As far as doing this generically goes, I'm thinking the right way would be to pin this to `InputControlLayout.isGenericTypeOfDevice`. I.e. when a device is removed, the next device of the same generic type should be made current. So if you remove something that is a `Mouse`, the next device that is a `Mouse` is made current.

Problem is we don't want to load layouts here and we don't currently have a way of quickly determining the name of the first base layout that has `InputControlLayoutAttribute.isGenericTypeOfDevice` set. In other words, we'd first have to make this query quick and easy.

jimon: `Mouse` inherits from `Pointer` and both `Mouse` and `Pointer` at a same time have `current`, so calling `device.MakeCurrent()` when it's a mouse will reset the `Pointer`. Because of this it's not a simple task of just traversing class hierarchies, code needs to understand semantics of where the `current` lives in types and the side effects of calling `MakeCurrent()`.

In this particular scenario I would like to focus on the actual bug: user just disables `OnScreenButton` and their `Keyboard.current` gets corrupted (by design). So we only need to fix it for `Keyboard`, and for `Keyboard` the heuristic of "set whatever first Keyboard you have because it will be the default system keyboard" makes a lot of sense. And after that we can figure out some functionality improvements later on (aka feature requests / refactoring / etc).

Assets/Tests/InputSystem/Plugins/OnScreenTests.cs
jimon: I'll add a keyboard test separately. But would like to keep bug specific test, just so we keep sure that something as disabling `OnScreenButton` doesn't break keyboard API's.

jimon: Added `Devices_RemovingKeyboardMakesNextKeyboardCurrent`

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1565,17 +1565,21 @@ public static InputDevice GetDevice(string nameOrLayout)
...
+
+        public static InputDevice GetDevice(Type type)
+        {
+            InputDevice result = null;
             var lastUpdateTime = -1.0;
             foreach (var device in devices)
             {
-                var deviceOfType = device as TDevice;
-                if (deviceOfType == null)
+                if (device.GetType() != type)
```
Rene-Damm: Should use `IsAssignableFrom` to take inheritance into account.

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -4180,4 +4180,20 @@ public unsafe void Devices_CanSetIMECursorPositionOnKeyboard()
         keyboard.SetIMECursorPosition(Vector2.one);
         Assert.That(commandWasSent, Is.True);
     }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_RemovingKeyboardMakesNextKeyboardCurrent()
```
Rene-Damm, line 4186: Ideally should check for inheritance. Maybe `Devices_RemovingDevice_MakesNextDeviceOfTypeCurrent` using a Pointer and a Mouse device.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1565,17 +1565,21 @@ public static InputDevice GetDevice(string nameOrLayout)
         public static TDevice GetDevice<TDevice>()
             where TDevice : InputDevice
         {
-            TDevice result = null;
+            return (TDevice)GetDevice(typeof(TDevice));
+        }
+
+        public static InputDevice GetDevice(Type type)
```
Rene-Damm, line 1587: Docs.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: The new `RemoveDevice` behavior is probably worth documenting as well. Something like

```
/// Removing a device that is the current device of its type (see <see cref="InputDevice.MakeCurrent"/>)
/// automatically makes the next device of a compatible type current. For example, removing a <see cref="Mouse"/>
/// may thus make a <see cref="Touchscreen"/> the current <see cref="Pointer.current"/>.
```

Rene-Damm: Hmm, actually just realized that is not what the behavior does. Probably what it should do but not sure we really need to bother.

jimon: fixed

Assets/Tests/InputSystem/CoreTests_Devices.cs
jimon: fixed

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -1562,20 +1562,40 @@ public static InputDevice GetDevice(string nameOrLayout)
             return s_Manager.TryGetDevice(nameOrLayout);
         }
 
+        ////REVIEW: this API seems inconsistent with GetDevice(string); both have very different meaning yet very similar signatures
```
Rene-Damm, line 1565: 👍

Rene-Damm (approved):

PR #1313, opened 2021-04-08
Syncing pointer position when allocating a new pointer in ui input module

jimon:
### Description

If a click event comes from a touch with new id, the gameobject raycast code looks in a wrong position (at default, `(0, 0)`) because we don't sync touch position. This is visible when using simulated touch as it generates new touch id every time.

### Changes made

Added explicit position sync.

### Notes

This change fixes the bug by introducing technical debt, a more proper fix would be to redesign how ui input module deals with data, as currently it's unclear if the data gets synchronized correctly at all, but due lack of time/priorities it's not possible right now.

### Checklist

Before review:

- [X] Changelog entry added.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -1604,7 +1604,7 @@ namespace UnityEngine.InputSystem.UI
             m_CurrentPointerId = pointerId;
             m_CurrentPointerIndex = index;
             m_CurrentPointerType = pointerType;
-            
+
             if (point != null && point.action != null)
             {
                 var state = m_PointerStates[m_CurrentPointerIndex];
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -1540,6 +1540,13 @@ namespace UnityEngine.InputSystem.UI
                 m_CurrentPointerId = pointerId;
                 m_CurrentPointerType = pointerType;
 
+                if (point != null && point.action != null)
+                {
+                    var state = m_PointerStates[m_CurrentPointerIndex];
+                    state.screenPosition = point.action.ReadValue<Vector2>();
+                    m_PointerStates[m_CurrentPointerIndex] = state;
+                }
+
                 return m_CurrentPointerIndex;
             }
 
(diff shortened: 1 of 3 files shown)


andrew-oc (commented):
Confused by this one. The test that was added passes even if the changes to InputSystemUIInputModule are reverted. Doesn't the InputSystemUIInputModule.OnPoint method already sync the screen position in the same way as the changes?

jimon:
@andrew-oc good catch! I _think_ what happens is that ui module reuses touch entries, so if touch id doesn't change .. it kinda uses old position value. Not 100% sure this is correct, but that could explain why this was working for normal touch screens but fails for emulated ones, because emulated one creates a new touch id for every press IIRC

I need to adjust the test it seems

jimon:
@andrew-oc updated the test, now it does multiple attempts, without the fix second attempt will fail

andrew-oc (approved):

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1313) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1313) before we can accept your contribution.<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1313) it.</sub>

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1540,6 +1540,13 @@ private int GetPointerStateIndexFor(InputControl control)
                 m_CurrentPointerId = pointerId;
                 m_CurrentPointerType = pointerType;
 
+                if (point != null && point.action != null)
+                {
+                    var state = m_PointerStates[m_CurrentPointerIndex];
+                    state.screenPosition = point.action.ReadValue<Vector2>();
```
Rene-Damm, line 1546: This won't work reliably due to the nature of pass-through actions. The read will go to the last active control, not necessarily to the one that correlates with the control we're looking for here.

Rene-Damm:
Should be fixed by #1449.

PR #1314, opened 2021-04-12
FIX: UI stops working when multiple scenes with event systems are additively loaded and unloaded

andrew-oc:
### Description

Load a scene that contains an InputSystemUIInputModule that uses the default actions setup. Then additively load a second scene that also has an InputSystemUIInputModule with the default setup and unload the first scene. The first InputSystemUIInputModule will be destroyed and disable all of the input actions that the second input module is still using.

### Changes made

Added an InputAction reference counting solution to InputSystemUIInputModule where any time an InputSystemUIInputModule enables an action, it increments a reference for that action. When an InputSystemUIInputModule is destroyed, it will only disable input actions that have no more references. 

### Notes

The InputSystemUIInputModule used to have the concept of ownership of input actions to try and prevent this problem but it would fail for the case where the module was destroyed after another one had been created because ownership couldn't float between them. This might have been a reasonable fix for the issue here, but it would not have handled the case where an input action of the input module was assigned to after the module was enabled.

### Checklist

Before review:

- [x] Changelog entry added.
- [x] Tests added

### DIFF
--- a/Assets/Tests/InputSystem/Assets/UIInputModuleTestScene.unity
+++ b/Assets/Tests/InputSystem/Assets/UIInputModuleTestScene.unity
@@ -1,0 +1,209 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!29 &1
+OcclusionCullingSettings:
+  m_ObjectHideFlags: 0
+  serializedVersion: 2
+  m_OcclusionBakeSettings:
+    smallestOccluder: 5
+    smallestHole: 0.25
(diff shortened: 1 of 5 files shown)


Rene-Damm (approved):
LGTM.

Guess the problem here is the collapsing of configuration and runtime data into one showing its ugly side.

Wonder if there's a non-breaking way we can move away from that. Something where we have a controller object that for now is a step towards separation but remains optional and in addition to what's already there. And then later, when we can break, we move over entirely. Downside is it adds even more to an already crowded API surface.

PR #1315, opened 2021-04-12
FIX: clickCount not incremented correctly (case 1317239).

Rene-Damm:
### Description

Fixes [1317239](https://issuetracker.unity3d.com/issues/eventdata-dot-clickcount-doesnt-increase-when-clicking-repeatedly-in-the-new-input-system) ([internal](https://fogbugz.unity3d.com/f/cases/1317239/)).

### Changes made

Main problem was that the code before ran the `newPressed == lastPress` check before looking for the `OnPointerClick` handler (in the mistaken believe it would actually *invoke* the handler). This meant that if there was no `OnPointerDown` handler and only an `OnPointerClick` handler, click detection wouldn't work.

Other than this, I moved the click count around so that `clickCount` and `clickTime` are now actually tied to `OnPointerClick`. Meaning, you only get a click if the press-release sequence actually does form a click.

Some other minor massaging.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.cs
@@ -1789,12 +1789,12 @@ internal class UITests : CoreTestsFixture
         scene.uiModule.point = InputActionReference.Create(actions.UI.Point);
         scene.uiModule.leftClick = InputActionReference.Create(actions.UI.Click);
 
-        Set(mouse.position, scene.From640x480ToScreen(100, 100), queueEventOnly : true);
+        Set(mouse.position, scene.From640x480ToScreen(100, 100), queueEventOnly: true);
 
         yield return null;
 
         scene.ClearEvents();
-        Press(mouse.leftButton, queueEventOnly : true);
+        Press(mouse.leftButton, queueEventOnly: true);
 
         yield return null;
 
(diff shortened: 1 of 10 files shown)


Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -494,131 +504,76 @@ public IEnumerator UI_CanDriveUIFromPointer(string deviceLayout, UIPointerType p
...
+                    // PointerClick.
+                    OneEvent("type", EventType.PointerClick),
+                    OneEvent("clickTime", clickTime),
+                    OneEvent("clickCount", 1),
+                    OneEvent("pointerDrag", scene.leftGameObject),
+                    OneEvent("pointerPress", scene.leftGameObject),
+                    OneEvent("rawPointerPress", scene.leftGameObject),
+                    OneEvent("useDragThreshold", false)
+                )
+            );
```
Rene-Damm, line 551: Just some more replacing of older test code with the newer event check mechanism.

andrew-oc (approved):
The fix looks fine. I'm not a fan of the extra state that was added to ButtonState to make this work though. It doesn't feel like part of button state. Its just shoehorned in there for convenience. I think a better approach would be to move this state to a class that only does click tracking. The input module could keep an instance of that class around and forward presses and releases to it. Would have the added benefit of simplifying the logic in ProcessPointerButton.

PR #1317, opened 2021-04-15
Setting default input actions asset in InputSystemUIInputModule (fixes 1323566)

jimon:
### Description

When adding ui module with `AddComponent<InputSystemUIInputModule>` in `Awake` of some script present in the scene, deserialization of default references of ui module happens after `OnEnable` is called. Which makes `InputSystemUIInputModule` not work correctly, e.g. setting `module.actionsAsset = defaultReference` would be a no-op even if module went through `Awake`/`OnEnable` with `m_ActionsAsset` being `null`

### Changes made

- Setting default asset from `Reset()` instead of relying on script file settings in inspector.
- Added `AssignDefaultActions` for convenience of the user, otherwise they would need to set all actions manually from script.

### Notes

This is not possible to test from unit test. The behavior is only reproducible when pressing play button in editor, as part of entering playmode we load the scene and call `Awake` on everything, but editor is not in play loop yet, hence why setting default refences is still enabled at that point. Unit tests run in play mode instead.

### Checklist

- [X] Changelog entry added.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -6,6 +6,9 @@ using UnityEngine.InputSystem.Controls;
 using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.InputSystem.Utilities;
 using UnityEngine.Serialization;
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
 
 ////FIXME: The UI is currently not reacting to pointers until they are moved after the UI module has been enabled. What needs to
 ////       happen is that point, trackedDevicePosition, and trackedDeviceOrientation have initial state checks. However, for touch,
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1204,8 +1204,35 @@ public InputActionReference trackedDeviceSelect
...
+            // then deserialization of default references happens after Awake/OnEnable,
+            // and our logic relies on values of serialized fields not change after Awake/OnEnable.
+            // If we detect a potential situation like this, set the default input actions by hand to prevent it.
+            if (m_ActionsAsset != null || m_PointAction != null || m_MoveAction != null || m_SubmitAction != null ||
+                m_CancelAction != null || m_LeftClickAction != null || m_MiddleClickAction != null ||
+                m_RightClickAction != null || m_ScrollWheelAction != null || m_TrackedDevicePositionAction != null ||
+                m_TrackedDeviceOrientationAction != null) return;
+            var defaultActions = new DefaultInputActions();
+            m_ActionsAsset = defaultActions.asset;
+            m_PointAction = InputActionReference.Create(defaultActions.@UI.Point);
```
andrew-oc: What's with the verbatims?

andrew-oc (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1197,13 +1200,47 @@ public InputActionReference trackedDevicePosition
             set => SwapAction(ref m_TrackedDevicePositionAction, value, m_ActionsHooked, m_OnTrackedDevicePositionDelegate);
         }
 
+        /// <summary>
+        /// Assigns default input actions asset and input actions, similar to how defaults are assigned when creating UI module in editor.
+        /// Useful for creating <see cref="InputSystemUIInputModule"/> in runtime.
```
andrew-oc: "...*at* runtime"

andrew-oc (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jimon: N/A anymore

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jimon: fixed

andrew-oc (approved):

PR #1326, opened 2021-05-05
FIX: InputUser.pairedDevices getting corrupted (case 1327628).

Rene-Damm:
### Description

Fixes [1327628](https://issuetracker.unity3d.com/issues/input-system-devices-are-reassigned-to-the-wrong-users-after-adding-a-new-device) ([internal](https://fogbugz.unity3d.com/f/cases/1327628/)).

Pairing and unpairing devices from `InputUser`s had the potential for corrupting lists of `InputUser.pairedDevices` on users due to a nonsensical implementation of `ArrayHelpers.MoveSlice` (which slides a window of elements around in an array without requiring a temporary memory allocation).

### Changes made

Replaced the implementation of `ArrayHelpers.MoveSlice` entirely with one that recognizes that if it isn't an absolutely straightforward swapping of two array regions, then it is a simple rotation of elements within a window into the array.

Could be there's a good and proper algorithm for this that I'm ignorant of and that my Google fu failed to turn up. Had to turn this around in my head a couple times until I came up with the current solution. No proof that it's correct.

### Notes

I noticed that the existing tests in `ArrayHelperTests` were wrongly using `Is.EquivalentTo` (ignores element ordering) when they really should be using `Is.EqualTo`. Fixed that. Means that the existing test for `MoveSlice` was bogus.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UserTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UserTests.cs
@@ -1177,6 +1177,33 @@ internal class UserTests : CoreTestsFixture
         Assert.That(user.pairedDevices, Is.EquivalentTo(new[] { gamepad }));
     }
 
+    // https://fogbugz.unity3d.com/f/cases/1327628/
+    [Test]
+    [Category("Users")]
+    public void Users_WhenAddingDevicesToUsers_PairedDevicesOfExistingUsersAreUnaffected()
+    {
+        var user1 = InputUser.CreateUserWithoutPairedDevices();
+
+        var user1pad1 = InputSystem.AddDevice<Gamepad>("user1pad1");
+        var user1pad2 = InputSystem.AddDevice<Gamepad>("user1pad2");
+
+        InputUser.PerformPairingWithDevice(user1pad1, user1);
+        InputUser.PerformPairingWithDevice(user1pad2, user1);
+
+        var user2pad = InputSystem.AddDevice<Gamepad>("user2pad");
+        var user2 = InputUser.PerformPairingWithDevice(user2pad);
+
+        var user3pad = InputSystem.AddDevice<Gamepad>("user3pad");
+        var user3 = InputUser.PerformPairingWithDevice(user3pad);
+
+        var user1pad3 = InputSystem.AddDevice<Gamepad>("user1pad3");
+        InputUser.PerformPairingWithDevice(user1pad3, user1);
+
+        Assert.That(user1.pairedDevices, Is.EquivalentTo(new[] { user1pad1, user1pad2, user1pad3 }));
+        Assert.That(user2.pairedDevices, Is.EquivalentTo(new[] { user2pad }));
+        Assert.That(user3.pairedDevices, Is.EquivalentTo(new[] { user3pad }));
+    }
+
     [Test]
     [Category("Users")]
     [Ignore("TODO")]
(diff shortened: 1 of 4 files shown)


andrew-oc (approved):
This looks like a solid implementation but I'm wondering if it's solving a problem we don't really have. The only place that MoveSlice is used is when adding devices to users right? And the bit of logic it implements there only ever moves a slice to the end of the array i.e. it doesn't have to deal with wrap around. So I'll approve the PR because it looks like it fixes the bug, but I'll suggest that we actually kill MoveSlice in this form in this PR and implement MoveSliceToEnd or something similar on the basis that it will be simpler code to maintain and will be more performant. Of course if I'm wrong about how AddDeviceToUser actually works, feel free to ignore this. That code is a bit gnarly!

Rene-Damm:
> implement MoveSliceToEnd or something similar on the basis that it will be simpler code to maintain

Not sure it'll win us that much. Though I may totally be missing a trivially simple way to do this. In my head the "move slice to end" corresponds to the main path handled by `MoveSlice`. Elements may need to be shifted down while others need to be appended. Doing that with only single element temporary storage seems to pretty much arrive at the same code. But... maybe I'm blinded to the obvious here?

andrew-oc:
Yeah, you're right. Not really a performance win, but simplifying the code and making the code in AddDeviceToUser more intention revealing (make it obvious that it's just moving the slice to the end) are still wins.

Rene-Damm:
Gah, saw the comment before the edit and for a moment was hoping "haha, there *is* a much simpler way to do this!" Tried it out locally and probably came to the same conclusion you did.

Simplification RE the intent of `AddDeviceToUser` makes sense. In this case, given the implementation of the moving code will come out looking kinda identical, think I'll give up on this one.

andrew-oc:
Actually, wish I'd left the comment in now:) You can drop destinationIndex arg and then it's just

```c#
	        var arrayLength = array.Length - 1;
	        for (int i = count - 1; i >= 0; i--)
	        {
		        Swap(ref array[sourceIndex + i], ref array[arrayLength - count + i]);
	        }
```
But yeah, not a huge win

Rene-Damm:
> Actually, wish I'd left the comment in now:) You can drop destinationIndex arg and then it's just
> 
> ```cs
> 	        var arrayLength = array.Length - 1;
> 	        for (int i = count - 1; i >= 0; i--)
> 	        {
> 		        Swap(ref array[sourceIndex + i], ref array[arrayLength - count + i]);
> 	        }
> ```
> 
> But yeah, not a huge win

Wait, that doesn't seem to work. If we take

```
-------------------------
| 1 | 2 | 3 | 4 | 5 | 6 |
-------------------------
```

and say we want to move two elements at index 1 to the end. Then the loop above would do two swaps (`count` - 1) and give

```
-------------------------
| 1 | 5 | 6 | 4 | 2 | 3 |
-------------------------
```

andrew-oc:
Well it was just an example :) I mean what you had in the first if branch is basically the easiest way but you can do it without the destination index if it was a 'move to end' style thing.

Rene-Damm:
Ah heh, I see. Misunderstood :)

PR #1333, opened 2021-05-17
CHANGE: Register controls using static type references

andrew-oc:
### Description

Made some changes to enable control and device registrations be made using statically typed references to the control and state types. This should enable us in a later version to remove the Preserve attributes on all device types. Testing shows that all of the preserved types are adding about 60k to the size of a built IL2CPP player. So this PR doesn't do much to address that directly, but it does make control layout creation a little more efficient

### Changes made

Added a new method to InputManager to register controls that provide their own methods for creating InputControlLayouts. The method is called RegisterControlLayout<TControl>. The InputControlLayout creation method can be passed as a method group to be called later when a device is created. If no creation method is passed, a default one is inserted that creates a default InputControlLayout instance for the control being registered.

### Notes

* The new RegisterControlLayout<TControl> API is not exposed publicly yet intentionally while I wait and see if it was a good idea or not.
* Although this task was ostensibly about removing reflection use, the new API still uses reflection to build the list of child controls because testing this against the FromJson code path showed reflection to be about 4 times faster! The new path should use less memory though as it doesn't instantiate 'state' types to read the FourCC from them, and doesn't use Activator.CreateInstance anywhere. This seemed like the better route given the other two alternatives i.e. loading layouts from serialized json and loading from precompiled layouts. The former is much slower, and the latter adds thousands of lines of code to the code base.
* InputControlLayout constructor is now internal so controls and devices shipping with the framework can call it directly but this wouldn't be available for users.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
@@ -924,6 +924,16 @@ namespace UnityEngine.InputSystem.Layouts
             return layout;
         }
 
+        public void BuildChildControlsFrom<T>()
+        {
+            m_Controls = GetControlItems(typeof(T), name).ToArray();
+        }
+
+        public void SetStateFormat(FourCC format)
+        {
+            m_StateFormat = format;
+        }
+
         public string ToJson()
         {
             var layout = LayoutJson.FromLayout(this);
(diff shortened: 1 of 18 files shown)


Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -924,6 +924,16 @@ public static InputControlLayout FromType(string name, Type type)
             return layout;
         }
 
+        public void BuildChildControlsFrom<T>()
```
Rene-Damm: internal?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -924,6 +924,16 @@ public static InputControlLayout FromType(string name, Type type)
...
+        {
+            m_Controls = GetControlItems(typeof(T), name).ToArray();
+        }
+
+        public void SetStateFormat(FourCC format)
```
Rene-Damm: internal?

Packages/com.unity.inputsystem/InputSystem/Devices/Gamepad.cs
```diff
@@ -713,6 +713,13 @@ public virtual void SetMotorSpeeds(float lowFrequency, float highFrequency)
             m_Rumble.SetMotorSpeeds(this, lowFrequency, highFrequency);
         }
 
+        public static InputControlLayout CreateLayout(string name)
```
Rene-Damm: Shouldn't all the `CreateLayout` methods be internal?

Rene-Damm (commented):
Some random comments. Conversation ongoing.

andrew-oc:
Yep, all of those look like they should be internal. Not sure if I had some plan there, but since none of this is exposed to user code anyway, yep, fixed.

andrew-oc:
Closing this for now. It doesn't really add much value until we can delete the Preserve attributes, and the breaking changes we make doing that will probably make this PR obsolete.

PR #1335, opened 2021-05-18
FIX: Callback ordering issues when adding/removing during execution (case 1322530).

Rene-Damm:
Fixes [1322530](https://issuetracker.unity3d.com/issues/inputsystems-events-are-not-called-the-order-they-were-added-when-they-are-modified-in-the-middle-of-the-call-by-other-listener) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1322530/)).

### Description

Adding and removing callbacks from within callbacks isn't handled properly in the code. Some stuff in `DelegateHelpers` but falls over sideways quickly. Has come up a number of times in reports. There was a user PR about this, too (wasn't the right fix).

### Changes made

Bit the bullet and added a `CallbackArray` version that made a copy of the list before going through callbacks.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1055,8 +1002,7 @@ private void NotifyUsageChanged(InputDevice device)
             InputActionState.OnDeviceChange(device, InputDeviceChange.UsageChanged);
 
             // Notify listeners.
-            for (var i = 0; i < m_DeviceChangeListeners.length; ++i)
-                m_DeviceChangeListeners[i](device, InputDeviceChange.UsageChanged);
+            DelegateHelpers.InvokeCallbacksSafe(ref m_DeviceChangeListeners, device, InputDeviceChange.UsageChanged, "InputSystem.onDeviceChange");
```
andrew-oc, line 1006: I know the last argument here is just for logging and profiling, but is it right? They're all onDeviceChange?

andrew-oc (commented):
A neat solution. I'm wondering about the PrepareExecution method and paying the cost of the array copy (potentially, depending on number of events) whether the delegates are modified or not. Did you consider as an alternative changing CallbackArray so it sets a flag when delegates are being enumerated and any delegates that are added or removed while the flag is active get added to a different buffer maintained by CallbackArray. When enumeration finishes, any additions or removals get added into the main array. Basically you only pay the small performance hit when you actually change the delegates.

Rene-Damm:
> A neat solution. I'm wondering about the PrepareExecution method and paying the cost of the array copy (potentially, depending on number of events) whether the delegates are modified or not. Did you consider as an alternative changing CallbackArray so it sets a flag when delegates are being enumerated and any delegates that are added or removed while the flag is active get added to a different buffer maintained by CallbackArray. When enumeration finishes, any additions or removals get added into the main array. Basically you only pay the small performance hit when you actually change the delegates.

Did consider it but for some reason deemed it the worse option. Actually tried it out now and doesn't look that bad. Let me know what you think.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: Yup, these are correct. It's all just different InputDeviceChange enum values going out on the same `onDeviceChange` API endpoint.

andrew-oc (approved):

PR #1340, opened 2021-05-21
CHANGE: Added isEnabledByDefault constructor option to generated action sets.

Walter-Hulsebos:
### Description

Changed code generation of actions sets so they now have an optional `isEnabledByDefault` parameter and check in it's constructor.
It will then automatically call `.Enable()` at the end of the constructor if set to true.

It's handy to have to ability to enable and disable all InputActionMaps.
But there should be an option to have it enabled by default - for when you don't need the control and want to minimize code bloat.

### Changes made

- Added optional `isEnabledByDefault` parameter to generated action set's C# class.
- Added call to `.Enable()` at end of generated constructor when isEnabledByDefault is **true**.
- Re-generated existing actions sets to implement the new functionality.

### Notes

- This requires no change to existing API apart from re-generating existing action sets.
The constructor is functionally the same and sending the parameter is purely optional and defaults to false.

- When using `(isEnabledByDefault: true)` it means you're restricted to  _where/when_ you're creating the action set; for instance you can't create it in `Reset()` anymore because it'll call .Enable() at a time it can't/shouldn't.

### Checklist

Before review:

- [x] Changelog entry added.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Samples/InGameHints/InGameHintsActions.cs
+++ b/Assets/Samples/InGameHints/InGameHintsActions.cs
@@ -17,10 +17,10 @@ using UnityEngine.InputSystem.Utilities;
 
 namespace UnityEngine.InputSystem.Samples.InGameHints
 {
-    public partial class @InGameHintsActions: IInputActionCollection2, IDisposable
+    public partial class @InGameHintsActions : IInputActionCollection2, IDisposable
     {
         public InputActionAsset asset { get; }
-        public @InGameHintsActions()
+        public @InGameHintsActions(bool isEnabledByDefault = false)
         {
             asset = InputActionAsset.FromJson(@"{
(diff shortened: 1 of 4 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1340) <br/>All committers have signed the CLA.

ekcoh:
If work is resumed on this PR it must be updated with latest changes from develop to avoid causing CI issues.

PR #1344, opened 2021-05-27
FIX: PlayerInputManager join action not triggering

andrew-oc:
### Description

Fixes  [1260625](https://issuetracker.unity3d.com/product/unity/issues/guid/1260625) ([Fogbugz](https://fogbugz.unity3d.com/f/cases/1260625/))

This fixes the following case. An input action asset defines an action called Fire that uses the space bar key and the gamepad south button. The PlayerInputManager join action is a reference to the Fire action. The first player hits the space bar key. This associates the keyboard with that user, and *disables input events from the gamepad* for the current action map. The second player now tries to join by pressing the gamepad south button, but that input is no longer active, so they can't join. This only affects PlayerInputManager when:
  * join behaviour is set to Join Players When Join Action Is Triggered
  * the join action is a reference

### Changes made

PlayerInputManager now always clones the input action asset if the join action is a reference.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Rene-Damm (approved):
Change LGTM.

API doc updating for the change would be great. `joinAction` unfortunately has nothing ATM. Would be great to add something and mention the cloning behavior in there.

PR #1346, opened 2021-06-02
FIX: Setting motor speeds and light bar color in quick succession on DualShock 4

andrew-oc:
### Description

[case 1271119](https://issuetracker.unity3d.com/product/unity/issues/guid/1271119/) ([Fogbugz](https://fogbugz.unity3d.com/f/cases/1271119))

This commit fixes the issue with calling SetMotorSpeeds followed by SetLightBarColor in quick succession on a DualShock 4 controller. The IOCTL interface currently only allows a single command at a time to be processed on the input device, so the second of these calls will usually fail. The fix contained here is simply to expose an extra method on the DualShock 4 interface called SetMotorSpeedsAndLightBarColor that combines both instructions into a single IOCTL command, but a more robust fix will be added later that deals with this issue universally at a lower level in the InputDevice hierarchy.

### Changes made

Added SetMotorSpeedsAndLightBarColor method to DualshockGamepadHID to set both haptics in one device command.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
```diff
@@ -292,6 +292,38 @@ public override void SetMotorSpeeds(float lowFrequency, float highFrequency)
...
+        /// <remarks>
+        /// Use this method to set both the motor speeds and the light bar color in the same call. This method exists
+        /// because it is currently not possible to process an input/output control (IOCTL) command while another one
+        /// is in flight. For example, calling <see cref="SetMotorSpeeds"/> immediately after calling
+        /// <see cref="SetLightBarColor"/> might result in only the light bar color changing. The <see cref="SetMotorSpeeds"/>
+        /// call could fail. It is however possible to combine multiple IOCTL instructions into a single command, which
+        /// is what this method does.
+        ///
+        /// See <see cref="Haptics.IDualMotorRumble.SetMotorSpeeds"/> and <see cref="IDualShockHaptics.SetLightBarColor"/>
+        /// for the respective documentation regarding setting rumble and light bar color.</remarks>
```
Rene-Damm, line 311: Probably good idea to add some short blurbs to the docs for both `SetMotorSpeeds` and `SetLightBarColor` about the limitation in call frequency.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ however, it has to be formatted properly to pass verification tests.
 
 - Fixed pairing devices to existing `InputUser`s potentially corrupting list of paired devices from other `InputUser`s ([case 1327628](https://issuetracker.unity3d.com/issues/input-system-devices-are-reassigned-to-the-wrong-users-after-adding-a-new-device)).
 - Fixed StackOverflowException caused by calling InputSystem.Update from inside an input action callback ([case 1316000](https://issuetracker.unity3d.com/issues/crash-when-adding-inputsystem-dot-update-to-inputsystem-command-handler-to-force-processing-an-event-and-sending-input)).
+- Fixed setting both the light bar and motor speeds simultaneously on a DualShock 4 controller ([case 1271119](https://fogbugz.unity3d.com/f/cases/1271119)).
```
Rene-Damm: Think we haven't actually really fixed the problem but rather provided a workaround. If you call the two separate methods independently, you still run into the same problem.

Should we move this to `Added` and say something like "Added workaround to ...."?

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
andrew-oc: Added some docs to the dualshock section.

Packages/com.unity.inputsystem/CHANGELOG.md
andrew-oc: Good call

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
```diff
@@ -292,6 +292,38 @@ public override void SetMotorSpeeds(float lowFrequency, float highFrequency)
...
+        public bool SetMotorSpeedsAndLightBarColor(float lowFrequency, float highFrequency, Color color)
+        {
+            var command = DualShockHIDOutputReport.Create();
+            command.SetMotorSpeeds(lowFrequency, highFrequency);
+            command.SetColor(color);
+
+            var result = ExecuteCommand(ref command);
```
ekcoh, line 318: Looks good to me, but out of curiosity, will the execute command be queuing these commands when used like this (two actions on the same command) to avoid the problem with hw/interface not supporting both light color and motor speed being modulated at the same time? @andrew-oc

ekcoh (approved):

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
andrew-oc: @ekcoh No, device commands are ultimately just bit masks, so it's possible to combine multiple instructions into a single one. This fix just sends one command with the set light bar and rumble bits both set.

PR #1349, opened 2021-06-04
FIX: InvalidOperationException thrown when selecting a preset created from an InputActionImporter

andrew-oc:
[case 1199544](https://issuetracker.unity3d.com/issues/input-system-properties-are-not-visible-and-invalidoperationexception-is-thrown-on-selecting-inputactionimporter-preset-asset) ([Fogbugz](https://fogbugz.unity3d.com/f/cases/1199544/))

### Description

The InputActionImporterEditor expected the assetTarget field to be set i.e. that the currently selected asset was an InputActionAsset, but when selecting a preset asset, this wasn't the case because presets don't save asset references. 

### Changes made

Added a warning message to the InputActionImporterEditor when the selected object is not an InputActionAsset and disabled the edit button in this case.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ however, it has to be formatted properly to pass verification tests.
 
 - Fixed pairing devices to existing `InputUser`s potentially corrupting list of paired devices from other `InputUser`s ([case 1327628](https://issuetracker.unity3d.com/issues/input-system-devices-are-reassigned-to-the-wrong-users-after-adding-a-new-device)).
 - Fixed StackOverflowException caused by calling InputSystem.Update from inside an input action callback ([case 1316000](https://issuetracker.unity3d.com/issues/crash-when-adding-inputsystem-dot-update-to-inputsystem-command-handler-to-force-processing-an-event-and-sending-input)).
+- Fixed InvalidOperationException when opening a preset created from an InputActionImporter ([case 1199544](https://issuetracker.unity3d.com/issues/input-system-properties-are-not-visible-and-invalidoperationexception-is-thrown-on-selecting-inputactionimporter-preset-asset)).
```
Rene-Damm: Minor... `InputActionImporter` is an internal class. "preset create from an .inputactions asset" probably clearer.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
```diff
@@ -235,6 +247,9 @@ private void OnDestroy()
         // Set asset would usually only be called when the window is open
         private void SetAsset(InputActionAsset asset)
         {
+            if (asset == null)
+                return;
```
Rene-Damm, line 259: This case looks malformed. .inputactions editor windows with no associated assets should get nuked, it seems. What's the flow where we run into this if?

andrew-oc: The OpenEditor method, which calls SetAsset, is public and takes the asset as an argument, so it seems better to check it for null. I didn't see that the editor window gets blown away if there's no asset. I was definitely able to open one with this being null.

ekcoh (approved):
Changes looks good to me.

PR #1350, opened 2021-06-09
XR define changes

mattdatwork:
### Description

Cleans up XR Defines so that no platform specific defines are resident in the input system, and we can compile in or out support for XR using project library includes, in this case `com.unity.modules.xr` 

we would like to move this to `com.unity.xr.management` in the future when this is default included for XR. 
`com.unity.modules.xr` is on by default in new projects. 

This will compile in code that was previously compiled out based on hardcoded platform defines. Specifically the call to ` XR.XRSupport.Initialize();` This function registers the XR Layout Builder, the `Pose` , `Bone` and `Eyes` types, and the base `XRHMD` and `XRController` layouts. The inclusion of platform specific (ie: WMR/Oculus) layouts are still gated on `ENABLE_VR` which is defined per platform in the global defines jam files.

A new define `UNITY_FORCE_INPUTSYSTEM_XR_OFF ` is provided so that we can exclude compilation on a per platform basis should including these types cause compile errors. Its hoped we can remove this in the future once we've verified this is ok everywhere.
 
### Changes made

1. removes ENABLE_XR_COMBINED_DEFINE and associated platform specific defines from InputSystem.cs and 
2. moves tests which were using ENABLE_XR_COMBINED_DEFINE to be ifdef'd around UNITY_EDITOR, as they are testing editor functionality
3. adds UNITY_FORCE_INPUTSYSTEM_XR_OFF define which an be used to compile out support for XR when using the new input system. 
4. changes UNITY_INPUT_SYSTEM_ENABLE_XR to UNITY_XR_AVAILABLE, which better outlines that XR support is available. UNITY_XR_AVAILABLE is true when `com.unity.modules.xr` is in the project manifest this on by default in new projects. 

### Notes

Testing process:

Used OpenXR Sample Project which uses the new input system to show controllers and HMD positions. 
compiled on pc / android / xbox one / tvos / webgl with, and without UNITY_FORCE_INPUTSYSTEM_XR_OFF. 

platforms which support XR (eg:  quest2 on android) were verified that the input system was enabled and disabled correctly.
platforms which do not support XR (xbox1, tvos) were verified that the project including calls to XR functionality compiled correctly in both cases.  (and webgl, which technically has ENABLE_VR true, but I have no way to test this)

the input test project was loaded, and the tests were run to confirm that everything worked as expected.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1350) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1350) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: Dmytro Ivanov<br/>:x: MattDalby<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1350) it.</sub>

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRLayoutBuilder.cs
```diff
@@ -1,4 +1,4 @@
-#if UNITY_INPUT_SYSTEM_ENABLE_XR && ENABLE_VR
```
jimon, line 1: Some places have `UNITY_FORCE_INPUTSYSTEM_XR_OFF` and some don't, is there a bigger picture where it's needed? or should it be in all places?

jimon (approved):

tdmowrer (commented):
From a look at the code changes, it doesn't look like these defines would directly affect AR Foundation, though there are always knock on effects by enabling/disabling code paths that should be tested.

Who is meant to define `UNITY_FORCE_INPUTSYSTEM_XR_OFF`? Is it provided so that the user can define it in their project? If so, how is it documented?

Assets/Tests/InputSystem/Editor/EditModeTests.cs
```diff
@@ -52,12 +48,12 @@ public void Reset()
     [TestCase(InputSettings.UpdateMode.ProcessEventsManually, InputUpdateType.Manual)]
     [TestCase(InputSettings.UpdateMode.ProcessEventsInDynamicUpdate, InputUpdateType.Dynamic)]
     [TestCase(InputSettings.UpdateMode.ProcessEventsInFixedUpdate, InputUpdateType.Fixed)]
-    #if !ENABLE_XR_COMBINED_DEFINE
-    [Ignore("Must be on an XR-supported platform to run this test.")]
-    #endif
+#if !UNITY_EDITOR
```
StayTalm: I don't think this is needed. This is an 'edit mode' test, in an assembly that's only loaded on the editor platform. I know you passed this by me, I forgot about the asmdef/editor test

https://github.com/Unity-Technologies/InputSystem/blob/develop/Assets/Tests/InputSystem/Editor/Unity.InputSystem.Tests.Editor.asmdef

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -2645,7 +2641,7 @@ internal static void Update(InputUpdateType updateType)
             }
         }
 
-#if UNITY_EDITOR && ENABLE_XR_COMBINED_DEFINE
+#if UNITY_EDITOR
```
StayTalm: @jimon
Pointing Dmytro this way. I landed this runUpdatesInEditMode property on the condition that it's something we use specifically for XR for now, and can expand it to other platforms if needed, and the feature is more fully tested for those platforms.

I think this should still have some kind of XR guard.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/GoogleVR.cs
```diff
@@ -1,4 +1,4 @@
-#if UNITY_INPUT_SYSTEM_ENABLE_XR && ENABLE_VR && !DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR && !PACKAGE_DOCS_GENERATION
+#if UNITY_XR_AVAILABLE && !DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR && !PACKAGE_DOCS_GENERATION && !UNITY_FORCE_INPUTSYSTEM_XR_OFF
```
StayTalm: These files for specific devices and their defines (e.g.: DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR) all pass the two year LTS mark. These can be safely removed, as all supported XR.SDK packages have device layouts in them.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
```diff
@@ -317,6 +318,7 @@ public static void Initialize()
 
             InputSystem.onFindLayoutForDevice += XRLayoutBuilder.OnFindLayoutForDevice;
 
+#if ENABLE_VR
 #if !DISABLE_BUILTIN_INPUT_SYSTEM_WINDOWSMR
```
StayTalm, line 321: These should all be safe to delete by now.  They pass the two year mark since all XR.SDK packages include their own layouts.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
```diff
@@ -317,6 +318,7 @@ public static void Initialize()
 
             InputSystem.onFindLayoutForDevice += XRLayoutBuilder.OnFindLayoutForDevice;
 
+#if ENABLE_VR
```
StayTalm, line 320: Why is this one ENABLE_VR, where all other checks are for UNITY_XR_AVAILABLE?

StayTalm (changes requested):
Some questions on defines, some extra info on what can now be safe to be deleted,
Main change needed is to XR-guard runUpdatesInEditMode, unless @jimon is ok without it.

bryandube (approved):

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
mattdatwork: the trick with this one is that we currently have no XR specific defines here other than ENABLE_VR which will be 1 in the editor anyway. there's no asmdef for the test project that i can see either. could add one? @jimon ?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/GoogleVR.cs
mattdatwork: deleted them all. :)

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
mattdatwork: removed with layouts.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
mattdatwork: removed with layouts

mattdatwork:
> From a look at the code changes, it doesn't look like these defines would directly affect AR Foundation, though there are always knock on effects by enabling/disabling code paths that should be tested.
> 
> Who is meant to define `UNITY_FORCE_INPUTSYSTEM_XR_OFF`? Is it provided so that the user can define it in their project? If so, how is it documented?

agreed. i'll figure out how to document this. @jimon / @Rene-Damm any ideas? is there a spot in the docs I can update.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
jimon: Yes true, this functionality should stay XR-only for near future, as running-in-editor was not properly tested on our side, we were thinking to give it a deeper pass and then enable as a proper feature.

@mattdatwork are XR modules always on in the editor? anything else we can use? I assume editor is not in XR mode by default.

StayTalm: What about adding it if XR Management is included

thep3000 (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRLayoutBuilder.cs
mattdatwork: think we've resolved that now.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
mattdatwork: editor is ENABLE_VR by default. which is why I left it on. 
the trick here is that theres no current _test_ asmdef. so we'd need to add one to put in any compile defines for the tests.

StayTalm: The condition here is **NOT** tests.
The condition is XR systems.  XR Management is the candidate I suggest.

mattdatwork: <facepalm> right. sorry thought this was the test file. we can use UNITY_XR_AVAILABLE then. which atm is the modules, but will eventually be xr mgt?

StayTalm (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
```diff
@@ -316,108 +317,8 @@ public static void Initialize()
             InputSystem.RegisterLayout<XRController>();
 
             InputSystem.onFindLayoutForDevice += XRLayoutBuilder.OnFindLayoutForDevice;
-
```
Rene-Damm, line 319: Discussed this with the team. In the end, agreed we should leave the registrations in as long as we have the device classes still around in the package. Both should be nuked together.

ATM, if you happen to have a project that does indeed reference these classes, then with the change here, you wouldn't get a compile error but the project likely wouldn't work either.

Think the chances of this making a difference are slim. In practice, we likely wouldn't break anyone by just nuking the whole thing now. But for the sake of limiting breakage, let's just keep it and then remove it en bloc when we break the API.

Rene-Damm (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
mattdatwork: I have a tested change that does remove the define. however that define is a manual "I absolutely need everything off" and isnt intended to be used outside a major problem, in which case the define doesnt quite do what it's supposed to do.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/GoogleVR.cs
```diff
@@ -1,4 +1,4 @@
-#if UNITY_INPUT_SYSTEM_ENABLE_XR && ENABLE_VR && !DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR && !PACKAGE_DOCS_GENERATION
+#if UNITY_XR_AVAILABLE && ENABLE_VR && !DISABLE_BUILTIN_INPUT_SYSTEM_GOOGLEVR && !PACKAGE_DOCS_GENERATION && !UNITY_XR_FORCE_OFF
```
jimon: Should this be `UNITY_XR_FORCE_OFF` or `UNITY_FORCE_INPUTSYSTEM_XR_OFF`?

PR #1358, opened 2021-06-17
FIX: 'Pointer should have exited' error on screen rotation.

Rene-Damm:
### Description

When rotating the screen on mobile, the UI would log a `Pointer should have exited all objects before removing` error. Didn't really impact functionality but it's annoying.

The root cause of this is that `InputSystemUIInputModule` will blindly go and create a pointer state for *any* activity on the pointer-related `InputAction`s. So, even if we just *cancel* all those actions because we are re-resolving bindings, `InputSystemUIInputModule` goes and creates pointer states.

### Changes made

There are two separate changes in this PR.

#### 1. No new PointerModel states when an action cancels

A simple change to `InputSystemUIInputModule.OnPoint` and related callbacks to make sure we are not creating new pointers when the actions are indeed not giving us any input.

#### 2. Initial State Checks

This one is less obvious.

When we disable and re-enable actions, we need to re-sync UI state. Actions do have a mechanism called "initial state check" that is available to `Value` type actions. However, the UI uses `Pass-Through` type actions, throughout.

This change adds the ability for `Button` and `Pass-Through` actions to manually enable this check. If enabled, the next input update will check the bound controls over those actions for their current value and trigger actions as needed.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -709,3 +709,7 @@ If you don't want your Action to perform disambiguation, you can set your Action
 Actions with the type set to [Value](Actions.md#value) perform an initial state check when they are first enabled, to check the current state of any bound Control, and to set the Action's value to the highest value of any bound Control.
 
 Actions with the type set to [Button](Actions.md#button) don't perform any initial state check, so that only buttons pressed after the Action was enabled have any effect on the Action.
+
+For both [Button](Actions.md#button) and [Pass-Through](Actions.md#pass-through) actions, initial state checks can be enabled explicitly in the action editor or by setting [`InputAction.wantsInitialStateCheck`](../api/UnityEngine.InputSystem.InputAction.html#UnityEngine_InputSystem_InputAction_wantsInitialStateCheck) to true.
```
ekcoh: Good that the added check functionality is referenced here, but maybe elaborate or link/ref to what "initial state checks" actually means. At least to me, it sounds a bit abstract, what "state" are we talking about, active state, pressed state? (Or did I overlooked where this is explained?)

Packages/com.unity.inputsystem/Documentation~/Pointers.md
```diff
@@ -14,7 +14,7 @@ Each of these types implements a common set of Controls. For a more detailed des
 |Control|Type|Description|
 |-------|----|-----------|
 |[`position`](../api/UnityEngine.InputSystem.Pointer.html#UnityEngine_InputSystem_Pointer_position)|[`Vector2Control`](../api/UnityEngine.InputSystem.Controls.Vector2Control.html)|The current pointer coordinates in window space.|
-|[`delta`](../api/UnityEngine.InputSystem.Pointer.html#UnityEngine_InputSystem_Pointer_delta)|[`Vector2Control`](../api/UnityEngine.InputSystem.Controls.Vector2Control.html)|The difference in `position` since the last frame.|
+|[`delta`](../api/UnityEngine.InputSystem.Pointer.html#UnityEngine_InputSystem_Pointer_delta)|[`Vector2Control`](../api/UnityEngine.InputSystem.Controls.Vector2Control.html)|Motion delta in pixels for the current frame/update. Resets to `(0,0)` every frame. Within a frame, accumulates. Meaning that if an input update contains several events for a pointer, all deltas from all events are aggregated.<br><br>Note that the resolution of deltas depends on the specific and/or platform.|
```
ekcoh: Minor, but phrasing suggestion to emphasize what happens more direct: "Provides motion delta in pixels accumulated (summed) over the duration of the current frame/update. Resets to `(0,0)` each frame. Note that the resolution of deltas depends on the specific platform."

You decide the best way to formulate this, but just wanted to provide an alternative suggestion. If you rather stick with the previous it seems like the and/or platform, there is only the platform mentioned?

ekcoh (approved):
Looks good to me (without fully grasping the impact of this PR). Some minor feedback on documentation.

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
Rene-Damm: Good point. Reworded the entire section. Hopefully clearer now.

Packages/com.unity.inputsystem/Documentation~/Pointers.md
Rene-Damm: Done.

andrew-oc (approved):

ekcoh:
Looked through changes again @Rene-Damm looks good. Confirmed that tests are also green locally using editor test run. Confirm that tests are also green (exit code 0) when run locally using UTR.

PR #1359, opened 2021-06-29
FIX: Incorrect usage of IndexOfReference

ITR13:
### Description

In InputUser.OnDeviceChange IndexOfReference is used to search through s_AllLostDevices and s_AllPairedDevices, if any device is found it will search again from the found index.
The fourth argument of IndexOfReference should be the remaining amount of indices to look through, but in this method the total count was passed. Since the array starts at 0 it is safe to just subtract the starting index from the total count.

### Changes made

Subtracted new starting index from total count. 

### Notes

I have also checked through all other uses of this method in the project, and could not find any other incorrect usages.
When a device is removed it currently loops through the entire list again whenever it finds a device, this could potentially be shortened by passing in `deviceIndex` (without +1 since it's removed) and `s_AllPairedDeviceCount - deviceIndex`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
@@ -1642,7 +1642,7 @@ namespace UnityEngine.InputSystem.Users
 
                         // Search for another user who had lost the same device.
                         deviceIndex =
-                            s_AllLostDevices.IndexOfReference(device, deviceIndex + 1, s_AllLostDeviceCount);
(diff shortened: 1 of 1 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1359) <br/>All committers have signed the CLA.

ekcoh:
Thanks for your contribution @ITR13. Your contribution is routed via https://github.com/Unity-Technologies/InputSystem/pull/1362. Your analysis of the problem and feedback is also highly appreciated.

PR #1362, opened 2021-07-06
FIX: Incorrect array indexing in InputUser.

ekcoh:
### Description

As highlighted in contribution by [Mikael Klages](https://github.com/Unity-Technologies/InputSystem/pull/1359) the array indexing in [InputUser.OnDeviceChange] is incorrect which is a valid statement. However in the proposed changes in contribution [1359](https://github.com/Unity-Technologies/InputSystem/pull/1359) it is suggested that remove functionality offset into the array, but this would better be avoided since the arrays iterated on are modified/shrinked as part of the loop. 

### Changes made

- As highlighted in original contribution [1359](https://github.com/Unity-Technologies/InputSystem/pull/1359) the code in handling of `InputDeviceChange.ConfigurationChanged` uses an incorrect range in the linear search. It has been modified to use a valid count and also to offset directly into the next user slice of devices since the action associated with this event is per user and should be executed twice for the same user (would only result in overhead).
- Added a clarification into handling of `InputDeviceChange.Added` event that no index offset is used since the array of devices iterated upon is lost devices which is modified inside the loop as part of `RemoveDeviceFromUser`. This is to avoid further confusion around indexing in this part of the code.

### Notes

- Extended test coverage of related ArrayHelpers.cs in ArrayHelpersTest.cs.
- Test case added to UserTests.cs: `Users_CanReacquireDevicesIfSingleDeviceIsPairedWithMultipleUsers()` which basically verifies that devices can be required but has been designed to trigger a scenario where incorrect offsets in ``InputDeviceChange.Added` or ``InputDeviceChange.Removed` would result in incorrect offsets.
- It was attempted during development of this PR to also provide test coverage of `InputDeviceChange.ConfigurationChanged` handler code but unfortunately source code seem to be non-testable in its current form since underlying triggers may not be mocked.

### Checklist

Before review:

- [x] Changelog entry added.
    - Functional tests `UserTests.cs#Users_CanReacquireDevicesIfSingleDeviceIsPairedWithMultipleUsers`, `ArrayHelpersTest.cs#Utilities_IndexOfReference__IsUsingReferenceEqualsAndConstrainedByStartIndexAndCount`.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UserTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UserTests.cs
@@ -1049,6 +1049,41 @@ internal class UserTests : CoreTestsFixture
         }));
     }
 
+    // Designed to cause indexing problems in InputUser code, see report/contrib:
+    // https://github.com/Unity-Technologies/InputSystem/pull/1359
+    [Test]
+    [Category("Users")]
+    public void Users_CanReacquireDevicesIfSingleDeviceIsPairedWithMultipleUsers()
+    {
+        // Arrange: Setup three users sharing a single device
+        var user1 = InputUser.CreateUserWithoutPairedDevices();
+        var user2 = InputUser.CreateUserWithoutPairedDevices();
+        var user3 = InputUser.CreateUserWithoutPairedDevices();
+
+        var keyboard = InputSystem.AddDevice<Keyboard>("keyboard");
+
+        InputUser.PerformPairingWithDevice(keyboard, user1);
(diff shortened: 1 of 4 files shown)


Rene-Damm (approved):
Awesome!

PR #1363, opened 2021-07-06
FIX: Submit and cancel UI actions now trigger on press instead of release

andrew-oc:
### Description

Submit and cancel UI actions fire on release, rather than press, so if the user holds the enter key, moves the selection to a new button, and then releases the enter key, the newly selected button receives the submit action. This was reported as a bug because it's easy to accidentally press the wrong button while quickly moving through a UI.

### Changes made

Submit and cancel actions now trigger on press.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -626,19 +626,19 @@ internal void ProcessNavigation(ref NavigationModel navigationState)
...
+                //       instead of *up*, so button *up* will come in *after* we have applied the state change.
                 var submitAction = m_SubmitAction?.action;
                 var cancelAction = m_CancelAction?.action;
 
                 var data = GetBaseEventData();
-                if (cancelAction != null && cancelAction.WasReleasedThisFrame())
+                if (cancelAction != null && cancelAction.WasPressedThisFrame())
                     ExecuteEvents.Execute(eventSystem.currentSelectedGameObject, data, ExecuteEvents.cancelHandler);
-                if (!data.used && submitAction != null && submitAction.WasReleasedThisFrame())
+                if (!data.used && submitAction != null && submitAction.WasPressedThisFrame())
```
Rene-Damm, line 644: Does the rebinding UI sample still work with this change? Does the rebinding triggered from UI still work even if it has the suppression of events turned off?

ekcoh (commented):
@Rene-Damm had a good point regarding sample that probably good to check. From my perspective @andrew-oc code looks good, but seems liek there's a conflict that needs to be resolved. A test case for the submit action would have been beneficial though for future regression.

Something not relevant for this review (code looks good to me but I might not yet see the full context) but for long term it might be beneficial to have automation tests running the samples to verify that they work at all times. This also reduces the need for duplicating that kind of code between samples and tests. From previous experiences "events/triggers" may be added to samples that make them easier to automate. Just food for thought for long-term....

Rene-Damm:
> Something not relevant for this review (code looks good to me but I might not yet see the full context) but for long term it might be beneficial to have automation tests running the samples to verify that they work at all times.

Now that you mention it, I remember that I actually did add a test back then (`Samples_RebindingUI_SuppressingEventsDoesNotInterfereWithUIInput`) which I think covers this -- or at least it refers to the original case that sparked the initial change here.

So, should be fine. Probably still worth manually verifying, though.

Overall, we do have a certain amount of automated test coverage for the samples. By far not for all and not for every aspect, though.

ekcoh:
> Overall, we do have a certain amount of automated test coverage for the samples. By far not for all and not for every aspect, though.

Awesome!

PR #1365, opened 2021-07-08
CHANGE: Touch OnPointerExit now deferred one frame (case 1347048).

Rene-Damm:
Fixes [1347048](https://issuetracker.unity3d.com/issues/input-system-ispointerovergameobject-returns-false-when-used-with-a-tap-interaction) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1347048/)).

### Description

`InputSystemUIInputModule` creates and removes pointers for touch input on the fly. So, finger goes down, pointer is created; finger moves around, pointer is updated; finger is lifted, pointer is removed. This is different from mouse and pen input where a pointer is created for the whole device and just sticks around.

This creates a problem with `IsPointerOverGameObject()` that relies on a) a pointer being there in the first place and b) it having a state of "over GameObject".

### Changes made

Pointer removal is now simply deferred one frame. So, the `OnPointerUp` event does happen in the same frame as before but the `OnPointerExit` now happens one frame later.

#### Potential for Regression

We now have pointer states that may "leak" into the next frame. If a platform reuses touch IDs, we may end up picking up a pointer record that before did not exist.

I believe this is not an issue. Even if a pointer record gets reused this way, the event sequence that we surface from the record should be unaffected.

#### Old Input System

I have not verified how this behaves in the old input system. I believe that in most but not all situations it would work as `StandaloneInputModule` uses `UnityEngine.Input.touches` which implicitly delays some touch endings by one frame. IIRC this only happens, however, if there's also a `Moved` in the same frame. In practice, that is likely to almost always be the case but in practice, it seems the same problem may occur with the old input system depending on frame vs input timing.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.cs
@@ -866,14 +866,14 @@ internal class UITests : CoreTestsFixture
         Assert.That(scene.leftChildReceiver.events[0].pointerData.pressPosition, Is.EqualTo(firstScreenPosition).Using(Vector2EqualityComparer.Instance));
         Assert.That(scene.leftChildReceiver.events[0].pointerData.clickTime, Is.EqualTo(clickTime));
         Assert.That(scene.leftChildReceiver.events[0].pointerData.clickCount, Is.EqualTo(clickCount));
-        Assert.That(scene.leftChildReceiver.events[0].pointerData.pointerEnter, Is.SameAs(isTouch ? null : scene.rightGameObject)); // Pointer-exit comes before pointer-up.
+        Assert.That(scene.leftChildReceiver.events[0].pointerData.pointerEnter, Is.SameAs(scene.rightGameObject)); // Pointer-exit comes before pointer-up.
         Assert.That(scene.leftChildReceiver.events[0].pointerData.pointerDrag, Is.SameAs(scene.leftGameObject));
         Assert.That(scene.leftChildReceiver.events[0].pointerData.pointerPress, Is.SameAs(scene.leftGameObject));
         Assert.That(scene.leftChildReceiver.events[0].pointerData.rawPointerPress, Is.SameAs(scene.leftGameObject));
         Assert.That(scene.leftChildReceiver.events[0].pointerData.lastPress, Is.Null);
         Assert.That(scene.leftChildReceiver.events[0].pointerData.dragging, Is.True);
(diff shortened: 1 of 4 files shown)


ekcoh (approved):
Code looks good to me based on what I was able to grasp. The only things that makes me wonder is whether the introduced delay will introduce inconsistencies or might break UI logic and/or tests where touch pointer UI is involved? It sounds like you do not think that would be a problem.

...realize now that its only the exit event that is affected if I understand you correctly and then its likely less critical. E.g. it would be a bit exotic to have logic based on touch exit that is dependent on 1 frame diff I guess.

Rene-Damm:
> Code looks good to me based on what I was able to grasp. The only things that makes me wonder is whether the introduced delay will introduce inconsistencies or might break UI logic and/or tests where touch pointer UI is involved? It sounds like you do not think that would be a problem.
> 
> ...realize now that its only the exit event that is affected if I understand you correctly and then its likely less critical. E.g. it would be a bit exotic to have logic based on touch exit that is dependent on 1 frame diff I guess.

Definitely can't say with certainty that it won't but yup, does seem unlikely. We just space out the event sequence from the same `PointerModel` state slightly different from before.

Main risk I see is us picking up the same `PointerModel` instance later and "reviving" it. But even there, I expect the outcome to be correct.

andrew-oc (approved):
Looks good. The only thing I'd suggest is to expand the comment [here](https://github.com/Unity-Technologies/InputSystem/pull/1365/files#diff-3b5d28ba317b2b6240ae00166a94786980b2126bde8a6d830b6c962581b44583R1887). It's not obvious from there why we don't process this if it was released this frame without doing a bunch of digging.

Rene-Damm:
> The only thing I'd suggest is to expand the comment [here](https://github.com/Unity-Technologies/InputSystem/pull/1365/files#diff-3b5d28ba317b2b6240ae00166a94786980b2126bde8a6d830b6c962581b44583R1887). It's not obvious from there why we don't process this if it was released this frame without doing a bunch of digging.

Good point. Fixed.

PR #1354, opened 2021-06-13
FIX: InputControlPathDrawer duplicated instances in lists

NibbleByte:
### Description
Using [InputControl] for string members (as control path) of classes that are in a list resulted in all entries having (seemingly) the same value in the Inspector.
This was caused due to a cache in the InputControlPathDrawer. It didn't account for reusing the property drawer instance in lists.

Repro steps:
1. Add this to any MonoBehaviour / ScriptableObject
`[InputControl]
public string[] BindingPath;`
2. Display the object in the Inspector. Add 3-4 entries.
3. Change the first entry
Result: all the other entries will display the same value.
Expected: Only first entry should be changed.

### Changes made
Removed the editor cache in the InputControlPathDrawer. I believe it won't have big performance hit.

### Notes
Haven't really tested other non-string usages of [InputControl]

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #1366, opened 2021-07-08
FIX: InputControlPathDrawer showing duplicate entries

andrew-oc:
User-contributed fix [#1354](https://github.com/Unity-Technologies/InputSystem/pull/1354)

### Description

Using [InputControl] for string members (as control path) of classes that are in a list resulted in all entries having (seemingly) the same value in the Inspector.
This was caused due to a cache in the InputControlPathDrawer. It didn't account for reusing the property drawer instance in lists.

Repro steps:

Add this to any MonoBehaviour / ScriptableObject
[InputControl] public string[] BindingPath;
Display the object in the Inspector. Add 3-4 entries.
Change the first entry
Result: all the other entries will display the same value.
Expected: Only first entry should be changed.

### Changes made

InputControlPathEditor now has a more stateless path where as much state as possible required for rendering is passed in as arguments to OnGUI.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 
 - Fixed pairing devices to existing `InputUser`s potentially corrupting list of paired devices from other `InputUser`s ([case 1327628](https://issuetracker.unity3d.com/issues/input-system-devices-are-reassigned-to-the-wrong-users-after-adding-a-new-device)).
+- Fixed `InputControlPathDrawer` drawing every entry in a collection of `InputControl`s as the first entry.
```
Rene-Damm: `InputControlPathDrawer` is an internal class. Better to describe in terms of what broke when you did what. Maybe something like
```
* Fixed inspector for arrays/lists of `[InputControl]` paths duplicating the value of the first element as the value of all others (...).
```

Or something like that.

Rene-Damm (approved):

ekcoh (approved):
Looks good to me, seems like one outstanding documentation feedback item from @Rene-Damm. Happy to see some valuable user-contributions making their way into future releases.

PR #1371, opened 2021-07-15
Compressing mouse move events (partially fixing 1281266)

jimon:
### Description

When using high frequency mouse too many move events are generated for us to handle, so processing time explodes to 1-2+ms (100+ms in worst cases).

### Changes made

This change _partially_ fixes the problem by combining all consecutive mouse move state events into the last one of them. This way if you have 100 mouse move events only last one will be present for rest of the code to process (move/scroll delta's are accumulated).

It accomplishes them by running event buffer through a utility bursted job that modifies states in place. Code comments hopefully contain enough information about specifics of the algorithm.

Due to us using burst for the first time I made escape hatches to completely bypass this change from user side just in case.

### Notes

- This change achieves approx 20x speedup, but cost of running a job is non zero, so speedup might be negligible or absent in some cases.

- If any mouse delta state events are present I currently disable the compression because it's tricky to support them when we modify data in-place (we can only shrink the buffer, never grow). None of real backends are using mouse delta state events, but our test fixture uses them quite a bit.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Events/CompressMouseMoveEvents.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Events/CompressMouseMoveEvents.cs
@@ -1,0 +1,257 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using Unity.Collections;
+using Unity.Collections.LowLevel.Unsafe;
+using Unity.Jobs;
+using Unity.Profiling;
+using UnityEngine.InputSystem.LowLevel;
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+#if !UNITY_INPUT_SYSTEM_DO_NOT_USE_BURST
+using Unity.Burst;
+#endif
+
+namespace UnityEngine.InputSystem
+{
+    // Utility to merges consecutive mouse move events into one.
+    // It operates directly on event buffer and does all modification in-place. It can only shrink event buffer.
+    // Use this define to completely disable any burst stuff, might be helpful if there are any issues with it.
+#if !UNITY_INPUT_SYSTEM_DO_NOT_USE_BURST
+    [BurstCompile(CompileSynchronously = true)]
+#endif
+    internal unsafe struct CompressMouseMoveEvents : IJob
+    {
+        private static ProfilerMarker s_ProfilerMarker = new ProfilerMarker("CompressMouseMoveEvents");
+
+        public static void ProcessEvents(InputUpdateType updateType, double processUntilTimestamp,
+            ref InputEventBuffer events)
+        {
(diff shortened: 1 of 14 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1371) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1371) before we can accept your contribution.<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1371) it.</sub>

Packages/com.unity.inputsystem/InputSystem/Events/CompressMouseMoveEvents.cs
```diff
@@ -0,0 +1,257 @@
...
+
+                if (updateType == InputUpdateType.Fixed && currentEvent->internalTime >= processUntilTimestamp)
+                {
+                    eventToProcessCount = i;
+                    break;
+                }
+            }
+            eventPtrs[eventToProcessCount] = (IntPtr)currentEvent;  // Remember the end pointer, will make life easier later.
+
+            // Step 2: compress move events in-place, mark redundant events for skipping
```
ekcoh, line 107: This seems like aggregation, maybe call this step aggregation since basically just summing events which have linearly additive components without losing information bits. Compression stage seems to come later.

ekcoh (approved):
20x speedup sounds good, but is this overall processing or on the function level. I would guess that the majority of processing cost would still be associated with lowest-level event dispatching, i.e. lots of non-batched events containing small amounts of data? Regardless, if this reduces processing overhead and/or latency, and not just moves the cost somewhere else its a good addition. Would have been interesting to have seen the diff in latency and/or total CPU time but maybe this is something that requires its own JIRA for future KPI monitoring / performance assesments.

ekcoh (changes requested):
Seems like branch is not building correctly and it probably needs to be looked into.

andrew-oc:
It was decided to go with the slightly slower but simpler (in terms of code) approach over [here](https://github.com/Unity-Technologies/InputSystem/pull/1380)

PR #1375, opened 2021-07-20
Tentative fix for auto switching property related to control schemes.

ekcoh:
### Description

This is a tentative fix for https://fogbugz.unity3d.com/f/cases/1240924/. What the case reports isn't really a bug but the behavior of auto-switching is not strictly defined. Created this pull request as an RFC since I am not really happy with the fix in this branch and I believe it would be more beneficial to revisit design/interface at a later point. 

### Changes made

If auto switch is on, (neverSwitchControllerScheme: false), and there is a single player, switching is enabled. If auto switch is off and defaultControllerScheme is null / any, an initial assignment to a viable controller scheme is granted. If auto switch is off and defaultControllerScheme is not null / any, only that controller scheme may be activated.

### Notes

Currently one test case is failing and would be happy for assistance on understanding exactly why this wouldn't be a valid case: XXX. I believe this might be related to device listener/monitoring tied to auto-switch functionality, but I might be wrong.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
@@ -781,6 +781,85 @@ internal class PlayerInputTests : CoreTestsFixture
         Assert.That(playerInput.devices, Is.EquivalentTo(new InputDevice[] { keyboard, mouse }));
     }
 
+    // Similar to PlayerInput_CanAutoSwitchControlSchemesInSinglePlayer() and 
+    // PlayerInput_AutoSwitchingControlSchemesInSinglePlayer_CanBeDisabled() but verifies the
+    // scenario when auto-switching is enabled after initial assignment attempt has taken
+    // place
+    // Related to https://fogbugz.unity3d.com/f/cases/1240924/
+    [Test]
+    [Category("PlayerInput")]
+    public void PlayerInput_CanAutoSwitchControlSchemesInSinglePlayer_IfChangingAutoSwitchAfterInitialActivation()
+    {
+        var gamepad = InputSystem.AddDevice<Gamepad>();
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        var mouse = InputSystem.AddDevice<Mouse>();
+
+        var go = new GameObject();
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -1500,6 +1507,8 @@ private bool TryToActivateControlScheme(InputControlScheme controlScheme)
                 return false;
             }
 
+            // Debug.Log($"Activated control scheme '{controlScheme.name}' for '{m_InputUser}'");
```
ekcoh, line 1510: This should be removed if PR is decided to be completed for merge

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -781,6 +781,85 @@ public void PlayerInput_AutoSwitchingControlSchemesInSinglePlayer_CanBeDisabled(
...
+
+        playerInput.neverAutoSwitchControlSchemes = false;
+        Press(gamepad.buttonSouth);
+
+        Assert.That(playerInput.devices, Is.EquivalentTo(new[] { gamepad }));
+        Assert.That(playerInput.user.controlScheme, Is.Not.Null);
+        Assert.That(playerInput.user.controlScheme.Value.name, Is.EqualTo("Gamepad"));
+        Assert.That(listener.messages, Is.EquivalentTo(new[]
+        {
+            ////TODO: reduce the steps in which PlayerInput updates the data to result in fewer re-resolves
```
ekcoh, line 816: Remove this TODO since there since test case was copy pasted and modified

ekcoh:
Decided that this fix is not fit for pre6

PR #1380, opened 2021-07-26
CHANGE: Improve performance when using high frequency mice

andrew-oc:
Alleviates the issues from [case 1281266](https://fogbugz.unity3d.com/f/cases/1281266/)
The original PR for this idea is [here](https://github.com/Unity-Technologies/InputSystem/pull/1371). The approach here is slower but simpler and more self contained in FastMouse.

### Description

When using high frequency mouse too many move events are generated for us to handle, so processing time explodes to 1-2+ms (100+ms in worst cases).

### Changes made

Modified FastMouse so that consecutive mouse move events are coalesced into a single device state update.  This way if you have 100 mouse move events only last one will be raised for the rest of the code to process (move/scroll delta's are accumulated).

### Notes

- Compression of mouse move events is disabled automatically when a composite binding of mouse position and any control from any other device is detected. This is because mouse compression happens inside FastMouse which only ever sees mouse events, so it can't detect multi-device data dependencies. 
- The user can turn compression back on manually after it has been disabled and that could result in missed callbacks.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
@@ -1,3 +1,5 @@
+using System;
+using System.Linq;
 using UnityEngine.InputSystem.Layouts;
 using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.InputSystem.Processors;
@@ -40,6 +42,16 @@ namespace UnityEngine.InputSystem
     /// <seealso cref="InputSystem.onSettingsChange"/>
     public partial class InputSettings : ScriptableObject
     {
+        public void Awake()
+        {
+            HookDisableMouseCompressionListener();
+        }
+
+        public void OnDisable()
+        {
+            UnhookDisableMouseCompressionListener();
+        }
+
         /// <summary>
         /// Determine how the input system updates, i.e. processes pending input events.
         /// </summary>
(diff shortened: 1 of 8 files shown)


Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -518,6 +530,119 @@ public ReadOnlyArray<string> supportedDevices
...
+                    var bindingPath = binding.path;
+                    if (string.Equals(bindingPath, "<mouse>/position", StringComparison.OrdinalIgnoreCase))
+                        mousePositionBindingFound = true;
+                    else if (!bindingPath.StartsWith("<mouse>", StringComparison.OrdinalIgnoreCase))
+                        otherDeviceBindingFound = true;
+                }
+
+                if (mousePositionBindingFound && otherDeviceBindingFound)
+                {
+                    Debug.LogWarning($"Mouse move event compression will be disabled because the mouse position control " +
```
jimon, line 636: I think this regresses UX too much, many games have camera character/controls with shortcuts, which quite often are done with mouse position composites. Plus this code unlikely to catch someone writing a custom merger/interaction.
If you _really_ want to go that route - remove the warning. There is no point to show warning if it will trigger in many game projects, nor the warning is actionable for game developers, they do as per the manual.

I would go for original PR proposal and propagate every every into `FastMouse` (by that point you could already just batch all of them) so `FastMouse` can decide for itself when to flush the accumulation.

PS. Doing that on `onActionChange` feels _very_ brittle. Are you sure this will work with all serialization/domain reload magic there?

PS2. You forgot mouse delta, mouse scroll, mouse scroll delta :)

andrew-oc: Thanks for the comments. Yeah, good point about the warning. 

For custom interactions (what's a custom merger?), I don't see what's missing. If the user adds an interaction to a composite control, it will still get called at the right times.

> I would go for original PR proposal and propagate every every into FastMouse (by that point you could already just batch all of them) so FastMouse can decide for itself when to flush the accumulation.

If we go that route though we're mostly back to the more complicated solution from before. onActionChange does seem to catch everything, since it will fire from enabling an asset, map, or singleton control and I don't know of any way to get actions to work without enabling them.

> You forgot mouse delta, mouse scroll, mouse scroll delta :)

I did! Thanks.

jimon: > If we go that route though we're mostly back to the more complicated solution from before

Not necessarily, IIRC `FastMouse.OnStateEvent` is already called for every event, so you might be able to copy your `Flush`  code and do that if non-mouse/non-state event arrived. That way checks for bindings are not necessary.

jimon: Though that being said I'm a bit worried that if someone is listening to events they will still get all of them. Mouse compression in editor does that on windows event basis, so it eliminates problems from the source.

andrew-oc: > > If we go that route though we're mostly back to the more complicated solution from before
> 
> Not necessarily, IIRC `FastMouse.OnStateEvent` is already called for every event, so you might be able to copy your `Flush` code and do that if non-mouse/non-state event arrived. That way checks for bindings are not necessary.

My impression was that each event is only processed by one device. We look up the device by the deviceId property in each event, so FastMouse only ever sees mouse events.

andrew-oc: > Though that being said I'm a bit worried that if someone is listening to events they will still get all of them. Mouse compression in editor does that on windows event basis, so it eliminates problems from the source.

Yeah, that's a good point. This PR doesn't do anything to improve things like listen for unpaired device activity which is something we already have performance issues around. So I suppose we can either try and optimize those places further, or go with the more complicated approach for this PR.

jimon:
Closing in favor of https://github.com/Unity-Technologies/InputSystem/pull/1387

PR #1382, opened 2021-07-31
NEW: InputSystem.onAnyButtonPress.

Rene-Damm:
### Description

This addresses a long-standing request and shortcoming in our current API. In the old input system, you can do

```CSharp
void Update()
{
    if (Input.anyKeyDown)
        Debug.Log("Some button was pressed");
}
```

In our current API, there's no easy way to do this and even with the `EnumerateChangedControls` API added in 1.1, it's still a PITA to do this simple thing.

### Changes made

Principally, there's two changes.

#### `InputSystem.onEvent`

`InputSystem.onEvent` is no longer a plain C# event but rather a property returning an `InputEventListener` struct now. The struct overloads the `+` and `-` operators, so:

```CSharp
InputSystem.onEvent += ...;
InputSystem.onEvent -= ...;
```

will continue to compile and work just fine.

However, `InputEventListener` is also an `IObservable<InputEventPtr>` which now acts as a springboard for newly added Rx-style APIs to very easily observe the event stream.

```CSharp
InputSystem.onEvent
    .ForDevice(Mouse.current)
    .Where(e => e.HasButtonPress())
    .Call(_ => Debug.Log("Yay!"));
```

#### `InputSystem.onAnyButtonPress`

While the change above makes it relatively straightforward to listen for button presses, this PR adds this API as a very simple wrapper in order to both make it even easier and also make it much more discoverable.

```CSharp
InputSystem.onAnyButtonPress
    .CallOnce(ctrl => Debug.Log($"Button {ctrl} was pressed"));
```

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,13 +10,6 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased]
 
-### Changed
-
-- Changed `TrackedPoseDriver` to use properties of type `InputActionProperty` rather than `InputAction` to allow more flexibility.
-- XRLayoutBuilder now supports `_` in sanitized names.
-- Added method `SetMotorSpeedsAndLightBarColor` as a workaround for setting both the light bar and motor speeds simultaneously on a DualShock 4 controller ([case 1271119](https://issuetracker.unity3d.com/issues/dualshock4-setlightbarcolor-and-setmotorspeeds-cannot-be-called-on-the-same-frame-using-input-system)).
-- Updated documentation for sensor WebGL support in 2021.2.
-
```
Rene-Damm: Moved this since there already is a CHANGED section.

jimon:
FWIW, it's still worse in UX compared to old system, where old one is just a boolean check, new one is callback

jimon (changes requested):
I think the code is fine, but what is the performance of this?
I can imagine GetFirstButtonPressOrNull grinds to a halt if we call it 100+ times a frame.

Code wise I would rather prefer data oriented design instead. This PR has unclear performance implications and will be hard to move to data based approach in the future.

UX wise I think having a callback is suboptimal, something like `InputSystem.anyButton.wasPressedThisFrame` would be miles ahead in terms of ergonomics. Can we make this into polling API instead? Can add events on top of polling API later.

PS. feel free to ignore my request for changes, I understand that I request essentially a complete rewrite, which might be too late at this point.

Rene-Damm:
> I can imagine GetFirstButtonPressOrNull grinds to a halt if we call it 100+ times a frame.

The underlying mechanism is relatively efficient (basically a memcmp against the default state overlaid by the noise mask). Not nearly as efficient as alternate approaches not operating on unprocessed data in the first place but definitely a great deal better than manually coded versions of detecting button presses on `InputSystem.onEvent`.

> will be hard to move to data based approach in the future.

I don't really see the current input system player code moving to a data-oriented approach at any point in the future. In my view, a DOTS-aligned version of input involves a separate runtime not tied to `UnityEngine.dll` and not involving pretty much any code of `UnityEngine.InputSystem` in the player. Might be too black&white.

> UX wise I think having a callback is suboptimal, something like `InputSystem.anyButton.wasPressedThisFrame` would be miles ahead in terms of ergonomics. Can we make this into polling API instead? Can add events on top of polling API later.

In this case, I'm strongly favoring *not* having a polling API. For this stuff here, if we can't determine when the result will actually be needed, it'll be very hard to avoid doing a bunch of needless work given the architecture we have. The only practical way I see out of that ATM is having an explicit Enable/Disable thing a la `EnhancedTouch`. Which I think would be great to avoid.

I do agree that a polling API would often be the most convenient. The strength I see in the approach used by this PR is that it's mostly just an extension to the `InputSystem.onEvent` facility to make it easier to use. Which, as a side-effect, provides a pretty simple way to poll for button presses.

jimon:
> In this case, I'm strongly favoring not having a polling API. For this stuff here, if we can't determine when the result will actually be needed, it'll be very hard to avoid doing a bunch of needless work given the architecture we have. 

I'm worried that this is not shared by our users and they will just register all callbacks upfront to get polling-esque API on their level and call it a day. Though having at least some way not to degrade performance is reasonable in given circumstances, so maybe from that perspective callback is better.

andybastable:
From my POV (added to this review as someone with end-user use case for this) I'd be happy with the proposed solution. Polling might make this marginally nicer to code, but a system event handler for this is still a million times nicer than having to hand roll logic for the "any key pressed" ourselves.

jimon (approved):
Did some profiling. Our direction at this point is to reduce amount of events coming in by doing event merging (currently only implemented for mouse), given that direction we will not have too many events to process, so performance implications of this PR are minimal to non-existent.

andrew-oc (approved):

ekcoh (approved):

PR #1394, opened 2021-08-25
FIX: Regression from suppressing same control bound repeatedly.

Rene-Damm:
### Description

#1256 regressed behavior for the XR team by no longer allowing the same control to be bound in *different* ways to the same action.

### Changes made

Removed the previous change and added code to

1) explicitly check for actuation from the *same* control in `ShouldIgnoreControlStateChange`, and to
2) filter out duplicates when setting up the data for `InputAction.controls` in `SetUpPerActionCachedBindingData`.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -180,6 +180,8 @@ partial class CoreTests
         var actionMap = new InputActionMap();
         var action3 = actionMap.AddAction("action3");
         var action4 = actionMap.AddAction("action4");
+        var action5 = actionMap.AddAction("action5");
+        var action6 = actionMap.AddAction("action6");
 
         action1.AddBinding("<Gamepad>/buttonSouth");
         action1.AddBinding("<Gamepad>/buttonSouth");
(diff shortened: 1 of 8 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -836,6 +836,23 @@ private unsafe void SetUpPerActionCachedBindingData()
...
                 m_SingletonAction.m_ControlStartIndex = 0;
                 m_SingletonAction.m_ControlCount = m_State?.totalControlCount ?? 0;
+
+                // Only complication, InputActionState allows a control to appear multiple times
+                // on the same action and InputAction.controls[] doesn't.
+                if (m_ControlsForEachAction.HaveDuplicateReferences(0, m_SingletonAction.m_ControlCount))
+                {
+                    var numControls = 0;
+                    var controls = new InputControl[m_SingletonAction.m_ControlCount];
+                    for (var i = 0; i < m_SingletonAction.m_ControlCount; ++i)
```
ekcoh, line 846: On this for loop I would suggest using braces to increase readability and reduce likelyhood of future error

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1378,7 +1379,12 @@ private bool ShouldIgnoreControlStateChange(ref TriggerState trigger, int action
                 // it drive the action - this is like a direction change on the same control.
                 if (bindingStates[trigger.bindingIndex].isPartOfComposite && triggerControlIndex == actionStateControlIndex)
                     return false;
-                if (trigger.magnitude > 0 && triggerControlIndex != actionState->controlIndex)
+                // If we do have an actuation on a control that isn't currently driving the action, flag the action has
+                // having multiple concurrent inputs ATM.
+                // NOTE: We explicitly check for whether it is in fact not the same control even if the control indices are different.
+                //       The reason is that we allow the same control, on the same action to be bound more than once on the same
+                //       action.
+                if (trigger.magnitude > 0 && triggerControlIndex != actionState->controlIndex && controls[triggerControlIndex] != controls[actionState->controlIndex])
```
ekcoh, line 1387: This seems complex, have we covered this in the test cases?

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
```diff
@@ -107,6 +107,27 @@ public static bool ContainsReference<TValue>(this TValue[] array, TValue value)
...
         }
 
+        public static bool ContainsReference<TFirst, TSecond>(this TFirst[] array, int startIndex, int count, TSecond value)
+            where TSecond : class
+            where TFirst : TSecond
+        {
+            return IndexOfReference(array, value, startIndex, count) != -1;
+        }
+
+        public static bool HaveDuplicateReferences<TFirst>(this TFirst[] first, int index, int count)
```
ekcoh, line 117: Is there a reason why we do linear O(N) search for duplicate references instead of using a hash-based container? This will of course do the job but seem like more complex and slower than using a data structure that supports the problem domain (have noticed this in multiple places)

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
Rene-Damm: Done

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -192,16 +194,30 @@ public void Actions_WhenSeveralBindingsResolveToSameControl_ControlIsAssociatedW
         action3.AddBinding("<Gamepad>/buttonSouth");
         action4.AddBinding("<Gamepad>/buttonSouth"); // Should not be removed; different action.
 
+        action5.AddBinding("<Gamepad>/buttonSouth", interactions: "press(behavior=0)");
```
ekcoh, line 197: Maybe add a one liner explaining what senario/expected behavior is intended with these scenarios since its not intuitively clear what to expect and why

ekcoh: ...another option would be to split the test into one-test per scenario

ekcoh (approved):
Overall looks good but the complex method handling the binding and multiple control to single action behavior is a bit hard to grasp (and was prior to this PR as well). Probably something we can address later, but just want to make it clear I might have overlooked something in the logic.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
Rene-Damm: Yup, the added case makes sure the newly added expression is required and the existing tests cover the rest.

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
Rene-Damm: Not sure what that would look like. The array that is being constructed is just a plain InputControl[] where controls are then referenced by index. To construct a temporary hashset for the array as it is being constructed seems like much work for questionable benefit.

In practice, there will be few actions with more than <10 controls bound to them (usually, will be one). So, linear search should do pretty well.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
ekcoh: Great!

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
ekcoh: OK, I see.

PR #1395, opened 2021-08-25
Fix input test fixture state

ekcoh:
### Description

This PR addresses problems related to `InputTestFixture` and its `SetUp`() and `TearDown`() where mixing test cases based on `InputTestFixture` (using mocked InputSystem) and regular test cases (using real InputSystem) would lead to static state leaking between test cases causing random failures and unexpected/undefined behavior. 

### Changes made

Static state in `InputUser`, `InputActionState`, `Touch` have been adressed and a save/revert mechanism have been introduced via `ISavedState` providing an opaque interface to restore a previously saved non-serialized state without knowing internals of the class from which the state was saved. The motivation for this pattern was that initialization/reset require knowledge and is different between classes and that restore might or might not need additional "static dispose" code. Unfortunately, there is currently no consistent "static disposal pattern" implemented in the repository that I am aware of, so that could be a future extension to better encapsulate and handle static state.

### Notes

This relates to, and have been tested/verified against https://issuetracker.unity3d.com/product/unity/issues/guid/1329015 (https://fogbugz.unity3d.com/f/cases/1329015/).

No tests of test harness was added as part of this fix. Extra care should be taken in review to check the removed clean-up code in InputSystem#Destroy() since this seems to relate to test cases but should not be needed with this fix (!?)

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
+## [1.1.0-verified]
+
+### Fixed
+
+- Fixed an issue where mixing test cases based on `InputTestFixture` (using mocked `InputSystem`) and regular test cases (using real `InputSystem`) would lead to static state leaking between test cases causing random failures and unexpected/undefined behavior.
+
 ## [1.1.0-pre.6] - 2021-08-23
 
 ### Fixed
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
-## [1.1.0-verified]
+## [Unreleased]
 
 ### Fixed
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -12,7 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 
-- Fixed an issue where mixing test cases based on `InputTestFixture` (using mocked `InputSystem`) and regular test cases (using real `InputSystem`) would lead to static state leaking between test cases causing random failures and unexpected/undefined behavior.
+- Fixed an issue where mixing test cases based on `InputTestFixture` (using mocked `InputSystem`) and regular test cases (using real `InputSystem`) would lead to static state leaking between test cases causing random failures and unexpected/undefined behavior ([case 1329015](https://issuetracker.unity3d.com/product/unity/issues/guid/1329015)).
 
 ## [1.1.0-pre.6] - 2021-08-23
 
--- a/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
@@ -3609,18 +3609,23 @@ namespace UnityEngine.InputSystem
         {
             Debug.Assert(s_SavedStateStack != null && s_SavedStateStack.Count > 0);
 
+            // Load back previous state.
+            var state = s_SavedStateStack.Pop();
+
+            state.inputUserState.StaticDisposeCurrentState();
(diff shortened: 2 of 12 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
+## [1.1.0-verified]
```
ekcoh: @Rene-Damm Should this be plain [1.1.0]?

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3519,7 +3520,6 @@ private static void Destroy()
             //       state repeatedly during tests but we want to not create InputSystemObject
             //       over and over.
 
-            InputActionState.ResetGlobals();
```
ekcoh, line 3522: @Rene-Damm With these changes, I think this is not needed. Do you agree?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Due to package verification, the latest version below is the unpublished version and the date is meaningless.
 however, it has to be formatted properly to pass verification tests.
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed an issue where mixing test cases based on `InputTestFixture` (using mocked `InputSystem`) and regular test cases (using real `InputSystem`) would lead to static state leaking between test cases causing random failures and unexpected/undefined behavior ([case 1329015](https://issuetracker.unity3d.com/product/unity/issues/guid/1329015)).
```
Rene-Damm, line 22: Nice description.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3553,6 +3553,10 @@ internal struct State
             [SerializeField] public string systemObject;
             #endif
             ////REVIEW: preserve InputUser state? (if even possible)
+            ///
```
Rene-Damm: No triple slash comments except for doctools.

Looking at the REVIEW comment, probably also good to clarify for this one that it relates to domain reloads. Maybe should be a 

```
////TODO: make these saved states capable of surviving domain reloads
```

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Never mind realized it should be [Unreleased]

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: Seems correct.

However, seems that the ordering has changed now. Seems we now *first* destroy `InputManager` and then reset the saved state. That means that `InputActionState` et al will now do its work against an `InputManager` that has technically been destroyed. Seems problematic.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Thanks, trying to focus on how the problem manifests itself instead of the nitty gritty details

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: I'll update the TODO/REVIEW and get rid of the slashes

ekcoh: Fixed

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: hmmm.... will take a look

ekcoh: Hopefully this works correctly now? @Rene-Damm 
This makes the whole interface/abstraction thing kind of redundant, but should probably be revisited together with an attempt to make life-time management consistent, do you agree?

ekcoh: The order is corrected and I realized I had done a mistake in the assignment and accidentally placed it within preprocessor guard for editor builds. Player tests work locally so expecting to see CI passed when I revisit tomorrow morning.

PR #1402, opened 2021-09-03
Fix - Avoid re-pairing devices when switching to the already active control scheme

ekcoh:
### Description

An issue related to callback/event not firing when switching control scheme was reported here: https://fogbugz.unity3d.com/f/cases/reply/1342297/
It seems like cancellation of the event is by design when unpairing and repairing devices, but what is clearly undesirable here is that devices gets unpaired and paired again when switching to the already active control scheme as being done in the provided repro project. 

### Changes made

Changes involve minor refactoring to `PlayerInput `and `InputUser `to not trigger device unpairing and pairing if the control scheme switching (explicit) is to the already equivalent control scheme and set of devices. In this case, the switching is instead just an early return indicating successful operation (true).

### Notes

A possible workaround for this problem before this fix landing in a public release is to similar to this PR check if the requested devices control scheme checked via `InputControlScheme.FindControlSchemeForDevices(...)` is the same control scheme as currently assigned to `PlayerInput`.

### Checklist

Before review:

- [X] Changelog entry added.
    - Functional tests: 1 existing functional test in PlayerInputTests updated, 1 new test cases in PlayerInputTest added to cover scenario. 1 new test case added to `ArrayHelperTests`
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -908,11 +908,13 @@ public bool SwitchCurrentControlScheme(params InputDevice[] devices)
                 throw new InvalidOperationException(
                     "Must set actions on PlayerInput in order to be able to switch control schemes");
 
+            // Find control scheme matching given devices in associated action asset
             var scheme = InputControlScheme.FindControlSchemeForDevices(devices, actions.controlSchemes);
-            if (scheme == null)
+            if (!scheme.HasValue)
```
andrew-oc, line 913: 'scheme' still needs a null check here

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -1930,6 +1927,31 @@ private void OnDeviceChange(InputDevice device, InputDeviceChange change)
...
 
+        private bool IsCurrentControlScheme(ref InputControlScheme scheme, InputDevice[] devices)
+        {
+            var isSameScheme = user.controlScheme.HasValue && user.controlScheme.Value.Equals(scheme);
+            return (isSameScheme &&
+                ArrayHelpers.EqualSets(user.pairedDevices.ToArray(), devices) &&
+                !user.hasMissingRequiredDevices);
+        }
+
+        private void SwitchControlSchemeInternalUnchecked(ref InputControlScheme controlScheme, params InputDevice[] devices)
```
andrew-oc: I don't quite understand why this method (and others) are called [Something]Unchecked? My first thought is that you're somehow doing unchecked array accesses inside them using the 'unchecked' keyword, but it's not that.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -1930,6 +1927,31 @@ private void OnDeviceChange(InputDevice device, InputDeviceChange change)
...
+        {
+            // Precondition: Control scheme and devices are correlated
```
andrew-oc: Why not Debug Assert your preconditions?

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
```diff
@@ -485,31 +485,39 @@ public ControlSchemeChangeSyntax ActivateControlScheme(string schemeName)
...
+                FindControlScheme(schemeName, ref scheme); // throws if not found
+            return ActivateControlScheme(scheme);
+        }
 
-                // Need actions to be available to be able to activate control schemes
-                // by name only.
-                if (s_GlobalState.allUserData[userIndex].actions == null)
+        private bool TryFindControlScheme(string schemeName, ref InputControlScheme scheme)
+        {
+            if (!string.IsNullOrEmpty(schemeName))
```
andrew-oc: Could invert if here

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -1930,6 +1927,31 @@ private void OnDeviceChange(InputDevice device, InputDeviceChange change)
             }
         }
 
+        private bool IsCurrentControlScheme(ref InputControlScheme scheme, InputDevice[] devices)
+        {
+            var isSameScheme = user.controlScheme.HasValue && user.controlScheme.Value.Equals(scheme);
+            return (isSameScheme &&
+                ArrayHelpers.EqualSets(user.pairedDevices.ToArray(), devices) &&
```
andrew-oc: Would `devices.All(user.pairedDevices.Contains)` do the same thing? It would avoid the new array allocation and all the extra code in EqualSets.

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
```diff
@@ -485,31 +485,39 @@ public ControlSchemeChangeSyntax ActivateControlScheme(string schemeName)
...
-            {
-                var userIndex = index; // Throws if user is invalid.
+                FindControlScheme(schemeName, ref scheme); // throws if not found
+            return ActivateControlScheme(scheme);
+        }
 
-                // Need actions to be available to be able to activate control schemes
-                // by name only.
-                if (s_GlobalState.allUserData[userIndex].actions == null)
+        private bool TryFindControlScheme(string schemeName, ref InputControlScheme scheme)
```
andrew-oc: `out` would be better than `ref` here

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
```diff
@@ -485,31 +485,39 @@ public ControlSchemeChangeSyntax ActivateControlScheme(string schemeName)
...
-                if (scheme == default)
-                    throw new ArgumentException(
-                        $"Cannot find control scheme '{schemeName}' in actions '{s_GlobalState.allUserData[userIndex].actions}'");
             }
 
-            return ActivateControlScheme(scheme);
+            return false;
+        }
+
+        internal void FindControlScheme(string schemeName, ref InputControlScheme scheme)
```
andrew-oc: `out` vs `ref` again

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
```diff
@@ -147,6 +147,25 @@ public static bool HaveEqualElements<TValue>(TValue[] first, TValue[] second, in
             return true;
         }
 
+        public static bool EqualSets<TValue>(TValue[] first, TValue[] second)
```
andrew-oc: If we do keep this method (see other comment above about replacing it with All) then the comment in the body should be promoted to an xml comment

andrew-oc (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
ekcoh: Sorry for the confusion, is an old habit to emphasize internal functions that doesn't implement preconditional contracts since they are already indirectly given by calling code (to avoid doing it twice). I will rename this in a coming commit.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
ekcoh: Very valid point, I will check if this is possible, it will be a quite costly assert though which is indirectly implied by the calling code, but can see if its feasible. Guess it could be an assert calling a static method that makes sure all devices are applicable for the control scheme and that they are actually available.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
ekcoh: I do not believe so since FindControlSchemeForDevices returns a nullable struct wrapping struct InputControlScheme, e.g. https://docs.microsoft.com/en-us/dotnet/api/system.nullable-1?view=net-5.0 struct. Hence scheme?.HasValue would not be valid code since value type. I believe Nullable<T> using comparison operator overload == is equal to HasValue but is there a code convention I am not aware of that we prefer using the operator overload instead?

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
ekcoh: Ended up with relaxing this to only require devices to not be null, replaced with assert and removed "Unchecked" part.

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
ekcoh: Done but increases lines of code and reduces readability in my personal taste, but updated with early-return pattern

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
ekcoh: Changed to ref

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
ekcoh: Changed to ref

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
ekcoh: I removed the method.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
ekcoh: Almost but creates some issues due to IReadOnlyArray and value types. Replaced by other comparison, see latest commit.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
andrew-oc: Yep, my bad. Didn't notice it was nullable

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
andrew-oc: I don't see it. Where is the value type in this? Seems like reference comparisons between devices will work well here, especially since the sample code in the documentation for SwitchCurrentControlScheme uses Keyboard.current and Mouse.current as the devices.

ekcoh: Hopefully this makes more sense to you now @andrew-oc , thanks for the discussion. Also realized that pairing and controller scheme switch should be uncorrelated actions. Unfortunately not possible to write a test for the case of doing unnecessary work on pairing since pairing/unpairing doesn't seem to be observable, but verified manually with user-submitted test project in additional to the automated tests.

andrew-oc (approved):

PR #1405, opened 2021-09-13
NEW: Actions now consume their inputs.

Rene-Damm:
### Description

Addresses the problem of binding to "SHIFT+B" and "B" and seeing both trigger when pressing "SHIFT+B".

Also, addresses the problem of "SHIFT+B" being fine if "SHIFT" is pressed *after* "B".

KeyboardShortcuts

### Changes made

State monitors are extended with the concept of "groups". Within a group, monitors are sorted by their monitor index.

Additionally, from within `NotifyControlStateChanged`, a handler can opt to "consume" the input by setting `InputEvent.handled` to `true`. When doing so, all remaining monitors in the same group are "silenced", i.e. have their `signalled` bit reset to `false`.

`InputActionState` leverages this by

1. computing a "complexity" metric for each control based on the way it is bound,
2. assigning the same group to all bindings to the same control, and
3. setting `InputEvent.handled` to `true` when an `InputAction` is performed.

This PR also fixes a problem discovered in the process where `InputAction.activeControl` will not track correctly when `ShouldIgnoreControlStateChange` is performing conflict resolution.

Finally, this PR addresses the problem of the current "modifier" style composites not enforcing an ordering on when the modifiers need to be pressed. This is fine for a binding like "CTRL+MouseDelta" but not fine for a binding like "CTRL+A" where you absolutely want to require that "CTRL" be pressed *before* "A".

To do so, it introduces "press times" made available through `InputBindingCompositeContext.GetPressTime(part)`. For `OneModifierComposite` and `TwoModifiersComposite`, logic has been added to only require an ordering if the composite is bound to a button. This means that "SHIFT+B" has to be pressed in order but "SHIFT+MouseDelta" does not. An explicit `overrideModifiersNeedToBePressedFirst` parameter has been added to override this default inference.

### Notes

* Right now, the concept of consumption is implicit. However, this can be extended to allow interactions and/or callbacks to also opt to consume or not consume the current input.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


andrew-oc (approved):

forestrf:
This fix actually broke functionality I was relying on. I was able to fix it by making better use of Action Maps (not exactly, they seem to be sometimes buggy when disabled action maps have the same bindings as active ones, preventing active actions from firing).
This problem was aggravated due to it being the default for at least a year and having a custom composite modifier that accounted for the order of the pressed buttons to avoid this problem.

While this new behaviour is the correct one and the one I prefer, I would consider adding to the docs a note somewhere that when updating from 1.3.x to 1.4.x (or whenever this was merged, sorry I don't know when it was) the behaviour of composites has changed and now not all actions will fire.

Thank you for this fix, it made me think there was a strange bug somewhere in the input system but the bug was in my code relying on a "bug". It took me a while to find this pull request and finally understand what was happening, as I even remember talks in the forum about the strange old composite behaviour, didn't expect it to get fixed one day.

PR #1407, opened 2021-09-16
FIX: HID device capabilities JSON being parsed more than once for same device

andrew-oc:
[1362733](https://fogbugz.unity3d.com/f/cases/1362733/)

### Description

On the AddAvailableDevicesThatAreNowRecognized path, TryFindMatchingControlLayout was called twice, which for HID devices, results in the device capabilities JSON being parsed both times.

### Changes made

Removed a redundant call to TryFindMatchingControlLayout when adding devices.

### Checklist

Before review:

- [x] Changelog entry added.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -1207,6 +1207,16 @@ namespace UnityEngine.InputSystem
             return device;
         }
 
+        public InputDevice AddDevice(InputDeviceDescription description, InternedString layout, string deviceName = null,
+            int deviceId = InputDevice.InvalidDeviceId, InputDevice.DeviceFlags deviceFlags = 0)
+        {
+            Profiler.BeginSample("InputSystem.AddDevice");
+            var device = AddDevice(layout, deviceId, deviceName, description, deviceFlags);
+            device.m_Description = description;
+            Profiler.EndSample();
+            return device;
+        }
+
         public void RemoveDevice(InputDevice device, bool keepOnListOfAvailableDevices = false)
         {
             if (device == null)
@@ -2648,17 +2658,27 @@ namespace UnityEngine.InputSystem
                     continue;
 
                 var layout = TryFindMatchingControlLayout(ref m_AvailableDevices[i].description, id);
-                if (IsDeviceLayoutMarkedAsSupportedInSettings(layout))
+                if (!IsDeviceLayoutMarkedAsSupportedInSettings(layout)) continue;
+
+                if (layout.IsEmpty())
                 {
-                    try
+                    // If it's a device coming from the runtime, disable it.
+                    if (id != InputDevice.InvalidDeviceId)
                     {
-                        AddDevice(m_AvailableDevices[i].description, false,
-                            deviceId: id,
-                            deviceFlags: m_AvailableDevices[i].isNative ? InputDevice.DeviceFlags.Native : 0);
+                        var command = DisableDeviceCommand.Create();
+                        m_Runtime.DeviceCommand(id, ref command);
                     }
-                    catch (Exception)
-                    {
-                    }
+
+                    continue;
+                }
+
+                try
+                {
+                    AddDevice(m_AvailableDevices[i].description, layout, deviceId: id,
+                        deviceFlags: m_AvailableDevices[i].isNative ? InputDevice.DeviceFlags.Native : 0);
+                }
+                catch (Exception)
+                {
                 }
             }
         }
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -1210,11 +1210,18 @@ namespace UnityEngine.InputSystem
         public InputDevice AddDevice(InputDeviceDescription description, InternedString layout, string deviceName = null,
             int deviceId = InputDevice.InvalidDeviceId, InputDevice.DeviceFlags deviceFlags = 0)
         {
-            Profiler.BeginSample("InputSystem.AddDevice");
-            var device = AddDevice(layout, deviceId, deviceName, description, deviceFlags);
-            device.m_Description = description;
-            Profiler.EndSample();
-            return device;
+            try
+            {
+                Profiler.BeginSample("InputSystem.AddDevice");
+
+                var device = AddDevice(layout, deviceId, deviceName, description, deviceFlags);
+                device.m_Description = description;
+                return device;
+            }
+            finally
+            {
+                Profiler.EndSample();
+            }
         }
 
         public void RemoveDevice(InputDevice device, bool keepOnListOfAvailableDevices = false)
@@ -2679,6 +2686,9 @@ namespace UnityEngine.InputSystem
                 }
                 catch (Exception)
                 {
+                    // the user might have changed the layout of one device, but others in the system might still have
+                    // layouts we can't make sense of. Just quietly swallow exceptions from those so as not to spam
+                    // the user with information about devices unrelated to what was actually changed.
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2648,17 +2658,27 @@ internal void AddAvailableDevicesThatAreNowRecognized()
...
+                        m_Runtime.DeviceCommand(id, ref command);
                     }
+
+                    continue;
+                }
+
+                try
+                {
+                    AddDevice(m_AvailableDevices[i].description, layout, deviceId: id,
+                        deviceFlags: m_AvailableDevices[i].isNative ? InputDevice.DeviceFlags.Native : 0);
```
Rene-Damm, line 2685: Unfortunately, I think this won't work correctly. Sometime in the past, the code used to just use the layout it had before the domain reload. That was okay when we actually took all layout information across the reload.

Now, we have a setup where all layout information is considered "thrown away" during domain reloads and allowed to change completely. That, however, also means we have to do the matching process from scratch as matchers are part of what is allowed to change.

andrew-oc: But this new code *does* do the matching process in the same way as before, it just does it once because it calls into the new AddDevice overload, which doesn't have a TryFindMatchingControlLayout call in it :confused:

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1207,6 +1207,16 @@ public InputDevice AddDevice(InputDeviceDescription description)
             return device;
         }
 
+        public InputDevice AddDevice(InputDeviceDescription description, InternedString layout, string deviceName = null,
+            int deviceId = InputDevice.InvalidDeviceId, InputDevice.DeviceFlags deviceFlags = 0)
```
ekcoh, line 1211: Out of curiosity, is null / InvalidDeviceId a reasonable default value(s), e.g. adding a device that is not valid is the "common" operations?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2648,17 +2658,27 @@ internal void AddAvailableDevicesThatAreNowRecognized()
...
+
+                    continue;
+                }
+
+                try
+                {
+                    AddDevice(m_AvailableDevices[i].description, layout, deviceId: id,
+                        deviceFlags: m_AvailableDevices[i].isNative ? InputDevice.DeviceFlags.Native : 0);
+                }
+                catch (Exception)
```
ekcoh, line 2687: Is this error commonly excepted or do we want to see it in e.g. Debug log at least? If intended to suppress maybe add a comment why we ignore the error?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1207,6 +1207,16 @@ public InputDevice AddDevice(InputDeviceDescription description)
             return device;
         }
 
+        public InputDevice AddDevice(InputDeviceDescription description, InternedString layout, string deviceName = null,
+            int deviceId = InputDevice.InvalidDeviceId, InputDevice.DeviceFlags deviceFlags = 0)
+        {
+            Profiler.BeginSample("InputSystem.AddDevice");
```
ekcoh: Since I noticed that AddDevice is called within try-catch below, does the Profiler.BeginSample() / Profiler.EndSample() pair need to a scoped RAII operation, e.g. using (new ScopedProfilerSample()) { ..... } in order to guarantee that EndSample() is called also if AddDevice throws? If AddDevice never throws, is the try-catch needed?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
andrew-oc: I don't know if it's reasonable, but it's consistent :smile:

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
andrew-oc: Wasn't sure, so left as is. Maybe @Rene-Damm knows?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
andrew-oc: It can throw further down the call tree. I'm not familiar with ScopedProfilerSample. We probably should be using ProfilerMarker here and Auto for scoping, but that's the newer profiling API and I'm not sure if there are any complications from switching to it.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: Adding a device without a device ID will generate one. Devices added in the native backend already come with a device ID. Devices existing purely on the managed side get one generated on demand.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: Ah, waddayaknow, can't even read me own code. You're right.

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: (only exception: for any kind of device, we take IDs across domain reloads; so in that case, no new IDs should be generated)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: Comment probably a good idea. Reason we just suppress the exception is that here in the `AddAvailableDevicesThatAreNowRecognized` method we're doing a kind of "you know, our layout stuff has changed so maybe we can make sense of some stuff now what we couldn't make sense of before... let's try". But that trying may still fail. So, thinking is, if you make a completely unrelated change to the layout information, you shouldn't get spammed with errors from an unrelated layout that happens to be cause issues.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Allright, thanks for taking your time to explain the why, all good with me then

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: There is no ScopedProfilerSample as far as I know it was more of "psuedo description" but what I meant was roughly (borrowing form C++)

```cs
class ScopedProfilerSample : IDisposable
{
    ScopedProfilerSample(string id) { this.id = id; profiler.BeginSample(id); }
    void Dispose() { profiler.EndSample(id); }
    private string id;
}
```

So you could use it as a scoped guard in a "using" statement (as all other RAII-like stuff with using statements)

ekcoh: Andrew I think it would make sense to ensure that Profiler.EndSample() is called also if AddDevice throws if AddDevice may throw. However, the description I assume should not be updated in that case so probably just EndSample in catch -> rethrow while the Add + description setter remains in "try" part?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: @andrew-oc Could we add a one-liner comment just explaining the why along these lines for future reference? An empty catch without explanation looks like "forgetting to add handling code"

ekcoh (changes requested):
Marked 2 small things I think would be beneficial to address

ekcoh (approved):
Looks good.

PR #1408, opened 2021-09-23
Handle all exceptions in OnUpdate and cleanup m_InputEventStream state to avoid locking up latter updates

jimon:
### Description

- Any unhandled exception in OnUpdate will lock up OnUpdate reentrance check and make input system stop working until domain reload / editor restart. To make matters worse the log will be full of `InvalidOperationException: Already have an event buffer set! Was OnUpdate() called recursively?` but the actual problem is the very first exception in the log, which can be hard to spot.

- Users are confused by input system reporting `...Exception while executing '...' callbacks of '...'` / `...Exception during event processing of ... update; resetting event buffer` / etc and report them as something input system is to be blamed for, this is because the actual exception is printed second.

### Changes made

- Capturing all exceptions with-in OnUpdate and clearing up `m_InputEventStream`.
- Changed order of printing to always print exception first, followed by our explanation.

### Notes

I also tested in manually in device tester, confirmed that system recovers from single exception throwed from OnStateEvent in FastMouse

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -2952,371 +2952,393 @@ namespace UnityEngine.InputSystem
             var isPlaying = gameIsPlaying;
             #endif
 
-            m_InputEventStream = new InputEventStream(ref eventBuffer, m_Settings.maxQueuedEventsPerUpdate);
-            var totalEventBytesProcessed = 0U;
-
-            InputEvent* skipEventMergingFor = null;
-
-            // Handle events.
-            while (m_InputEventStream.remainingEventCount > 0)
+            try
             {
-                if (m_Settings.maxEventBytesPerUpdate > 0 &&
-                    totalEventBytesProcessed >= m_Settings.maxEventBytesPerUpdate)
+                m_InputEventStream = new InputEventStream(ref eventBuffer, m_Settings.maxQueuedEventsPerUpdate);
+                var totalEventBytesProcessed = 0U;
+
+                InputEvent* skipEventMergingFor = null;
+
+                // Handle events.
+                while (m_InputEventStream.remainingEventCount > 0)
                 {
-                    Debug.LogError("Exceeded budget for maximum input event throughput per InputSystem.Update(). Discarding remaining events. "
-                        + "Increase InputSystem.settings.maxEventBytesPerUpdate or set it to 0 to remove the limit.");
-                    break;
-                }
-
-                InputDevice device = null;
-                var currentEventReadPtr = m_InputEventStream.currentEventPtr;
-
-                Debug.Assert(!currentEventReadPtr->handled, "Event in buffer is already marked as handled");
-
-                // In before render updates, we only take state events and only those for devices
-                // that have before render updates enabled.
-                if (updateType == InputUpdateType.BeforeRender)
-                {
-                    while (m_InputEventStream.remainingEventCount > 0)
+                    if (m_Settings.maxEventBytesPerUpdate > 0 &&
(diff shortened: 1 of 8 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1408) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1408) before we can accept your contribution.<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1408) it.</sub>

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2952,371 +2952,393 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             var isPlaying = gameIsPlaying;
             #endif
 
-            m_InputEventStream = new InputEventStream(ref eventBuffer, m_Settings.maxQueuedEventsPerUpdate);
-            var totalEventBytesProcessed = 0U;
+            try
```
ekcoh, line 2955: Not mainly introduced by this branch but this method is very long and hard to read and could probably benefit from some refactoring or dividing responsibility up between sub-functions. Maybe something to refactor later or what do you think?

ekcoh: A bit tricky to get an overview of the changes here so maybe splitting this method up would also make it easier to evaluate exception safety

ekcoh (approved):
Looks good to me, hard to assess exception safety contract of that long method though.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Seems like tests need some work before wrapping this up though (failing).

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2952,371 +2952,393 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
-            m_InputEventStream.Close(ref eventBuffer);
+                m_InputEventStream.Close(ref eventBuffer);
+            }
+            catch (Exception)
+            {
+                // We need to restore m_InputEventStream to a sound state
+                // to avoid failing recursive OnUpdate check next frame.
+                m_InputEventStream.CleanUpAfterException();
+                throw;
+            }
```
andrew-oc, line 3341: Hmm, I'm wondering now if it makes more sense to replace this catch with a finally block and move the input event stream close method into it, delete the CleanUpAfterException method, and let the NativeInputRuntime do its job and reset the event buffer if there's an exception.

jimon: Mmm, maybe, though data correctness after exception is not a given, so I would probably would still put a more exception friendly code branch in Close, e.g. don't reuse existing buffers and just reset ..

Though it's a bit of a slippery slope, our best effort here should be to recover, but we're likely can get in situations were the only recover procedure is to restart whole editor/player ...

PR #1411, opened 2021-10-04
NEW: Show single and multi-touch press controls in input action dropdown

andrew-oc:
[case 1357664](https://fogbugz.unity3d.com/f/cases/1357664/)

### Description

The input action asset editor doesn't allow for selecting a binding that supports multi-touch, and new users find it difficult to impossible to know the touch*/press path binding syntax.

### Changes made

Replaced the Press option in the input action dropdown editor with two new options:
* Press (Single touch)
* Press (Multi-touch)

The second option generates the touch*/press binding path.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Editor.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Editor.cs
@@ -2958,6 +2958,38 @@ partial class CoreTests
         EditorHelpers.RestartEditorAndRecompileScripts(dryRun: true);
     }
 
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputControlPicker_TouchscreenPickerContainsSingleAndMultiTouchControls()
+    {
+        var dropdown = new StubInputControlPickerDropdown(new InputControlPickerState(), _ => {});
+        var root = dropdown.BuildRoot();
+
+        Assert.That(() =>
+        {
+            var touchscreen = root.children.FirstOrDefault(c => c.name == "Touchscreen");
+            if (touchscreen == null)
+                return false;
+
+            return touchscreen.children.Any(c => c.name == "Press (Single touch)") &&
+            touchscreen.children.Any(c => c.name == "Press (Multi-touch)");
+        });
+    }
+
+    internal class StubInputControlPickerDropdown : InputControlPickerDropdown
+    {
+        public StubInputControlPickerDropdown(InputControlPickerState state, Action<string> onPickCallback,
+                                              InputControlPicker.Mode mode = InputControlPicker.Mode.PickControl)
+            : base(state, onPickCallback, mode)
+        {
+        }
+
+        public UnityEngine.InputSystem.Editor.AdvancedDropdownItem BuildRoot()
+        {
+            return base.BuildRoot();
+        }
+    }
+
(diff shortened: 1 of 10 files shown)


Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -2958,6 +2958,38 @@ public void Editor_CanRestartEditorThroughReflection()
...
+        var dropdown = new StubInputControlPickerDropdown(new InputControlPickerState(), _ => {});
+        var root = dropdown.BuildRoot();
+
+        Assert.That(() =>
+        {
+            var touchscreen = root.children.FirstOrDefault(c => c.name == "Touchscreen");
+            if (touchscreen == null)
+                return false;
+
+            return touchscreen.children.Any(c => c.name == "Press (Single touch)") &&
```
ekcoh, line 2986: Code looks clean and readable to me. However I have one question related to the general concept here, what is the difference with respect to single press touch and multi-touch. Is single press touch a trigger mapping only to a press with a single touch point and multi-touch is anything more than a single touch point for the press, e.g. N > 1, where N = 2, 3, 4....

ekcoh: Couldn't find any explanation in the docs around this other than "Using touch with Actions" under https://docs.unity3d.com/Packages/com.unity.inputsystem@1.1/manual/Touch.html either so maybe we are missing something there or did I only fail to find it? Does multi-touch here correspond to wild-card touches? Does this enforce pass-through as suggested in the manual?

ekcoh (approved):
Seems like there are failing tests that needs to be addressed. Apart from that looks good to me. Also see comment/questions regarding tying this improvement together with manual/docs.

Assets/Tests/InputSystem/CoreTests_Editor.cs
andrew-oc: > Code looks clean and readable to me. However I have one question related to the general concept here, what is the difference with respect to single press touch and multi-touch. Is single press touch a trigger mapping only to a press with a single touch point and multi-touch is anything more than a single touch point for the press, e.g. N > 1, where N = 2, 3, 4....

Single press maps to primary touch, which is just touch0, and multi-touch maps to any touch and so will also work for 1 touch. The difference is that primary touch can "migrate" to a different finger if the finger that was considered the first touch is released, while another is still held down.

andrew-oc: > Couldn't find any explanation in the docs around this other than "Using touch with Actions" under https://docs.unity3d.com/Packages/com.unity.inputsystem@1.1/manual/Touch.html either so maybe we are missing something there or did I only fail to find it? Does multi-touch here correspond to wild-card touches? Does this enforce pass-through as suggested in the manual?

That's also the only bit of documentation I found. I thought it did quite a good job of explaining the problem, but it's just a bit too hard to find, which is the reason for this PR. Yes, multi-touch corresponds to wild-card touches. I didn't make this force passthrough because there could be other bindings on the action that the user doesn't want to be passthrough. Not sure what the best solution is there. Filling in touch*/Press for the user sort of solves only half the problem.

Rene-Damm:
Nice.

PR #1415, opened 2021-10-06
FIX: VirtualMouseInput works correctly in the editor again

andrew-oc:
[case 1367553](https://fogbugz.unity3d.com/f/cases/1367553/)

### Description

VirtualMouseInput uses OnAfterUpdate callback to query the state of the device. This works well during dynamic or fixed updates but doesn't work so well for editor updates. This is because the state of the device in the editor state buffers will not have been updated when the game view has focus due to how we short circuit the event processing loop in focus code.

### Changes made

After update callbacks are no longer called during editor mode updates.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -3397,9 +3397,10 @@ namespace UnityEngine.InputSystem
 
         private void InvokeAfterUpdateCallback(InputUpdateType updateType)
         {
-            // don't invoke the after update callback if this is an editor update. Otherwise, any
-            // handlers for this delegate that query input state will get stale values.
-            if (updateType == InputUpdateType.Editor)
+            // don't invoke the after update callback if this is an editor update and the game is playing. We
+            // skip event processing when playing in the editor and the game has focus, which means that any
+            // handlers for this delegate that query input state during this update will get no values.
+            if (updateType == InputUpdateType.Editor && gameIsPlaying)
                 return;
 
             DelegateHelpers.InvokeCallbacksSafe(ref m_AfterUpdateListeners,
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3391,12 +3391,18 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
             m_CurrentUpdate = default;
         }
 
-        private void InvokeAfterUpdateCallback()
+        private void InvokeAfterUpdateCallback(InputUpdateType updateType)
         {
+            // don't invoke the after update callback if this is an editor update and the game is playing. We
+            // skip event processing when playing in the editor and the game has focus, which means that any
+            // handlers for this delegate that query input state during this update will get no values.
+            if (updateType == InputUpdateType.Editor && gameIsPlaying)
```
Rene-Damm, line 3403: Why the `gameIsPlaying` condition?

andrew-oc: After discussing with @jimon this seems a less blunt way of breaking the API contract, while still fixing the issue with virtual mouse. If anyone has code that depends on this in the editor outside of play mode, it should continue to work.

jimon (approved):

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jimon: It seems we only have this problem when we're in playmode in editor.
Otherwise there might other legit users of `onAfterUpdate` from editor mode only, no need to break them without a reason

PR #1416, opened 2021-10-07
Fixed case 1369081. When using SingleUnifiedPointer, the same pointer…

benoit-alain:
…Id should be sent to UIToolkit regardless of pointer type or finger id.

### Description

Fixes case 1369081. In the user's project, the button thinks it still has a finger on top of it because when a second finger is pressed somewhere else, it replaces the data from the first finger and a PointerUpEvent is never sent with the pointerId of the first finger. What this fix does is to pretend that all the input actions relating to pointers are occuring on the exact same pointerId, namely the mousePointerId. This pointerId seems to be the only suitable candidate because, despite the SingleUnifiedPointer option, InputSystemUIInputModule sends duplicate mouse-based events when touch occurs, making the mouse pointerId the lowest common denominator if we can only send one.

This PR will effectively make UI Toolkit disregard touch or pen information when it receives actions from touch or pen, due to the fact that they will all be treated like they came from a single, unified mouse. I'm not sure if that's a desirable property, but it makes sense when reading what the SingleUnifiedPointer documentation says; it looks like that's what this option is intending to do, if I understand it correctly.

### Changes made

When using the UIPointerBehavior.SingleUnifiedPointer option, pointer events sent to UI Toolkit panels through InputSystemUIInputModule now always have the same pointerId, regardless of the hardware component that produced them.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ however, it has to be formatted properly to pass verification tests.
...
 - Fixed `action.ReadValue` and others returning invalid data when used from `FixedUpdate` or early update when running in play mode in the editor ([case 1368559](https://issuetracker.unity3d.com/issues/enter-key-is-not-registered-when-using-waspressedthisframe-with-input-system-1-dot-1-1) [case 1367556](https://issuetracker.unity3d.com/issues/input-action-readvalue-always-returns-zero-when-called-from-fixedupdate) [case 1372830](https://issuetracker.unity3d.com/issues/querying-inputs-before-preupdate-dot-newinputupdate-returns-invalid-data-when-running-in-play-mode-in-editor)).
+- Fixed a problem with UI Toolkit buttons remaining active when multiple fingers are used on a touchscreen, using `InputSystemUIInputModule` with pointerBehavior set to `UIPointerBehavior.SingleUnifiedPointer`. UI Toolkit will now always receive the same pointerId when that option is in use, regardless of the hardware component that produced the pointer event. ([case 1369081](https://fogbugz.unity3d.com/f/cases/1369081)).
```
ekcoh: Thanks for submitting this fix @benoit-alain . We have some issues with UI tests since a recent trunk update which causes input UI CI tests to fail with assertion failures so will likely not merge this until CI is back on track, but could you update this internal URL to the public URL https://issuetracker.unity3d.com/product/unity/issues/guid/1369081 instead?

ekcoh (changes requested):
Changes look good. Will wait with merge though since a recent trunk update causes our UI tests to fail and this haven't been resolved yet. Great if you could fix URL so this can be approved.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: @benoit-alain Fixed the link myself

ekcoh (approved):

Rene-Damm (approved):

andrew-oc:
@benoit-alain Seems to be a test failure with this now on 2021.2, both locally and on Yamato, something to do with EndTouch not being processed correctly maybe. Do you think you could take a look before we merge this one?

benoit-alain:
> @benoit-alain Seems to be a test failure with this now on 2021.2, both locally and on Yamato, something to do with EndTouch not being processed correctly maybe. Do you think you could take a look before we merge this one?

OK, I will have a look as soon as possible (probably today) and let you know what I find or push a fix.

benoit-alain:
I was able to repro the failing test locally and pushed a fix which now passes locally on 2021.2. I don't think it's different enough that it would fail on trunk or other versions of the editor, but Yamato will tell us for sure.

PR #1423, opened 2021-10-15
Openxr integration prototype

andrew-oc:
A brief overview:
* we've introduced the concept of a global input action asset (InputSettings.globalInputActions), and the XR actions in there are the ones that are communicated to OpenXR
* communication is via json file that the XR backend will have to read before creating the session
there is a new binding path syntax specifically for XR paths, prefixed with "OpenXR:"
* OpenXRDevice is the main communication channel between the backend and the action system. It creates a mapping between OpenXR actions and input controls based on a hash of the action name. we expect the backend to send the correct identifier for an action in each state event and we look up the correct control to update from that


jfreire-unity:
@andrew-oc is this PR still relevant?

PR #1426, opened 2021-10-15
Fix for incorrect action display names in Inspector.

ekcoh:
### Description

Fixed an issue where serialized `InputAction` properties would have display name "Input Action" in the Inspector window instead of their given name.

### Changes made

Not sure the suggested fix is the correct or cleanest way to fix this but seems to do  the job. Better suggestions are more than welcome since I am not sure I understand the original intended design.

### Notes

Only verified by examining action asset editor and inspector.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionDrawerBase.cs
```diff
@@ -106,6 +106,8 @@ private static string GetPropertyTitle(SerializedProperty property)
 
             if (property.GetParentProperty() != null && property.GetParentProperty().isArray)
                 propertyTitleNumeral = $" {property.GetIndexOfArrayElement()}";
+            if (property.type == nameof(InputAction) && property.displayName != null && property.displayName.Length > 0)
+                return $"{property.displayName}";
```
andrew-oc: This bug just won't go away. This doesn't work for serialized input actions in a collection and shows the names as Element 0, Input Action, Input Action, Input Action...

andrew-oc (changes requested):

andrew-oc (approved):

PR #1429, opened 2021-10-18
FIX: Memory leak in InputAction.Enable (case 1367442).

Rene-Damm:
### Description

`InputAction.Enable` will often fail to reuse memory allocated prior and thus keep growing internal data structures. The current cleanup code that enables reuse sits in `ProcessStateChangeMonitors`. This is invoked from the event processing path. If no events were processed on the device that had the monitors, no cleanup would occur.

### Changes made

Adding new monitors will now explicitly check whether a cleanup is due and is safe to happen. This restricts failure to reuse to the case where we add monitors from within action processing.

### Notes

Actual leak is in state monitor code but I opted to keep the test at the action level to get more comprehensive coverage and ensure that repeated disabling and enabling of actions (without any other change) does not result in GC activity.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2130,6 +2139,31 @@ public void Clear()
...
+            public void CompactArrays()
+            {
+                for (var i = 0; i < count; ++i)
+                {
+                    var memoryRegion = memoryRegions[i];
+                    if (memoryRegion.sizeInBits != 0)
+                        continue;
+
+                    RemoveAt(i);
+                    --i;
```
ekcoh: Is this really correct? I would have expected a backwards loop starting from count-1 and RemoveAt(i) as you were iterating and avoid fiddling with i, or do I misterpret this?....
I was thinking this was intended? Not?

`
for (var i=count-1; i >= 0; --i)
{
   if (memoryRegions[i].sizeInBits == 0)
        RemoveAt(i);
}
`

ekcoh (changes requested):
Looks good good in general. Only one question/request regarding a loop.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
Rene-Damm: Was correct but your version is better. Changed.

ekcoh (approved):

ekcoh:
Approved, but seems to be test issues remaining, did not reavaluating that `count` property change some behavior?

PR #1430, opened 2021-10-19
NEW: Add generic RacingWheel class.

Rene-Damm:
### Description

Add a new `RacingWheel` generic device class.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
@@ -1,0 +1,248 @@
+using UnityEngine.InputSystem.Controls;
+using UnityEngine.InputSystem.Layouts;
+using UnityEngine.InputSystem.LowLevel;
+using UnityEngine.InputSystem.Utilities;
+
+namespace UnityEngine.InputSystem.LowLevel
+{
+    /// <summary>
+    /// Input state for a <see cref="RacingWheel"/>.
+    /// </summary>
+    /// <remarks>
+    /// <example>
+    /// <code>
+    /// var wheel = InputSystem.AddDevice&lt;RacingWheel&gt;();
+    ///
+    /// InputSystem.QueueStateEvent(wheel,
+    ///     new RacingWheelState
+    ///     {
+    ///         gear = 2,
+    ///         wheel = 0.5f, // Halfway right.
+    ///         throttle = 0.25f, // Quarter pressure on gas pedal.
+    ///     });
+    /// </code>
+    /// </example>
+    /// </remarks>
+    public struct RacingWheelState : IInputStateTypeInfo
+    {
+        public static FourCC Format => new FourCC('W', 'E', 'E', 'L');
+
+        [InputControl(name = "dpad", layout = "Dpad", usage = "Hatswitch", displayName = "D-Pad", format = "BIT", sizeInBits = 4, bit = 0)]
+        [InputControl(name = "dpad/up", format = "BIT", bit = (uint)RacingWheelButton.DpadUp, sizeInBits = 1)]
+        [InputControl(name = "dpad/right", format = "BIT", bit = (uint)RacingWheelButton.DpadRight, sizeInBits = 1)]
+        [InputControl(name = "dpad/down", format = "BIT", bit = (uint)RacingWheelButton.DpadDown, sizeInBits = 1)]
+        [InputControl(name = "dpad/left", format = "BIT", bit = (uint)RacingWheelButton.DpadLeft, sizeInBits = 1)]
+        [InputControl(name = "gearUp", layout = "Button", bit = (uint)RacingWheelButton.GearUp, displayName = "Next Gear")]
+        [InputControl(name = "gearDown", layout = "Button", bit = (uint)RacingWheelButton.GearDown, displayName = "Previous Gear")]
+        [InputControl(name = "menu", alias = "start", layout = "Button", bit = (uint)RacingWheelButton.Menu, displayName = "Menu")]
+        [InputControl(name = "view", alias = "select", layout = "Button", bit = (uint)RacingWheelButton.View, displayName = "View")]
+        public uint buttons;
+
+        /// <summary>
+        /// Current gear selected in gear shift.
+        /// </summary>
+        [InputControl(displayName = "Gear", layout = "Integer")]
+        public int gear;
+
+        /// <summary>
+        /// Position of wheel in [-1,1] range. -1 is all the way left, 1 is all the way right.
+        /// </summary>
+        [InputControl(displayName = "Wheel", layout = "Axis")]
+        public float wheel;
+
+        /// <summary>
+        /// Position of throttle/gas pedal in [0,1] range. 0 is fully depressed, 1 is pressed all the way down.
+        /// </summary>
+        [InputControl(displayName = "Throttle", alias = "gas", layout = "Axis")]
+        public float throttle;
+
+        /// <summary>
+        /// Position of brake pedal in [0,1] range. 0 is fully depressed, 1 is pressed all the way down.
+        /// </summary>
+        [InputControl(displayName = "Break", layout = "Axis")]
+        public float brake;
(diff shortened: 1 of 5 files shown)


Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+namespace UnityEngine.InputSystem
+{
+    /// <summary>
+    /// Base class for racing wheels.
+    /// </summary>
+    [InputControlLayout(stateType = typeof(RacingWheelState), displayName = "Racing Wheel", isGenericTypeOfDevice = true)]
+    public class RacingWheel : InputDevice
```
Rene-Damm, line 151: Biggest question for me is how we want this to relate to `Gamepad`. I think pretty much any wheel out there has controls (such as the face and shoulder buttons) to allow it to function mostly like a gamepad. But not every API (same [GameInput](https://docs.microsoft.com/en-us/gaming/gdk/_content/gc/reference/input/gameinput/structs/gameinputracingwheelstate)) defines wheels that way.

So, guess the question is how much we want to put in here in the abstract definition.

Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
```diff
@@ -0,0 +1,248 @@
...
+    ///     new RacingWheelState
+    ///     {
+    ///         gear = 2,
+    ///         wheel = 0.5f, // Halfway right.
+    ///         throttle = 0.25f, // Quarter pressure on gas pedal.
+    ///     });
+    /// </code>
+    /// </example>
+    /// </remarks>
+    public struct RacingWheelState : IInputStateTypeInfo
```
jimon, line 26: We likely don't need state struct, rather only `RacingWheel` device class, so users can write code against that, as device structs will come with platform specific devices. And HID already has it's own thing anyway.

Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
jimon: Likely shouldn't be a gamepad, but rather dpad from racing wheel should participate in navigation.

Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
```diff
@@ -0,0 +1,248 @@
...
+    {
+        /// <summary>
+        /// The D-Pad control on the wheel.
+        /// </summary>
+        public DpadControl dpad { get; protected set; }
+
+        /// <summary>
+        /// Button to shift gears up.
+        /// </summary>
+        public ButtonControl gearUp { get; protected set; }
```
jimon, line 161: Any idea how https://www.logitechg.com/en-us/products/driving/driving-force-shifter.941-000119.html is exposed on HID layer? Not sure if gear lever needs more work here or not

Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
Rene-Damm: Wondering about additional controls, though. Seems most sensible to add them here, too. Don't think the wheel should separately be represented as a gamepad but it should probably have ABXY face buttons and shoulder buttons instead of just the narrow set that's currently on the device.

Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
Rene-Damm: My assumption is this would use the `gear` control and not the `gearUp`/`gearDown` button controls corresponding to the commonly found levers/buttons on the wheel. But not sure.

Also not sure how it's exposed at the HID layer. My very limited experience so far with Logitech wheels has been that they masquerade as plain XInput devices and have no interesting properties whatsoever as HIDs and need interfacing through the Logitech SDK to get to the advanced stuff. But no clue about the gear shifters.

Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
ShaunHewitt: I agree any additional face buttons should exist with the race wheel device.

Packages/com.unity.inputsystem/InputSystem/Devices/RacingWheel.cs
Rene-Damm: My experience has been that every time we tried to hide the state struct or do away with this, it just ended up being painful because you had no way to send input to the device without using the tedious APIs that write values through individual controls.

It's like `GamepadState`. `Gamepad` has many different variations and possible states but just being able to spawn a gamepad and directly feed it input is quite useful. IMO better to treat all devices the same way there.

jimon: So let's say PS4/PS5 code will need more state, e.g. it will be like DualShock gamepad vs Gamepad, how would that work?

Rene-Damm: Same as for the current DualShock support. I.e. it'd just come up with its own state format.

jimon: Ok but then it's questionable, e.g. you will not able to queue RacingWheelState into device that is racing wheel but implements it's own state format.

ShaunHewitt:
How will the haptics (Force Feedback) side of things be handled? 

Will it be similar to how Gamepad derives from IDualMotorRumble,  maybe having IRaceWheelHaptics?

Rene-Damm:
> How will the haptics (Force Feedback) side of things be handled?
> 
> Will it be similar to how Gamepad derives from IDualMotorRumble, maybe having IRaceWheelHaptics?

Good question. My assumption is that wheel effects will not as readily fit into as simple a box as gamepad rumble. I'm thinking the best way might be to ignore them for now and then take it on as part of the bigger discussion around what to do with haptics. Without some more sophisticated API in place, trying to cast wheel haptics into some `IDualMotorRumble` style interface will probably lead to some pretty poor result, I'd think.

ShaunHewitt:
> Without some more sophisticated API in place, trying to cast wheel haptics into some `IDualMotorRumble` style interface will probably lead to some pretty poor result, I'd think.

Agreed. IDualMotorRumble isn't doing too well for controllers with more than 2 motors atm.

Rene-Damm:
Probably makes sense to close this one.

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1430) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1430) before we can accept your contribution.<br/><hr/>**Rene Damm** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1430) it.</sub>

timcannellunity3d (approved):

ekcoh:
If work would be resumed on this PR it needs to updated with latest changes from develop to not cause CI issues.

PR #1431, opened 2021-10-20
Temporarily disable UI tests from running against trunk

andrew-oc:
Set the TEMP_DISABLE_UI_TESTS_ON_TRUNK conditional compile flag to true when compiling against a version of Unity between 2022.1.0a12 and 2022.1.0a16 to avoid test failures when running through UTR due to a uGUI change.


jimon (approved):
Let's just not forget to clean up this later :)

ekcoh (approved):

PR #1432, opened 2021-10-20
FIX: API based adding of composite bindings

reapazor:
### Description

Composite bindings need to have their name set when created via API, otherwise, a null ref occurs when viewing in the action editor. It most likely doesn't occur when creating via the UI as it drops to renaming the composite when you add a new one.

### Changelist
* Automatically generate a name for composites being added via the API.
* Minor tweak for an ambiguous string compare


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1432) <br/>All committers have signed the CLA.

ekcoh:
Part of this change seem to overlap another pending PR: https://github.com/Unity-Technologies/InputSystem/pull/1425/files

reapazor:
Do you want me to incorporate/update to use that instead?

ekcoh:
@reapazor No worry we handle that in merge, that change haven't landed yet anyway. Regarding failed tests let it sit just briefly since we are resolving some other CI issues at the moment. I'll get back to you if any additional action is needed here.

Rene-Damm:
Thank you for the fix. Integrated and merged as part of #1477.

(Sorry, @reapazor, only belatedly realize you're Matthew Davey and not an outside contributor)

PR #1440, opened 2021-10-26
CHANGE: Make Keyboard ctrl/alt/shift writable (case 1340793).

Rene-Damm:
Fixes [1340793](https://issuetracker.unity3d.com/issues/on-screen-button-errors-on-mouse-down-slash-up-when-its-control-path-is-set-to-control-keyboard) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1340793/)).

### Description

`DiscreteButtonControl`, which we use for the synthetic ctrl/alt/shift combined controls on the keyboard, throws `NotSupportedException` when calling `WriteValueIntoState`.

While for uses such as the  PS4 dpad this is indeed justified (it's an enumeration where each button has multiple pressed states so writing a single button in isolation is undefined), for the keyboard modifier keys, this operation is in fact unambiguous.

### Changes made

Added a `DiscreteButtonControl.writeMode` parameter that allows toggling on a simple write path. It'll simply write `nullValue` for "off" and `maxValue` for on. This is sufficient for the case of keyboard modifiers but not sufficient for PS4 dpad controls.

### Notes

Also fixed two secondary issues I discovered

* `DeltaStateEvent.From` only took the bit offset but not the bit size into account for bit-addressed controls.
* Both `StateEvent` and `DeltaStateEvent` do not ensure the resulting buffer has a length of a multiple of 4 (which could lead to invalid memory accesses in the event code).

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


jimon (approved):
Ok so it's an opt-in, and DualShock dpad is not supported.

Rene-Damm:
> Ok so it's an opt-in, and DualShock dpad is not supported.

Yup, exactly. Don't think the DualShock case is solvable as long as individual controls can be written to. But maybe I'm just not seeing what could be a perfectly decent solution here.

PR #1443, opened 2021-10-27
FIX: Navigation not being confined to multiplayer playerRoot (case 1306361).

Rene-Damm:
Fixes [1306361](https://issuetracker.unity3d.com/issues/input-system-ui-input-module-lets-the-player-navigate-across-other-canvases) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1306361/)).

### Description

When setting up a multiplayer UI using `MultiplayerEventEvent`, pointer input is confined to the `playerRoot` of each player. However, navigation input -- itself much more relevant to multiplayer input compared to pointer input -- is allowed to escape a `playerRoot` and thus makes it possible for one player to move onto the UI of another player.

### Changes made

The navigation mechanism built into uGUI's `Selectable` is not overridable or extensible. We thus cannot limit it to parts of the UI. Discussing this with @pcosgrave.

This PR attempts a workaround by relying on the `CanvasGroup` mechanism that allows toggling the `interactable` property off for an entire hierarchy and thus all `Selectable` objects in the hierarchy.

### Notes

`CanvasGroup` is not a fast mechanism. The underlying implementation relies on `BroadcastMessage` to reach `Selectable` objects. Might be good enough as a stopgap but likely, we want a more lightweight mechanism to get `Selectable` to only pick certain UI objects.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MultiplayerEventSystem.cs
```diff
@@ -19,14 +19,80 @@ public class MultiplayerEventSystem : EventSystem
...
+            InitializeCanvasGroup();
+        }
+
+        private void InitializeCanvasGroup()
+        {
+            if (m_PlayerRoot != null)
+            {
+                m_CanvasGroup = m_PlayerRoot.GetComponent<CanvasGroup>();
+                if (m_CanvasGroup == null)
+                    m_CanvasGroup = m_PlayerRoot.AddComponent<CanvasGroup>();
```
ekcoh, line 70: Is this add free some side-effects? I assume CanvasGroup has other implications? Reading through the docs I guess this could work just fine if "part of solution" for local multiplayer UI and having two "systems" isn't a viable solution in itself to create two navigation trees that are disjoint.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MultiplayerEventSystem.cs
```diff
@@ -19,14 +19,80 @@ public class MultiplayerEventSystem : EventSystem
...
+            {
+                m_CanvasGroup = null;
+            }
+        }
+
+        protected override void OnDisable()
+        {
+            var index = s_MultiplayerEventSystems.IndexOfReference(this);
+            if (index != -1)
+                s_MultiplayerEventSystems.EraseAtWithCapacity(ref s_MultiplayerEventSystemCount, index);
```
ekcoh, line 86: In case CanvasGroup have been added by OnEnable(), should OnDisable() remove it in that case for symmetry? The user didn't intentionally added it.

ekcoh (approved):
Looks good to me.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/MultiplayerEventSystem.cs
Rene-Damm: There's a range of implications and side-effects -- most importantly, it has pretty poor performance. This is more a workaround than a proper solution. Unfortunately, with the tools we have available ATM, it doesn't seem like there's a better solution for the moment. Overall, think this is part of some extensions to uGUI that are needed to support this use case properly. Most of `MultiplayerEventSystem` is a workaround.

PR #1446, opened 2021-10-27
FIX: Missed inputs with multiple interactions and inputs (case 1364667).

Rene-Damm:
Fixes [1364667](https://issuetracker.unity3d.com/issues/new-input-system-stops-working-after-pressing-2-keyboard-buttons-at-the-same-time) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1364667/)).

### Description

When an action is bound to multiple controls and there's concurrent inputs on more than one control, conflict resolution kicks in. So, if you bind an action to the A and S key and then press both, there's code to track those concurrent actuations. And if you then release only one key, this code will go and keep tracking the key that is still pressed.

This works fine if there's no interactions in play. But if there are, the current code leads to inconsistent outputs as only the control changes are tracked.

### Changes made

The code now also performs switching of interactions. So, if we move from one control to another and that switch also moves to a different binding, the code transitions from the current interaction (if any) on the old binding to whatever in-progress interaction on the new binding (if any).

### Notes

This PR also fixes a similar problem with `InputAction.IsPressed`/`WasPressedThisFrame`/`WasReleasedThisFrame` not correctly tracking when there's multiple concurrent inputs.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1392,6 +1392,30 @@ private bool ShouldIgnoreControlStateChange(ref TriggerState trigger, int action
...
+                            ResetInteractionState(trigger.mapIndex, actionState->bindingIndex, actionState->interactionIndex);
+
+                        // If there's an interaction in progress on the new binding, let
+                        // it drive the action.
+                        var bindingState = &bindingStates[bindingWithHighestActuation];
+                        var interactionCount = bindingState->interactionCount;
+                        var interactionStartIndex = bindingState->interactionStartIndex;
+                        for (var i = 0; i < interactionCount; ++i)
+                        {
+                            if (!interactionStates[interactionStartIndex + i].phase.IsInProgress())
```
ekcoh, line 1410: Since we compute interactionStartIndex + i 3 times within this for maybe compute once as a local and reuse?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1179,7 +1182,7 @@ private static bool ShouldIgnoreControlStateChangeOnCompositeBinding(BindingStat
         /// from all controls. However, it would not necessarily result in more useful conflict resolution and would
         /// at the same time be much more expensive.
         /// </remarks>
-        private bool ShouldIgnoreControlStateChange(ref TriggerState trigger, int actionIndex)
+        private bool IsConflictingInput(ref TriggerState trigger, int actionIndex)
```
ekcoh, line 1185: Its a bit confusing that a Is-prefixed function updates state so maybe call this function something else since it does checking and modifies state?

ekcoh (approved):

andrew-oc (approved):
LGTM

PR #1447, opened 2021-10-28
FIX: Incorrect writes into stick half-axes (case 1336240).

Rene-Damm:
Fixes [1336240](https://issuetracker.unity3d.com/issues/inputtestfixture-tests-return-inverted-values-when-pressing-gamepads-left-or-down-joystick-buttons) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1336240/)).

### Description

We don't "unprocess" values when we write them. This is a general problem. So if you do this in a test on current `develop`

```CSharp
Set(gamepad.leftStick.left, 1f);
```

the resulting stick value will be `Vector2.right` instead of `Vector2.left`.

### Changes made

Decided to not tackle the bigger issue but rather just fix this for `AxisControl` only. Unlike some other controls, `AxisControl` does most of its processing through built-in parameters rather than relying on processor stacks. So I just added code to do "unprocessing" based on those parameters (minus clamping as we don't know what the unclamped value should be).

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Controls.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Controls.cs
@@ -317,6 +317,36 @@ partial class CoreTests
             Is.EqualTo(new AxisDeadzoneProcessor().Process(0.5f)));
     }
 
+    // https://fogbugz.unity3d.com/f/cases/1336240/
+    [Test]
+    [Category("Controls")]
+    public void Controls_CanWriteIntoHalfAxesOfSticks()
+    {
+        // Disable deadzoning.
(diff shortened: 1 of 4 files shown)


jimon (approved):
Seems ok to solve this particular problem. In the future let's take this into account during backend refactoring

PR #1449, opened 2021-10-28
FIX: Successive UI clicks not working with TouchSimulation (case 1330014).

Rene-Damm:
Fixes [1330014](https://issuetracker.unity3d.com/issues/onclick-isnt-triggered-on-the-second-click-when-the-mouse-isnt-moved-and-simulate-touch-input-from-mouse-or-pen-is-enabled) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1330014/)).

### Description

With `TouchSimulation` on, clicking the same UI button without moving the mouse would not result in clicks. Debugging reveals that `InputSystemUIInputModule` would in fact generate `OnClick` events but would do so at position `(0,0)`. This is because without input on the `Point` action, the code does not know the position of the pointer.

### Changes made

When we allocate a new pointer state for touches, we now also sync the current touch position.

### Notes

This fixes the immediate problem but there's three underlying issues that likely need to be addressed and found better solutions for.

1. Currently the only way to make sure that initial values are making it through is initial state checks.
2. For pass-through actions collecting to N inputs, there's currently no way to poll current values other than just going directly to bound controls and reading them out one-by-one.
3. The pointer-related inputs are all on separate actions when really, we want a composite pointer input that binds together position and button states from one particular source. Actions currently have no way of doing that. You can bind inputs together in a binding composite. But, in the presence of pass-through input sinks as well as multiple devices of the same type (e.g. `<Mouse>/position` may yield arbitrary mouse position inputs depending on the number of mice present), composites alone are not enough to set up independent input channels.

In fact, if actions had a polling API for pass-through style input sinks, a great deal of complexity in `InputSystemUIInputModule` would simply get dropped. And the problem in the original ticket here along with it.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


jimon (approved):
Seems almost exactly the same as https://github.com/Unity-Technologies/InputSystem/pull/1313 which I haven't landed due to broken tests, so let's close both tickets (they are probably duplicates) when this lands

PR #1450, opened 2021-11-02
Fixing the problem with ui input inspector not showing action references for runtime asset

jimon:
### Description

When two player inputs with two ui input modules point to the same asset, one of them will get a runtime created copy of actions asset (because of `m_Actions = Instantiate(m_Actions);`). That makes one of ui input modules inspector showing everything as "None" because `GetAllActionsFromAsset` was returning empty array.

### Changes made

I've thought about it, and it doesn't seem to make sense to resolve `InputActionReference` for a runtime created asset, because to be able to resolve it later on one would need the asset instance. So I made inspector to fallback to "read-only" mode.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
@@ -135,6 +135,7 @@ namespace UnityEngine.InputSystem
 
             m_Asset = asset;
             m_ActionId = action.id.ToString();
+            name = GetDisplayName(action);
 
             ////REVIEW: should this dirty the asset if IDs had not been generated yet?
         }
@@ -157,6 +158,19 @@ namespace UnityEngine.InputSystem
             }
 
             return base.ToString();
+        }
+
+        private static string GetDisplayName(InputAction action)
+        {
+            return !string.IsNullOrEmpty(action?.actionMap?.name) ? $"{action.actionMap?.name}/{action.name}" : action?.name;
+        }
+
+        /// <summary>
+        /// Return a string representation useful for showing in UI.
+        /// </summary>
+        internal string ToDisplayName()
+        {
+            return string.IsNullOrEmpty(name) ? GetDisplayName(action) : name;
         }
 
         /// <summary>
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
@@ -161,7 +161,7 @@ namespace UnityEngine.InputSystem.UI.Editor
                     if (m_ReferenceProperties[i].objectReferenceValue != null &&
                         (m_ReferenceProperties[i].objectReferenceValue is InputActionReference reference))
                         retrievedName = MakeActionReferenceNameUsableInGenericMenu(reference.ToDisplayName());
-                
+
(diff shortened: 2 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
```diff
@@ -159,6 +160,19 @@ public override string ToString()
             return base.ToString();
         }
 
+        private static string GetNiceName(InputAction action)
```
ekcoh: Minor: Not sure whats so nice about it :) but maybe GetUserFriendlyName  or GetDisplayName is more conventional?

ekcoh (approved):
Probably good to get someone elses eyes on this as well (since I am note sure I understand everything that is going on here) but looks good to me

andrew-oc (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
```diff
@@ -122,17 +130,34 @@ public override void OnInspectorGUI()
...
+                    if (EditorGUI.EndChangeCheck())
+                        m_ReferenceProperties[i].objectReferenceValue =
+                            index > 0 ? m_AvailableActionReferencesInAssetDatabase[index - 1] : null;
+                }
+            }
+            else
+            {
+                // Somehow we have an asset but no asset references from the database, pull out references manually and show them in read only UI
+                EditorGUILayout.HelpBox("Showing fields as read-only because current action asset seems to be created by a script and assigned programmatically.", MessageType.Info);
+                for (var i = 0; i < numActions; i++)
```
Rene-Damm, line 158: Probably better in disabled block (`EditorGUI.BeginDisabled`/`EndDisabled` or `DisabledScope`) to actually make read-only. Bit confusing that the UI is actually not read-only. Also, this way the codepaths can probably remain the same.

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1450) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1450) before we can accept your contribution.<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1450) it.</sub>

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
jimon: Made `EditorGUI.BeginDisabledGroup`, looks much better now 👍 

Haven't managed to combine both codepathes though, they do seem to be very different indeed, I'll keep them separate instead.

PR #1451, opened 2021-11-02
FIX: Special chars in HID names causing rebinding to not work (case 1335465).

Rene-Damm:
Fixes [1335465](https://issuetracker.unity3d.com/issues/general-certain-hid-joysticks-fail-to-send-input-on-windows-even-though-interactive-rebind-works) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1335465/)).

### Description

In control paths, characters like (/, <, >, {, and }) take on special meaning. Backslashes are used to escape this.

However, the current code takes many strings (such as product names) as is and puts them into control paths where they are then read incorrectly.

### Changes made

The code now escapes and unescapes strings in several places. Also, the / character takes on special meaning only while not inside a {..} or <..> group.

### Notes

`InputControlPath` is still waiting for a proper refactor which I chose not to roll into this bugfix. There's the `PathParser` code now and `MatchPathComponent` should be brought over to it. All code operating on path structures should be rolled over into using `PathParser`.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -97,6 +97,19 @@ public static class InputControlPath
 
         public const char Separator = '/';
 
+        // We consider / a reserved character in control names. So, when this character does creep
```
ekcoh, line 100: What if the original string contains space?

ekcoh: Maybe its less likely that this string UTF-8(?) from lower character table is not present or are we sure there is no "ids" containing space?

ekcoh (approved):
Approved but one question/concern regarding using space as replacement char, maybe we already got it covered?

Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -1312,5 +1311,54 @@ public void Utilities_CanRecognizeVendorDefinedUsages()
...
         Assert.That(HID.UsageToString((HID.UsagePage) 0xff01, 0x33), Is.Null);
     }
+
+    // https://fogbugz.unity3d.com/f/cases/1335465/
+    [Test]
+    [Category("Devices")]
+    [TestCase('/')]
+    [TestCase('\\')]
+    [TestCase('>')]
+    [TestCase('<')]
```
ekcoh: Maybe add a test case containing a space?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -912,22 +912,24 @@ internal static string BuildPath(this InputControl control, string deviceLayout,
             var device = control.device;
 
             builder.Append('<');
-            builder.Append(deviceLayout);
+            builder.Append(deviceLayout.Replace("\\", "\\\\").Replace(">", "\\>"));
```
andrew-oc: Curious why we don't use our own StringHelper.Escape method for all this escaping?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
Rene-Damm: Spaces are fine. They have no special meaning in control paths.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
Rene-Damm: Fixed

Assets/Tests/InputSystem/Plugins/HIDTests.cs
Rene-Damm: Done

andrew-oc (approved):

PR #1456, opened 2021-11-04
Fixing device current getting set when a state event with no changes is sent

jimon:
### Description

Some backends send a state event every once in a while (every frame, twice a frame, etc) without even checking if anything changed. That is currently making devices from these backends as `.current`.

### Changes made

Always memcmp new state with previous one, and only make device current if anything changed (or non-noisy data changed for noisy devices).

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -4033,15 +4033,6 @@ partial class CoreTests
         Assert.That(gamepad1.rightTrigger.noisy, Is.False);
         Assert.That(Gamepad.current, Is.SameAs(gamepad2));
 
-        var receivedSettingsChange = false;
-        InputSystem.onSettingsChange += () => receivedSettingsChange = true;
-
-        // Enable filtering. Off by default.
-        InputSystem.settings.filterNoiseOnCurrent = true;
-
-        Assert.That(InputSystem.settings.filterNoiseOnCurrent, Is.True);
-        Assert.That(receivedSettingsChange, Is.True);
-
         // Send delta state without noise on first gamepad.
         InputSystem.QueueDeltaStateEvent(gamepad1.leftStick, new Vector2(0.123f, 0.234f));
         InputSystem.Update();
@@ -4075,13 +4066,6 @@ partial class CoreTests
         Assert.That(Gamepad.current, Is.SameAs(gamepad1));
     }
 
-    [Test]
-    [Category("Devices")]
-    public void Devices_FilteringNoiseOnCurrentIsTurnedOffByDefault()
-    {
-        Assert.That(InputSystem.settings.filterNoiseOnCurrent, Is.False);
-    }
-
     // We currently do not read out actual values during noise detection. This means that any state change on a control
     // that isn't marked as noisy will pass the noise filter. If, for example, the sticks are wiggled but they are still
     // below deadzone threshold, they will still classify as carrying signal. To do that differently, we would have to
@@ -4092,8 +4076,6 @@ partial class CoreTests
     {
         var gamepad1 = InputSystem.AddDevice<Gamepad>();
         InputSystem.AddDevice<Gamepad>();
-
-        InputSystem.settings.filterNoiseOnCurrent = true;
 
         // Actuate leftStick below deadzone threshold.
         InputSystem.QueueStateEvent(gamepad1, new GamepadState { leftStick = new Vector2(0.001f, 0.001f)});
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -3975,7 +3975,7 @@ partial class CoreTests
 
         Assert.That(Gamepad.current, Is.Not.SameAs(gamepad1));
 
-        InputSystem.QueueStateEvent(gamepad1, new GamepadState());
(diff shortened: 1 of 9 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1456) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1456) before we can accept your contribution.<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1456) it.</sub>

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -3975,7 +3975,7 @@ public void Devices_AreMadeCurrentWhenReceivingStateEvent()
 
         Assert.That(Gamepad.current, Is.Not.SameAs(gamepad1));
 
-        InputSystem.QueueStateEvent(gamepad1, new GamepadState());
+        InputSystem.QueueStateEvent(gamepad1, new GamepadState().WithButton(GamepadButton.A));
```
Rene-Damm, line 3978: Ideally, also test the opposite case here. That just sending a state event only shouldn't result in a change.

Rene-Damm: Ah doh, nvm, that's a separate test down there. Perfect.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3702,18 +3702,13 @@ internal unsafe bool UpdateState(InputDevice device, InputEvent* eventPtr, Input
...
-                    !MemoryHelpers.MemCmpBitRegion(deviceStatePtr, statePtr,
-                        0, stateSize * 8, mask: noiseMask);
-            }
+            var noiseMask = device.noisy && m_Settings.filterNoiseOnCurrent
+                ? (byte*)InputStateBuffers.s_NoiseMaskBuffer + deviceStateOffset
+                : null;
+            // Compare the current state of the device to the newly received state but overlay
+            // the comparison by the noise mask.
+            var makeDeviceCurrent = !MemoryHelpers.MemCmpBitRegion(deviceStatePtr, statePtr,
+                0, stateSize * 8, mask: noiseMask);
```
Rene-Damm, line 3726: Hmm, I wonder whether we should `if` the check here behind `m_Settings.filterNoiseOnCurrent`. I.e. like

```
var makeDeviceCurrent = true;
if (m_Settings.filterNoiseOnCurrent)
{
            var noiseMask = device.noisy && m_Settings.filterNoiseOnCurrent
                ? (byte*)InputStateBuffers.s_NoiseMaskBuffer + deviceStateOffset
                : null;
            // Compare the current state of the device to the newly received state but overlay
            // the comparison by the noise mask.
            makeDeviceCurrent = !MemoryHelpers.MemCmpBitRegion(deviceStatePtr, statePtr,
                0, stateSize * 8, mask: noiseMask);
}
```

The intention behind `filterNoiseOnCurrent` is to disable any kind of filtering on `current`.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ however, it has to be formatted properly to pass verification tests.
 
 - Fixed writing values into the half-axis controls of sticks (such as `Gamepad.leftStick.left`) producing incorrect values on the stick ([case 1336240](https://issuetracker.unity3d.com/issues/inputtestfixture-tests-return-inverted-values-when-pressing-gamepads-left-or-down-joystick-buttons)).
 - Fixed setting size of event trace in input debugger always growing back to largest size set before.
+- Fixed a device becoming `.current` (e.g. `Gamepad.current`, etc) when sending a new state event that contains no control changes (case 1377952).
```
ekcoh, line 31: Any public issue tracker link available?

ekcoh (commented):
Will just comment since @Rene-Damm had some questions

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jimon: `filterNoiseOnCurrent` is false by default though, maybe we can make it true by default and then I can wrap whole machinery under this setting? e.g. I mean I want memcmp checks to be on by default

Packages/com.unity.inputsystem/CHANGELOG.md
jimon: Not for this one

andrew-oc (approved):
The code looks good but I'm wondering about whether this is a good behaviour change. Do any devices exist that will continuously send input events for held controls? In those cases, you kind of want to keep that device current even though the state hasn't changed. It feels like this change here would be better under device local control, where we can say ok, we know this particular device sends random events, so we're going to do the more involved memcmp every time, and the global behaviour remains 'set device current on any non-noisy input'.

PR #1457, opened 2021-11-04
FIX: InputSystemUIInputModule shows incorrect bindings after pressing the PlayerInput 'Fix UI Input Module' button

andrew-oc:
### Description

Pressing the 'Fix UI Input Module' button in the PlayerInput component causes all bindings in the related InputSystemUIInputModule component to display as 'None'.

### Changes made

When we use the InputSystemUIInputModuleEditor.ReassignActions method, we create all new instances of InputActionReferences, and then later when displaying the binding dropdowns, we compare those instances to the instances assigned to the InputSystemUIInputModule component using reference equality, which always fail. Changed the comparison to work of the wrapped InputActions instead.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
```diff
@@ -127,7 +127,22 @@ public override void OnInspectorGUI()
                 if (m_AvailableActionsInAsset == null)
                     continue;
 
-                var index = Array.IndexOf(m_AvailableActionsInAsset, m_ReferenceProperties[i].objectReferenceValue) + 1;
+                // find the input action reference from the asset that matches the input action reference from the
+                // InputSystemUIInputModule that is currently selected. Note we can't use reference equality of the
+                // two InputActionReference objects here because in ReassignActions above, we create new instances
+                // every time it runs.
+                var index = 0;
+                var inputAction = ((InputActionReference)m_ReferenceProperties[i]?.objectReferenceValue)?.action;
```
ekcoh: Just a thought, and minor, but if objectReferenceValue and action here are both possible to be null, maybe skip the whole for-loop thing if inputAction is null?

ekcoh (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
andrew-oc: Done

Rene-Damm (approved):

PR #1460, opened 2021-11-08
Resync all devices after entering play mode

jimon:
### Description

As part of fixing [case 1231907](https://issuetracker.unity3d.com/issues/mouse-coordinates-reported-as-00-until-the-first-move) I've noticed that even if I implemented sync for the mouse it still doesn't work, it's because we are dropping all events generated between exiting edit mode and entering play mode here:

```
                    // In the editor, we discard all input events that occur in-between exiting edit mode and having
                    // entered play mode as otherwise we'll spill a bunch of UI events that have occurred while the
                    // UI was sort of neither in this mode nor in that mode. This would usually lead to the game receiving
                    // an accumulation of spurious inputs right in one of its first updates.
                    //
                    // NOTE: There's a chance the solution here will prove inadequate on the long run. We may do things
                    //       here such as throwing partial touches away and then letting the rest of a touch go through.
                    //       Could be that ultimately we need to issue a full reset of all devices at the beginning of
                    //       play mode in the editor.
#if UNITY_EDITOR
                    if ((updateType & InputUpdateType.Editor) == 0 &&
                        InputSystem.s_SystemObject.exitEditModeTime > 0 &&
                        currentEventTimeInternal >= InputSystem.s_SystemObject.exitEditModeTime &&
                        (currentEventTimeInternal < InputSystem.s_SystemObject.enterPlayModeTime ||
                         InputSystem.s_SystemObject.enterPlayModeTime == 0))
                    {
                        m_InputEventStream.Advance(false);
                        continue;
                    }
#endif
```

Currently devices are only resynced when they're recreating during domain reload, and that means the resync state event currently is generated during domain reload, hence it's ignored.

### Changes made

Ideally the ignoring logic should not ignore resyncs, but that implies some sort of flag or metadata added to events. But even if we do this, it's unclear what to do with a proper input event that might come after resync event. Maybe instead all listeners should be skipped during domain reload? Likely something else will break then.

I couldn't really figure out how to properly fix it at this stage, feels like we need to separate "calculating control values" step from "calling callbacks" step more cleanly, then we could store all updates but not trigger any UI stuff.

So for now I did a very hacky fix to just resync the universe in a frame after play mode, I think this effectively means that first scene update after play mode will have no data. Not ideal, but almost fixes the problem.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -2526,27 +2526,40 @@ namespace UnityEngine.InputSystem
             #endif
         }
 
+#if UNITY_EDITOR
         private void SyncAllDevicesWhenEditorIsActivated()
         {
-            #if UNITY_EDITOR
             var isActive = m_Runtime.isEditorActive;
             if (isActive == m_EditorIsActive)
                 return;
 
             m_EditorIsActive = isActive;
             if (m_EditorIsActive)
-            {
-                for (var i = 0; i < m_DevicesCount; ++i)
-                {
-                    // When the editor comes back into focus, we actually do want resets to happen
-                    // for devices that don't support syncs as they will likely have missed input while
-                    // we were in the background.
-                    if (!m_Devices[i].RequestSync())
-                        ResetDevice(m_Devices[i], issueResetCommand: true);
-                }
-            }
-            #endif
+                SyncAllDevices();
         }
+
+        private void SyncAllDevices()
+        {
+            for (var i = 0; i < m_DevicesCount; ++i)
+            {
+                // When the editor comes back into focus, we actually do want resets to happen
+                // for devices that don't support syncs as they will likely have missed input while
+                // we were in the background.
+                if (!m_Devices[i].RequestSync())
+                    ResetDevice(m_Devices[i], issueResetCommand: true);
+            }
+        }
+
+        internal void SyncAllDevicesAfterEnteringPlayMode()
+        {
+            // Because we ignore all events between exiting edit mode and entering play mode,
+            // that includes any potential device resets/syncs/etc,
+            // we need to resync all devices after we're in play mode proper.
+            ////TODO: this is a hacky workaround, implement a proper solution where events from sync/resets are not ignored.
+            SyncAllDevices();
+        }
+
+#endif
 
         private void WarnAboutDevicesFailingToRecreateAfterDomainReload()
         {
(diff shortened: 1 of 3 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1460) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1460) before we can accept your contribution.<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1460) it.</sub>

ekcoh (approved):
Looks good, do we need an internal ticket on making a proper solution for the TODO filed as part of closing this?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2526,28 +2526,41 @@ private void RestoreDevicesAfterDomainReloadIfNecessary()
...
             var isActive = m_Runtime.isEditorActive;
             if (isActive == m_EditorIsActive)
                 return;
 
             m_EditorIsActive = isActive;
             if (m_EditorIsActive)
+                SyncAllDevices();
+        }
+
+        private void SyncAllDevices()
```
andrew-oc, line 2541: Any reason that SyncAllDevices isn't internal and used instead of SyncAllDevicesAfterEnteringPlayMode?

jimon: I think `SyncAllDevices` is way too leaky, like, sync why? And allowing it to be called from other places might have unexpected consequences.
In this case `SyncAllDevicesAfterEnteringPlayMode` is very specific about why, and in the future refactoring we can easier reimplement it some other way, or nuke it all together.
But agree this is all purely syntax sugar.

andrew-oc: Nice

andrew-oc (approved):

PR #1462, opened 2021-11-08
FIX: Broken UI input when switching between assets (case 1371332).

Rene-Damm:
Fixes [1371332](https://issuetracker.unity3d.com/issues/ui-navigation-stops-working-after-adding-two-input-devices-to-a-scene) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1371332/)).

### Description

When assigning `InputSystemUIInputModule.actionsAsset` while actions were disable `HookActions` was not called which subsequently led to broken input.

### Changes made

Pulled `HookActions` out of the `if` block.

### Notes

Also discovered that `IsAnyActionEnabled` returned `true` if any of the actions was `null`.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1406,16 +1406,16 @@ private bool HasNoActions()
...
+            return (m_PointAction?.action?.enabled ?? false) &&
+                (m_LeftClickAction?.action?.enabled ?? false) &&
+                (m_RightClickAction?.action?.enabled ?? false) &&
+                (m_MiddleClickAction?.action?.enabled ?? false) &&
+                (m_MoveAction?.action?.enabled ?? false) &&
+                (m_SubmitAction?.action?.enabled ?? false) &&
+                (m_CancelAction?.action?.enabled ?? false) &&
+                (m_ScrollWheelAction?.action?.enabled ?? false) &&
+                (m_TrackedDeviceOrientationAction?.action?.enabled ?? false) &&
+                (m_TrackedDevicePositionAction?.action?.enabled ?? false);
```
andrew-oc: Hmmm, this was possibly correct before, but the name of the method was wrong. The logic of "if any action is unassigned when I set a new asset on the input module, but all other actions are enabled, enable all actions" seems better than the alternative. Maybe the method could be renamed to IsAnyActionEnabledOrUnassigned?

Rene-Damm: Yup, you're right. Classic mistake of reading the booleans wrong. Still the previous version was wrong, too. If you have *no* actions at all, it means it *will* automatically enable actions when you assign some.

I think this should read

```
            return (m_PointAction?.action?.enabled ?? false) ||
                (m_LeftClickAction?.action?.enabled ?? false) ||
                (m_RightClickAction?.action?.enabled ?? false) ||
                (m_MiddleClickAction?.action?.enabled ?? false) ||
                (m_MoveAction?.action?.enabled ?? false) ||
                (m_SubmitAction?.action?.enabled ?? false) ||
                (m_CancelAction?.action?.enabled ?? false) ||
                (m_ScrollWheelAction?.action?.enabled ?? false) ||
                (m_TrackedDeviceOrientationAction?.action?.enabled ?? false) ||
                (m_TrackedDevicePositionAction?.action?.enabled ?? false);
```

Rene-Damm: Removed the method altogether and changed the logic of the `actionsAsset` setter.

PR #1464, opened 2021-11-09
FIX: Calling IsPressed() on device being true (case 1374024).

Rene-Damm:
Fixes [1374024](https://issuetracker.unity3d.com/issues/inputcontrol-dot-ispressed-always-returns-true-when-using-new-input-system) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1374024/)).

### Description

`IsPressed` is an extension method for `InputControl`s that checks whether actuation levels are above button press thresholds. As it is defined for any `InputControl`, it also works for an `InputDevice`. The current implementation returns true as long as the device is not in its default state.

### Changes made

Made the check more nuanced to handle "EvaluateMagnitude() has no defined result" better.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
@@ -41,6 +41,7 @@ namespace UnityEngine.InputSystem
             return null;
         }
 
+        ////REVIEW: This ist too high up in the class hierarchy; can be applied to any kind of control without it being readily apparent what exactly it means
         /// <summary>
         /// Check whether the given control is considered pressed according to the button press threshold.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -99,8 +100,14 @@ public static bool IsActuated(this InputControl control, float threshold = 0)
...
             if (magnitude < 0)
             {
-                ////REVIEW: we probably want to do a value comparison on this path to compare it to the default value
-                return true;
+                // We know the control is not in default state but we also know it doesn't support
+                // magnitude. So, all we can say is that it is actuated. Not how much it is actuated.
+                //
+                // If we're looking for a specific threshold here, consider the control to always
+                // be under. But if not, consider it actuated "by virtue of not being in default state".
+                if (Mathf.Approximately(threshold, 0))
```
ekcoh, line 108: Why not just `return Mathf.Approximately(threshold, 0) ?` instead of evaluating twice?

ekcoh (approved):

PR #1465, opened 2021-11-10
FIX: Touch.activeTouches not canceled on focus loss (case 1364017).

Rene-Damm:
Fixes [1364017](https://issuetracker.unity3d.com/issues/input-system-new-input-system-registering-active-touches-when-app-loses-focus) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1364017/)).

### Description

When focus is lost, devices are reset and thus get reset to default state. However, for `Touchscreen` this is undesirable as touches should cancel cleanly, not just disappear.

### Changes made

Modified `Touchscreen` to have its own reset logic.

### Notes

There's a bigger problem in here around how we cannot tell why a specific state change happens. So `Touchscreen.Reset()` has to jump through some hoops an actually construct an event that, in the end, isn't actually used other than to check whether there is an event.

Recorded a Tech Debt ticket for this ([ISX-930](https://jira.unity3d.com/browse/ISX-930)).

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
@@ -1160,30 +1160,20 @@ internal class EnhancedTouchTests : CoreTestsFixture
         {
             // When running in the background, next update after focus loss sees touches cancelled
             // and update after that sees them gone.
-
             InputSystem.Update(InputUpdateType.Dynamic);
-
-            Assert.That(Touch.activeTouches, Has.Count.EqualTo(1));
-            Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Canceled));
-
-            InputSystem.Update();
-
-            Assert.That(Touch.activeTouches, Is.Empty);
         }
         else
         {
             // When not running in the background, the same thing happens but only on focus gain.
-
             runtime.PlayerFocusGained();
-
             InputSystem.Update();
-
-            Assert.That(Touch.activeTouches, Has.Count.EqualTo(1));
-            Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Canceled));
-
-            InputSystem.Update();
-
-            Assert.That(Touch.activeTouches, Is.Empty);
         }
+
+        Assert.That(Touch.activeTouches, Has.Count.EqualTo(1));
+        Assert.That(Touch.activeTouches[0].phase, Is.EqualTo(TouchPhase.Canceled));
+
+        InputSystem.Update();
+
+        Assert.That(Touch.activeTouches, Is.Empty);
     }
 }
--- a/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
@@ -1140,4 +1140,50 @@ internal class EnhancedTouchTests : CoreTestsFixture
         Assert.That(Touchscreen.current.touches[0].isInProgress, Is.True);
         Assert.That(Touchscreen.current.touches[0].position.ReadValue(), Is.EqualTo(new Vector2(123, 234)));
     }
+
+    [Test]
+    [Category("EnhancedTouch")]
+    [TestCase(true)]
+    [TestCase(false)]
+    public void EnhancedTouch_ActiveTouchesGetCanceledOnFocusLoss_WithRunInBackgroundBeing(bool runInBackground)
+    {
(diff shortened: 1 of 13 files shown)


Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
```diff
@@ -1140,4 +1140,50 @@ public unsafe void EnhancedTouch_TouchSimulation_DisablesPointerDevicesWithoutDi
...
+        runtime.PlayerFocusLost();
+
+        if (runInBackground)
+        {
+            // When running in the background, next update after focus loss sees touches cancelled
+            // and update after that sees them gone.
+
+            InputSystem.Update(InputUpdateType.Dynamic);
+
+            Assert.That(Touch.activeTouches, Has.Count.EqualTo(1));
```
ekcoh: Move 2*Asserts + Update() + final Assert outside if-else if equal stimuli/behavior is required/expected to avoid duplicated code?

ekcoh: By the way I liked this test case, easy to follow and understand what is expected and is going on.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -13,17 +13,19 @@ however, it has to be formatted properly to pass verification tests.
...
 - Fixed an issue where UI button clicks could be ignored by `InputSystemUIInputModule` if modifying on-screen devices from Update() callbacks [1365070](https://issuetracker.unity3d.com/product/unity/issues/guid/1365070).
+- Fixed `Touch.activeTouches` still registering touches after the app loses focus ([case 1364017](https://issuetracker.unity3d.com/issues/input-system-new-input-system-registering-active-touches-when-app-loses-focus)).
 
 #### Actions
 
 - Fixed an issue where `InputAction.Enable` would not reuse memory allocated prior and thus lead to memory leaks ([case 1367442](https://issuetracker.unity3d.com/issues/input-system-puts-a-lot-of-pressure-on-the-garbage-collector-when-enabling-and-disabling-inputactionmaps)).
+- Fixed an issue where serialized `InputAction` properties would have display name "Input Action" in the Inspector window instead of their given name. ([case 1367240](https://issuetracker.unity3d.com/product/unity/issues/guid/1367240)).
```
ekcoh: How come this one is a diff on this branch?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -13,17 +13,19 @@ however, it has to be formatted properly to pass verification tests.
 ### Changed
 
 - The artificial `ctrl`, `shift`, and `alt` controls (which combine the left and right controls into one) on the keyboard can now be written to and no longer throw `NotSupportedException` when trying to do so ([case 1340793](https://issuetracker.unity3d.com/issues/on-screen-button-errors-on-mouse-down-slash-up-when-its-control-path-is-set-to-control-keyboard)).
+- A device reset (such as when focus is lost) on `Touchscreen` will now result in all ongoing touches getting cancelled instead of all touches being simply reset to default state.
```
ekcoh, line 19: Is this really both a fix and a change? Wasn't this adressed as incorrect behavior which to me would be a fix compared to a change?

ekcoh (changes requested):
Overall looks very good but some outstanding questions and suggestions before approval.

jimon (approved):

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
Rene-Damm: Done

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Not sure what happened (probably me effing up some merge or stash) but indeed should not be here. Good catch. Removed.

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: Figured that since from a user POV this affects two different APIs (`Touchscreen` vs `Touch.activeTouches`), I'd call out the change in behavior of the underlying API separately. Might be superfluous.

PR #1468, opened 2021-11-15
FIX: Device lost to InputUser after layout changes (case 1347320).

ekcoh:
Fixes [1347320](https://issuetracker.unity3d.com/issues/cinemachine-camera-cannot-be-controlled-using-a-controllers-right-stick) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1347320)).

### Description

Fixed a problem with `InputUser` where devices would be removed and not added again after layout overrides preventing certain devices, e.g. gamepads to not work correctly when associated with action map bindings tied to `PlayerInput`. ([case 1347320](https://issuetracker.unity3d.com/product/unity/issues/guid/1347320)).

Considered overriding equals/hashcode for InputDevice to make it comparable on ID, but decided in the end to not introduce comparison outside reference comparison since there currently is not true uniqueness of InputDevice. The suggested fix is more surgical and has less potential side-effects until InputDevice equality can be well-defined.

The fix addresses reference comparison which is valid for most InputDevice use-cases except layout overrides where a new device inheriting the device ID is used. From what I could see this is the only valid case for this, hence a adapted tracking of lost devices makes sense.

### Changes made

Replaced reference comparison of lost devices to be ID based instead. This covers "lost device" and "recreated device".
Extended ArrayHelpers since range based predicate-based IndexOf was missing. 

Added test cases to cover ArrayHelpers and the particular failure scenario.

### Notes

Can be verified via submitted unit tests or by using repro project from related issue ticket: https://issuetracker.unity3d.com/product/unity/issues/guid/1347320

### Checklist

Before review:

- [x] Changelog entry added.
    - Functional tests `PlayerInput_WhenOverridingDeviceLayout_LostDeviceShouldBeResolvedAndRepaired`, `Utilities_IndexOfPredicate__IsUsingPredicateForEqualityAndConstraintedByStartIndexAndCount`.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Rene-Damm:
Could you try this one out with unplugging one controller (e.g. Xbox) and plugging another one in (PS4)? My guess is you'll end up in the same state where the player has a lost device and won't switch to the PS4 controller. If so, that'd be the root of the problem that needs fixing.

ekcoh:
> Could you try this one out with unplugging one controller (e.g. Xbox) and plugging another one in (PS4)? My guess is you'll end up in the same state where the player has a lost device and won't switch to the PS4 controller. If so, that'd be the root of the problem that needs fixing.

Will try to test this scenario

jimon (approved):

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
```diff
@@ -176,6 +176,20 @@ public static int IndexOf<TValue>(this TValue[] array, Predicate<TValue> predica
             return -1;
         }
 
+        public static int IndexOf<TValue>(this TValue[] array, Predicate<TValue> predicate, int startIndex = 0, int count = -1)
+        {
+            if (array != null)
```
andrew-oc: A minor thing but inverting this if would make things just a tiny bit nicer

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
```diff
@@ -176,6 +176,20 @@ public static int IndexOf<TValue>(this TValue[] array, Predicate<TValue> predica
...
+            {
+                var end = startIndex + (count < 0 ? array.Length - startIndex : count);
+                for (var i = startIndex; i < end; ++i)
```
andrew-oc: Could do with some bounds checks on these indexes

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
ekcoh: Sure thing, ended up with this due to personal preference but will change since I anyway need to add more changes to fix the problem for real (pending)

Packages/com.unity.inputsystem/InputSystem/Utilities/ArrayHelpers.cs
ekcoh: Yup, would you suggest I fix all of them (previously existing) since this is lacking for all pre-existing helpers? Incorrect indexing would result in exception right? And assert would also result in exception but do you suggest checking twice?

andrew-oc: Probably should be done for all of them that directly index the array, but I'd be ok if that was done in a separate PR so it doesn't muddy this one. Assert will only result in exceptions during development though right, which is where you want them. I don't really understand what you mean by checking twice. I'd add debug asserts for startIndex and end being in range is all.

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
```diff
@@ -1506,10 +1506,17 @@ private static void OnDeviceChange(InputDevice device, InputDeviceChange change)
...
-                    // Could be a previously lost device. Could affect multiple users. Repeatedly search in
-                    // s_AllLostDevices until we can't find the device anymore.
-                    var deviceIndex = s_GlobalState.allLostDevices.IndexOfReference(device, s_GlobalState.allLostDeviceCount);
-                    while (deviceIndex != -1)
+                    int FindLostDevice(int deviceIndex, int deviceId)
+                    {   // Note that we compare by Device ID since the lost device may be another instance
+                        // since its recreated during layout override. For other cases its the same reference.
+                        return s_GlobalState.allLostDevices.IndexOf((d) => d.deviceId == device.deviceId,
+                            deviceIndex, s_GlobalState.allLostDeviceCount);
+                    }
```
Rene-Damm: This local method will likely lead to unnecessary garbage on the GC heap (as it references the `device` local variable).

Packages/com.unity.inputsystem/InputSystem/Plugins/Users/InputUser.cs
```diff
@@ -1506,10 +1506,17 @@ private static void OnDeviceChange(InputDevice device, InputDeviceChange change)
                 // New device was added. See if it was a device we previously lost on a user.
                 case InputDeviceChange.Added:
                 {
-                    // Could be a previously lost device. Could affect multiple users. Repeatedly search in
-                    // s_AllLostDevices until we can't find the device anymore.
-                    var deviceIndex = s_GlobalState.allLostDevices.IndexOfReference(device, s_GlobalState.allLostDeviceCount);
-                    while (deviceIndex != -1)
+                    int FindLostDevice(int deviceIndex, int deviceId)
+                    {   // Note that we compare by Device ID since the lost device may be another instance
+                        // since its recreated during layout override. For other cases its the same reference.
```
Rene-Damm: I think this needs to compare *both* by reference and by device ID. We may also have the situation that the *same* `InputDevice` instance is coming back with a *new* device ID (happens when a device is retained on the disconnected list and then pulled back out when the native backend reports the same device under a new ID).

Rene-Damm: Probably intended to use `deviceId` argument.

Realize now that the anonymous function at line 1512 also creates garbage. Don't know... maybe it doesn't matter on this path.

PR #1471, opened 2021-11-16
Adding better support for Switch Pro controller

jimon:
### Description

Switch controllers generally have simple and full modes, our existing support relied on a controller being in simple mode, but any app (browser, steam, etc) could switch it to full mode and then we were not compatible with data layout at all. Also our previous implementation was lacking initial handshake.

### Changes made

Implemented support for handshake, simple and full modes.

### Notes

Needs some extra testing, I verified it works on Windows/macOS and USB/BT.
It seems controller is time sensitive to handshake sequence, so I implemented handshake restart in case if we're getting stuck somewhere (editor not refreshing, lost focus, etc)

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
@@ -1,5 +1,6 @@
 #if UNITY_EDITOR || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_WSA || PACKAGE_DOCS_GENERATION
 using System;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using UnityEngine.InputSystem.Controls;
 using UnityEngine.InputSystem.Layouts;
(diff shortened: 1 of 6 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1471) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1471) before we can accept your contribution.<br/><hr/>**Dmytro Ivanov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1471) it.</sub>

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
```diff
@@ -15,94 +16,109 @@ namespace UnityEngine.InputSystem.Switch.LowLevel
     /// <summary>
     /// Structure of HID input reports for Switch Pro controllers.
     /// </summary>
-    [StructLayout(LayoutKind.Explicit, Size = 20)]
+    [StructLayout(LayoutKind.Explicit, Size = 6)]
```
jimon, line 19: Should be 5

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
```diff
@@ -113,12 +129,352 @@ namespace UnityEngine.InputSystem.Switch
...
+        }
+
+        private static readonly SwitchMagicOutputReport.CommandIdType[] s_HandshakeSequence = new[]
+        {
+            SwitchMagicOutputReport.CommandIdType.Status,
+            SwitchMagicOutputReport.CommandIdType.Handshake,
+            SwitchMagicOutputReport.CommandIdType.Highspeed,
+            SwitchMagicOutputReport.CommandIdType.Handshake,
+            SwitchMagicOutputReport.CommandIdType.ForceUSB
+            ////TODO: Should we add a step to revert back to simple interface?
```
mattdatwork, line 175: should we move this todo to a jira ticket to track?

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
```diff
@@ -113,12 +129,352 @@ namespace UnityEngine.InputSystem.Switch
...
+                return;
+
+            m_HandshakeTimer = currentTime;
+
+            var command = s_HandshakeSequence[m_HandshakeStepIndex];
+
+            // Native backend rejects one of the commands based on size of descriptor.
+            // So just report both at a same time.
+            ////TODO: fix this.
+            var commandBt = SwitchMagicOutputHIDBluetooth.Create(command);
```
mattdatwork, line 217: should we move this to jira as a ticket to fix at some point? does it need fixing? what's the outcome of leaving as is vs fixing?

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
```diff
@@ -113,12 +129,352 @@ namespace UnityEngine.InputSystem.Switch
...
+            [FieldOffset(7)] public byte left1;
+            [FieldOffset(8)] public byte left2;
+            [FieldOffset(9)] public byte right0;
+            [FieldOffset(10)] public byte right1;
+            [FieldOffset(11)] public byte right2;
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public SwitchProControllerHIDInputState ToHIDInputReport()
+            {
+                ////TODO: calibration curve
```
mattdatwork, line 391: move to jira? what kind of calibration curve do you want to use here?

mattdatwork (approved):

ekcoh (approved):

ekcoh:
Agree with @mattdatwork  that TODOs might fit into a JIRA ticket tied to Device Support epic rather than being forgotten TODOs in code.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
jimon: Added ISX-955

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
jimon: ISX-956

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
jimon: ISX-957

Meorge:
Thank you for linking this fork/pull request to me! I was able to use it to add Joy-Con and Pro Controller support, including HD Rumble, IMU data, and controller colors, in my repository here: https://github.com/Meorge/JoyConUnityInputSystem

If you think it makes sense, I could fork off of this branch and add my changes, so that the Input System also supports Joy-Cons and the extra controller features.

jimon:
@Meorge I think I'll go merge this first, and if you can base your changes against this PR we might be able to incorporate them later.

One of the concerns I have right now is if you notice, `PreProcessEvent` can only shrink payloads or keep them same size, so size of `SwitchProControllerHIDInputState` is 6 bytes here is not a coincidence, it has to be that small to incorporate simple report which is 12 bytes .. So that might be one area that I need to refactor later on.

Meorge:
> @Meorge I think I'll go merge this first, and if you can base your changes against this PR we might be able to incorporate them later.

Sounds good to me! I've got some other things I need to work on, but I'll keep an eye on this and perhaps work on my own PR once this has been merged.

> One of the concerns I have right now is if you notice, `PreProcessEvent` can only shrink payloads or keep them same size, so size of `SwitchProControllerHIDInputState` is 6 bytes here is not a coincidence, it has to be that small to incorporate simple report which is 12 bytes .. So that might be one area that I need to refactor later on.

The approach I took to this in my code was to not continue processing simple input reports (i.e. return `false` when they were received in `IEventPreProcessor.PreProcessEvent()`), but before that, send a request to the controller to switch to the full input mode. If my understanding is correct, [according to this table](https://github.com/dekuNukem/Nintendo_Switch_Reverse_Engineering/blob/master/bluetooth_hid_notes.md#standard-input-report-format), only processing full reports (with the ID of `0x30`) would expand the maximum payload size to 49 bytes.

jimon:
@Meorge According to this https://github.com/libsdl-org/SDL/blob/main/src/joystick/controller_type.h#L576-L597 there are bunch of controllers that are input only controllers, and their state struct https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_switch.c#L118-L124 seems to be 7 bytes long.

And tbh I'm not sure about switching controllers to full mode, because then code that tries to read it as a HID gamepad starts spamming completely random data. E.g. Unity GetButton + Steam, because Steam switching the gamepad into full mode, and that breaks down HID layer ... So maybe keeping us into whatever mode the gamepad happens to be could be more pragmatic. SDL2 does switch to full mode but they restore to simple mode at shutdown, which is also interesting.

Rene-Damm (approved):

PR #1474, opened 2021-11-23
FIX: Timers on interactions sometimes triggering erroneously (case 1251231).

Rene-Damm:
Fixes [1251231](https://issuetracker.unity3d.com/issues/input-system-composites-hold-interaction-can-be-performed-when-no-keys-are-hold) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1251231/)).

### Description

Say you have a "Hold" on a "Shift+B" binding like so:

```
        var action = new InputAction(interactions: "hold(duration=2)");
        action.AddCompositeBinding("OneModifier")
            .With("Modifier", "<Keyboard>/shift")
            .With("Button", "<Keyboard>/b");
```

Then depending on the order that shift and B were pressed in, you'd sometimes see the action spuriously getting triggered when no input was received.

The problem is that the current `InputActionState.StartTimeout` and `InputActionState.StopTimeout` code assume that the `bindingIndex` and `controlIndex` passed to `StopTimeout` is the same as that passed to `StartTimeout`. If that is not the case, an existing timeout will not be removed.

### Changes made

Removed the explicit `bindingIndex` and `controlIndex` parameters. We already have the assumption that any given `interactionIndex` can have only one associated timer running. So instead, we now retain the `monitorIndex` and store it in the `InteractionState`. Thus, `StopTimeout` works regardless of what the currently active control is.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions_Interactions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions_Interactions.cs
@@ -462,6 +462,64 @@ internal partial class CoreTests
         }
     }
 
+    // https://fogbugz.unity3d.com/f/cases/1251231/
+    [Test]
+    [Category("Actions")]
+    public void Actions_HoldInteraction_CanBePerformedWhenInvolvingMoreThanOneControl()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        InputSystem.AddDevice<Mouse>();
+
+        // Add several bindings just to ensure that if conflict resolution is in the mix,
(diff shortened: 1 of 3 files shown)


andrew-oc (approved):
Very nice

PR #1475, opened 2021-11-25
FIX: InputSystem.onAnyButtonPress now filters out non-state or delta state events

andrew-oc:
Case [1376034](https://issuetracker.unity3d.com/product/unity/issues/guid/1376034/) [(Fogbugz)](https://fogbugz.unity3d.com/f/cases/1376034/)

### Description

InputSystem.onAnyButtonPress relies on EnumerateControls, which only works with state and delta state events, and throws exceptions otherwise. When onAnyButtonPress is hooked up and any other type of event comes through, players get exceptions.

### Changes made

Changed onAnyButtonPress to filter out everything except state and delta state events.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
@@ -1099,6 +1099,9 @@ namespace UnityEngine.InputSystem
         /// east (bit position 5), so if both buttons were pressed in the given event, button north would be returned.</remarks>
         public static InputControl GetFirstButtonPressOrNull(this InputEventPtr eventPtr, float magnitude = -1, bool buttonControlsOnly = true)
         {
+            if (eventPtr.type != StateEvent.Type || eventPtr.type != DeltaStateEvent.Type)
+                return null;
+
             if (magnitude < 0)
                 magnitude = InputSystem.settings.defaultButtonPressPoint;
 
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
@@ -1068,9 +1068,9 @@ namespace UnityEngine.InputSystem
         /// <summary>
         /// Return true if the given <paramref name="eventPtr"/> has any <see cref="Input"/>
         /// </summary>
-        /// <param name="eventPtr"></param>
-        /// <param name="magnitude"></param>
-        /// <param name="buttonControlsOnly"></param>
+        /// <param name="eventPtr">An event. Must be a <see cref="StateEvent"/> or <see cref="DeltaStateEvent"/>.</param>
+        /// <param name="magnitude">The threshold value that a button must be actuated by to be considered pressed.</param>
+        /// <param name="buttonControlsOnly">Whether the method should only consider button controls. See <see cref="InputControl.isButton"/>.</param>
         /// <returns></returns>
         /// <exception cref="ArgumentNullException"><paramref name="eventPtr"/> is a <c>null</c> pointer.</exception>
         /// <exception cref="ArgumentException"><paramref name="eventPtr"/> is not a <see cref="StateEvent"/> or <see cref="DeltaStateEvent"/> -or-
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -2401,7 +2401,9 @@ public static InputEventListener onEvent
         /// <seealso cref="ButtonControl.isPressed"/>
         /// <seealso cref="onEvent"/>
         public static IObservable<InputControl> onAnyButtonPress =>
-            onEvent.Select(e => e.GetFirstButtonPressOrNull()).Where(c => c != null);
+            onEvent
+                .Where(evt => evt.type == StateEvent.Type || evt.type == DeltaStateEvent.Type)
```
Rene-Damm: Minor but this will perform a null check twice. Better to fetch the type here only once.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -2401,7 +2401,9 @@ public static InputEventListener onEvent
...
+                .Select(e => e.GetFirstButtonPressOrNull()).Where(c => c != null);
```
Rene-Damm, line 2405: Wondering, is it better to solve this here for `onAnyButonPress` or should `GetFirstButtonPressOrNull` simply return null for any event that isn't a state/delta event?

Rene-Damm: Thinking about it, think I'd prefer the latter and just allow `GetFirstButtonPressOrNull` to be called on any event.

andrew-oc: Yep, that's much better

Rene-Damm: Nice!

Just saw that the lazy butt who added the method (bet the history shows it was me) didn't properly document it. Would you mind adding the docs for `eventPtr`, `magnitude`, and `buttonControlsOnly`?

andrew-oc: Haha, no probs

Rene-Damm (approved):

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -1083,9 +1083,9 @@ public static bool HasButtonPress(this InputEventPtr eventPtr, float magnitude =
         /// <summary>
         /// Get the first pressed button from the given event or null if the event doesn't contain a new button press.
         /// </summary>
-        /// <param name="eventPtr"></param>
-        /// <param name="magnitude"></param>
-        /// <param name="buttonControlsOnly"></param>
+        /// <param name="eventPtr">An event. Must be a <see cref="StateEvent"/> or <see cref="DeltaStateEvent"/>.</param>
```
Rene-Damm, line 1088: Seeing now that there is actually a exception doc for this. Let's remove the `ArgumentException` doc RE `StateEvent`/`DeltaStateEvent` and add something here that states that if the event is not of this type, the method will always return `null`.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -1083,9 +1083,9 @@ public static bool HasButtonPress(this InputEventPtr eventPtr, float magnitude =
...
+        /// <param name="magnitude">The threshold value that a button must be actuated by to be considered pressed.</param>
+        /// <param name="buttonControlsOnly">Whether the method should only consider button controls. See <see cref="InputControl.isButton"/>.</param>
```
Rene-Damm: Maybe

```Whether the method should only consider <see cref="ButtonControl"/>s. Otherwise, any <see cref="InputControl"/> that has an actuation (see <see cref="InputControl.EvaluateMagnitude"/>) equal to or greater than the given <paramref name="magnitude"/> will be considered a pressed button. This is `true` by default. ```

`isButton` is not public API.

File: Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs [hunk 5]
```diff
@@ -1083,9 +1083,9 @@ public static bool HasButtonPress(this InputEventPtr eventPtr, float magnitude =
         /// <summary>
         /// Get the first pressed button from the given event or null if the event doesn't contain a new button press.
         /// </summary>
-        /// <param name="eventPtr"></param>
-        /// <param name="magnitude"></param>
-        /// <param name="buttonControlsOnly"></param>
+        /// <param name="eventPtr">An event. Must be a <see cref="StateEvent"/> or <see cref="DeltaStateEvent"/>.</param>
+        /// <param name="magnitude">The threshold value that a button must be actuated by to be considered pressed.</param>
```
Rene-Damm: ```The threshold value that a control must be actuated by (see <see cref="InputControl.EvaluateMagnitude"/>) to be considered pressed. If not given, defaults to <see cref="InputSettings.defaultButtonPressThreshold"/>.```

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -1097,6 +1097,9 @@ public static bool HasButtonPress(this InputEventPtr eventPtr, float magnitude =
         /// east (bit position 5), so if both buttons were pressed in the given event, button north would be returned.</remarks>
         public static InputControl GetFirstButtonPressOrNull(this InputEventPtr eventPtr, float magnitude = -1, bool buttonControlsOnly = true)
         {
+            if (eventPtr.type != StateEvent.Type && eventPtr.type != DeltaStateEvent.Type)
+                return null;
```
Rene-Damm, line 1109: Let's add the same change to `GetAllButtonPresses`.

PR #1481, opened 2021-12-09
adds tracking flags to the tracked pose driver.

mattdatwork:
### Description

Adds optional tracking state flags to the Tracked Pose Driver.
When no action reference / inline action to the tracking state flags (int) is used, the Tracked Pose Driver operates as per current.
When an action is provided, we check the flags and only apply the position/rotation components that were actually indicated as correct by the tracking state flags.

This will resolve snap to zero issues when using the TPD

### Changes made

Change the trackedposedriver.cs file to allow additional Tracking State Flag actions.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -18,6 +18,7 @@ however, it has to be formatted properly to pass verification tests.
 - Deprecated `InputSettings.filterNoiseOnCurrent`. Now noise filtering is always enabled. Device only will become `.current` if any non-noise control have changed state.
 - A device reset (such as when focus is lost) on `Touchscreen` will now result in all ongoing touches getting cancelled instead of all touches being simply reset to default state.
 - Calling `InputTestFixture.Press`, `InputTestFixture.Set`, etc. from within a `[UnityTest]` will no longer immediately process input. Instead, input will be processed like it normally would as part of the Unity player loop.
+- Added `TrackingStateFlags` to the tracked pose driver packaged with the input system. Assigning this to an IntegerControl which provides TrackingState information will now correctly only apply those elements that actually recieved reliable data. If a TrackingStateFlags is not provided, it is assumed both position and rotation are valid.
 
 ### Fixed
 
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
@@ -1,5 +1,6 @@
 using System;
 using UnityEngine.InputSystem.LowLevel;
+using UnityEngine.XR;
 
 namespace UnityEngine.InputSystem.XR
 {
@@ -144,32 +145,8 @@ namespace UnityEngine.InputSystem.XR
         }
 
 
-        /// <summary>
-        /// Bitflag enum which represents what data was set on an associated Pose struct
-        /// This matches UnityXRInputTrackingStateFlags 
-        /// </summary>
-
-        [Flags]
-        protected enum PoseDataFlags
-        {
-            /// <summary>
-            /// No data was actually set on the pose
-            /// </summary>
-            NoData = 0,
-            /// <summary>
-            /// If this flag is set, position data was updated on the associated pose struct
-            /// </summary>
-            Position = 1 << 0,
-            /// <summary>
-            /// If this flag is set, rotation data was updated on the associated pose struct
-            /// </summary>
-            Rotation = 1 << 1,
-        };
-
         Vector3 m_CurrentPosition = Vector3.zero;
         Quaternion m_CurrentRotation = Quaternion.identity;
-        bool m_HasTrackingStateFlags = false;
-        PoseDataFlags m_CurrentTrackingStateFlags = PoseDataFlags.NoData;
         bool m_RotationBound;
         bool m_PositionBound;
 
@@ -300,7 +277,7 @@ namespace UnityEngine.InputSystem.XR
         /// <summary>
         /// This function is called when the object becomes enabled and active.
         /// </summary>
-        protected void OnEnable()
+        protected void OnEnable() 
         {
             InputSystem.onAfterUpdate += UpdateCallback;
             BindActions();
@@ -354,18 +331,24 @@ namespace UnityEngine.InputSystem.XR
             }
         }
 
-        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation, bool hasTrackingStateFlagInfo, PoseDataFlags trackingStateFlags)
+
+        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation)
+        {
+            SetLocalTransform(newPosition, newRotation, false, InputTrackingState.None);
+        }
+
+        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation, bool hasTrackingStateFlagInfo, InputTrackingState trackingStateFlags)
         {
             if (hasTrackingStateFlagInfo)
             {
-                if ((trackingStateFlags & PoseDataFlags.Position) == PoseDataFlags.Position &&
+                if ((trackingStateFlags & InputTrackingState.Rotation) == InputTrackingState.Rotation &&
                     m_TrackingType == TrackingType.RotationAndPosition ||
                     m_TrackingType == TrackingType.RotationOnly)
                 {
                     transform.localRotation = newRotation;
                 }
 
-                if ((trackingStateFlags & PoseDataFlags.Rotation) == PoseDataFlags.Rotation &&
+                if ((trackingStateFlags & InputTrackingState.Position) == InputTrackingState.Position &&
                     m_TrackingType == TrackingType.RotationAndPosition ||
                     m_TrackingType == TrackingType.PositionOnly)
                 {
@@ -398,17 +381,19 @@ namespace UnityEngine.InputSystem.XR
 
         protected virtual void PerformUpdate()
         {
+            InputTrackingState currentTrackingStateFlags = InputTrackingState.None;
+            bool hasTrackingStateFlags = false;
             if (m_TrackingStateFlags != null && m_TrackingStateFlags.action != null && m_TrackingStateFlags.action.bindings.Count > 0)
             {
-                m_CurrentTrackingStateFlags = (PoseDataFlags)m_TrackingStateFlags.action.ReadValue<int>();
-                m_HasTrackingStateFlags = true;
+                currentTrackingStateFlags = (InputTrackingState)m_TrackingStateFlags.action.ReadValue<int>();
+                hasTrackingStateFlags = true;
             }
             else
             {
-                m_HasTrackingStateFlags = false;
+                hasTrackingStateFlags = false;
             }
 
-            SetLocalTransform(m_CurrentPosition, m_CurrentRotation, m_HasTrackingStateFlags, m_CurrentTrackingStateFlags);
+            SetLocalTransform(m_CurrentPosition, m_CurrentRotation, hasTrackingStateFlags, currentTrackingStateFlags);
         }
 
         #region DEPRECATED
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
@@ -81,6 +81,28 @@ namespace UnityEngine.InputSystem.XR
             BeforeRender,
         }
 
+        /// <summary>
+        /// Bitflag enum which represents what data was set on an associated Pose struct
+        /// This matches UnityXRInputTrackingStateFlags 
+        /// </summary>
+        [Flags]
+        protected enum PoseDataFlags
+        {
+            /// <summary>
+            /// No data was actually set on the pose
+            /// </summary>
+            None = 0,
+            /// <summary>
+            /// If this flag is set, position data was updated on the associated pose struct
+            /// </summary>
+            Position = 1 << 0,
+            /// <summary>
+            /// If this flag is set, rotation data was updated on the associated pose struct
+            /// </summary>
+            Rotation = 1 << 1,
+        };
+
+
         [SerializeField]
         UpdateType m_UpdateType = UpdateType.UpdateAndBeforeRender;
         /// <summary>
@@ -141,14 +163,24 @@ namespace UnityEngine.InputSystem.XR
         public InputActionProperty trackingStateFlags
         {
             get => m_TrackingStateFlags;
-            set { m_TrackingStateFlags = value; }
+            set {
+                if (Application.isPlaying)
+                    UnbindTrackingState();
+
+                m_TrackingStateFlags = value;
+
+                if (Application.isPlaying && isActiveAndEnabled)
+                    BindTrackingState();
+            }
         }
 
 
         Vector3 m_CurrentPosition = Vector3.zero;
         Quaternion m_CurrentRotation = Quaternion.identity;
+        PoseDataFlags m_CurrentTrackingStateFlags = 0;
         bool m_RotationBound;
         bool m_PositionBound;
+        bool m_TrackingStateBound;
 
 
         void BindActions()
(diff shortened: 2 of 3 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1481) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1481) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1481) it.</sub>

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -320,19 +354,40 @@ protected virtual void OnBeforeRender()
             }
         }
 
-        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation)
+        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation, bool hasTrackingStateFlagInfo, PoseDataFlags trackingStateFlags)
```
chris-massie: This is a `protected virtual` method, so you can't just update the signature without it being a breaking change for existing users since their code wouldn't compile.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -135,11 +135,45 @@ public InputActionProperty rotationInput
...
+        }
+
+
+        /// <summary>
+        /// Bitflag enum which represents what data was set on an associated Pose struct
+        /// This matches UnityXRInputTrackingStateFlags 
+        /// </summary>
+
+        [Flags]
+        protected enum PoseDataFlags
```
chris-massie: Should `UnityEngine.XR.InputTrackingState` from the XR Module be used instead since that's the actual tracking enum being used? That would require adding a dependency on the XR Module, but then users would have access to all of the information.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -135,11 +135,45 @@ public InputActionProperty rotationInput
...
+            Position = 1 << 0,
+            /// <summary>
+            /// If this flag is set, rotation data was updated on the associated pose struct
+            /// </summary>
+            Rotation = 1 << 1,
+        };
+
         Vector3 m_CurrentPosition = Vector3.zero;
         Quaternion m_CurrentRotation = Quaternion.identity;
+        bool m_HasTrackingStateFlags = false;
```
chris-massie: There's no reason to have this as a class field since it is only used locally.

chris-massie (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -320,19 +354,40 @@ protected virtual void OnBeforeRender()
...
 
-        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation)
+        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation, bool hasTrackingStateFlagInfo, PoseDataFlags trackingStateFlags)
         {
-            if (m_TrackingType == TrackingType.RotationAndPosition ||
-                m_TrackingType == TrackingType.RotationOnly)
+            if (hasTrackingStateFlagInfo)
             {
-                transform.localRotation = newRotation;
+                if ((trackingStateFlags & PoseDataFlags.Position) == PoseDataFlags.Position &&
```
vrdave-unity: Is this supposed to be `PoseDataFlags.Rotation` (swap logic with line 368)?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -320,19 +354,40 @@ protected virtual void OnBeforeRender()
...
+                    m_TrackingType == TrackingType.RotationAndPosition ||
+                    m_TrackingType == TrackingType.RotationOnly)
+                {
+                    transform.localRotation = newRotation;
+                }
+
+                if ((trackingStateFlags & PoseDataFlags.Rotation) == PoseDataFlags.Rotation &&
```
vrdave-unity: (See comment for line 361)

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
vrdave-unity: This method signature doesn't look like it has to change since you are using member variables to track `m_HasTrackingStateFlags` and `m_CurrentTrackingStateFlags`, correct? Are there other consumers of this function that will change the transform data without changing the 'new' internal state of this object?

bryandube (approved):
Aside from being sad that we adding another possible inline action it looks like the other reviewers caught everything that I noticed already.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
caryhaynie: Two things:
* Can't `hasTrackingStateInfo` be derived from `trackingStateFlags`? (eg, `hasTrackingStateInfo = trackingStateFlags != PoseDataFlags.None;`
* You could also just overload the method and have to 2-param version call into the 4-param version (using the member variables @vrdave-unity alluded to above to supply the additional parameters?)

mattdatwork: good call. adding a facade call through.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -343,7 +398,17 @@ bool HasStereoCamera()
 
         protected virtual void PerformUpdate()
```
caryhaynie, line 448: (I realize this isn't part of this actual PR, but I wanted to ask about it)
Why is this method virtual? Is there case where the user will want to completely ignore this behavior and do something entirely different instead? Additionally, because this method uses private variables, isn't it functionally impossible for a user to implement a replacement correctly?

caryhaynie:
> Aside from being sad that we adding another possible inline action ...

👍 for a new yamato job: "Make sure @bryandube is not saddened by this change." 🤣

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
mattdatwork: added a pass through.

and i need the additional flag because i want tracking to work if there's no action providing tracking state flags. a None will cause nothing to be applied.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
mattdatwork: just allows us to call the same code in update/onbefore render. figured it might be nice for people to override this. but yeah. lack of overall protected et al is probably not great.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -354,18 +331,24 @@ protected virtual void OnBeforeRender()
             }
         }
 
-        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation, bool hasTrackingStateFlagInfo, PoseDataFlags trackingStateFlags)
+
+        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation)
```
caryhaynie, line 418: probably want to add an [Obsolete] attribute here, pointing people to the new override.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using UnityEngine.InputSystem.LowLevel;
+using UnityEngine.XR;
```
caryhaynie, line 3: as @chris-massie mentioned, need to add a dependency on `com.unity.modules.xr` for this.

caryhaynie: (in package.json, just to be clear)

Rene-Damm (changes requested):
Missing tests and changelog.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -343,7 +381,19 @@ bool HasStereoCamera()
 
         protected virtual void PerformUpdate()
         {
-            SetLocalTransform(m_CurrentPosition, m_CurrentRotation);
+            InputTrackingState currentTrackingStateFlags = InputTrackingState.None;
+            bool hasTrackingStateFlags = false;
```
ekcoh: Minor, but why not direct evaluate here to reduce this code? E.g.
`bool hasTrackingStateFlags = m_TrackingStateFlags?.action?.bindings.Count > 0;
var currentTrackingStateFlags = hasTrackingStateFlags ? (InputTrackingState)m_TrackingStateFlags.action.ReadValue<int>() : InputTrackingState.None;`
or even evaluate in call to SetLocalTransform and skip the locals?

ekcoh: Maybe easier to debug this way....

ekcoh (commented):
Looks good to me, missing changelog/tests though similar to what was already highlighted by @Rene-Damm

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -342,8 +446,17 @@ bool HasStereoCamera()
         }
 
         protected virtual void PerformUpdate()
-        {
-            SetLocalTransform(m_CurrentPosition, m_CurrentRotation);
+        {            
+            PoseDataFlags trackingStateFlags = m_CurrentTrackingStateFlags;
```
vrdave-unity, line 450: I suggest changing the name of this local variable to `updatedTrackingStateFlags` to be a bit more specific and not conflict with the globally scoped variable name.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -342,8 +446,17 @@ bool HasStereoCamera()
...
+            if (m_TrackingStateFlags != null && m_TrackingStateFlags.action != null && m_TrackingStateFlags.action.bindings.Count > 0)
+            { 
+            }   
+            else
+            {
+                trackingStateFlags = PoseDataFlags.Position | PoseDataFlags.Rotation;
+            }
```
vrdave-unity, line 457: I feel like the following is a bit clearer and simpler for readability: 
```
if (m_TrackingStateFlags == null || m_TrackingStateFlags.action == null || m_TrackingStateFlags.action.bindings.Count == 0)
    trackingStateFlags = PoseDataFlags.Position | PoseDataFlags.Rotation;
```

Is there a performance advantage to the existing code that is relevant?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ however, it has to be formatted properly to pass verification tests.
 - Deprecated `InputSettings.filterNoiseOnCurrent`. Now noise filtering is always enabled. Device only will become `.current` if any non-noise control have changed state.
 - A device reset (such as when focus is lost) on `Touchscreen` will now result in all ongoing touches getting cancelled instead of all touches being simply reset to default state.
 - Calling `InputTestFixture.Press`, `InputTestFixture.Set`, etc. from within a `[UnityTest]` will no longer immediately process input. Instead, input will be processed like it normally would as part of the Unity player loop.
+- Added `TrackingStateFlags` to the tracked pose driver packaged with the input system. Assigning this to an IntegerControl which provides TrackingState information will now correctly only apply those elements that actually recieved reliable data. If a TrackingStateFlags is not provided, it is assumed both position and rotation are valid.
```
unity-andrewc, line 21: recieved -> received

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -80,6 +81,28 @@ public enum UpdateType
...
+            None = 0,
+            /// <summary>
+            /// If this flag is set, position data was updated on the associated pose struct
+            /// </summary>
+            Position = 1 << 0,
+            /// <summary>
+            /// If this flag is set, rotation data was updated on the associated pose struct
+            /// </summary>
+            Rotation = 1 << 1,
+        };
```
unity-andrewc, line 103: Remove semicolon

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -135,10 +158,30 @@ public InputActionProperty rotationInput
             }
         }
 
+        [SerializeField]
+        InputActionProperty m_TrackingStateFlags;
+        public InputActionProperty trackingStateFlags
+        {
+            get => m_TrackingStateFlags;
+            set {
```
unity-andrewc, line 166: Move opening curly brace to new line

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -135,10 +158,30 @@ public InputActionProperty rotationInput
...
+                if (Application.isPlaying)
+                    UnbindTrackingState();
+
+                m_TrackingStateFlags = value;
+
+                if (Application.isPlaying && isActiveAndEnabled)
```
unity-andrewc, line 172: Are there any side effects with assuming that we should bind tracking state if this line is true? Should we instead and with a local bool that we set if we unbound the tracking state above?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -135,10 +158,30 @@ public InputActionProperty rotationInput
...
+                    BindTrackingState();
+            }
+        }
+
+
         Vector3 m_CurrentPosition = Vector3.zero;
         Quaternion m_CurrentRotation = Quaternion.identity;
+        PoseDataFlags m_CurrentTrackingStateFlags = 0;
```
unity-andrewc, line 180: `0` -> `PoseDataFlags.None`

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -266,7 +360,7 @@ protected virtual void Awake()
         /// <summary>
         /// This function is called when the object becomes enabled and active.
         /// </summary>
-        protected void OnEnable()
+        protected void OnEnable()
```
unity-andrewc, line 363: Undo whitespace change

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -320,20 +414,30 @@ protected virtual void OnBeforeRender()
...
 
+
         protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation)
         {
-            if (m_TrackingType == TrackingType.RotationAndPosition ||
-                m_TrackingType == TrackingType.RotationOnly)
+            SetLocalTransform(newPosition, newRotation, PoseDataFlags.Position | PoseDataFlags.Rotation);
+        }
+
+        protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRotation,PoseDataFlags trackingStateFlags)
```
unity-andrewc, line 423: Missing space between final comma and `PoseDataFlags`

unity-andrewc (changes requested):
There's a lot of unnecessary new-lines in this PR (two consecutive line breaks, empty lines between closing curly braces) that would be nice to remove as well

chris-massie:
This PR should be closed and replaced with the implementation done in https://github.com/Unity-Technologies/InputSystem/pull/1594

ekcoh:
PR referenced by @chris-massie https://github.com/Unity-Technologies/InputSystem/pull/1594 have been merged so closing this PR since that is obsolete now.

PR #1483, opened 2022-01-04
Add missing mapIndex init when creating a new TriggerState

russellquinn:
### Description

My project (Unity 2020.3.25f1 and InputSystem 1.2.0) was experiencing lots of out-of-bound errors and exceptions related to actions not being valid in a map. I investigated the Package code and saw `mapIndex` was always zero. I traced this back to `ChangePhaseOfInteraction` in `InputActionState.cs`. There are two different times that a `new TriggerState` is created. The second instance, is missing the initialization of `mapIndex = trigger.mapIndex`.

### Changes made

I added in `mapIndex = trigger.mapIndex` to give `mapIndex` a valid value. It fixed my crashes.

### Notes

I tested this by cloning the InputSystem repo, creating a new branch from the 1.2.0 release commit (to verify that my change was the fix, and that it hadn't been resolved since), changed the line, removed the official InputSystem package from my project and then added a Git Package in Package Manager, pointing to my clone and branch. It fixed the crashes.

I think it might be the solution to this post too: https://forum.unity.com/threads/map-index-on-trigger-does-not-correspond-to-map-index-of-trigger-state.1145087/

@Rene-Damm It looks like the original code was added in this commit: https://github.com/Unity-Technologies/InputSystem/commit/7e6490903117a600299099bf3235ec4b3a2548b4

**UPDATE: See my comment below for further investigation into what bug this actually fixes.**

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -1791,6 +1791,7 @@ namespace UnityEngine.InputSystem
                                     controlIndex = interactionStates[index].triggerControlIndex,
                                     bindingIndex = trigger.bindingIndex,
                                     interactionIndex = index,
+                                    mapIndex = trigger.mapIndex,
                                     time = interactionStates[index].performedTime, // Time when the interaction performed.
                                     startTime = startTime,
                                 };


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1483) <br/>All committers have signed the CLA.

russellquinn:
**UPDATE:** After using my fixed version of the Input System in my game, I started getting a lot of double-firing actions. I eventually tracked this down to me having added the _same_ Interaction on both Action and Binding. 

From the documentation (below), this seems to be intended behavior? ("the Input System processes the ones from the binding first" and then, I assume, the ones from the Action?)

> **Interactions applied to Actions**
> Interactions on Actions work very similar to Interactions on Bindings, but they affect all Controls bound to an Action, not just the ones coming from a specific Binding. If there are Interactions on both the Binding and the Action, the Input System processes the ones from the binding first.

Indeed, when I removed the Interaction from the Binding, leaving it only on the Action, everything worked as expected.

All this is to say, I think the bug fixed in this Pull Request, **broke this intended behavior**. Meaning, if you have Interactions on both Actions and Bindings (on an ActionMap that isn't index 0), only one Interaction is fired. If others made the same implementation mistake as me (adding multiple Interactions) then this bug fix will cause double-firing actions. A note about this should probably be added to any release notes.

With hindsight, it was this implementation mistake in my game, that raised the original "out-of-bound errors and exceptions related to actions not being valid in a map," due to the bug in the Input System code. Expected behavior, was that both Interactions should have fired.

**In other words, this Pull Request "Fixes crash when multiple Interactions are present on both Binding and/or Action in ActionMaps other than the first."**

(The bug is not present in the first ActionMap, as it has index 0, which is the default value of the `mapIndex` variable that is not correctly initialized.)

russellquinn:
Unity bug report submitted (1392559)

russellquinn:
This bug now has an entry in the Unity Issue Tracker:

https://issuetracker.unity3d.com/issues/map-index-on-trigger-and-indexoutofrangeexception-are-thrown-when-using-interaction-on-both-binding-and-its-parent-action

Rene-Damm:
Good fix. Thank you! Will route this in via a branch in the repo and separate PR shortly (to add tests and CHANGELOG and get our CI to run; the usual dance).

> All this is to say, I think the bug fixed in this Pull Request, **broke this intended behavior**. Meaning, if you have Interactions on both Actions and Bindings (on an ActionMap that isn't index 0), only one Interaction is fired. If others made the same implementation mistake as me (adding multiple Interactions) then this bug fix will cause double-firing actions. A note about this should probably be added to any release notes.
> 
> With hindsight, it was this implementation mistake in my game, that raised the original "out-of-bound errors and exceptions related to actions not being valid in a map," due to the bug in the Input System code. Expected behavior, was that both Interactions should have fired.

Will add something to the docs.

Will also bring this up in the ongoing discussions around improving the action editor/stuff. Its one of the cases that it makes quite non-obvious. At the very least, the interactions added to an action should probably also appear on bindings (but grayed out) in the position they would be added to the stack. Though ideally, interactions applied to actions should probably behave independently from interactions applied to individual bindings.

Rene-Damm:
Routed in via #1491.

russellquinn:
> Good fix. Thank you! 

No problem!

> Will also bring this up in the ongoing discussions around improving the action editor/stuff. Its one of the cases that it makes quite non-obvious. At the very least, the interactions added to an action should probably also appear on bindings (but grayed out) in the position they would be added to the stack. Though ideally, interactions applied to actions should probably behave independently from interactions applied to individual bindings.

Yeah, it's kind of confusing behavior and the fact it's been broken for some ActionMaps past index 0 and will suddenly start working is… extra confusing. Your idea of listing any Interactions in a parent Action on the Binding view sounds like a great solution. It could just be a text label that shows the Interactions that the Binding will also trigger, to expose the behavior. That would have been enough to stop me making this mistake.

PR #1485, opened 2022-01-05
Fixed bug where 'Keyboard/escape' would match 'Keyboard/e', and the k…

steinbitglis:
Fixed bug where 'Keyboard/escape' would match 'Keyboard/e', and the key 'e' would therefore not be mappable.

### Problem

When using RebindingOperation().WithCancelingThrough("\<Keyboard\>/escape"), the operation will be treated as cancelled, if the user presses the e button.

### Changes made

The custom string comparison was written so that only the shortest string length was used during comparison. And while the index of the compared character would happily skip ahead if any "escaped characters", no index was tracking the string that supposedly didn't have that same escaped character.

So previously  'abc\n'  would match 'abc*n', (and the fourth character isn't compared) but now instead 'abc\n' will match 'abcn', because the escaping character i taken out of the equation.

More importantly   'AB' will no longer match 'A', and vice versa.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
@@ -1404,22 +1404,21 @@ namespace UnityEngine.InputSystem
                 var pathElementLength = pathElement.length;
                 var elementLength = element.Length;
 
-                // `element` is expected to not include escape sequence. `pathElement` may.
-                // So if `element` is longer than `pathElement`, the two can't be a match.
-                if (elementLength > pathElementLength)
-                    return false;
-
-                for (var i = 0; i < pathElementLength && i < elementLength; ++i)
+                for (int i = 0, j = 0;;i++, j++)
                 {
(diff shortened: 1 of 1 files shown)


Rene-Damm:
Awesome. Thank you! Routed in via #1492.

JosephWhi98:
Any idea when this will be in a release version?

AdamTheD:
Does this solve all instances of this problem?

For instance I'm pretty sure "a" and "anyKey" are matching, along with things like "r" and "rightStick".

steinbitglis:
It has been merged a long time ago. In the pull request linked by Rene-Damm

ekcoh:
The fix was part of release version 1.4.0, but I recommend 1.4.1 since 1.4.0 had a regression for touch behavior. Its available as pre-release and should soon be published.

SHADE-Dev01:
> Does this solve all instances of this problem?
> 
> For instance I'm pretty sure "a" and "anyKey" are matching, along with things like "r" and "rightStick".

It seems like it. We had an issue where "q" was being shadowed by "quote", and applying this change to our local Input System package (1.3.0) fixed it.

PR #1486, opened 2022-01-09
Fixed a bug causing layout overrides to crash the editor with stackov…

ekcoh:
### Description

* Fixed an issue where a layout-override registered via `InputSystem.RegisterLayoutOverride(...)` would cause the editor to malfunction or crash if the layout override had a name already used by an existing layout. (case 1377685).
* Fixed an issue where attempting to replace an existing layout-override by using an existing layout-override name didn't work as expected and would instead aggregate overrides instead of replacing them when an override with the given name already exists.

### Changes made

Added test cases to cover use-cases mentioned in related bug issue.

### Notes

Seems like override replacements haven't been clearly documented what the expected behaviour is and layout code still contains many TODOs, so we should remedy if not reflecting expected behaviour.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/Tests/InputSystem/CoreTests_Layouts.cs
```diff
@@ -896,6 +896,78 @@ public void Layouts_CanApplyOverridesToExistingLayouts()
...
+                ""extend"" : ""Mouse"",
+                ""controls"" : [
+                    { ""name"" : ""extraControl"", ""layout"" : ""Button"" }
+                ]
+            }
+        ");
+
+        //var device = InputSystem.AddDevice<Mouse>();
+
+        //Assert.That(device["extraControl"], Is.TypeOf<ButtonControl>());
```
Rene-Damm: No commented out code pls. (except in exceptional cases and with comment)

Assets/Tests/InputSystem/CoreTests_Layouts.cs
```diff
@@ -896,6 +896,78 @@ public void Layouts_CanApplyOverridesToExistingLayouts()
...
+                ""extend"" : ""Mouse"",
+                ""controls"" : [
+                    { ""name"" : ""extraControl"", ""layout"" : ""Button"" }
+                ]
+            }
+        ";
+
+        Assert.That(() => InputSystem.RegisterLayoutOverride(json, "Mouse"),
+            Throws.Exception.With.Message.Contain("Layout overrides must have unique names"));
+    }
```
Rene-Damm, line 962: Could the last two tests be one using `[TestCase(...)]`?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased]
 
 ### Fixed
+* Fixed an issue where a layout-override registered via `InputSystem.RegisterLayoutOverride(...)` would cause the editor to malfunction or crash if the layout override had a name already used by an existing layout. (case 1377685).
```
Rene-Damm, line 33: Nitpick: empty line between header and start of list.

Rene-Damm: For case, link to case on issue tracker pls.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -354,16 +354,34 @@ public void RegisterControlLayout(string json, string name = null, bool isOverri
...
+                                                $"since a layout named '{internedLayoutName}' already exist. Layout overrides must " +
+                                                $"have unique names with respect to existing layouts.");    
+                }
+            }
+
             m_Layouts.layoutStrings[internedLayoutName] = json;
             if (isOverride)
             {
-                m_Layouts.layoutOverrideNames.Add(internedLayoutName);
+                m_Layouts.layoutOverrideNames.Add(internedLayoutName); // set semantics
```
Rene-Damm: What does "set semantics" mean?

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: It was reported internally so there is unfortunately no issue tracker link

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -354,16 +354,34 @@ public void RegisterControlLayout(string json, string name = null, bool isOverri
...
                 for (var i = 0; i < baseLayouts.length; ++i)
                 {
                     var baseLayoutName = baseLayouts[i];
                     m_Layouts.layoutOverrides.TryGetValue(baseLayoutName, out var overrideList);
-                    ArrayHelpers.Append(ref overrideList, internedLayoutName);
+                    if (!isReplacement)
+                        ArrayHelpers.Append(ref overrideList, internedLayoutName);
```
jimon, line 382: I'm a bit confused, what kind of a layout will be an override but not "replacement"?

jimon (approved):
If it works :)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: I should maybe get rid of it, the code is valid since the underlying container is a set (dictionary) if it was e.g. a list it would be a bug

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: The regular ones? As I understand this the override basically adds to the base layout as a diff while a non-override would replace an existing layout and a replacement of an existing override is a replacement and a previous override? :) At least that's how I understood @Rene-Damm when I asked about how this is supposed to work. I wonder if all of this is needed though....

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: It was more a highlight of an assumption on the container type

Assets/Tests/InputSystem/CoreTests_Layouts.cs
ekcoh: Yes, fixed

Assets/Tests/InputSystem/CoreTests_Layouts.cs
ekcoh: Removed, mistake

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: removed it

PR #1490, opened 2022-01-13
CHANGE: Behavior between button release threshold and zero (case 1393330).

Rene-Damm:
Addresses [1393330](https://issuetracker.unity3d.com/issues/input-system-reading-a-gamepad-trigger-as-a-button-produces-inaccurate-press-slash-release-states) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1393330/)).

### Description

When a button, after a press, is released below its release threshold, we invariably go to `canceled`. This leads to slightly strange behavior when the button is not yet fully released (common with gamepad triggers). So, a release to 0.3 would trigger `canceled` and a subsequent release to 0.2 would trigger `started` again.

### Changes made

A button that is released but not yet fully depressed now makes the action go back to `started`.

`canceled` will only be transitioned to on full release (`actuation==0`).

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
+++ b/Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
@@ -838,27 +838,27 @@ namespace UnityEngine.InputSystem
                 if (value != null)
                 {
                     var val = eventPtr.ReadValueAsObject();
-                    if (value is float f)
+                    if (val is float f)
                     {
-                        if (!Mathf.Approximately(f, Convert.ToSingle(val)))
+                        if (!Mathf.Approximately(f, Convert.ToSingle(value)))
                             return false;
                     }
-                    else if (value is double d)
+                    else if (val is double d)
                     {
-                        if (!Mathf.Approximately((float)d, (float)Convert.ToDouble(val)))
+                        if (!Mathf.Approximately((float)d, (float)Convert.ToDouble(value)))
                             return false;
                     }
-                    else if (value is Vector2 v2)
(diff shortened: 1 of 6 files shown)


Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
```diff
@@ -838,27 +838,27 @@ private bool Verify(InputActionTrace.ActionEventPtr eventPtr)
...
                     }
-                    else if (value is Vector3 v3)
+                    else if (val is Vector3 v3)
                     {
-                        if (!Vector3EqualityComparer.Instance.Equals(v3, val.As<Vector3>()))
+                        if (!Vector3EqualityComparer.Instance.Equals(v3, value.As<Vector3>()))
                             return false;
                     }
-                    else if (!value.Equals(val))
+                    else if (!val.Equals(value))
```
Rene-Damm, line 861: This is a minor tweak to the framework. Previously we used the passed in value to determine the value type of the comparison. It makes more sense to let the actual control type determine the value type so I went and swapped the roles.

```
// Before: Fails if the control happens to be a float control.
// Now: 0 will automatically be cast to a float because the control is a float control.
Assert.That(trace, Canceled(action, value: 0);
```

jimon (approved):

PR #1493, opened 2022-01-13
FIX: Input from other devices not working when DualShock 4 controller is connected

andrew-oc:
[case 1365891](https://issuetracker.unity3d.com/issues/input-from-the-keyboard-is-not-working-when-the-dualshock-4-controller-is-connected)

### Description

The sensors on the DualShock 4 have not yet been mapped to controls and marked as noisy, so whenever those sensors send events, which is all the time, the system has no way to know that the events represent noisy data. This means the DualShock 4 constantly sets itself as the current device, even if the user is using a different device.

### Changes made

Added the ability to mark a device as noisy even if none of the device's controls are marked as noisy.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
@@ -475,6 +475,11 @@ namespace UnityEngine.InputSystem.Layouts
             }
         }
 
+        /// <summary>
+        /// Mark the input device created from this layout as noisy, irrespective of whether or not any
+        /// of its controls have been marked as noisy.
+        /// </summary>
+        /// <seealso cref="InputControlLayoutAttribute.isNoisy"/>
         public bool isNoisy
         {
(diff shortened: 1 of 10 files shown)


Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs
```diff
@@ -475,6 +475,18 @@ internal set
             }
         }
 
+        public bool isNoisy
```
Rene-Damm, line 483: Needs xmldoc. (yeah, I know, I'm one to talk..... :))

Rene-Damm (approved):

PR #1494, opened 2022-01-14
CHANGE: Move initialization to SubsystemRegistration (case 1292358).

Rene-Damm:
Fixes [1392358](https://issuetracker.unity3d.com/issues/inputsystem-does-not-initialize-properly-in-a-build-when-accessed-early) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1392358/)).

### Description

Any access of `UnityEngine.InputSystem` APIs would run the initialization code. This code would trigger an immediate "null" update meant to flush out device registrations from the runtime. However, if this "null" update happened to get triggered before the runtime even did its own device discovery, then nothing would get flushed out and no devices would be available until the first frame (i.e. no devices in `Start` and `Awake`).

### Changes made

Moved initialization to `SubsystemRegistration` step and separated it from the initial "null" update which gets triggered `BeforeSceneLoad` like before.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


andrew-oc (approved):
Looks fine but just to point out in case it was an oversight that there will be two initial updates now because InitializeInPlayer also calls RunInitialUpdate.

Rene-Damm:
> Looks fine but just to point out in case it was an oversight that there will be two initial updates now because InitializeInPlayer also calls RunInitialUpdate.

Dang, I could swear I had removed that call. Will fix.

PR #1495, opened 2022-01-18
FIX: D-pad inputs sometimes getting skipped over (case 1389858).

Rene-Damm:
Fixes [1389858](https://issuetracker.unity3d.com/issues/onmove-debugs-different-value-than-input-debugger-when-quickly-pressing-two-different-d-pad-buttons-in-a-row) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1389858/)).

### Description

When binding an action to multiple controls, one of which was a D-Pad, inputs would sometimes get skipped over.

We perform a check on whether the new magnitude is *increasing* over what we already have. But this we should only do on controls *other* than the one already driving the action. Instead, we'd do

1. Dpad(-1,0) [left] -> magnitude(1) -> don't skip
2. Dpad(1,0) [right] -> magnitude(1) == magnitude(1) -> skip (... facepalm... )

### Changes made

Removed stupid.

### Notes

Also refactored some minor bits in the `IsConflictingInput` method. Also made some checks of "is this the same control?" 'more correct'.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
+++ b/Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
@@ -21,7 +21,7 @@
 * [Details](#details)
   * [Binding resolution](#binding-resolution)
     * [Choosing which Devices to use](#choosing-which-devices-to-use)
-  * [Disambiguation](#disambiguation)
+  * [Conflict resolution](#conflict-resolution)
   * [Initial state check](#initial-state-check)
 
 An [`InputBinding`](../api/UnityEngine.InputSystem.InputBinding.html) represents a connection between an [Action](Actions.md) and one or more [Controls](Controls.md) identified by a [Control path](Controls.md#control-paths). An Action can have an arbitrary number of Bindings pointed at it. Multiple Bindings can reference the same Control.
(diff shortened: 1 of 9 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1435,20 +1443,13 @@ private bool IsConflictingInput(ref TriggerState trigger, int actionIndex)
 
             // If we're not really effecting any change on the action, ignore the control state change.
             // NOTE: We may be looking at a control here that points in a completely direction, for example, even
-            //       though it has the same magnitude. However, we require a control to *higher* absolute actuation
+            //       though it has the same magnitude. However, we require a control to *increase* absolute actuation
             //       before we let it drive the action.
-            if (Mathf.Approximately(trigger.magnitude, actionState->magnitude))
+            if (!isControlCurrentlyDrivingTheAction && Mathf.Approximately(trigger.magnitude, actionState->magnitude))
```
Rene-Damm, line 1448: Sidenote: if you press d-pad *left* on one gamepad and then press d-pad *right* on a second gamepad, the second input will be *ignored*. The current code makes it a requirement that the level of actuation is higher. So the d-pad *right* press will only come through once you release the d-pad *left* input.

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -680,15 +680,40 @@ You can override this behavior by restricting [`InputActionAssets`](../api/Unity
...
+4. Right trigger is released.
+
+Applying the "rule of maximum actuation", this leads to the following sequence of changes on the Action:
+
+1. Action is `started` and then `performed`. Value is 1, Control is left trigger.
+2. Nothing happens. The right trigger is not actuated enough for it to override the input on the left trigger.
+3. Action is `performed`. Value is 0.6, Control is right trigger. This is because now the left trigger has fallen below the level of the right trigger and thus the latter is chosen to now "drive" the action.
+4. Action is `canceled` as no more active inputs are feeding into the Action.
+
+Note that when a Control is part of a Composite, the "rule of maximum actuation" is applied to the Composite as a whole, not to the individual Controls bound as part of it. So, a WASD keyboard binding, for example, has a single value of actuation corresponding to the magnitude of the resulting vector.
```
Rene-Damm, line 705: Excluded the case here of the same control bound multiple times to the same action. We have some support for this but the case is not well-defined so I opted to not "codify" it any further here in the docs. Should be sufficiently obscure not to be a noticeable omission.

jimon (approved):

PR #1496, opened 2022-01-19
FIX: Exceptions when accessing actions during startup (case 1378614).

Rene-Damm:
Fixes [1378614](https://issuetracker.unity3d.com/issues/input-system-nullreferenceexception-error-is-thrown-when-using-input-actions-in-builds) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1378614/)).

### Description

Another small startup problem. All automatic initialization of the input system is keyed to accessing `UnityEngine.InputSystem.InputSystem` and thus triggering its static ctor. However, when just going straight to `InputAction`s, there's no access to `InputSystem` happening. And thus you get exceptions.

### Changes made

The problem is implicitly fixed by #1494. However, I opted to make a further change by ensuring that `InputBindingResolver` taps `InputSystem` and triggers initialization, if necessary.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
@@ -3175,6 +3175,13 @@ namespace UnityEngine.InputSystem
             #endif
         }
 
+        // Initialization is triggered by accessing InputSystem. Some parts (like InputActions)
+        // do not rely on InputSystem and thus can be accessed without tapping InputSystem.
+        // This method will explicitly make sure we trigger initialization.
+        internal static void EnsureInitialized()
+        {
+        }
+
 #if UNITY_EDITOR
         internal static InputSystemObject s_SystemObject;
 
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3176,6 +3176,13 @@ private static void RunInitializeInPlayer()
             #endif
         }
 
+        // Initialization is triggered by accessing InputSystem. Some parts (like InputActions)
+        // do not rely on InputSystem and thus can be accessed without tapping InputSystem.
+        // This method will explicitly make sure we trigger initialization.
+        internal static void EnsureInitialized()
```
ekcoh, line 3182: Not sure I see why invoking this empty method makes a difference?

Is it possible to add some kind of test case to verify this initialization scenario?

ekcoh (approved):
Approved, if possible would be good to have a test case verifying the initialization scenario.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
Rene-Damm: Calling a method from a class ensures its class constructor is called. `InputSystem` initialization is triggered via this mechanism in general (i.e. call any of its public methods -> initialization automatically happens before actual method invocation).

Rene-Damm: Testing is a little tricky because we cannot use our existing test infrastructure as it automatically ensures initialization. The "automatic initialization through cctor" is mostly implicitly tested by having players where the input system works at all. In light of that, opted to forego adding a separate test setup for this.

PR #1500, opened 2022-01-26
CHANGE: Adding or removing device no longer temp-disables actions (case 1379932).

Rene-Damm:
Fixes [1379932](https://issuetracker.unity3d.com/issues/inputactionreferences-reading-resets-when-inputactionmap-has-an-action-for-the-other-hand-and-that-hand-starts-slash-stops-tracking) ([FogBugz](https://fogbugz.unity3d.com/f/cases/1379932/)).

### Description

In current `develop`, we handle two cases that lead to bindings needing to be resolved more than once.

1. The controls targeted by the bindings have changed.
2. Actions and/or maps have been added or removed.

For case 2, we require actions to be disabled.

However, case 1 we allow to happen while actions may be enabled (and possibly even in progress). The sequence it triggers is:

1. We disable all the actions in the map or asset.
2. We resolve all bindings from scratch.
3. We re-enable all actions that we disabled and turn on "initial state checks" for them.
4. In the next input update, we check all the controls bound to the actions we re-enabled and thus trigger actions according to current control states.

The big downside of this is that if you simply add or remove a device, the actions that are affected by that will get interrupted.

For XR this is especially bad as we remove a controller when we lose tracking and re-add when we regain tracking. So if you press a button on the left hand controller and then the right hand controller comes online, there's a flicker of activity on the left hand button.

### Changes made

Introduce a third codepath that is specific to adding and removing devices. This codepath will only touch an action if its active control is disappearing. For example, if you press a button and then the controller's batteries die, the action gets cancelled. But if you simply add or remove devices without affecting any active controls, the change is transparent.

Due to how `InputActionState` packs all its state into arrays, this isn't a trivial change. Also, I went in and more rigorously channeled the different cases down the respective codepaths. And added more test coverage for that.

### Notes

Sorry for the huge PR :/

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
+++ b/Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
@@ -20,6 +20,7 @@
 * [Control schemes](#control-schemes)
 * [Details](#details)
   * [Binding resolution](#binding-resolution)
+    * [Binding resolution while Actions are enabled](#binding-resolution-while-actions-are-enabled)
     * [Choosing which Devices to use](#choosing-which-devices-to-use)
   * [Disambiguation](#disambiguation)
   * [Initial state check](#initial-state-check)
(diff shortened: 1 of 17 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1095,6 +1095,22 @@ public unsafe bool IsPressed()
             return false;
         }
 
+        /// <summary>
```
andrew-oc, line 1098: Is this method really necessary when the inProgress property exists?

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -665,6 +666,18 @@ If there are multiple Bindings on the same Action that all reference the same Co
...
+
+Adding or removing a device &ndash; either [globally](../api/UnityEngine.InputSystem.InputSystem.html#UnityEngine_InputSystem_InputSystem_devices) or to/from the [device list](../api/UnityEngine.InputSystem.InputActionAsset.html#UnityEngine_InputSystem_InputActionAsset_devices) of an Action &ndash; will remain transparent __except__ if an Action is in progress and it is the device of its [active Control](../api/UnityEngine.InputSystem.InputAction.html#UnityEngine_InputSystem_InputAction_activeControl) that is being removed. In this case, the Action will automatically be [cancelled](../api/UnityEngine.InputSystem.InputAction.html#UnityEngine_InputSystem_InputAction_canceled).
+
+Modifying the [binding mask](../api/UnityEngine.InputSystem.InputActionAsset.html#UnityEngine_InputSystem_InputActionAsset_bindingMask) or modifying any of the Bindings (such as through [rebinding](#interactive-rebinding) or by adding or removing bindings) will, however, lead to all enabled Actions being temporarily disabled and then re-enabled and resumed.
```
Rene-Damm, line 679: Missed these changes in my initial commit.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -1050,18 +1105,64 @@ private unsafe void SetUpPerActionControlAndBindingArrays()
...
+            LazyResolveBindings(fullResolve: true);
+        }
+
+        internal void OnBindingModified()
+        {
+            ClearCachedActionData();
+            LazyResolveBindings(fullResolve: true);
+        }
+
+        internal void OnDeviceAddedOrRemoved()
```
Rene-Damm: Ooops, should remove this.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -1107,18 +1208,56 @@ internal void ResolveBindingsIfNecessary()
...
+        //      * Actions can be in enabled state.
+        //      * Actions *can* be in an in-progress state.
+        //        IF the control currently driving the action is on a device that is no longer usable with the action, the
+        //        action is CANCELLED. OTHERWISE, the action will be left as is and keep being in progress from its active control.
+        //      * A device CONFIGURATION change will NOT go down this path (e.g. changing the Keyboard layout). This is because
+        //        any binding path involving display names may now resolve to different controls -- which may impact currently
+        //        active controls of in-progress actions.
+        //      * A change in the USAGES of a device will NOT go down this path either. This is for the same reason -- i.e. an
+        //        active control may no longer match the binding path it matched before. If, for example, we switch the left-hand
+        //        and right-hand roles of two controllers, will will go down path (3) and not (2).
```
andrew-oc, line 1244: "...*we* will go down path..."

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -1133,21 +1272,20 @@ internal void ResolveBindingsIfNecessary()
...
-
             // Make sure that if we trigger callbacks as part of disabling and re-enabling actions,
             // we don't trigger a re-resolve while we're already resolving bindings.
             using (InputActionRebindingExtensions.DeferBindingResolution())
             {
                 // In case we have actions that are currently enabled, we temporarily retain the
                 // UnmanagedMemory of our InputActionState so that we can sync action states after
                 // we have re-resolved bindings.
-                var tempMemory = new InputActionState.UnmanagedMemory();
+                var oldMemory = new InputActionState.UnmanagedMemory();
```
andrew-oc, line 1284: I sort of also went back and forth on a previous PR deciding whether to change this name, but I was thinking about something more descriptive, like previousExecutionState or something, maybe even changing the name of the type to ActionExecutionState or similar. Not sure oldMemory is much of an improvement :smile:

andrew-oc (approved):
So, insofar as it's possible to tell on a PR of this size and complexity, this looks good.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
Rene-Damm: Absolutely not. Me being confused. Will remove.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
Rene-Damm: No, I'm talking about Will here. Kidding. Will fix will.

PR #1501, opened 2022-01-27
FIX: Processors are now always applied when reading the value of an action.

andrew-oc:
Case [1293728](https://issuetracker.unity3d.com/product/unity/issues/guid/1293728/)([Fogbugz](https://fogbugz.unity3d.com/f/cases/1293728/))

### Description

The original bug report suggested that the dead zone processor was interacting with the normalize processor and causing it to malfunction. The actual issue was that calling InputAction.ReadValue() when no binding/control is actuated always returns default(T), where in this case the user had a normalize processor on a stick X axis binding that they wanted to return 0.5 at the default state, instead of 0.

### Changes made

During binding resolution, and when action state is reset, the binding index for each action is set to point to the first binding in that actions' list of bindings. Additionally, when calling InputAction.ReadValue<T>(), if the action is not currently in progress, the default value of T is run through the processors of the active binding.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -29,7 +29,41 @@ using Is = UnityEngine.TestTools.Constraints.Is;
 // in terms of complexity.
 partial class CoreTests
 {
-    #if UNITY_EDITOR
+    [Test]
+    [Category("Actions")]
+    public void Actions_WhenBindingsAreResolved_BindingIndexPointsToFirstBindingInAction()
+    {
+        InputSystem.AddDevice<Gamepad>();
+
+        var map = new InputActionMap("map");
+        map.AddAction("action2", binding: "<Gamepad>/buttonNorth");
+        var action1 = map.AddAction("action1", binding: "<Gamepad>/leftStick/x", processors: "normalize(min=-1,max=1,zero=-1)");
+        action1.AddBinding("<Gamepad>/rightStick/x", processors: "normalize(min=0,max=1)");
+
+        map.Enable();
+
+        Assert.That(action1.ReadValue<float>(), Is.EqualTo(0.5f));
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_WhenResetActionState_BindingIndexPointsToFirstBindingInAction()
+    {
+        InputSystem.AddDevice<Gamepad>();
+
+        var map = new InputActionMap("map");
+        map.AddAction("action2", binding: "<Gamepad>/buttonNorth");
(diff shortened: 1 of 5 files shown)


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -29,7 +29,41 @@
...
+        action1.AddBinding("<Gamepad>/rightStick/x", processors: "normalize(min=0,max=1)");
+
+        map.Enable();
+
+        Assert.That(action1.ReadValue<float>(), Is.EqualTo(0.5f));
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_WhenResetActionState_BindingIndexPointsToFirstBindingInAction()
```
Rene-Damm: The binding index aspect seems like implementation detail where the connection to the test contents isn't quite clear.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -980,21 +980,13 @@ object ICloneable.Clone()
...
-                    var controlIndex = actionStatePtr->controlIndex;
-                    if (controlIndex != InputActionState.kInvalidIndex)
-                        result = state.ReadValue<TValue>(actionStatePtr->bindingIndex, controlIndex);
-                }
-            }
+            if (state == null) return default(TValue);
 
-            return result;
+            var actionStatePtr = &state.actionStates[m_ActionIndexInState];
+            return phase.IsInProgress()
```
Rene-Damm: `actionStatePtr->phase`.

Sidenote: These days, with ref vars now being supported, a whole bunch of the direct pointer code I wrote is probably better written as ref vars.

Rene-Damm (approved):
Some minor stuff.

Assets/Tests/InputSystem/CoreTests_Actions.cs
andrew-oc: Yeah, I felt the same. I couldn't see a clearer way to test it though. It's a roundabout way of testing the binding index by using the processors attached to different bindings, which is horribly brittle in itself. Can you think of a better way?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
andrew-oc: 👍

Assets/Tests/InputSystem/CoreTests_Actions.cs
Rene-Damm: I think the test contents are fine. Just that the names seem misleading. Maybe `Actions_ReadingValueRightAfterEnabling_AppliesProcessorsFromFirstBinding` or something.

Issue #121, opened 2018-05-22
[iOS][Touch][Stylus] Pressure value is not normalized


renaldasrep:
Found using c798ef8
This is also the case with old Input on iOS, android seems to not support stylus on old input though (does not register touches made with it).
Pressure range is 0 -> 4.166667 on iOS with stylus (iPad pro + iPen):
<img width="999" alt="screen shot 2018-05-22 at 14 45 04" src="https://user-images.githubusercontent.com/39190411/40360447-ed982052-5dce-11e8-95d6-3abe1ff8a2d2.png">

Pressure range is 0 -> 1 on Android with stylus (Samsung Galaxy Note 8 SM-N950F):
<img width="544" alt="screen shot 2018-05-22 at 14 40 25" src="https://user-images.githubusercontent.com/39190411/40360452-f0f9b878-5dce-11e8-8a11-dbe32a962874.png">

todi1856:
On both iOS and Android, the normal range is [0..1], value 1.0 is when you touch screen normally without applying extra weight,  you can go above 1.0 when you're putting pressure with heavy weight, like in the example above.

In short the behavior is expected.

renaldasrep:
Quite a lot of weight needs to be applied to get 1.0 on android though, and that is the upper limit, you cannot go over it (at least not on the Galaxy Note 8), can this be tuned somehow or is there nothing that could be done?

todi1856:
So there's a difference here, touch pressure and pen pressure.

I think we should agree on that touch pressure can go over 1.0, while pen pressure cannot go over 1.0

todi1856:
@renaldasrep could you test latest develop branch, this should be fine now

todi1856:
Fixed in 2018.3.0a12 @renaldasrep

todi1856:
Also fixed in 2018.2.11f1. I think @renaldasrep already tested it, so this can be closed @Rene-Damm

PR #1516, opened 2022-03-15
NEW: Can dynamically get/set parameters on actions.

Rene-Damm:
### Description

Adds much requested functionality for dynamically getting and setting parameters as used by composites, interactions, and processors on actions.

Adds two new APIs with various overloads.

```C#
// Get parameter.
action.GetParameterValue("duration");
action.GetParameterValue("tap:duration");
action.GetParameterValue("duration", InputBinding.MaskByGroup("Gamepad"));
action.GetParameterValue("duration", new InputBinding { path = "<Gamepad>/buttonSouth" });
action.GetParameterValue((TapInteraction x) => x.duration);

// Set parameter.
action.ApplyParameterOverride("duration", 0.4f);
action.ApplyParameterOverride("tap:duration", 0.4f);
action.ApplyParameterOverride("tap:duration", 0.4f, InputBinding.MaskByGroup("Gamepad"));
action.ApplyParameterOverride("tap:duration", 0.4f, new InputBinding { path = "<Gamepad>/buttonSouth" });
action.ApplyParameterOverride((TapInteraction x) => x.duration, 0.4f);

// Can also set at level of map or asset.
action.actionMap.ApplyParameterOverride("duration", 0.4);
action.actionMap.asset.ApplyParameterOverride("duration", 0.4);
```

### Changes made

- Added a new `InputActionParameters.cs` which adds the respective methods to the `InputActionRebindingExtensions` static class.
- Parameter overrides are stored on `InputActionMap` and `InputActionAsset`.
- `InputBindingResolver` now takes overrides into account.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -69,11 +69,32 @@ however, it has to be formatted properly to pass verification tests.
...
 
-### Added
-- Added support for SteelSeries Nimbus+ gamepad on Mac (Addition contributed by [Mollyjameson](https://github.com/MollyJameson)).
+#### Actions
+
+- Added new APIs for getting and setting parameter values on interactions, processors, and composites.
+  ```CSharp
+  // Get parameter.
+  action.GetParameterValue("duration");     // Any "duration" value on any binding.
+  action.GetParameterValue("tap:duration"); // "duration" on "tap" interaction on any binding.
```
ekcoh, line 94: What is the motivation behind selecting ':' to describe hierarchy here, is it consistent with any other custom string-based access scheme we have? Colon would to me suggest that these are prototypes/templates rather than instances which doesn't seem to be the case. Maybe '.' would be more expected since this is how you access in code?

ekcoh: Would e.g. action.GetParameterValue&lt;TapInteraction>("duration") would have been a valid alternative to action.GetParameterValue("tap:duration")? Is strings more convenient here you think that actual types? Just saying that if string based there needs to be somewhere to lookup the valid values here and it would be good to provide docs for how to approach figuring out what to write in that string and what naming scheme/syntax rules apply.

ekcoh: E.g. another question relating to ambiguity is, should I expect "tap:duration" to affect taps, multi-tap and slow-tap or only what is corresponding to TapInteraction? I see we document "duration" is corresponding to parameter name while "tap" do not correspond to class name. Hence above comment.

ekcoh: another one, if I set this and add processors/interactions after the e.g. parameter override has been applied, is it also applied to the newly added interactions/processors? This is not clear to me, but I might have overlooked some comment. Would recommend stating this clearly in docs since its instance based this might not be intuitive.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+        /// of the action's current bindings.
+        /// </summary>
+        /// <param name="action">Action on whose bindings to look for the value of the given parameter.</param>
+        /// <param name="name">Name of the parameter to get the value of. Case-insensitive. This can either be just the name of the
+        /// parameter (like <c>"duration"</c> or expressed as <c>nameof(TapInteraction.duration)</c>) or can be prefixed with the
+        /// type of object to get the parameter value from. For example, <c>"tap:duration"</c> will specifically get the <c>"duration"</c>
+        /// parameter from the object registered as <c>"tap"</c> (which will usually be <see cref="Interactions.TapInteraction"/>).</param>
+        /// <param name="bindingMask">Optional mask that determines on which bindings to look for objects with parameters. If used, only
+        /// bindings that match (see <see cref="InputBinding.Matches"/>) the given mask will be taken into account.</param>
+        /// <returns>The current value of the given parameter or <c>null</c> if the parameter could be found.</returns>
```
ekcoh: typo: "null if the parameter could be found" (?) assuming missing NOT

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+        /// When the binding above is applied to an action, the <c>duration</c> parameter from the <c>Hold</c> interaction can be
+        /// queried like so:
+        ///
+        /// <example>
+        /// <code>
+        /// action.GetParameterValue("duration") // Returns 0.6
+        /// </code>
+        /// </example>
+        ///
+        /// Note that if there are multiple objects on the action that use the same parameter name, the value of the <em>first</em> parameter
```
ekcoh, line 52: Why is this? (Maybe I find out by continue reading this PR :) )

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+        /// to look for processors, interactions, and composites on.</param>
+        /// <returns>The current value of the given parameter or <c>null</c> if the parameter could be found.</returns>
+        /// <remarks>
+        /// This method is a variation of <see cref="ApplyParameterOverride(UnityEngine.InputSystem.InputActionMap,string,UnityEngine.InputSystem.Utilities.PrimitiveValue,UnityEngine.InputSystem.InputBinding)"/>
+        /// to specifically target a single binding by index. Otherwise, the method is identical in functionality.
+        /// </remarks>
+        /// <exception cref="ArgumentNullException"><paramref name="action"/> is <c>null</c> -or- <exception cref="name"> is <c>null</c></exception></exception>
+        public static PrimitiveValue? GetParameterValue(this InputAction action, string name, int bindingIndex)
+        {
+            if (action == null)
```
ekcoh, line 120: Not really specific to this branch but something I reflected on recently. Is there a conventional reason that we check for null and throw ArgumentNullException instead of, as in this method letting #121 below throw NullReferenceException (which it does without checks)? I understand the check should exist for a method that doesn't directly use the reference, but here we do.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+                throw new ArgumentNullException(nameof(action));
+            if (string.IsNullOrEmpty(name))
+                throw new ArgumentNullException(nameof(name));
+            if (bindingIndex < 0)
```
ekcoh, line 124: Do we need to check bindingIndex here, wouldn't action.BindingIndexOnActionToBindingIndexOnMap throw the same for invalid index? Quick glance suggests this.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+        /// <code>
+        /// // Apply a parameter override only to action1.
+        /// map.ApplyBindingOverride("clamp:min", 0.25f, new InputBinding { action = action1.name });
+        ///
+        /// // Apply a parameter override only to a specific binding path.
+        /// map.ApplyBindingOverride("clamp:min", 0.4f, new InputBinding { path = "&lt;Gamepad&gt;/leftTrigger" });
+        /// </code>
+        /// </example>
+        ///
+        /// If there are multiple overrides that could be applied to a specific parameter, the most specific override
```
ekcoh: I find this paragraph a bit confusing, maybe possible to rephrase this a bit to make it easier to grasp? We are talking about two overrides "duration", but what do we mean by "an override for "duration" specifically on bindings to the "look" action exists"?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+        /// is chosen. For example, if both an override for "duration" and an override for "duration" specifically on bindings
+        /// to the "look" action exists, then the latter is used for all bindings to the "look" action and the former is used
+        /// for all other bindings. For parameter overrides of equal specificity, the behavior is undecided.
+        ///
+        /// Note that parameter overrides stay in place on the map. Like binding overrides, however, they are not
+        /// automatically persisted and thus need to be reapplied when actions are loaded from assets. This will, however, be applied
+        /// automatically to bindings added to the action in the future as well as whenever bindings are resolved to controls.
```
ekcoh, line 442: I guess this answers my question stated in the beginning of this PR.... that overrides are applied to processors/interactions added after the override was set.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+        }
+
+        /// <summary>
+        /// Set the value of the given parameter on the <see cref="InputBindingComposite"/>, <see cref="IInputInteraction"/>,
+        /// and <see cref="InputProcessor"/> objects found on the <see cref="InputActionMap.bindings"/> of each of the <see cref="InputActionAsset.actionMaps"/>
+        /// in <paramref name="asset"/>.
+        /// </summary>
+        /// <param name="asset">An <c>.inputactions</c> asset on whose <see cref="InputActionMap.bindings"/> to look for objects to set
+        /// the parameter value on.</param>
+        /// <param name="name">Name of the parameter to get the value of. Case-insensitive. This can either be just the name of the
```
ekcoh, line 468: We have a lot of redundant docs (more maintenence) between these methods. Do you know if there is any way to cross-reference or document method groups with the doc generator (docfx)?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+        /// This method is a variation of <see cref="ApplyParameterOverride(InputActionMap,string,PrimitiveValue,InputBinding)"/> which
+        /// allows specifying a binding by index. It otherwise behaves identically to that method.
+        /// </remarks>
+        public static void ApplyParameterOverride(this InputAction action, string name, PrimitiveValue value, int bindingIndex)
+        {
+            if (action == null)
+                throw new ArgumentNullException(nameof(action));
+            if (string.IsNullOrEmpty(name))
+                throw new ArgumentNullException(nameof(name));
+            if (bindingIndex < 0)
```
ekcoh, line 677: Similar precondition comments as previously in this PR

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+                m_BindingMask = parameter.bindingMask;
+                Reset();
+            }
+
+            private bool MoveToNextBinding()
+            {
+                // Find a binding that matches our mask.
+                while (true)
+                {
+                    ++m_BindingCurrentIndex;
```
ekcoh, line 809: minor: but why not setting m_BindingCurrentIndex to zero in Reset() and simplify this to e.g. 
`for(; m_BindingCurrentIndex != m_BindingEndIndex; ++m_BindingCurrentIndex)` - e.g. get rid of while true, the first paragraph here? Wonder whether "// Minus one to account for first MoveNext()." would be needed for enumeration pattern here?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
```diff
@@ -0,0 +1,805 @@
...
+            public Parameter Current => new Parameter
+            {
+                instance = m_CurrentObject,
+                field = m_CurrentParameter,
+                bindingIndex = m_BindingCurrentIndex,
+            };
+
+            object IEnumerator.Current => Current;
+
+            public void Dispose()
```
ekcoh, line 945: Intentionally empty dispose?

ekcoh (changes requested):
Lots of code here. Overall looks very good and if this have been asked for by users for a long time its great to get it in. At the same time it seems like the actual effects of overrides on map/asset level might be hard to predict due to rules tied to more-specific overrides etc. 

Great to see tests for new functionality and great job on docs. I have added a couple of questions and some suggested changes. 

A general question is that doc also states that if the parameter is ambiguous the most specific or first parameter is overridden. What does this mean if e.g. an action has an interaction and a processor with the same parameter but different semantics. Is interactions evaluated before processors or what are the rules tied to this?

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: > Would e.g. action.GetParameterValue<TapInteraction>("duration")

I considered this and discarded it in favor of the current approach. Interactions, processors, and composites are referred to in the API by their registration names, not by their implementation type -- which really is an implementation detail.

> if string based there needs to be somewhere to lookup the valid values here and it would be good to provide docs for how to approach figuring out what to write in that string and what naming scheme/syntax rules apply.

The docs mention this is the same as the registration name.

```
// "tap" referred to by name.
new InputAction(binding: "<Gamepad>/buttonSouth", interactions: "tap(duration=0.4)"
```

> should I expect "tap:duration" to affect taps, multi-tap and slow-tap or only what is corresponding to TapInteraction?

Only a single interaction (`TapInteraction`) has "tap" as its registration name.

>  if I set this and add processors/interactions after the e.g. parameter override has been applied, is it also applied to the newly added interactions/processors?

Yes, stated in the docs (API and manual).

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: Yup. Mandated by interface but there's nothing to dispose here.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: That doesn't work correctly. The iterator must have an initial state of "one before first element" and `MoveNext()` must make progress. The suggested for loop and similar patterns won't.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: Not aware of a mechanism. I'm sure there is something of some sort but I'd be surprised if the benefit outweighs the cost.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: There's many cases where more than one interaction, processor, and/or composite object exists on an action that uses a certain parameter "x". Most obvious case is something like this

```
var action = new InputAction();
action.AddBinding("<Gamepad>/buttonSouth", interactions: "tap(duration=0.5)");
action.AddBinding("<Keyboard>/space", interactions: "tap(duration=0.8)");
```

Now doing this is ambiguous:

```
action.GetParameterValue("duration");
```

There's two `TapInteraction` objects on there and either matches the binding mask (none).

I think throwing or anything else isn't useful behavior here. So it just returns whatever it comes across first. The user can easily resolve this by making the query more specific.

```
action.GetParameterValue("duration", new InputBinding { path = "<Gamepad>/buttonSouth" });
```

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: OK, guess I have overlooked where we describe currently available interaction string names in the manual/docs then, e.g. a list containing "tap" (among others) as a reference list of other valid options, where is this documented currently?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
ekcoh: I realized this further down in the review, thanks for explaining

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
ekcoh: ok

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
ekcoh: ah, yes, got it

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
ekcoh: ok, yeah I didn't say anything to get rid of either unless nulling references was needed but I guess not

Packages/com.unity.inputsystem/CHANGELOG.md
Rene-Damm: It's documented in `Interactions.md`, though probably not nearly as clearly as it could be.

As per @andrew-oc's suggestion I added overloads that allow using C# expressions to specify this information instead:

```
action.GetParameterValue((TapInteraction x) => x.duration);

action.ApplyParameterOverride((TapInteraction x) => x.duration, 0.4f);
```

That has a number of advantages. Gets rid of the registration name. Provides a value type and thus type-safety. And gets rid of the string parameter name.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: Good catch. Fixed.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: Personally, one of the things I'm flip-flopping on. In general, I try to throw argument exceptions directly from the method that receives the problematic argument. If you throw from an inner method, it's less clear of who's to blame -- was the argument already wrong in the initial method or did it only "become wrong" along the way.

```
// I expect GetParameterValue() to be the method throwing here.
// But if BindingIndexOnActionToBindingIndexOnMap throws, then who's to blame;
// my code or GetParameterValue() doing something wrong?
action.GetParameterValue("foo", 123213123);
```

BUT.... I know I'm inconsistent with this (like with just about everything).

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: See above.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
Rene-Damm: Took a stab at rewording this. Hopefully clearer now.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Thats nice

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionParameters.cs
ekcoh: Yeah that approach is fine, but then I would argue that everything internal should be asserted instead so there is one layer on the error checking cake. Nothing that needs actining now, but maybe an interesting topic as a policy for tech talk or something.... there usually a ton of habit and culture tied into it :)

ekcoh (approved):

PR #1520, opened 2022-03-18
FIX: Ensure FocusChanged respects IgnoreFocus in all cases

jamesmcgill:
### Description

Ensure OnFocusChanged returns early in all cases when IgnoreFocus is set.
This is to ensure that if runInBackground is somehow set to false but the playerLoop actually continues to run anyway, that we have this escape hatch to completely prevent the state changes. 
In this particular instance the Android backend is reporting that runInBackground is false but then expects the Head Tracking to continue running in the background. Until that is resolved we can use this escape hatch.

### Changes made
Decouple the IgnoreFocus behavior so it's no longer dependant on runInBackground.
I've moved it at the very top of the function so that nothing will be executed if IgnoreFocus is set to true.

### Notes
Please note I do not update m_HasFocus in this case. This is used in many places so maybe there are unintentional side-effects I'm not aware of.
I did this because OnUpdate() will early-out if focus is lost (canFlushBuffer will be true there).

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -4669,7 +4669,7 @@ partial class CoreTests
 
         // If we're set up to process input even when in the background, make sure
         // this is intact.
-        if (appRunInBackground || kIsEditor)
+        if (appRunInBackground || kIsEditor || backgroundBehavior == InputSettings.BackgroundBehavior.IgnoreFocus)
         {
             // Queue a change on each device.
             Set(mouse.position, new Vector2(333, 444), queueEventOnly: true);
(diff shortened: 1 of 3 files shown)


Rene-Damm (changes requested):
Test and changelog pls :)

jamesmcgill:
Went with https://github.com/Unity-Technologies/InputSystem/pull/1521 instead

PR #1521, opened 2022-03-21
FIX: Ensure update always continues when BackgroundBehavior.IgnoreFocus is set

jamesmcgill:
### Description

Ensure that that the input update loop continues to run for the case that `runInBackground == false` and `BackgroundBehaviour.IgnoreFocus` is set.

Input processing currently stops because OnUpdate() will early-out if `m_HasFocus == false && runInBackground == false`.  (canFlushBuffer will be true there).
In this particular instance the Android backend is reporting that runInBackground is false but then expects the Head Tracking to continue running in the background. Until that is resolved we can force `BackgroundBehaviour.IgnoreFocus` to allow updates to continue.

### Changes made

- onUpdate (and some other places) shouldn't use m_HasFocus directly but instead query the gameHasFocus property.
- Tests modified to ensure devices sync on focus gained in `runInBackground==false` case, but not the `true` case. (As per Behavior matrix in documentation)

### Notes
- This changes the behavior described in the matrix. Now for IgnoreFocus with `runInBackground==false` it should technically say 'Input is processed normally while the app is not in the foreground'

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -54,6 +54,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed "STAT event with state format TOUC cannot be used with device 'Touchscreen:/Touchscreen'" when more than max supported amount of fingers, currently 10, are present on the screen at a same time (case 1395648).
 - Fixed missing tooltips in PlayerInputManagerEditor for the Player Limit and Fixed Splitscreen sizes labels ([case 1396945](https://issuetracker.unity3d.com/issues/player-input-manager-pops-up-placeholder-text-when-hovering-over-it)).
 - Fixed DualShock 4 controllers not working in some scenarios by adding support for extended mode HID reports ([case 1281633](https://issuetracker.unity3d.com/issues/input-system-dualshock4-controller-returns-random-input-values-when-connected-via-bluetooth-while-steam-is-running), case 1409867).
+- Fixed `BackgroundBehavior.IgnoreFocus` so that it keeps the input update alive if runInBackground is false ([case 1400456](https://issuetracker.unity3d.com/issues/xr-head-tracking-lost-when-lost-focus-with-action-based-trackedposedriver-on-android)).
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed "STAT event with state format TOUC cannot be used with device 'Touchscreen:/Touchscreen'" when more than max supported amount of fingers, currently 10, are present on the screen at a same time (case 1395648).
 - Fixed missing tooltips in PlayerInputManagerEditor for the Player Limit and Fixed Splitscreen sizes labels ([case 1396945](https://issuetracker.unity3d.com/issues/player-input-manager-pops-up-placeholder-text-when-hovering-over-it)).
 - Fixed DualShock 4 controllers not working in some scenarios by adding support for extended mode HID reports ([case 1281633](https://issuetracker.unity3d.com/issues/input-system-dualshock4-controller-returns-random-input-values-when-connected-via-bluetooth-while-steam-is-running), case 1409867).
+- Fixed `BackgroundBehavior.IgnoreFocus` so that it keeps the input update alive if runInBackground is false ([case 1400456](https://issuetracker.unity3d.com/issues/xr-head-tracking-lost-when-lost-focus-with-action-based-trackedposedriver-on-android)).
```
Rene-Damm: Recommend inverting to describe the wrong behavior that was exhibited and is now fixed. Something sorta lik

```
- Fixed `BackgroundBehavior.IgnoreFocus` incorrectly getting ignored when `Application.runInBackground` was false (...).
```

Rene-Damm (approved):

PR #1523, opened 2022-03-23
Star complaince docs - batch 1

ekcoh:
### Description

An initial batch of adding missing API docs.

### Changes made

Adding various missing API docs to reduce number of PVS validation errors.

### Notes

Good to check extra carefully (CommonUsages) @Rene-Damm if this is correct.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -171,6 +216,17 @@ public void SubscribeTo(InputAction action)
             m_SubscribedActions.AppendWithCapacity(action);
         }
 
+        /// <summary>
+        /// Subscribe to a all actions part of <paramref name="actionMap"/>.
```
jamesmcgill: typo here, should probably be 'subscribe to all actions in'

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -258,6 +331,12 @@ public unsafe void RecordAction(InputAction.CallbackContext context)
             context.ReadValue(valueBuffer, valueSizeInBytes);
         }
 
+        /// <summary>
+        /// Clears all recorded data.
+        /// </summary>
+        /// <remarks>
+        /// Note that this doesn't affect currently monitored actions being recorded.
```
jamesmcgill: Maybe something like 'Note that this doesn't affect which actions are subscribed to be monitored, only the recorded event data so far'

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -446,14 +544,37 @@ public IInputInteraction interaction
                 }
             }
 
+            /// <summary>
+            /// Gets a timestamp corresponding to the time at wich the event was generated at.
```
jamesmcgill: small typo

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
```diff
@@ -11,25 +11,82 @@ public static class CommonUsages
...
+        /// Example: Right stick on a gamepad.
         /// </remarks>
         public static readonly InternedString Secondary2DMotion = new InternedString("Secondary2DMotion");
 
+        /// <summary>
+        /// Primary action control.
+        /// </summary>
+        /// <remarks>
+        /// Example: Primary mouse button (left button on right-handed configuration, right button on left-handed configuration),
+        /// south-button on a a gamepad.
```
jamesmcgill: small typo

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
```diff
@@ -11,25 +11,82 @@ public static class CommonUsages
...
+        /// </remarks>
         public static readonly InternedString PrimaryAction = new InternedString("PrimaryAction");
+
+        /// <summary>
+        /// Secondary action control.
+        /// </summary>
+        /// <remarks>
+        /// Example: Secondary mouse button (right button on right-handed configuration, left button on left-handed configuration),
+        /// east-button on a a gamepad.
```
jamesmcgill: small typo 'a a'

Packages/com.unity.inputsystem/InputSystem/InputExtensions.cs
```diff
@@ -80,6 +80,7 @@ public static bool IsModifierKey(this Key key)
         /// Check if a <see cref="Key"/> enum value represents key generating text input.
         /// </summary>
         /// <param name="key">The key enum value you want to check.</param>
+        /// <returns><c>true</c> if <paramref name="key"/> represents a key generating text input, else <c>false</c>.</returns>
         /// <remarks>
         /// Returns true if this key is a key generating non-whitespace character input, false otherwise.
```
jamesmcgill: This line should probably be removed now or the content moved into the <returns> block

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -374,6 +386,18 @@ public float defaultSlowTapTime
             }
         }
 
+        /// <summary>
+        /// Default minimum duration required of a press-and-release interaction to evaluate to a hold-interaction.
+        /// </summary>
+        /// <value>The default minimum duration that the button-like input control must remain in pressed state for the interaction to evaluate to a hold-interaction.</value>
+        /// <remarks>
+        /// A slow-tap-interaction is considered as a press-and-release sequence on a button-like input control.
```
jamesmcgill: Should be 'hold interaction'

jamesmcgill (approved):
Looks good.
Some small typos. I can fix these.
The commit message should start with DOCS: but this can be added at the merge step

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -428,12 +514,24 @@ public unsafe struct ActionEventPtr
...
             internal ActionEvent* m_Ptr;
 
+            /// <summary>
+            /// Gets the <see cref="InputAction"/> associated with this action event.
+            /// </summary>
             public InputAction action => m_State.GetActionOrNull(m_Ptr->bindingIndex);
 
+            /// <summary>
+            /// Gets the <see cref="InputActionPhase"/> associated with this action event.
+            /// </summary>
```
Rene-Damm, line 524: Xref `InputAction.phase`.

Rene-Damm: Xref `InputAction.CallbackContext.phase`.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -98,19 +98,46 @@ public sealed class InputActionTrace : IEnumerable<InputActionTrace.ActionEventP
...
+        /// Returns the number of events in the associated event buffer.
+        /// </summary>
         public int count => m_EventBuffer.eventCount;
 
+        /// <summary>
+        /// Constructs a new default initialized <c>InputActionTrace</c>.
+        /// </summary>
+        /// <remarks>
+        /// When using this constructor no action(s) are being recorded.
+        /// Traced actions needs to be explicitly setup after object creation to record actions.
```
duckets: Guidance:

- We try to talk to the user directly, i.e. "you".
- Wherever possible we try to say what is performing actions (a.k.a. "active voice").
- Because this overload contrasts with the other overloads that begin recording, it's good to mention those too.

Suggestion:

        /// When you use this constructor, the new InputActionTrace object does not start recording any actions.
        /// To record actions, you must explicitly set them up after creating the object.
        /// Alternatively, you can use one of the other constructor overloads which begin recording actions immediately.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -156,6 +190,17 @@ public void UnsubscribeFromAll()
                 UnsubscribeFrom(m_SubscribedActionMaps[m_SubscribedActionMaps.length - 1]);
         }
 
+        /// <summary>
+        /// Subscribe to <paramref name="action"/>.
+        /// </summary>
+        /// <param name="action">The action to be recorded.</param>
+        /// <remarks>
+        /// Note that there is no guard that prevents subscribing to the same action multiple times.
+        /// This needs to be handled by the user.
```
duckets: Guidance:

- For notes, we use **Note:** (in bold) then a full sentence.
- We address the reader directly as "you", because the reader is "our user", but the reader also has their users in mind, i.e. the people using the software that they are using Unity to create! So "user" can be ambiguous. Eg. "user input" can mean the user playing the game, not the person making the game.
- If there are potential unwanted side effects we should state them explicitly rather than implicitly (in this case, events would be duplicated in the buffer)

Suggestion:

        /// **Note:** This method does not prevent you from subscribing to the same action multiple times.
        /// If you subscribe to the same action multiple times, your event buffer will contain duplicate entries.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -156,6 +190,17 @@ public void UnsubscribeFromAll()
                 UnsubscribeFrom(m_SubscribedActionMaps[m_SubscribedActionMaps.length - 1]);
         }
 
+        /// <summary>
+        /// Subscribe to <paramref name="action"/>.
```
duckets: Guidance: 

We use the "3rd person singular" tense, instead of the "imperative mood" tense. The simplest way to remember this is that the content should sound like "does x", rather than "do x".

Suggestion:

Should be: "**Subscribes to**" instead of "**Subscribe to**"

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -171,6 +216,17 @@ public void SubscribeTo(InputAction action)
             m_SubscribedActions.AppendWithCapacity(action);
         }
 
+        /// <summary>
+        /// Subscribe to all actions contained within <paramref name="actionMap"/>.
```
duckets: Same tense issue as before, should be "Subscribes to".

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -171,6 +216,17 @@ public void SubscribeTo(InputAction action)
...
+        /// </summary>
+        /// <param name="actionMap">The action-map containing all actions to be recorded.</param>
+        /// <remarks>
+        /// Note that there is no guard that prevents subscribing to the same action multiple times.
+        /// This needs to be handled by the user.
```
duckets: as above

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -184,6 +240,16 @@ public void SubscribeTo(InputActionMap actionMap)
             m_SubscribedActionMaps.AppendWithCapacity(actionMap);
         }
 
+        /// <summary>
+        /// Unsubscribes from an action previously subscribed to.
```
duckets: Guidance:

Contradicts the note below. 

Suggestion:

Unsubscribes from an action, if that action was previously subscribed to.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -184,6 +240,16 @@ public void SubscribeTo(InputActionMap actionMap)
...
+        /// </summary>
+        /// <param name="action">The action to unsubscribe from.</param>
+        /// <remarks>
+        /// Note that this method has no side-effects if the action has not previously been subscribed to by this instance.
```
duckets: Guidance:

- Note formatting
- Active voice, talking to the reader directly

Suggestion:

        /// **Note:** This method has no side effects if you attempt to unsubscribe from an action that you have not previously subscribed to.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -201,6 +267,13 @@ public void UnsubscribeFrom(InputAction action)
                 m_SubscribedActions.RemoveAtWithCapacity(index);
         }
 
+        /// <summary>
+        /// Unsubscribes from all actions included in <paramref name="actionMap"/>.
+        /// </summary>
+        /// <param name="actionMap">The action-map containing actions to be recorded.</param>
```
duckets: We're unsubscribing here, so "to be recorded" seems incorrect. is this a copy & paste mistake?

Suggestion:

        /// <param name="actionMap">The action-map containing actions to be unsubscribed from.</param>

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -258,6 +331,12 @@ public unsafe void RecordAction(InputAction.CallbackContext context)
             context.ReadValue(valueBuffer, valueSizeInBytes);
         }
 
+        /// <summary>
+        /// Clears all recorded data.
+        /// </summary>
+        /// <remarks>
+        /// Note that this doesn't affect which actions are currently subscribed for monitoring
```
duckets: Guidance:
- **Note:** formatting
- We can describe "doesn't affect" in a bit more detail

Suggestion:

        /// **Note:** This method does not unsubscribe any actions that the instance is listening to, so after clearing the recorded data, new input on those subscribed actions will continue to be recorded.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -446,14 +546,37 @@ public IInputInteraction interaction
                 }
             }
 
+            /// <summary>
+            /// Gets a timestamp corresponding to the time at which the event was generated.
```
duckets: Guidance:

- Because this is a property we don't need to say "gets"
- It's useful to mention the units in the summary. (the summary appears in more places, such as parent pages & IDE tooltips, than the rest of the description or remarks content)

Suggestion

            /// The time, in seconds since your game or app started, that the event occurred.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -446,14 +546,37 @@ public IInputInteraction interaction
...
+            /// </summary>
+            /// <remarks>
+            /// Times are in seconds and progress linearly in real-time. The timeline is the
+            /// same as for <see cref="Time.realtimeSinceStartup"/>.
```
duckets: The extra detail about the time progressing linearly is - I think - intended to say that it is not affected by Time.timescale, right? If so we should be explicit about that.

Suggestion:

            /// The time recorded for events is the same as <see cref="Time.realtimeSinceStartup"/>, which is not affected by variations in <see cref="Time.timeScale"/>.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -446,14 +546,37 @@ public IInputInteraction interaction
...
+            /// </remarks>
             public double time => m_Ptr->baseEvent.time;
 
+            /// <summary>
+            /// Gets the timestamp for when the <see cref="phase"/> transitioned into <see cref="InputActionPhase.Started"/>.
```
duckets: Guidance:

- "Gets" is a small redundancy in the sentence. It's okay to just say what the property contains:

Suggestion:

            /// The time at which the recorded action's <see cref="phase"/> transitioned to <see cref="InputActionPhase.Started"/>

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -446,14 +546,37 @@ public IInputInteraction interaction
...
+            /// </summary>
             public double startTime => m_Ptr->startTime;
 
+            /// <summary>
+            /// Gets the duration (in seconds) that has elapsed between when this event was generated and when the
```
duckets: Minor style adjustment:

            /// The duration, in seconds, that has elapsed between when this event was generated and when the

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -446,14 +546,37 @@ public IInputInteraction interaction
...
+            /// action <see cref="phase"/> transitioned to <see cref="InputActionPhase.Started"/> and has remained active.
+            /// </summary>
             public double duration => time - startTime;
 
+            /// <summary>
+            /// Gets the size in bytes of the value associated with this action event.
```
duckets: Style:

            /// The size, in bytes, of the value associated with this action event.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -488,6 +611,15 @@ public object ReadValueAsObject()
                 return control.ReadValueFromBufferAsObject(valuePtr, valueSizeInBytes);
             }
 
+            /// <summary>
+            /// Reads the value associated with this event into the contiguous memory buffer defined by <c>[buffer, buffer + bufferSize)</c>.
+            /// </summary>
+            /// <param name="buffer">Pointer to the contiguous memory buffer to write value data to.</param>
+            /// <param name="bufferSize">The size in bytes of the contiguous buffer pointed to by <paramref name="buffer"/>.</param>
```
duckets: Style:

    The size, in bytes, of the (...)

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
```diff
@@ -501,6 +633,12 @@ public void ReadValue(void* buffer, int bufferSize)
                 UnsafeUtility.MemCpy(buffer, m_Ptr->valueData, valueSizeInBytes);
             }
 
+            /// <summary>
+            /// Reads the value associated with this event as an object of type <typeparamref name="TValue"/>.
+            /// </summary>
+            /// <typeparam name="TValue">The event value type to be used.</typeparam>
+            /// <returns>Object of type <typeparamref name="TValue"/>.</returns>
+            /// <exception cref="InvalidOperationException">In case the size of <typeparamref name="TValue"/> do not match the size of the value associated with this event.</exception>
```
duckets: Grammar:

"do not match" should be "does not match"

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
```diff
@@ -11,25 +11,82 @@ public static class CommonUsages
...
         /// Secondary 2D motion control.
         /// </summary>
         /// <remarks>
-        /// Example: right stick on gamepad.
+        /// Example: Right stick on a gamepad.
         /// </remarks>
         public static readonly InternedString Secondary2DMotion = new InternedString("Secondary2DMotion");
 
+        /// <summary>
+        /// Primary action control.
```
duckets: Guidance: maybe too short? 

Suggestion: (if correct)
        /// The primary action control on any input device, such as a gamepad, mouse, or keyboard.

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
```diff
@@ -11,25 +11,82 @@ public static class CommonUsages
...
+        /// </summary>
+        /// <remarks>
+        /// Example: Primary mouse button (left button on right-handed configuration, right button on left-handed configuration),
+        /// south-button on a gamepad.
+        /// </remarks>
         public static readonly InternedString PrimaryAction = new InternedString("PrimaryAction");
+
+        /// <summary>
+        /// Secondary action control.
```
duckets: Same as above, and for subsequent similar summaries below.

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
```diff
@@ -11,25 +11,82 @@ public static class CommonUsages
...
+        /// </summary>
+        /// <remarks>
+        /// Example: Secondary mouse button (right button on right-handed configuration, left button on left-handed configuration),
+        /// east-button on a gamepad.
+        /// </remarks>
         public static readonly InternedString SecondaryAction = new InternedString("SecondaryAction");
+
+        /// <summary>
+        /// Primary trigger control.
```
duckets: /// The primary trigger control on input devices with triggers.

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
```diff
@@ -11,25 +11,82 @@ public static class CommonUsages
...
+        /// <summary>
+        /// A modifier action control that modifies usage of other controls.
+        /// </summary>
+        /// <remarks>
+        /// Example: Keyboard modifier keys like CTRL, SHIFT, ALT, OPTION, etc.
+        /// </remarks>
+        public static readonly InternedString Modifier = new InternedString("Modifier");
+
+        /// <summary>
+        /// Spatial position control.
```
duckets: /// The spatial position control on input devices with spatial tracking.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -44,7 +44,7 @@ namespace UnityEngine.InputSystem
     public partial class InputSettings : ScriptableObject
     {
         /// <summary>
-        /// Determine how the input system updates, i.e. processes pending input events.
+        /// Determines how the input system handle updates, i.e. how pending input events are processed.
```
duckets: Guidance:
- grammar: "handle" should "handles"
- I notice below the 'value' says _when_, so perhaps this should say "when" too?
- "determines" isn't completely clear whether the user can choose this. If so, should be something more like "Allows you to control how (...)

Suggestion:

        /// Allows you to control how the input system handles updates. In other words, how and when pending input events are processed.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -361,6 +361,18 @@ public float defaultTapTime
             }
         }
 
+        /// <summary>
+        /// Default minimum duration required of a press-and-release interaction to evaluate to a slow-tap-interaction.
```
duckets: Does this allow the user to specify a new default? if so:

    /// Allows you to specify the default minimum duration required of a press-and-release interaction to evaluate to a slow-tap-interaction.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -374,6 +386,18 @@ public float defaultSlowTapTime
             }
         }
 
+        /// <summary>
+        /// Default minimum duration required of a press-and-release interaction to evaluate to a hold-interaction.
```
duckets: similar to above

Packages/com.unity.inputsystem/InputSystem/Utilities/ReadOnlyArray.cs
```diff
@@ -69,6 +69,12 @@ public TValue[] ToArray()
             return result;
         }
 
+        /// <summary>
+        /// Returns the index of the first element for which the given <paramref name="predicate"/> is true (if any).
+        /// </summary>
+        /// <param name="predicate">The unary predicate to be evaluated for each element.</param>
```
duckets: "unary predicate" could be a little too complex for less experienced coders. Can we link somewhere to explain, or give some useful examples?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Makes sense, updated based on suggestion.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Noted, updated this and similar occurrences I could identify.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Thanks for guiding. Updated based on your recommendation.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Fixed

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Fixed

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Thanks, updated based on suggestion

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Thanks, updated based on suggestion

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Copy-paste error, fixed

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Good suggestion, accepted as-is

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Makes sense, updated based on suggestion.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Yes, basically reflecting properties from Time.realtimeSinceStartup. Updated based on suggestion.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Since this is also reflecting to the same timeline I assume the other part is needed as well (as mentioned above)?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Thanks, figured based on the other suggestions

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Fixed

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Fixed

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionTrace.cs
ekcoh: Thanks, fixed

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
ekcoh: Its correct, updated.

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
ekcoh: Updated

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
ekcoh: Updated

Packages/com.unity.inputsystem/InputSystem/Controls/CommonUsages.cs
ekcoh: Updated, also updated orientation which is similar.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
ekcoh: Good suggestion, updated based on your suggestion.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
ekcoh: Yes, used your suggestion

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
ekcoh: Changed similarly

Packages/com.unity.inputsystem/InputSystem/Utilities/ReadOnlyArray.cs
ekcoh: Hmm.... since we are in C# predicate is also a concept so removed the unary (mathematical) aspect here and rewrote the description to make it clearer. Better?

ekcoh: Not sure if you mean we should link to C# reference, e.g. https://docs.microsoft.com/en-us/dotnet/api/system.predicate-1?view=net-6.0? Or stay with current explanation? 

Added an example to the doc.

duckets: Actually, now that it's simply "predicate", I think that's precise enough for users to google "c# predicate" on their own if they need to :)

duckets (approved):

PR #1524, opened 2022-03-23
FIX: Input devices sometimes marked as disabled on reconnect

unity-cwells:
### Description

If a `InputDevice` is disconnected while an app is out-of-focus and the the device's `InputDevice.DeviceFlags.DisabledWhileInBackground` flag is set, the flag will remain set upon reconnect even if the app has since regained focus. This results in the device incorrectly appearing as disabled until the next focus change event. This PR corrects this by clearing the `InputDevice.DeviceFlags.DisabledWhileInBackground` on reconnection of a device. If the app is still out-of-focus during reconnect and the flag is valid, this will be detected and set as normal during the call to `InputManager.AddDevice()`. 

### Changes made

Fixed an issue where `InputDevice.enabled` could return false when a device was disconnected while an application was out-of-focus and then re-connected when in-focus

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Rene-Damm (changes requested):
Needs test.

PR #1409, opened 2021-09-27
re-adds ENABLE_VR to specific XR C# files.

mattdatwork:
### Description

Fix for compile error when _only_ installing players that do not define ENABLE_VR.

### Changes made

This puts ENABLE_VR back into 
InputSystem.cs
PoseControl.cs
GenericXRDevices.cs
XRLayoutBuilder.cs
XRSupport.cs

### Checklist

Before review:

- [x] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [N/A] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [N/A] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #1525, opened 2022-03-23
Add support for Game Core platforms to XR classes

chris-massie:
### Description

This update modifies some of the preprocessor define guards for the XR classes to include the Game Core platforms. `ENABLE_VR` is not defined on that platform, but the XR module assembly is available with limited functionality when the module is enabled. The types that the XR classes in Input System depends on is available in that assembly even when `ENABLE_VR` is not defined.

### Changes made

Changed many `ENABLE_VR` checks to `(ENABLE_VR || UNITY_GAMECORE)`. I also removed `ENABLE_AR` from some of the QA behaviors since that define is not used in the types it was guarding against.

### Notes

Ideally, there would be a Unity define when `com.unity.modules.xr` is enabled in a project _and_ its assembly is available on the platform, but there isn't one at the moment, so I opted to add `UNITY_GAMECORE` rather than remove `ENABLE_VR` and opt out platforms like `UNITY_STADIA` where the assembly is unavailable.

This is a follow up change to the PR where the `ENABLE_VR` check was added to fix Stadia: https://github.com/Unity-Technologies/InputSystem/pull/1409

This is also related to https://github.com/Unity-Technologies/InputSystem/pull/1350 for define changes. This PR reverts the change to the `versionDefines` done with that PR since I feel it actually made it less clear. Changing from `UNITY_XR_AVAILABLE` to `UNITY_INPUT_SYSTEM_ENABLE_XR` makes it match the naming style of the other module checks, and since the XR module can be enabled in a project but unavailable on a platform, calling it the former makes it misleading. That PR is also the only place which explains the undocumented `UNITY_FORCE_INPUTSYSTEM_XR_OFF` define, which I don't know if it's actually used anywhere.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


mtschoen-unity (approved):
I haven't tested this but I've been chatting with Chris about this on and off. It seems like we have lost some institutional knowledge about exactly how ENABLE_VR is used and what it's for. This fix seems to work in the meantime but I wonder if we still even need ENABLE_VR at all.

keveleigh:
> but I wonder if we still even need ENABLE_VR at all

I like this thought for sure 😁 unless there's some core engine dependency that breaks when the flag isn't set, it feels like allowing these types to be serialized/referenced regardless of my current editor target would be ideal, whether they're actually "supported" on that platform or not.

PR #1529, opened 2022-04-06
FIX: Workaround for System.CodeDom test failures relating to Unity 2022 and .NET Standard

ekcoh:
### Description

We have suffered from failing CI on tests using System.CodeDom (https://docs.microsoft.com/en-us/dotnet/api/system.codedom?view=dotnet-plat-ext-6.0) on Unity 2022.x. This was previously suppressed and unresolved after conversations and previous investigations. Looking into this again it turns out that failure is due to cross referenced types from net-standard even when configured for net-framework. This cross-reference is within Unity and will be investigated further separately.

Symptoms was failures like:

```
The type `System.IEquatable`1<UnityEngine.Vector2>' is defined in an assembly that is not referenced. Consider adding a reference to assembly `netstandard, Version=2.1.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' in C:\Users\unity\AppData\Local\Temp\fnfqb8m3.0.cs:207
The type `System.IFormattable' is defined in an assembly that is not referenced. Consider adding a reference to assembly `netstandard, Version=2.1.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' in C:\Users\unity\AppData\Local\Temp\fnfqb8m3.0.cs:207
```

which where hidden from logs due to incorrect formatting of codedom compiler error messages.

Another symptom was that switching API profile for the Input project would result in compilation failure due to missing reference on net-standard.

### Changes made

- Fixed string formatting of codedom compiler failures so that all errors are included in the test assertion failure and hence logs.
- Explicitly load netstandard assembly (workaround) for Unity 2022 and add it to list of referenced assemblies from CodeDom compiler. This makes compilation possible.
- Removed previous symbols to disable editor tests on trunk.
- Added preprocessor defines to conditionally include these tests when not using .NET standard (default for our project). This yields successful compilation also when configured for .NET standard but results in less test coverage.

### Notes

Cross-references should be investigated separately and may eliminate the need for this workaround in the future.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -59,7 +59,6 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed missing tooltips in PlayerInputManagerEditor for the Player Limit and Fixed Splitscreen sizes labels ([case 1396945](https://issuetracker.unity3d.com/issues/player-input-manager-pops-up-placeholder-text-when-hovering-over-it)).
 - Fixed DualShock 4 controllers not working in some scenarios by adding support for extended mode HID reports ([case 1281633](https://issuetracker.unity3d.com/issues/input-system-dualshock4-controller-returns-random-input-values-when-connected-via-bluetooth-while-steam-is-running), case 1409867).
 - Fixed `BackgroundBehavior.IgnoreFocus` having no effect when `Application.runInBackground` was false ([case 1400456](https://issuetracker.unity3d.com/issues/xr-head-tracking-lost-when-lost-focus-with-action-based-trackedposedriver-on-android)).
-- Fixed a problem that would result in compilation errors if Input System test project was modified to .NET Standard API profile or test failures if tested against Unity 2022.
 
 #### Actions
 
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed missing tooltips in PlayerInputManagerEditor for the Player Limit and Fixed Splitscreen sizes labels ([case 1396945](https://issuetracker.unity3d.com/issues/player-input-manager-pops-up-placeholder-text-when-hovering-over-it)).
 - Fixed DualShock 4 controllers not working in some scenarios by adding support for extended mode HID reports ([case 1281633](https://issuetracker.unity3d.com/issues/input-system-dualshock4-controller-returns-random-input-values-when-connected-via-bluetooth-while-steam-is-running), case 1409867).
 - Fixed `BackgroundBehavior.IgnoreFocus` having no effect when `Application.runInBackground` was false ([case 1400456](https://issuetracker.unity3d.com/issues/xr-head-tracking-lost-when-lost-focus-with-action-based-trackedposedriver-on-android)).
+- Fixed a problem that would result in compilation errors if Input System test project was modified to .NET Standard API profile or test failures if tested against Unity 2022.
```
Rene-Damm: No changelog needed. Change isn't going in  package.

Rene-Damm (approved):
CHANGELOG can be removed. Otherwise LGTM.

jamesmcgill (approved):

PR #1531, opened 2022-04-07
Implement touch API

jamesmcgill:
### Description

Implement legacy Touch API using new input system

### Changes made

- Populate touch functions
- Add some basic tests

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_API_Touch.cs
+++ b/Assets/Tests/InputSystem/CoreTests_API_Touch.cs
@@ -1,10 +1,10 @@
 using System.Collections;
 using NUnit.Framework;
 using UnityEngine;
+using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.TestTools;
 using InputSystem = UnityEngine.InputSystem.InputSystem;
 using Touchscreen = UnityEngine.InputSystem.Touchscreen;
-using EnhancedTouch = UnityEngine.InputSystem.EnhancedTouch;
 
 partial class CoreTests
 {
@@ -70,8 +70,6 @@ partial class CoreTests
             Assert.That(first.type, Is.EqualTo(TouchType.Direct));
             Assert.That(first.altitudeAngle, Is.EqualTo(0));
             Assert.That(first.azimuthAngle, Is.EqualTo(0));
-            Assert.That(first.radius, Is.EqualTo(0));
-            Assert.That(first.radiusVariance, Is.EqualTo(0));
         }
         InputSystem.Update();
 
@@ -89,15 +87,7 @@ partial class CoreTests
             Assert.That(middle.rawPosition, Is.EqualTo(new Vector2(1.0f, 2.0f)));
             Assert.That(middle.deltaPosition, Is.EqualTo(new Vector2(14.0f, 23.0f)));
             Assert.That(middle.deltaTime, Is.EqualTo(0.1f).Within(float.Epsilon));
-            Assert.That(middle.tapCount, Is.EqualTo(0));
             Assert.That(middle.phase, Is.EqualTo(TouchPhase.Moved));
-            Assert.That(middle.pressure, Is.EqualTo(1.0f));
-            Assert.That(middle.maximumPossiblePressure, Is.EqualTo(1.0f));
-            Assert.That(middle.type, Is.EqualTo(TouchType.Direct));
-            Assert.That(middle.altitudeAngle, Is.EqualTo(0));
-            Assert.That(middle.azimuthAngle, Is.EqualTo(0));
-            Assert.That(middle.radius, Is.EqualTo(0));
-            Assert.That(middle.radiusVariance, Is.EqualTo(0));
         }
         InputSystem.Update();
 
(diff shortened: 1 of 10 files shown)


Assets/Tests/InputSystem/CoreTests_API_Touch.cs
```diff
@@ -0,0 +1,138 @@
...
+            first.azimuthAngle == second.azimuthAngle &&
+            first.radius == second.radius &&
+            first.radiusVariance == second.radius;
+    }
+
+    [UnityTest]
+    [Category("API")]
+    public IEnumerator API_CanReadTouchEventsThroughTouchesAPI()
+    {
+        Assert.That(Input.touchSupported, Is.False);
```
Rene-Damm, line 36: Prob good to try the other related APIs here before Touchscreen is added to make sure we get the expected defaults.

Packages/com.unity.input/Runtime/Input.cs
```diff
@@ -647,27 +648,125 @@ public static bool GetKeyUp(KeyCode key)
...
+            {
+                case InputSystem.TouchPhase.Began: return TouchPhase.Began;
+                case InputSystem.TouchPhase.Moved: return TouchPhase.Moved;
+                case InputSystem.TouchPhase.Ended: return TouchPhase.Ended;
+                case InputSystem.TouchPhase.Canceled: return TouchPhase.Canceled;
+                case InputSystem.TouchPhase.Stationary: return TouchPhase.Stationary;
+                    //InputSystem.TouchPhase.None  // These are filtered out and shouldn't be seen
+            }
+            throw new ArgumentOutOfRangeException(nameof(phase), $"Not expected phase value: {phase}");
+        }
```
Rene-Damm: Makes me wonder whether we should go and unify the two. We could add `None` to the trunk side version and kill the `InputSystem` side `TouchPhase`. Binary breaking change but just seems confusing we not only have two such enums now but are actively using both.

Anyway, not directly related to this PR. Just something for us to consider, I think.

Assets/Tests/InputSystem/CoreTests_API_Touch.cs
Rene-Damm: Also, test out of range access with `GetTouch`.

Packages/com.unity.input/Runtime/Input.cs
```diff
@@ -647,27 +648,125 @@ public static bool GetKeyUp(KeyCode key)
...
+            touch.radiusVariance = variance;
+
+            return touch;
         }
 
         public static Touch GetTouch(int index)
         {
-            return default;
+            if (index >= touchCount)
+                throw new IndexOutOfRangeException();
```
Rene-Damm: `IndexOutOfRangeException` should not be thrown this way. Use `ArgumentOutOfRangeException` instead.

Rene-Damm: (also need to make sure we are throwing a type equivalent to what the old implementation throws; can be subtype, though)

Packages/com.unity.input/Runtime/Input.cs
```diff
@@ -647,27 +648,125 @@ public static bool GetKeyUp(KeyCode key)
...
+            }
+        }
+        public static Touch[] touches
+        {
+            get
+            {
+                var result = new List<Touch>();
+                if (Touchscreen.current != null)
+                {
+                    var convertedTouches = from t in Touchscreen.current.touches
```
Rene-Damm: This needs to use the `EnhancedTouch` API. Polling `Touchscreen` directly will not give the right results. It'll drop touches and lead to incorrect touch sequences.

Packages/com.unity.input/Runtime/Input.cs
jamesmcgill: I changed it to match what I found here: https://github.cds.internal.unity3d.com/unity/unity/blob/ccbb0cecf33d617817244907b804def62b34e6b9/Runtime/Input/InputBindings.h

I.e. ArgumentException("Invalid touch index.");

jamesmcgill: Sorry I mean: throw new ArgumentOutOfRangeException("Invalid touch index.");

Rene-Damm: First arg must be parameter name.

```
throw new ArgumentOutOfRangeException(nameof(index));
```

Packages/com.unity.input/Runtime/Input_Touch.cs
```diff
@@ -0,0 +1,110 @@
...
+                EnhancedTouch.Touch.s_GlobalState.playerState.UpdateActiveTouches();
+                return EnhancedTouch.Touch.s_GlobalState.playerState.activeTouchCount;
+            }
+        }
+        public static Touch[] touches
+        {
+            get
+            {
+                if (!EnhancedTouchSupport.enabled)
+                    EnhancedTouchSupport.Enable();
```
Rene-Damm: This needs to be enabled in `OnEnteringPlayMode`.

Packages/com.unity.input/Runtime/Input_Touch.cs
```diff
@@ -0,0 +1,110 @@
...
+                throw new ArgumentOutOfRangeException(nameof(index), "Invalid touch index.");
+
+            return ToTouch(EnhancedTouch.Touch.activeTouches[index]);
+        }
+
+        public static int touchCount
+        {
+            get
+            {
+                EnhancedTouch.Touch.s_GlobalState.playerState.UpdateActiveTouches();
```
Rene-Damm: Why this one?

Packages/com.unity.input/Runtime/Input_Touch.cs
```diff
@@ -0,0 +1,110 @@
...
+                return EnhancedTouch.Touch.s_GlobalState.playerState.activeTouchCount;
```
Rene-Damm: This should use `Touch.activeTouches.Count`.

jamesmcgill: I'm currently seeing a memory leak somewhere with this function. 
I found that `Touch.activeTouches.Count` was calling `UpdateActiveTouches()` and then creating a temporary List, then returning the size of the list.
It actually has nothing to do with `UpdateActiveTouches()` creating a temporary list, but it was my first thought. I left it here because I thought it would be cheaper, but yeah, this is using the insides directly. I will change it back to use the list Count.

Packages/com.unity.input/Runtime/Input_Touch.cs
jamesmcgill: See below (was needed to mimic `Touch.activeTouches.Count` behaviour)

Packages/com.unity.input/Runtime/Input.cs
```diff
@@ -66,6 +67,10 @@ public static partial class Input // Partial to allow us to inject project-speci
 
         internal static void OnEnteringPlayMode()
         {
+            // EnhancedTouch is required for the Touch API
+            if (!EnhancedTouchSupport.enabled)
+                EnhancedTouchSupport.Enable();
```
Rene-Damm: ATM this thing functions like a counter. So just calling Enable here and Disable in Reset prob makes sense.

PR #1540, opened 2022-05-09
FIX: Composite touchscreen controls not firing action after enabling (ISXB-98)

jamesmcgill:
### Description

After the first enabling of an action map that is bound to composite touchscreen controls, the action would not fire if the touchscreen was not in the default state at this point. Which would happen if the screen was ever touched prior to enabling the action. (This is because Touchscreens are never in the default state if they were ever used. E.g. the position never returns to default and is essentially always actuated forever).

The issue was due to sorting behaviour introduced in https://github.com/Unity-Technologies/InputSystem/pull/1405
This sorting would reverse the order of the Touchscreen controls which just happened to work before because the Press Control was first in the list (after sorting everything was reversed; device order was Touch 4, Touch3, Touch 2 etc, and controls within are reversed, with Pressed now last).
This fragile behaviour of the Touchscreen controls always being actuated and the order mattering is not addressed in this PR. What this PR does, is restore the previous ordering but still doing the sorting. As sorting by composite complexity is required for the composite PR #1405 to work.  

### Changes made

- Change sorting to be only based on complexity but otherwise be stable sorting (e.g. no reversing of controls)
- Revert unit test changes that were done in PR #1405 which were only due to the ordering change.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -10151,7 +10151,7 @@ partial class CoreTests
         Assert.That(values.Count, Is.EqualTo(prepopulateTouchesBeforeEnablingAction ? 2 : 0)); // started+performed arrive from OnBeforeUpdate
         values.Clear();
 
-        // Now subsequent touches are ignored
+        // Now subsequent touches should not be ignored
         BeginTouch(200, new Vector2(1, 1));
         Assert.That(values.Count, Is.EqualTo(1));
         Assert.That(values[0].InputId, Is.EqualTo(200));
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -10044,6 +10044,117 @@ partial class CoreTests
             Assert.AreEqual(40.0f, magnitudes[5]);
         }
     }
+    
+    // Straight from demo project
+    public struct PointerInput
+    {
+        public bool Contact;
+        public int InputId;
+        public Vector2 Position;
+        public Vector2? Tilt;
+        public float? Pressure;
+        public Vector2? Radius;
+        public float? Twist;
+    }
+
+    // Straight from demo project
+    public class PointerInputComposite : InputBindingComposite<PointerInput>
+    {
+        [InputControl(layout = "Button")]
+        public int contact;
+
+        [InputControl(layout = "Vector2")]
+        public int position;
+
+        [InputControl(layout = "Vector2")]
+        public int tilt;
+
+        [InputControl(layout = "Vector2")]
+        public int radius;
+
+        [InputControl(layout = "Axis")]
+        public int pressure;
+
+        [InputControl(layout = "Axis")]
+        public int twist;
+
+        [InputControl(layout = "Integer")]
+        public int inputId;
+
+        public override PointerInput ReadValue(ref InputBindingCompositeContext context)
+        {
+            var contact = context.ReadValueAsButton(this.contact);
+            var pointerId = context.ReadValue<int>(inputId);
+            var pressure = context.ReadValue<float>(this.pressure);
+            var radius = context.ReadValue<Vector2, Vector2MagnitudeComparer>(this.radius);
+            var tilt = context.ReadValue<Vector2, Vector2MagnitudeComparer>(this.tilt);
+            var position = context.ReadValue<Vector2, Vector2MagnitudeComparer>(this.position);
+            var twist = context.ReadValue<float>(this.twist);
+
+            return new PointerInput
+            {
+                Contact = contact,
+                InputId = pointerId,
+                Position = position,
+                Tilt = tilt != default ? tilt : (Vector2?)null,
+                Pressure = pressure > 0 ? pressure : (float?)null,
+                Radius = radius.sqrMagnitude > 0 ? radius : (Vector2?)null,
+                Twist = twist > 0 ? twist : (float?)null,
+            };
+        }
+    }
+    
+    // https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-98
+    [Test]
+    [Category("Actions")]
+    [TestCase(true)]
+    [TestCase(false)]
+    public void Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works(bool prepopulateTouchesBeforeEnablingAction)
+    {
+        InputSystem.RegisterBindingComposite<PointerInputComposite>();
+
+        InputSystem.AddDevice<Touchscreen>();
+
+        var actionMap = new InputActionMap("test");
+        var action = actionMap.AddAction("point", InputActionType.Value);
+        for (var i = 0; i < 5; ++i)
+            action.AddCompositeBinding("PointerInput") // todo original binding had Touch0 .. Touch4 name set
+                .With("contact", $"<Touchscreen>/touch{i}/press")
+                .With("position", $"<Touchscreen>/touch{i}/position")
+                .With("radius", $"<Touchscreen>/touch{i}/radius")
+                .With("pressure", $"<Touchscreen>/touch{i}/pressure")
+                .With("inputId", $"<Touchscreen>/touch{i}/touchId");
+
+        var values = new List<PointerInput>();
+        action.started += ctx => values.Add(ctx.ReadValue<PointerInput>());
+        action.performed += ctx => values.Add(ctx.ReadValue<PointerInput>());
+        action.canceled += ctx => values.Add(ctx.ReadValue<PointerInput>());
+        
+        if (!prepopulateTouchesBeforeEnablingAction) // normally actions are enabled before any control actuations happen
+            actionMap.Enable();
+
+        // Start 5 touches, so we fill all slots [touch0, touch4] in Touchscreen with some valid touchId
+        for(var i = 0; i < 2; ++i)
+            BeginTouch(100 + i, new Vector2(100 * (i + 1), 100 * (i + 1)));
+        for(var i = 0; i < 2; ++i)
+            EndTouch(100 + i, new Vector2(100 * (i + 1), 100 * (i + 1)));
+        Assert.That(values.Count, Is.EqualTo(prepopulateTouchesBeforeEnablingAction ? 0 : 3));
+        values.Clear();
+
+        // Now when enabling actionMap ..
+        actionMap.Enable();
+        // On the following update we will trigger OnBeforeUpdate which will rise started/performed
+        // from InputActionState.OnBeforeInitialUpdate as controls are "actuated"
+        InputSystem.Update();
+        Assert.That(values.Count, Is.EqualTo(prepopulateTouchesBeforeEnablingAction ? 2 : 0)); // started+performed arrive from OnBeforeUpdate
+        values.Clear();
+
+        // Now subsequent touches are ignored
+        BeginTouch(200, new Vector2(1, 1));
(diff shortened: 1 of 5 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1540) <br/>All committers have signed the CLA.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10045,6 +10048,117 @@ public void Actions_RebindingCandidatesShouldBeSorted_IfAddingNewCandidate()
         }
     }
 
+    // Straight from demo project
```
ekcoh, line 10051: Which demo project? Bug repro project? If this is the case it might be good to include bug issue number in case we need to trace it at some point. I noticed its in the changelog, but maybe add a link to public issue tracker here?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10045,6 +10048,117 @@ public void Actions_RebindingCandidatesShouldBeSorted_IfAddingNewCandidate()
...
+                Position = position,
+                Tilt = tilt != default ? tilt : (Vector2?)null,
+                Pressure = pressure > 0 ? pressure : (float?)null,
+                Radius = radius.sqrMagnitude > 0 ? radius : (Vector2?)null,
+                Twist = twist > 0 ? twist : (float?)null,
+            };
+        }
+    }
+
+    // https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-98
```
ekcoh, line 10109: Excellent you added this because I was writing a comment about it where it said "demo" project, either leave it or call it "repro project" for clarity

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10045,6 +10048,117 @@ public void Actions_RebindingCandidatesShouldBeSorted_IfAddingNewCandidate()
...
+    public void Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works(bool prepopulateTouchesBeforeEnablingAction)
+    {
+        InputSystem.RegisterBindingComposite<PointerInputComposite>();
+
+        InputSystem.AddDevice<Touchscreen>();
+
+        var actionMap = new InputActionMap("test");
+        var action = actionMap.AddAction("point", InputActionType.Value);
+        for (var i = 0; i < 5; ++i)
+            action.AddCompositeBinding("PointerInput") // todo original binding had Touch0 .. Touch4 name set
```
ekcoh: TODO intended?, not sure what this is referencing?

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
```diff
@@ -261,8 +261,15 @@ public void SortMonitorsByIndex()
                 // Insertion sort.
                 for (var i = 1; i < signalled.length; ++i)
                 {
-                    for (var j = i; j > 0 && listeners[j - 1].monitorIndex < listeners[j].monitorIndex; --j)
+                    for (var j = i; j > 0; --j)
                     {
+                        // Sort by complexities only to keep the sort stable
+                        // i.e. don't reverse the order of controls which have the same complexity
+                        var firstComplexity = (int)((listeners[j - 1].monitorIndex >> 48) & 0xff);
```
ekcoh: Not really following what is going on here, what are these rshift by 48 and mask byte? Maybe wort a comment what we have stored in MSB byte or maybe this is documented on the member?

ekcoh (approved):
@jamesmcgill Great description of the issue and fix in this PR and walking the reader through what was causing the problem. Impressed you were able to decipher this issue. Is there something regarding this behavior we should add or are missing in docs in your opinion?

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
ekcoh: I guess its some kind of priority combined with index?

jamesmcgill: I found that monitorIndex isn't really an index but some sort of bit packed field.
It's in InputActionState::ToCombinedMapAndControlAndBindingIndex().

Another alternative here would be to make InputActionState::SplitUpMapAndControlAndBindingIndex  public instead of private. And then extend it to pull out the complexity too. Then the code above would be simpler to read. The downside would be that there would be a bunch of extra unpacking and temporary variables needed here.  Probably this is packed for performance reasons so I wanted the simplest code.

Assets/Tests/InputSystem/CoreTests_Actions.cs
jamesmcgill: I actually don't know either. @jimon wrote the test. But probably I should tidy it up a bit before merging.

jamesmcgill:
Thanks for the review @ekcoh !
I've made some changes (cleanup the test, put the bit unpacking in a function, CHANGELOG etc.)

I found one place that the sorting was mentioned in the docs and I edited that part. But generally I think the sorting is an internal detail so I don't think we will need to add anything (especially as this PR just partially restores the previous behaviour rather than add something new).

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1333,11 +1333,18 @@ private void OnBeforeInitialUpdate()
...
-        // monitors. While we could look up map and binding indices from control indices, keeping
-        // all the information together avoids having to unnecessarily jump around in memory to grab
-        // the various pieces of data.
-
+        /// <summary>
+        /// Pack the mapIndex, controlIndex and bindingIndex components into a single monitor index value.
+        /// </summary>
+        /// <param name="mapIndex">Will hold the extracted mapIndex value after the function completes.</param>
+        /// <param name="controlIndex">Will hold the extracted controlIndex value after the function completes.</param>
+        /// <param name="bindingIndex">Will hold the extracted bindingIndex value after the function completes.</param>
```
andrew-oc: I don't think these param comments are correct. The arguments are input only, and are not changed by the method. These three values are packed together into the returned long. Ah, I see. The same comments are on the SplitUpMapAndControlBindingIndex method. Copy paste error?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1358,6 +1372,15 @@ private long ToCombinedMapAndControlAndBindingIndex(int mapIndex, int controlInd
             mapIndex = (int)((mapControlAndBindingIndex >> 40) & 0xff);
         }
 
+        /// <summary>
+        /// Extract the 'complexity' component from the provided bit packed argument (monitor index).
+        /// </summary>
+        /// <param name="mapControlAndBindingIndex">Represents a monitor index, which is a bit packed field containing multiple components.</param>
+        static public int GetComplexityFromMonitorIndex(long mapControlAndBindingIndex)
```
andrew-oc: Modifiers are in the wrong order. Should be public static int...Actually, it could be internal too. This is definitely not something we want to expose.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10045,6 +10048,116 @@ public void Actions_RebindingCandidatesShouldBeSorted_IfAddingNewCandidate()
...
+
+        // Now when enabling actionMap ..
+        actionMap.Enable();
+        // On the following update we will trigger OnBeforeUpdate which will rise started/performed
+        // from InputActionState.OnBeforeInitialUpdate as controls are "actuated"
+        InputSystem.Update();
+        Assert.That(values.Count, Is.EqualTo(prepopulateTouchesBeforeEnablingAction ? 2 : 0)); // started+performed arrive from OnBeforeUpdate
+        values.Clear();
+
+        // Now subsequent touches are ignored
```
jimon: `// Now subsequent touches should not be ignored`
:)

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
```diff
@@ -261,8 +261,15 @@ public void SortMonitorsByIndex()
...
                 {
-                    for (var j = i; j > 0 && listeners[j - 1].monitorIndex < listeners[j].monitorIndex; --j)
+                    for (var j = i; j > 0; --j)
                     {
+                        // Sort by complexities only to keep the sort stable
+                        // i.e. don't reverse the order of controls which have the same complexity
+                        var firstComplexity = InputActionState.GetComplexityFromMonitorIndex(listeners[j - 1].monitorIndex);
+                        var secondComplexity = InputActionState.GetComplexityFromMonitorIndex(listeners[j].monitorIndex);
+                        if (firstComplexity >= secondComplexity)
+                            break;
```
andrew-oc, line 271: If we remove sorting by monitorIndex, which includes the binding index in the lowest bits, we'll end up processing bindings in potentially a different order than they are specified in the input asset. I can't decide if that's ok or not. Wasn't there some discussion recently in devs-input about a team using the explicit order of bindings to fall through to less specific bindings if more specific ones weren't available? Can this break that?

andrew-oc (changes requested):

Assets/Tests/InputSystem/CoreTests_Actions.cs
jamesmcgill: Good catch thanks!

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
jamesmcgill: The sorting was merged on the 24th March and it effectively was reversing the order they would have been in before. So bindings with the sorting enabled were now in descending order e.g. 4,3,2,1. This PR change would keep them in whatever order they were before 24th March (which I think would be ascending). 

I'm not sure about the order in the asset or about the binding fall through discussion, but I'm thinking unless these things were post 24th March, then they might have also been broken by adding the sort, and this PR should probably help restore those too?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
jamesmcgill: Cheers for the review!
Yeah looks like I forgot to change them after copying them. Good catch.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
jamesmcgill: Yep, thanks for pointing that out

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
ekcoh: XR team is using it for this use-case and might have a project to test with, contact @chris-massie may provide insight

andrew-oc: Yep, good point.

andrew-oc (approved):

PR #1543, opened 2022-06-01
Handling InputActionAsset in InputActionTrace

jimon:
### Description

If  binding resolution triggers (due to adding new device, etc) when `InputActionTrace` is used on `InputAction` from `InputActionAsset`, we fail with debug exception that the provided object in `onActionChange` is unexpected.

Because in this case `onActionChange` is invoked with `InputActionAsset` object and not `InputAction` or `InputActionMap`, which is not handled by `InputActionTrace`.

### Changes made

- Added branch in `InputActionTrace` to handle `InputActionAsset`, and clone all relevant states from the asset.
- Modified the test to trigger the branch.

### Notes

I don't know why we have `m_ActionMapStateClones`, they don't seem to be used? Are they needed because `InputAction` indirectly uses memory from them?

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -2004,8 +2004,6 @@ partial class CoreTests
     [Category("Actions")]
     public void Actions_CanCreateActionAssetWithMultipleActionMaps()
     {
-        var gamepad = InputSystem.AddDevice<Gamepad>();
-
         var asset = ScriptableObject.CreateInstance<InputActionAsset>();
 
         var map1 = new InputActionMap("map1");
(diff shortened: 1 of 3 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1543) <br/>All committers have signed the CLA.

jamesmcgill (approved):
lgtm

andrew-oc (approved):
PR looks fine. Wanted to comment real quick that the m_ActionMapStateClones is needed so we can replay events later against the same set of bindings that were active when the events happened. If we didn't save them, all the binding indices and trigger states indices would potentially be wrong during a replay.

PR #1324, opened 2021-04-28
CHANGE: Rework focus handling.

Rene-Damm:
### Description

The focus handling we have currently on `develop` is fairly simplistic. It has caused both buggy behavior for users (see [121479](https://issuetracker.unity3d.com/issues/inputsystem-dot-lowlevel-dot-stateevent-dot-from-device-pointer-values-are-reset-and-dont-recover-when-clicking-outside-of-the-game-window) and [1222305](https://issuetracker.unity3d.com/issues/input-system-gamepad-stick-values-are-cached-when-changing-editor-window-focus)) and problems for other Unity components such as XR and render streaming.

This PR attempts to rework the behavior to both provide a default that hopefully works for the widest range of users out of the box (including XR) and at the same time provides options to flexibly adapt the default behavior to less common use cases.

### Changes made

The key change is to differentiate between `Application.runInBackground=false` and `Application.runInBackground=true` and to consider the property always `true` when in the editor.

If running in background is `false`, the behavior is simple:

* No change on focus loss (under the assumption that we don't advance the simulation while in the background anyway).
* Sync devices on focus gain (fetching latest state should automatically update all actions; devices that don't support syncs get reset).

If running in background is `true` (including *always* in the editor), the behavior is governed by the new `InputSettings.backgroundBehavior` setting:

* `ResetAndDisableNonBackgroundDevices`: Everything is keyed on `InputDevice.canRunInBackground`. Devices that support it don't get touched. Eveything else is disabled and reset while we're in the background. And then re-enabled and sync'd when we get back into focus. __This is the default__.
* `ResetAndDisableAllDevices`: Same but consider *no* device as being able to run in the background. So, essentially input is 'passivated' while we don't have focus.
* `IgnoreFocus`: Do nothing. Aimed at use cases like render streaming that basically run apps in an automation setting.

In the editor, we have the additional complications of a) supporting edit mode as its own thing and b) of Unity "simulating" `Application.isFocused` based on Game View focus but otherwise running continuously regardless of focus.

So, an additional new setting `InputSettings.editorInputBehaviorInPlayMode` governs what we consider "background" when we are in play mode (outside of play mode input always goes to the editor):

* `PointersAndKeyboardsRespectGameViewFocus`: Pointer and keyboard input is tied to Game View focus (i.e. goes to editor when `Application.isFocused==false` and goes to player otherwise). Other input simply always goes to the player regardless of focus. Means no `EditorWindow` can consume gamepad input, for example. __This is the default__.
[...]

PR #1544, opened 2022-06-01
FIX: UI sometimes ignoring the first mouse click event after losing and regaining focus

jamesmcgill:
### Description

We introduced a fix for unwanted click events occurring if Alt-tabbing out and then back in this PR: https://github.com/Unity-Technologies/InputSystem/pull/1324/
That solution was to ignore the first mouse click after returning focus, however this behaviour was happening all the time and causing wanted mouse clicks to be ignored.
This PR attempts to address this by only invoking this behaviour only for the case where a button click is in progress, so we still block the unwanted click on return for that case.

### Changes made

- check if a click was in progress and only perform the ignore click fix then
- added a test which exposes the issue

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.cs
@@ -3649,6 +3649,7 @@ internal class UITests : CoreTestsFixture
         scene.eventSystem.SendMessage("OnApplicationFocus", true);
 
         yield return null;
+        Assert.That(clicked, Is.False); // No spurious click events during focus changes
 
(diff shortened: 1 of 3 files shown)


jimon (approved):

ekcoh (approved):
LGTM, good catch.

PR #1547, opened 2022-06-07
UI controls in MultiplayerEventSystems should not change colour unexpectedly

andrew-oc:
[Jira](https://jira.unity3d.com/browse/ISXB-112)
[Issue Tracker](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-112)

### Description

This fixes an issue where UI controls would change colour when using MultiplayerEventSystems. The bug was caused by this previous fix (https://github.com/Unity-Technologies/InputSystem/pull/1443) that attempted to keep user navigation from jumping between different player UIs in local multiplayer games, which worked by making sure all UI controls were inside CanvasGroup controls, one for each player UI, and toggling the CanvasGroup interactable property between true and false, depending on what UI was currently updating. Unfortunately toggling the interactable property makes all child UI controls animate between active and disabled states, which changes their colour when transitions have been set.

### Changes made

This fix moves the logic to keep navigation inside a players UI to the InputSystemUIInputModule component, and reverts the previous CanvasGroup related changes to MultiplayerEventSystem. A 'localMultiPlayerRoot' property has been added that can be set when running a local multiplayer game and when executing the Move action, we first speculatively perform the navigation and disallow that move (don't execute the move event) if the target Selectable is outside the player root.

### Notes

With this change, InputSystemUIInputModule takes a reference to UnityEngine.UI, which I think should be ok.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.cs
@@ -2526,9 +2526,7 @@ internal class UITests : CoreTestsFixture
         Assert.That(players[1].eventSystem.currentSelectedGameObject, Is.SameAs(players[1].leftGameObject));
 
         Assert.That(players[0].leftChildReceiver.events, Is.Empty);
-        Assert.That(players[0].rightChildReceiver.events,
-            EventSequence(
-                OneEvent("type", EventType.Move))); // OnMove will still get called to *attempt* a move.
+        Assert.That(players[0].rightChildReceiver.events, Is.Empty);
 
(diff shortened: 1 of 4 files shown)


jamesmcgill (approved):
This looks much better

PR #1549, opened 2022-06-15
Only generate docs once per run when executing API tests

andrew-oc:
### Description

This is a simple change to make docfx run only once for any single test run. 

### Notes

This could have been done as a one time test fixture setup but we don't generate docs for all API tests and we don't want to pay the cost of docs generation for tests that don't care about it.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/Tests/InputSystem/APIVerificationTests.cs
```diff
@@ -412,9 +416,17 @@ private static bool IgnoreMethodForDocs(MethodDefinition method)
     }
 
     #if HAVE_DOCTOOLS_INSTALLED
-    ////TODO: move this to a fixture setup so that it runs *once* for all API checks in a test run
     private static string GenerateDocsDirectory(out string log)
     {
+        // only generate the docs once for any test run. This isn't in a one-time setup fixture method
```
jimon: Question, would it make more sense to move out docs tests into separate test class and do this under `[OneTimeSetUp]`?

I'm not gonna be too annoying about this, but we use NUnit for a reason, otherwise every test would have some custom voodoo setup code around it.

PS. Fine merging it any way

andrew-oc: Actually yes it would. Good idea.

jimon (approved):

PR #1552, opened 2022-06-28
Fix controller id leaks

damianh2k:
### Description

Fix id leaks from backend switch&rapid plug/unplug

### Requirements

native changes in 2020.3/input/fix-controller-id-leaks

### Changes made

these are an overview. detail in Notes below
- wgi can generate multiple add/remove gamepad events within one tick.
  'discovered' are always sent before any 'remove' events which can lead to leaks
  in both m_Devices[] and m_DisconnectedDevices[] due to this reordering.
  so: rework 'discovered' callback into an event.
  InputSystem sends native code an ioctl to query if this new mode is available,
  if it is, we adjust our hooks, otherwise we should work as before.
- 'removed' events during edit mode to play mode transition were dropped (leak).
  also caps for xinput and wgi differ, so the device matching fails anyway.
  fix: allow removed events & make directx caps consistent with wgi

### Notes

- leaks are visible in Input Debugger
- preventing code from filtering insert/removed events causes spurious
  events when backend switches. 
  if this is a problem, suggestions for a fix are in the comments of
  https://github.cds.internal.unity3d.com/unity/unity/commit/ebb779c08de5e26bbc89bd11d14c0ad1f5e8dfbf

detail from original changelists:

**1/fix for InputSystem device leaks from WGI connections**

requires https://github.cds.internal.unity3d.com/unity/unity/commit/6c968ed4f3e597fa6b42ed7ef8ab494b8164b8e2

_Symptom_
- Inactive ids leaked into the m_Devices[] and m_DisconnectedDevices[] arrays
- This is fairly reproducible by jiggling physical USB connectors in and out

_Cause_
- Gamepad discoveries aren’t sent as events (with SendInputEventsToScript()), 
  but are sent separately before (with SendDeviceDiscoveriesToScript()), 
  meaning chronological order is not maintained.
- The WGI thread often generates multiple events for a single gamepad between
  the main thread sending these out leading to cases like
  - Add 1, Remove 1, Add 2: interpreted as Add 1, Add 2, Remove 1. ‘Remove 1’ 
    moves the device to m_DisconnectedDevices, but that’s too late for ‘Add 2’ 
    to be able to recycle it, so we leak it to m_DisconnectedDevices
  - Remove 1, Add 2: Interpreted as Add 2 (doesn’t recycle) then Remove 1 
    (also leaks to m_DisconnectedDevices)

_Fix_
- Native code still sends connections using old SendDeviceDiscoveriesToScript()
  but also queue them as events
- InputSystem sends an IOCTL to the native code to see if it supports the event
  mechanism. 
  If so we don't, nothing changes.
  If we do, unhook the previous mechanism and listen for events instead. 
  When we receive the event, we replicate the previous call.
- Also fixes some unterminated strings which cause problems with my USB speaker

**2/ fix for InputSystem device leaks on backend switch**

requires https://github.cds.internal.unity3d.com/unity/unity/commit/ebb779c08de5e26bbc89bd11d14c0ad1f5e8dfbf

_Symptom_
- device ids leaked for any controllers connected when we switch backends

_Cause_
- Code near "There's a chance the solution here will prove inadequate on 
  the long run" filters out these messages as they come between ending 
  edit mode and entering play mode.

_Fix#1_
- Alter this filtering so it ignores "remove" messages in InputManager.
- However, this presents a new issue; when we get the inserts 
  (note: requires previous fix) the capabilities of the pad have changed 
  due to XInput and WGI differences so it isn't able to match the controller 
  and we leak it into m_DisconnectedDevices[]

_Fix #2_
- Alter XInput capabilities to match WGI ones
  (see code: i can't see a way to do the reverse)

_Note_
- The other fix i tried for this was to extend the DeviceRemoveEvent 
  message with an optional bool indicating "a matching controller won't come 
  back, don't add it to m_DisconnectedDevices[]".
  this also works but isn't as clean.
- This does mean we generate Removed/Disconnect/Add/Reconnected messages
  during this transition

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/Commands/SupportsConnectionsAsEventsCommand.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/Commands/SupportsConnectionsAsEventsCommand.cs
@@ -1,0 +1,33 @@
+using System.Runtime.InteropServices;
+using UnityEngine.InputSystem.Utilities;
+
+namespace UnityEngine.InputSystem.LowLevel
+{
+    /// <summary>
+    // Check if native code supports sending connections as events.
+    // Send it to deviceId 0 as it's a special "global" IOCTL that gets routed internally.
+    /// </summary>
+    [StructLayout(LayoutKind.Explicit, Size = kSize)]
+    internal struct SupportsConnectionsAsEventsCommand : IInputDeviceCommandInfo
+    {
+        public static FourCC Type { get { return new FourCC('C', 'N', 'E', 'V'); } }
+
+        internal const int kSize = InputDeviceCommand.kBaseCommandSize;
+
+        [FieldOffset(0)]
+        public InputDeviceCommand baseCommand;
+
+        public FourCC typeStatic
+        {
+            get { return Type; }
+        }
+
+        public static SupportsConnectionsAsEventsCommand Create()
+        {
+            return new SupportsConnectionsAsEventsCommand
+            {
+                baseCommand = new InputDeviceCommand(Type, kSize),
+            };
+        }
+    }
+}
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/Commands/SupportsConnectionsAsEventsCommand.cs.meta
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/Commands/SupportsConnectionsAsEventsCommand.cs.meta
@@ -1,0 +1,11 @@
+fileFormatVersion: 2
+guid: ad900433ff7ae6d4b9d682da3e36430b
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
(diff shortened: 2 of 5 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1552) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1552) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1552) it.</sub>

PR #1561, opened 2022-07-13
FIX: first player's UI controls stop working after a second player joins (ISXB-125)

jamesmcgill:
### Description

For multiplayer games where 2 or more players are being spawned from PlayerInputManager and each had their own UIInputModule, the players who joined later would disable the UIInputModule actions of the existing players.
This turned out to be just because 1) internally UIInputModule holds references to the actions which are then cloned and shared amongst the other players 2) setting new actions on UIInputModule causes the existing (shared) actions to be unhooked and disabled.

### Changes made

- Prevent a case when a UIInputModule is cloned in a disabled state would still attempt to disable (shared) actions when assigning a new actions asset. 
- Prevent double disabling/enabling occurring when assigning action assets (When SwapAction and UpdateReferenceForNewAction are combined in property assignment).

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -1401,6 +1401,52 @@ namespace UnityEngine.InputSystem.UI
             trackedDevicePosition = default;
         }
 
+        /// <summary>
+        /// Clones all the underlying action assignments, the <see cref="actionsAsset"/> as well as all individual
+        /// actions such as <see cref="leftClick"/>.
+        /// </summary>
+        /// <remarks>
+        /// This performs a deep copy of the actions, ensuring that multiple instances have their own instances and
+        /// will not interfere with one another's actions.
+        /// This should be called after instantiating clones that e.g. should be used by different players.
+        /// Without calling this the stored <see cref="InputActionReference"/> will effectively act as if they are shallow copies.
+        /// Calls to <see cref="AssignDefaultActions"/>, <see cref="UnassignActions"/> and <see cref="actionsAsset"/> would
+        /// affect all instances if the actions are not cloned first.
+        /// </remarks>
+        public void CloneActions()
+        {
+            if (m_ActionsAsset == null)
+                return;
+
+            var oldActions = m_ActionsAsset;
+            m_ActionsAsset = Instantiate(m_ActionsAsset);
+
+            void CheckAndReplace(ref InputActionReference actionRef, int map, int action)
+            {
+                if (actionRef != null && actionRef.action == oldActions.actionMaps[map].actions[action])
+                    actionRef = InputActionReference.Create(m_ActionsAsset.actionMaps[map].actions[action]);
+            }
+
+            for (var actionMap = 0; actionMap < oldActions.actionMaps.Count; actionMap++)
+            {
+                for (var binding = 0; binding < oldActions.actionMaps[actionMap].bindings.Count; binding++)
+                    m_ActionsAsset.actionMaps[actionMap].ApplyBindingOverride(binding, oldActions.actionMaps[actionMap].bindings[binding]);
+
+                for (var action = 0; action < oldActions.actionMaps[actionMap].actions.Count; action++)
+                {
+                    CheckAndReplace(ref m_CancelAction, actionMap, action);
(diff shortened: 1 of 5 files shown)


andrew-oc (commented):
I wonder if it would be simpler to always clone the actions in PlayerInput.InitializeActions? That would also fix this problem and we could finally get rid of that review comment from there :)

jamesmcgill:
Good idea.
I didn't get the part about the review comment though. Maybe you can explain to me after lunch?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1542,14 +1542,20 @@ private void EnableInputAction(InputActionReference inputActionReference)
             action.Enable();
         }
 
-        private static void DisableInputAction(InputActionReference inputActionReference)
+        private void DisableInputAction(InputActionReference inputActionReference, bool onDisableComponent = false)
```
andrew-oc: Is isComponentDisabling is a better name for this argument?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -2227,11 +2229,7 @@ private InputActionReference UpdateReferenceForNewAsset(InputActionReference act
             if (newAction == null)
                 return null;
 
-            var reference = InputActionReference.Create(newAction);
-            if (oldActionEnabled)
-                EnableInputAction(reference);
-
-            return reference;
+            return InputActionReference.Create(newAction);
```
andrew-oc, line 2232: Do we not still want the functionality where actions that were enabled before assigning a new asset are still enabled afterwards?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
andrew-oc: Wondering if we should rename this method to TryDisableInputAction now. It's become very conditional.

andrew-oc (commented):
Nice small fix in the end. Added some comments but nothing major.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jamesmcgill: sounds good

jamesmcgill: also good

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jamesmcgill: Actually no. This was the 'real' bug here. The other changes were for another bug that got me mixed up for a while.

`UpdateReferenceForNewAsset` is called only in conjunction with Action property Setters. These setters call `SwapAction` which also does this Disable/Enable dance. So what was happening was double refCount decrements and then double refCount increments. And in the sample project the double decrement would disabled the control early (took me a while to make a Unit test for exactly this scenario as it needs to be at refCount==2 before hitting this code, otherwise the double decrement wouldn't matter much).

jamesmcgill:
Thanks for the review and yeah thanks for pushing for the simpler fix, it turned out to be nice and small.

I do wonder though if in addition to these fixes, we should add the explict refCount guard I had previously. It adds some state but I don't think there are many of these UIInputModules in a scene.
My thinking is that I've explicitly plugged 2 ways it can go wrong with this, but there might easily be more hidden ways and the heavy handed solution should prevent them.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
andrew-oc: Ah, nice!

PR #1569, opened 2022-08-10
NEW: Improve performance when reading control values

andrew-oc:
### Description

This commit introduces the concept of control value caching where the system marks controls that have been changed as stale and when reading control values, either through InputControl.Readvalue or InputAction.ReadValue, only stale controls read from the underlying storage and process the result before returning it. Unchanged controls simply return the cached value to the caller.

### Changes made

The InputDeviceBuilder now creates a binary tree of ControlBitRangeNode elements to store the bit ranges that controls map to in the state memory of the device. The tree is stored in the device itself and when state is changed, either through a state event or through InputState.Change, it is traversed to figure out what controls have been changed. Changed controls are marked as stale and when reading the value of the control, only stale controls go to the device storage and pull out the value and run it through the control's processor stack. 

### Notes

The InputControl<TValue>.value property is ref readonly to improve performance when using some of the larger input control state structs, especially ones from XR like Bone and Eyes. This necessitated a change to those structs where all the properties have been made into fields so the readonly aspect can be enforced by the compiler at the call site, so now they're straight POCOs. 

This PR also lays the ground work for some further optimizations, namely:

- removal of state monitors
  State monitors exist to watch sections of memory and fire off input action processing when those sections change. They are implemented in O(n) time as a straight linear walk through all monitors which can get expensive for fully loaded action maps. With this PR, we can now tell what controls have changed given a piece of new state memory, so state monitors might not be necessary.
- improved performance of any code that uses the InputControlEnumerator, such as rebinding, InputSystem.onEvent callbacks, and especially listening for unpaired device activitiy.
- makes it possible to correctly fix wasPressed/ReleasedThisFrame

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -1896,6 +1897,26 @@ public DeviceBuilder WithStateOffsetToControlIndexMap(uint[] map)
...
+            {
+                var sizeOfNode = UnsafeUtility.SizeOf<InputDevice.ControlBitRangeNode>();
+                var numNodes = controlTreeNodes.Length / sizeOfNode;
+                device.m_ControlTreeNodes = new InputDevice.ControlBitRangeNode[numNodes];
+                fixed (byte* nodePtr = controlTreeNodes)
+                {
+	                for (var i = 0; i < numNodes; i++)
+	                {
+		                device.m_ControlTreeNodes[i] = *(InputDevice.ControlBitRangeNode*)(nodePtr + i * sizeOfNode);
+		                // device.m_ControlTreeNodes[i] = MemoryMarshal.Read<InputDevice.ControlBitRangeNode>(
```
ekcoh: Intentionally uncommented code or should it be removed?

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -708,6 +708,41 @@ internal bool disabledWhileInBackground
...
+            // is no child. The right child is always at the next index.
+            public short leftChildIndex;
+
+            // each node can point at multiple controls (because multiple controls can use the same range in memory and
+            // also because of overlaps in bit ranges). The control indicies for each node are stored contiguously in the
+            // m_ControlTreeIndicies array on the device, which acts as an indirection table, and these two values tell
+            // us where to start for each node and how many controls this node points at. This is an unsigned short so that
+            // we could in theory support devices with up to 65553 controls.
+            public ushort controlStartIndex;
+            public byte controlCount;
```
ekcoh, line 751: Is there a good reason to keep controlCount as a byte here since it will anyway get padded to the size of an ushort?

ekcoh (approved):
Code looks good but cannot say I follow on the design/reasoning behind the ControlBitRangeNode. I guess the ControlBitRangeNode is indirectly tested by code using it, but might be beneficial to test ControlBitRangeNode in isolation as well with some unit tests? Might also make it easier to grasp what is going on with the bits there.

ekcoh:
Out of curiosity, what performance improvements have you observed with relation to these changes? Based off the description it sound like this is an enabler for further improvements.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
andrew-oc: Should have been removed.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
andrew-oc: I was going for the smallest possible size here to keep the entire tree of ranges in L1, but I forgot to add the Pack attribute! Good catch.

andrew-oc:
> Out of curiosity, what performance improvements have you observed with relation to these changes? Based off the description it sound like this is an enabler for further improvements.

Micro-benchmarks showed about a 2x performance improvement when reading the value of simple controls with no processors, and that gets better the more complex a control is i.e. more child controls, especially AxisControl, and better again the more processors a control has. One real world scenario that should give us a solid performance improvement from this PR alone though is having something like a four way composite attached to 16 different controls (something that's given us problems before, particularly on Switch). In the past, we'd read the value of each of those 16 controls at least twice per composite ReadValue call (once for the value and once for the magnitude check). Now we'll only read the value of the composite parts that have actually changed.

Rene-Damm:
Beautiful work.

andrew-oc:
> Beautiful work.

:bow:

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -708,6 +709,55 @@ internal bool disabledWhileInBackground
...
+        internal ControlBitRangeNode[] m_ControlTreeNodes;
+
+        // An indirection table for control bit range nodes to point at zero or more controls. Indices are used to
+        // point into the m_ChildrenForEachControl array.
+        internal ushort[] m_ControlTreeIndices;
+
+        // When a device gets built from a layout, we create a binary tree from its controls where each node in the tree
+        // represents the range of bits that cover the left or right section of the parent range. For example, starting
+        // with the entire device state block as the parent, where the state block is 100 bits long, the left node will
+        // cover from bits 0-50, and the right from bits 50-100. For the left node, we'll get two more child nodes where
```
jimon: nitpick: 0-50 and 51-99, because in 100 bit log state block bit indices are from 0 to 99 inclusive

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -906,5 +956,130 @@ internal static TDevice Build<TDevice>(string layoutName = default, string layou
...
+            }
+
+            if (rightNode.leftChildIndex != -1)
+                WriteChangedControlStatesInternal(statePtr, stateSizeInBits, deviceStatePtr,
+                    rightNode, leftNode.endBitOffset);
+        }
+
+        private static unsafe bool HasDataChangedInRange(byte* deviceStatePtr, void* statePtr, uint startOffset, uint sizeInBits)
+        {
+            if (sizeInBits == 1)
```
jimon, line 1129: Is this branch necessary? Feels like MemCmp should contain single bit optimization instead

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -991,17 +1022,73 @@ internal virtual IEnumerable<object> GetProcessors()
     public abstract class InputControl<TValue> : InputControl
         where TValue : struct
     {
+        /// <summary>
+        /// Returns the current value of the control after processors have been applied.
+        /// </summary>
+        /// <remarks>
+        /// Note that this property implements caching to avoid applying processors when the underlying
+        /// control has not changed. With this in mind, be aware of processors that use global state, such
+        /// as the <see cref="Processors.AxisDeadzoneProcessor"/>. Unless the control has been actuated, the processors
```
jimon: Feels like we would benefit from a diagnostic mode where we always run full value read from with-in value and compare it to cached value, that way we would catch situations like changing globals in processors and make it easier to diagnose them

jimon (approved):
LGTM, nice work 👍 
I have some feeling that maybe there are more places where `MarkAsStale` should be called, and maybe we forgot some, so maybe it makes sense to add an escape hatch just-in-case, e.g. some internal feature flag that always forces use of uncached values.

jimon (changes requested):
Would probably be a better idea to add an opt-in or opt-out flag to how useCachedValue is computed so we are able to hot-fix it by disabling the feature if it breaks in the wild

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
andrew-oc: It should but it has too many arguments that also need to be checked before you can decide to do the single bit optimization and you end up with more branches instead of fewer.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
andrew-oc: Agree and I started writing this but then it got out of hand. Will add it as a separate PR at some point

jimon (approved):

Assets/Tests/InputSystem/APIVerificationTests.cs
```diff
@@ -517,22 +537,27 @@ private static string FilterIgnoredChanges(string line)
         }
     }
 
+    /// <summary>
+    /// Use a scoped exclusion property to exclude members of a type from API verification when the member's names are not
+    /// unique in the entire project and you don't want to exclude the unrelated members. This type will scope the exlusion
```
jamesmcgill, line 527: Teeny-tiny typo "exlusion"

ProjectSettings/ProjectSettings.asset
```diff
@@ -621,7 +621,7 @@ PlayerSettings:
   webGLThreadsSupport: 0
   webGLWasmStreaming: 0
   scriptingDefineSymbols:
-    1: UNITY_ENABLE_STEAM_CONTROLLER_SUPPORT
+    1: UNITY_ENABLE_STEAM_CONTROLLER_SUPPORT;UNITY_INPUT_SYSTEM_CONTROL_VALUE_CACHING
```
jamesmcgill: Intentional I assume.
Would this mean CI and/or local TestRunner tests would have this enabled?
What about packaged Sample projects. Would they be affected?
Any other implications?

jamesmcgill (approved):
Nice! +1000 bonus points for fixing WasPressedThisFrame

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -991,17 +1022,77 @@ internal virtual IEnumerable<object> GetProcessors()
...
+        /// Also note that this property returns the result as ref readonly. If custom control states are in use, i.e.
+        /// any controls not shipped with the Input System package, be careful of accidental defensive copies
+        /// <see href="https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#avoid-defensive-copies"/>.
+        /// </remarks>
+        /// <seealso cref="ReadValue"/>
+        public virtual ref readonly TValue value
+        {
+            get
+            {
+#if UNITY_INPUT_SYSTEM_CONTROL_VALUE_CACHING
```
ekcoh: New API so not guarding for existing behavior

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -991,17 +1022,77 @@ internal virtual IEnumerable<object> GetProcessors()
...
         /// <remarks>
         /// This can only be called on devices that have been added to the system (<see cref="InputDevice.added"/>).
+        ///
+        /// This call does not use caching and can be quite a bit less performant than using the <see cref="value"/>
+        /// property but has the advantage that it will always apply the controls processor stack.
         /// </remarks>
+        /// <seealso cref="value"/>
         public TValue ReadValue()
         {
             unsafe
```
jimon: Let's get ReadValue optimized as well

Packages/com.unity.inputsystem/InputSystem/Controls/InputProcessor.cs
```diff
@@ -76,6 +76,27 @@ internal static Type GetValueTypeFromType(Type processorType)
...
+            CacheResult = 0,
+
+            /// <summary>
+            /// Process value every call to <see cref="InputControl{TValue}.ReadValue()"/> even if unprocessed value has not been changed.
+            /// </summary>
+            EvaluateOnEveryRead = 1
+        }
+
+        /// <summary>
+        /// Caching policy of the processor. Override this property to provide a different value.
```
Rene-Damm, line 97: Should mention that `CacheResult` is the default.

PR #1572, opened 2022-08-18
Users can be removed without other users' assigned devices changing

andrew-oc:
[case ISXB-274](https://jira.unity3d.com/browse/ISXB-274)

### Description

This fixes an error where users paired devices could change when removing a different user. This only occurred when two users shared one or more devices. The code for removing assigned devices from a user was performing a reference equality search over the list of all devices in the system to find the index of the one to be removed. This would break when two users shared the same device (because the reference is the same) and the user being removed had been added *after* the user sharing the device.

### Changes made

Changed the device search so it only looks in the section of the devices array allocated to the user who we are trying to modify.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.


ekcoh (approved):
LGTM!

PR #1579, opened 2022-08-31
Improve performance of HID descriptor parsing and device matching

andrew-oc:
### Description

During domain reloads, and during InputTestFixture test teardown, all device descriptions are parsed to see if the system can add any devices that it knows about but couldn't add previously. For HID devices, that means parsing the HID descriptor, which on later version of Unity, contains quite a lot of data. On machines with many HID devices attached, this parsing can get quite expensive.

### Changes made

This PR replaces the parsing in the HID descriptor FromJson method with a custom predictive parser that depends on the descriptor being well formatted. If it fails, then processing falls back on the previous method. On 2022.1, this change resulted in full test runs going from 144 seconds down to 110. On previous Unity versions the test run improvement is only about two seconds as the HID descriptors have less data in them in those versions. Domain reload time is still improved though.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


jimon (approved):
Tbh, not happy with this solution, we should just have a fast json parser period, or not use json for this to begin with.

But a good solution today is better than a great solution tomorrow.

PR #1582, opened 2022-09-16
Fixing devices not being removed when unplugged during domain reload

jimon:
### Description

Device removal events where ignored during domain reload.

By pure luck, the way how Windows editor HID handling was implemented, meant that HID removal events where postponed after domain reload happened (due to windows message queue is not pumped during domain reload).

But on macOS situation is different and device removal events happen as expected during domain reload.

### Changes made

Only ignore state and delta state events that happened during domain reload, and process everything else.

PS. not sure if we do want to ignore text events as well then?


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1582) <br/>All committers have signed the CLA.

jamesmcgill (approved):

PR #1583, opened 2022-09-22
Improving state debugger to show live values, binary, etc

jimon:
### Description

Improved our state debugger window.

### Changes made

Added support for:
- Live refresh
- Binary view
- Highlighting changed values
- Showing Noise/Reset masks


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1583) <br/>All committers have signed the CLA.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputStateWindow.cs
```diff
@@ -330,18 +410,31 @@ private static class Styles
                 fontSize = EditorStyles.boldFont.fontSize - 2,
                 normal = new GUIStyleState { textColor = Color.black }
             };
+
+            public static GUIStyle hexLabel = new GUIStyle
+            {
+                fontStyle = FontStyle.Normal,
+                font = Font.CreateDynamicFontFromOSFont("Courier", EditorStyles.label.fontSize + 2),
```
ekcoh: Related to comment of Courier in sync, see https://www.grputland.com/2012/08/font-stacks-that-look-similar-in.html#cour20120806 for platform specific options for linux, mac, win if there is a desire to use something else. However I think this is fine as is and I think its the right call to align with other editor monospaced representations if there is nothing bundled.

ekcoh (approved):
LGTM, I think its a neat and useful improvement to the debugger. As a potential future additional improvement it might be useful to "fade out / mix back the color" changed bits/bytes over time (e.g. 2 seconds) - this doesn't need to be considered here.

PR #1585, opened 2022-09-22
Filtering out most of jittery bits on some gamepads

jimon:
### Description

Most gamepads have noisy stick controls. And this makes them become `Gamepad.current` when user doesn't actuate any controls, which in result is problematic if two or more gamepads connected to the system.

### Changes made

- Added notion of `jitterMask`, which can mark standalone bits in the control state as noise.
- Filtered out lower 3 bits on axes of DualShock4/DualSense/SwitchPro controllers.
- Moved Switch Pro controller to different update tick, so we don't set it current on every event (this is a weird sideeffect of `IInputStateCallbackReceiver` contract, and I couldn't fix it properly because resetting of delta controls depends on it apparently).

### Notes

Unfortunately this is not a complete fix, but it does improve the situation by quite a bit.
A proper fix would be doing LPF filtering on incoming data to observe magnitude of change, rather than individual bits.
This fix breaks where value goes between 0x7f and 0x80, this is because all 8 bits then change, and filtering on just noisy bits is no longer effective.

### DIFF
--- a/Assets/Tests/InputSystem/SwitchTests.cs
+++ b/Assets/Tests/InputSystem/SwitchTests.cs
@@ -82,6 +82,61 @@ internal class SwitchTests : CoreTestsFixture
 
     [Test]
     [Category("Devices")]
+    public void Devices_SwitchProAxisJitter_DoesntMakeDeviceCurrent()
+    {
+        var device1 = InputSystem.AddDevice<SwitchProControllerHID>();
+        var device2 = InputSystem.AddDevice<SwitchProControllerHID>();
+
+        InputSystem.QueueStateEvent(device1,
+            new SwitchProControllerHIDInputState
+            {
+                leftStickX = 0x7f,
+                leftStickY = 0x7f,
+                rightStickX = 0x7f,
+                rightStickY = 0x7f,
+            });
+        InputSystem.QueueStateEvent(device2,
+            new SwitchProControllerHIDInputState
+            {
+                leftStickX = 0x7f,
+                leftStickY = 0x7f,
+                rightStickX = 0x7f,
+                rightStickY = 0x7f,
+            });
+        InputSystem.Update();
+        Assert.That(Gamepad.current, Is.EqualTo(device2));
+
+        var jitterMask = 0b111;
+
+        InputSystem.QueueStateEvent(device1,
+            new SwitchProControllerHIDInputState
+            {
+                leftStickX = (byte)(0x7f - jitterMask),
+                leftStickY = (byte)(0x7f - jitterMask),
+                rightStickX = (byte)(0x7f - jitterMask),
+                rightStickY = (byte)(0x7f - jitterMask),
+            });
+        InputSystem.Update();
+        Assert.That(Gamepad.current, Is.EqualTo(device2));
+
+        InputSystem.QueueStateEvent(device1,
+            new SwitchProControllerHIDInputState
+            {
+                leftStickX = (byte)(0x7f - jitterMask - 1),
+                leftStickY = (byte)(0x7f - jitterMask - 1),
+                rightStickX = (byte)(0x7e - jitterMask - 1),
+                rightStickY = (byte)(0x7e - jitterMask - 1),
+            });
+        InputSystem.Update();
+        Assert.That(Gamepad.current, Is.EqualTo(device1));
+
+        ////TODO: 7f <-> 80 jitter does break this, because it changes all the bits:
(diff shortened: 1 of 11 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1585) <br/>All committers have signed the CLA.

Assets/Tests/InputSystem/SwitchTests.cs
```diff
@@ -80,6 +80,61 @@ private static SwitchProControllerHIDInputState StateWithButton(SwitchProControl
...
+            {
+                leftStickX = (byte)(0x7f - jitterMask - 1),
+                leftStickY = (byte)(0x7f - jitterMask - 1),
+                rightStickX = (byte)(0x7e - jitterMask - 1),
+                rightStickY = (byte)(0x7e - jitterMask - 1),
+            });
+        InputSystem.Update();
+        Assert.That(Gamepad.current, Is.EqualTo(device1));
+
+        ////TODO: 7f <-> 80 jitter does break this, because it changes all the bits:
```
ekcoh, line 133: Based on the InputControl configuration at the bottom of this PR it seems like the jitterMask is set to 0x7. Wouldn't this imply that this behavior you mention related to comparing binary rather that than numerical difference imply that such a  "false positive" would occur for a "staircase model" for any single bit change outside the jitter mask e.g. 0x7 to 0x8 (and the opposite), as well 0xf to 0x10 (and the opposite) etc, having certain points along the value axis that will trigger this when transitioning over a multiple of 8? 

I think an interesting test case for this is sweeping a "representative value" e.g. x by incrementing it by 1 in a loop and have a value y = x + error where error is coming from a e.g. gaussian distribution or at least random distribution which never exceeds e.g. the jitterMask value range. It might be that the starcaising here is not possible to solve without LPF as you mention so maybe such a test case would only fail? A test case with known seend random generators can be useful though when testing models trying to mitigate or manage noise to create invariant tests where e.g. seed would typically be test parameter into the test. It seems like this test case is verifying a fixed specific case but it might be desirable to complement with a test case utilizing a psuedorandom element. Especially if implementation changes over time such a test case would still be valid.

ekcoh (approved):
LGTM and happy to hear it improves the situation .I have one suggestion mainly related to how the feature is verified/checked which might not be applicable given current tools at hand.

Assets/Tests/InputSystem/SwitchTests.cs
```diff
@@ -80,6 +80,61 @@ private static SwitchProControllerHIDInputState StateWithButton(SwitchProControl
...
+            });
+        InputSystem.Update();
+        Assert.That(Gamepad.current, Is.EqualTo(device2));
+
+        var jitterMask = 0b111;
+
+        InputSystem.QueueStateEvent(device1,
+            new SwitchProControllerHIDInputState
+            {
+                leftStickX = (byte)(0x7f - jitterMask),
```
jamesmcgill, line 114: I don't think I understand what's going on here. Why device1 didn't become current after this?
The result is 0b0111'1000, so there are 4 set bits outside of the jitter range that are still set and should still be picked up as a change? Or am I missing something?

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
```diff
@@ -147,7 +147,7 @@ namespace UnityEngine.InputSystem.Switch
     /// A Nintendo Switch Pro controller connected to a desktop mac/windows PC using the HID interface.
     /// </summary>
     [InputControlLayout(stateType = typeof(SwitchProControllerHIDInputState), displayName = "Switch Pro Controller")]
-    public class SwitchProControllerHID : Gamepad, IInputStateCallbackReceiver, IEventPreProcessor
+    public class SwitchProControllerHID : Gamepad, IInputUpdateCallbackReceiver, IEventPreProcessor
```
andrew-oc, line 150: Are we removing public methods in here?

andrew-oc (commented):
Cool implementation. Was the intention to remove jitter when a control is at its rest point, or jitter all along its entire range of motion? If I understand the PR correctly, it will lower the sensitivity of a jittery control so the device only becomes current when the control is moved in increments of jitterMask. Is that right? Do we want that?

jimon:
Closing in favour of #1611

PR #1586, opened 2022-09-24
Fix issues with OnScreenStick when used with PlayerInput and auto-switching of devices

andrew-oc:
### Description

Trying to use OnScreenStick in a scene with a PlayerInput component set to automatically switch devices (or any code that pairs and unpairs devices) causes the stick to jitter. This happens because the PlayerInput component sees the input from the OnScreenStick as coming from an unpaired device and tries to pair it and that device switching causes active actions to be cancelled, which means the OnScreenStick would receive a pointer up event even though the physical control being used to move the stick hasn't been released.

### Changes made

Added a new mode to OnScreenStick where it can use a set of unique input action instances to get pointer down, move, and up events. In this mode, UI events from IPointerDownHandler, IDragHandler, and IPointerUpHandler are skipped, and movement is not cancelled if devices are paired or unpaired because the input actions don't share state with any others.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.


ekcoh (approved):
Code looks good, notice this adds a feature so should go into next minor release. While reviewing this PR I realize that when we have a one device driving another like in this scenario, e.g. the existing auto-paring/auto-switch causes issues since it is viewing these as a flat structure of devices. However if the auto-pairing/switching code could determine that the "onstick device" is driven by the other underlying physical device, e.g. mouse or touch and this is instead conceptually reasoned about as a multi-interface device or a hierarchy, this could maybe be avoided without additional settings or what do you think? I realize there is a lot of work/design already done here, so mainly wanted to share thoughts to get your perspective.

I'll mark the PR with a "feature" label to make it more visible that its a feature and not a fix on existing API surface.

PR #1590, opened 2022-10-05
Not adding temp settings object to preloaded assets, and actually setting dirty flag so editor saves asset on exit

jimon:
### Description

- Adding a temp setting object to preloaded assets would add `{fileID: 0}` to saves project settings preloaded assets field, that then deserializes to null object, which we don't have a good way to deal with ... e.g. do we remove it? do we ignore it? what else?
- IL2CPP build pipeline forcefully saves project settings between `OnPreprocessBuild` and `OnPostprocessBuild`, so if we modify preloaded assets in both, what happens is project settings get saves before the build, then we modify it after the build but because dirty flag is cleared, project setting never get saved unless someone modifies them. So we get a dangling asset reference there.

### Changes made

- Ignoring temp setting object all together, there seem to be no point to try to add it to preloaded assets as there is no corresponding asset AFAIK.
- Leaving dirty flags as-is, that way project settings will be at least saved on unity shutdown.

### Notes

Maybe we need to force save project settings after the build?

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsBuildProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsBuildProvider.cs
@@ -16,11 +16,10 @@ namespace UnityEngine.InputSystem.Editor
             if (InputSystem.settings == null)
                 return;
 
-            // If there is no asset and we operate on temporary object,
-            // adding it would result in preloadedAssets containing null object "{fileID: 0}",
-            // which then we can't easily remove from the list in post build.
+            // If we operate on temporary object instead of input setting asset,
+            // adding temporary asset would result in preloadedAssets containing null object "{fileID: 0}".
             // Hence we ignore adding temporary objects to preloaded assets.
-            if (string.IsNullOrEmpty(AssetDatabase.GetAssetPath(InputSystem.settings)))
+            if (!EditorUtility.IsPersistent(InputSystem.settings))
                 return;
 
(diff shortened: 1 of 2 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1590) <br/>All committers have signed the CLA.

jamesmcgill (approved):
lgtm

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsBuildProvider.cs
```diff
@@ -9,65 +9,44 @@ namespace UnityEngine.InputSystem.Editor
...
             if (InputSystem.settings == null)
                 return;
 
-            var wasDirty = IsPlayerSettingsDirty();
-            m_SettingsAddedToPreloadedAssets = null;
+            // If there is no asset and we operate on temporary object,
+            // adding it would result in preloadedAssets containing null object "{fileID: 0}",
+            // which then we can't easily remove from the list in post build.
+            // Hence we ignore adding temporary objects to preloaded assets.
+            if (string.IsNullOrEmpty(AssetDatabase.GetAssetPath(InputSystem.settings)))
```
karljj1: Try using this instead https://docs.unity3d.com/ScriptReference/EditorUtility.IsPersistent.html

karljj1 (approved):

PR #1594, opened 2022-10-11
Add Tracking State support to TrackedPoseDriver

chris-massie:
### Description

This adds support for constraining the `TrackedPoseDriver` to only update the Transform if the XR device reports the position and rotation values are valid. An XR device may indicate that the position and/or rotation values are not valid if tracking is lost, so users should typically not update the Transform to match garbage data and instead keep it unchanged.

This is a replacement for the old PR https://github.com/Unity-Technologies/InputSystem/pull/1481 which had many open comments.

### Changes made

Added `public bool ignoreTrackingState` and `public InputActionProperty trackingStateInput` properties to `TrackedPoseDriver`, which requires a bump in minor version. The `Reset` method also applies expected control types of the input actions for ease of configuration by users.

### Notes

The default value of `ignoreTrackingState = true` keeps the existing functionality where the updating of the Transform is unconstrained by Tracking State.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
@@ -362,37 +362,31 @@ namespace UnityEngine.InputSystem.XR
 
         void OnPositionPerformed(InputAction.CallbackContext context)
         {
-            Debug.Assert(m_PositionBound, this);
             m_CurrentPosition = context.ReadValue<Vector3>();
         }
 
         void OnPositionCanceled(InputAction.CallbackContext context)
         {
-            Debug.Assert(m_PositionBound, this);
             m_CurrentPosition = Vector3.zero;
         }
 
         void OnRotationPerformed(InputAction.CallbackContext context)
         {
-            Debug.Assert(m_RotationBound, this);
             m_CurrentRotation = context.ReadValue<Quaternion>();
         }
 
         void OnRotationCanceled(InputAction.CallbackContext context)
         {
-            Debug.Assert(m_RotationBound, this);
             m_CurrentRotation = Quaternion.identity;
         }
 
         void OnTrackingStatePerformed(InputAction.CallbackContext context)
         {
-            Debug.Assert(m_TrackingStateBound, this);
             m_CurrentTrackingState = (TrackingStates)context.ReadValue<int>();
         }
 
         void OnTrackingStateCanceled(InputAction.CallbackContext context)
         {
-            Debug.Assert(m_TrackingStateBound, this);
             m_CurrentTrackingState = TrackingStates.None;
         }
 
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
@@ -542,7 +542,6 @@ namespace UnityEngine.InputSystem.XR
         /// Must support reading a value of type <see cref="Vector3"/>.
         /// </summary>
         /// <seealso cref="positionInput"/>
-        [Obsolete("positionAction has been deprecated. Use positionInput instead.")]
         public InputAction positionAction
         {
             get => m_PositionInput.action;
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -186,10 +290,24 @@ void BindRotation()
...
+        void BindTrackingState()
         {
-            UnbindPosition();
-            UnbindRotation();
+            if (m_TrackingStateBound)
+                return;
+
+            var action = m_TrackingStateInput.action;
+            if (action == null)
+                return;
```
vrdave-unity, line 300: There may be a scenario where a user externally sets `trackingStateInput = null`
```suggestion
            if (m_TrackingStateInput == null)
                return;

            var action = m_TrackingStateInput.action;
            if (action == null)
                return;
```

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -226,39 +344,75 @@ void UnbindRotation()
             m_RotationBound = false;
         }
 
+        void UnbindTrackingState()
+        {
+            if (!m_TrackingStateBound)
+                return;
```
vrdave-unity, line 350:
```suggestion
            if (!m_TrackingStateBound || m_TrackingStateInput == null)
                return;
```

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -287,21 +445,100 @@ protected void OnDisable()
...
+            if (trackingStateAction == null || trackingStateAction.m_BindingsCount == 0)
+            {
+                // Treat an Input Action Reference with no reference the same as
+                // an enabled Input Action with no authored bindings, and allow driving the Transform pose.
+                m_CurrentTrackingState = TrackingStates.Position | TrackingStates.Rotation;
+                return;
+            }
+
+            // Grab state.
+            var actionMap = trackingStateAction.GetOrCreateActionMap();
```
vrdave-unity, line 501: If we cannot assume m_TrackingStateInput is properly set, we need some extra null checks here unfortunately...
```suggestion
            if (m_TrackingStateInput != null && m_TrackingStateInput.action != null && !trackingStateAction.enabled)
            {
                // Treat a disabled action as the default None value for the ReadValue call
                m_CurrentTrackingState = TrackingStates.None;
                return;
            }
           
            if (m_TrackingStateInput == null || m_TrackingStateInput.action == null || trackingStateAction.m_BindingsCount == 0)
            {
                // Treat an Input Action Reference with no reference the same as
                // an enabled Input Action with no authored bindings, and allow driving the Transform pose.
                m_CurrentTrackingState = TrackingStates.Position | TrackingStates.Rotation;
                return;
            }

            // Grab state.
            var trackingStateAction = m_TrackingStateInput.action;
            var actionMap = trackingStateAction.GetOrCreateActionMap();
```

vrdave-unity (changes requested):
Looks pretty straight-forward. I had a couple of nitpicks with potential null ref cases.

vrdave-unity (approved):

unity-andrewc (approved):
I didn't have anything to add beyond Dave's feedback

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -287,21 +445,100 @@ protected void OnDisable()
...
+            actionMap.ResolveBindingsIfNecessary();
+            var state = actionMap.m_State;
+
+            // Get list of resolved controls to determine if a device actually has tracking state.
```
jimon, line 505: Do I understand that this logic is to check if there is any active binding on the action @chris-massie ?
Maybe a more robust way would be to identify bound devices or pose controls?

jimon:
Gonna rebase on latest develop to get up-to-date CI

jimon:
CI failures are expected and unrelated to this PR

Issue #1559, opened 2022-07-08
[API Proposal] Basic input slice of the new high level API


andrew-oc:
### Background and Motivation

This API is an effort to simplify using the Input System for new users, small projects, and/or experienced users/teams starting new projects. It should be possible for all users to implement basic input functionality in a zero-configuration, everything-just-works, environment. New users should have to learn a minimum number of new concepts to get started, without having to read any documentation, but existing users of the legacy input manager should feel some level of familiarity with the interface. The intent is also that it is possible to transition with minimum effort from this high level API into the lower level APIs piecemeal as a game progresses.

This slice of the high level API includes methods for users to get basic input only (read: no actions, maps, processors, interactions, bindings, assets, control schemes, action types, disambiguation, callbacks, composites, or any other low level concepts are involved).

### Proposed API

```csharp
namespace UnityEngine.InputSystem
{
  public static partial class Input
  {
    public static IReadOnlyList<Gamepad> gamepads { get; }
    public static IReadOnlyList<Joystick> joysticks { get; }
    public static ReadOnlyArray<InputSlot> gamepadSlotEnums { get; }
    public static int maxGamepadSlots { get; }

    public static bool IsPressed(Inputs input, InputSlot slot = InputSlot.All);
    public static bool WasPressedThisFrame(Inputs input, InputSlot slot = InputSlot.All);
    public static bool WasReleasedThisFrame(Inputs input, InputSlot slot = InputSlot.All);

    public static bool IsPressed(GamepadButton button, InputSlot slot = InputSlot.All);
    public static bool WasPressedThisFrame(GamepadButton button, InputSlot slot = InputSlot.All);
    public static bool WasReleasedThisFrame(GamepadButton button, InputSlot slot = InputSlot.All);

    public static bool IsPressed(JoystickButton button, InputSlot slot = InputSlot.All);
    public static bool WasPressedThisFrame(JoystickButton button, InputSlot slot = InputSlot.All);
    public static bool WasReleasedThisFrame(JoystickButton button, InputSlot  slot = InputSlot.All);

    public static float GetAxis(Inputs input, InputSlot slot = InputSlot.All);
    public static float GetAxis(Inputs negativeAxis, Inputs positiveAxis);

    public static Vector2 GetAxis(Inputs left, Inputs right, Inputs up, Inputs down);
    public static Vector2 GetAxisRaw(Inputs left, Inputs right, Inputs up, Inputs down);
[...]

PR #1595, opened 2022-10-18
First pass on querying controls in basic api

jimon:
Still think we should try method overloads from https://github.com/Unity-Technologies/InputSystem/issues/1559#issuecomment-1248152830

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/API/Input.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/API/Input.cs
@@ -620,7 +620,7 @@ namespace UnityEngine.InputSystem.HighLevel
         /// <summary>
         ///     Is the indicated control currently pressed.
         /// </summary>
-        /// <param name="input">Control from Inputs enum</param>
+        /// <param name="input">Control from Inputs enum.</param>
         /// <returns>True if control is currently pressed, false if control is not pressed or not available (device disconnected, etc).</returns>
         /// <remarks>
         ///     This will look at all devices of the appropriate type (which will depend on the specified Inputs)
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -254,48 +258,224 @@ public static partial class Input
         /// </summary>
         public static IReadOnlyList<Gamepad> gamepads { get; }
 
+#endif
+
+        private static ButtonControl GetButtonControl(Inputs input)
+        {
+            // WHAT THE FCK!? or .. can we do better?
```
jamesmcgill: A minor improvement (for enums in the keyboard range at least) might be to move the key mapping out of here and into a dedicated function. Eg.
`
if (input < Inputs.Key_MAX) {
    return Keyboard.current != null ? Keyboard.current[KEYMAP_FUN(input)] : null;
}
`

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -254,48 +258,224 @@ public static partial class Input
...
         /// </summary>
         /// <param name="input"></param>
         /// <returns></returns>
         /// <remarks>
-        /// This will look at all devices of the appropriate type (which will depend on the specified Inputs)
-        /// and return true if the control was released in the current frame on any of them.
+        ///     This will look at all devices of the appropriate type (which will depend on the specified Inputs)
+        ///     and return true if the control was released in the current frame on any of them.
         /// </remarks>
         public static bool IsControlUp(Inputs input)
```
jamesmcgill, line 707: Still got confused by the naming here (but maybe just me) as it doesn't imply it's temporal in nature.
I.e. this is going to return false the majority of the time, even when keys are not being pressed (and are 'up').

jamesmcgill (approved):
nice!

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
ekcoh: Agree with single null check and early return on that.

But why arent we just setting the enum values to the enum value of the underlying mapped type? If you want a translation table for some reason thats fine, but maybe generate it in that case? Or put it in an extension method for the enum and return just value (and then translations can be added there on a need basis)?

andrew-oc (changes requested):
Xmldoc comments should be updated for public methods as part of this PR

PR #1597, opened 2022-10-20
First pass on GetAxis

jimon:


Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -133,4 +134,41 @@ public void HighLevelAPI_CanQueryControl()
             Assert.That(Input.IsControlUp(input), Is.False, $"Input '{input}' should be 'not up'");
         }
     }
+
+    [Test]
+    [Category("HighLevelAPI")]
+    public void HighLevelAPI_CanQueryGetAxis()
```
jamesmcgill, line 140: Maybe add some keyboard and mouse examples to the test

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -731,28 +731,48 @@ public static bool IsControlUp(Inputs input)
             }
         }
 
-        // RE-ENABLE ME WHEN YOU GONNA IMPLEMENT ME
-#if false
         /// <summary>
-        /// For single analogue inputs
+        /// Returns actuation value for single analogous controls.
```
jamesmcgill: small typo: should be `analogue`

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -731,28 +731,48 @@ public static bool IsControlUp(Inputs input)
...
         /// <returns>A value between 0 at un-actuated to 1 at fully actuated.</returns>
-        /// <exception cref="NotImplementedException"></exception>
         public static float GetAxis(Inputs input)
         {
-            throw new NotImplementedException();
+            var deviceType = GetDeviceTypeForInput(input);
+
+            switch (deviceType)
+            {
+                case InputDeviceType.Keyboard:
```
jamesmcgill, line 746: All the case expressions look similar, except for a device type and GetXControl function.
Maybe `switch` could condensed down to a single expression?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -765,7 +785,15 @@ public static float GetAxis(Inputs negativeAxis, Inputs positiveAxis)
         /// <returns></returns>
         public static Vector2 GetAxis(Inputs left, Inputs right, Inputs up, Inputs down)
         {
-            throw new NotImplementedException();
+            var vec2 = GetAxisRaw(left, right, up, down);
+            var currentMag = vec2.magnitude;
+            var min = InputSystem.settings.defaultDeadzoneMin;
+            var max = InputSystem.settings.defaultDeadzoneMax;
+            if (currentMag < min)
+                return Vector2.zero;
```
jamesmcgill: Maybe redundant if `InverseLerp` will return zero anyway?

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Not sure, how do we go from runtime enum variable to a type that can be used in a generic?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Good catch!

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -789,7 +837,28 @@ public static Vector2 GetAxisRaw(Inputs left, Inputs right, Inputs up, Inputs do
...
+                case 0: return joystick.stick;
+                default:
+                    throw new ArgumentException($"Unexpected joystickAxis value '{joystickAxis}'");
+            }
+        }
+
+        private static IEnumerable<Joystick> GetJoysticksForSlot(JoystickSlot joystickSlot)
+        {
+            // TODO implement me properly
+            return InputSystem.devices.OfType<Joystick>();
```
andrew-oc, line 887: I feel that the design should really be one device per slot since that's how the user to device abstraction will work. A user is assigned a device slot, not to a device, so they can just use whatever device is in that slot. What's the idea behind having multiple devices per slot?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -802,9 +871,12 @@ public static Vector2 GetAxis(GamepadAxis stick, GamepadSlot gamepadSlot = Gamep
         /// </remarks>
         public static Vector2 GetJoystickAxis(int joystickAxis, JoystickSlot joystickSlot = JoystickSlot.Any)
         {
-            throw new NotImplementedException();
+            return GetJoysticksForSlot(joystickSlot)
+                .Aggregate(Vector2.zero, (current, joystick) => current + GetJoystickStickControl(joystick, joystickAxis).ReadValue());
```
andrew-oc: This Aggregate call will allocate every time because of the closure around joystickAxis. Not a problem if we move to one device per slot, but will need to be fixed if we don't.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Fixed

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Hm I have to keep a similar check because `currentMag` can be 0.0f and that we need to early out to avoid division by zero

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -765,7 +785,15 @@ public static float GetAxis(Inputs negativeAxis, Inputs positiveAxis)
         /// <returns></returns>
         public static Vector2 GetAxis(Inputs left, Inputs right, Inputs up, Inputs down)
         {
-            throw new NotImplementedException();
+            var vec2 = GetAxisRaw(left, right, up, down);
```
andrew-oc: Do I remember correctly that our coding convention says no abbreviations?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -731,28 +731,48 @@ public static bool IsControlUp(Inputs input)
...
         {
-            throw new NotImplementedException();
+            var deviceType = GetDeviceTypeForInput(input);
+
+            switch (deviceType)
+            {
+                case InputDeviceType.Keyboard:
+                    return InputSystem.devices
+                        .OfType<Keyboard>()
+                        .Max(x => Mathf.Clamp01(GetKeyboardButtonControl(x, input).ReadValue()));
```
andrew-oc: All of these Max lambdas will allocate on every call. Think we might have to do it the old fashioned way

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: It's just a stub to support passing `JoystickSlot.All`, for which we need to iterate through all devices. Otherwise it's one device per slot.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Is there a non-allocating LINQ or should I write a for-loop?

jimon:
Let's merge that as-is and fix any further comments in next PRs

PR #1598, opened 2022-10-21
First pass on implementing gamepad trigger press point and deadzone settings

jimon:


Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -849,8 +897,29 @@ private static float GetInputSystemDefaultDeadZone()
...
+                case GamepadSlot.Slot4:
+                case GamepadSlot.Slot5:
+                case GamepadSlot.Slot6:
+                case GamepadSlot.Slot7:
+                case GamepadSlot.Slot8:
+                case GamepadSlot.Slot9:
+                case GamepadSlot.Slot10:
+                case GamepadSlot.Slot11:
+                case GamepadSlot.Slot12:
+                    return s_GamepadConfigs[(int)gamepadSlot].deadzone;
```
ekcoh: I still do not see what we gain here by having GamepadSlot as an enum compared to plain integer? The only special thing seems to be "All" which could be a constant used as default argument? basically the enum prevents any kind of loop/enumeration processing on top which is a likely use-case?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -948,7 +1057,32 @@ public static bool IsGamepadButtonUp(GamepadButton input, GamepadSlot gamepadSlo
         /// </remarks>
         public static void SetGamepadTriggerPressPoint(float pressPoint, GamepadSlot gamepadSlot = GamepadSlot.All)
         {
-            throw new NotImplementedException();
+            // TODO remove me when init is implemented
+            InitializeGamepadConfigs();
+
+            switch (gamepadSlot)
```
ekcoh, line 1028: Similar comment as above to slot enum vs integer

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -849,8 +897,29 @@ private static float GetInputSystemDefaultDeadZone()
...
+                case GamepadSlot.Slot6:
+                case GamepadSlot.Slot7:
+                case GamepadSlot.Slot8:
+                case GamepadSlot.Slot9:
+                case GamepadSlot.Slot10:
+                case GamepadSlot.Slot11:
+                case GamepadSlot.Slot12:
+                    return s_GamepadConfigs[(int)gamepadSlot].deadzone;
+                case GamepadSlot.All:
+                    return s_GamepadConfigs[0].deadzone; // TODO what should it mean?
```
ekcoh: Indeed, feels like its not applicable if this is an aggregated gamepad shouldn't each individual be used for each individual gamepad and aggregation be on top of those processed results?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -937,6 +1006,46 @@ public static bool IsGamepadButtonUp(GamepadButton input, GamepadSlot gamepadSlo
         {
             throw new NotImplementedException();
         }
+#endif
+        
+        private static GamepadSlot FindSlotForGamepad(Gamepad gamepad)
+        {
+            // TODO implement me
```
ekcoh: Waiting for "stable arrays" for this one?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -937,6 +1006,46 @@ public static bool IsGamepadButtonUp(GamepadButton input, GamepadSlot gamepadSlo
...
+                case GamepadSlot.Slot6:
+                case GamepadSlot.Slot7:
+                case GamepadSlot.Slot8:
+                case GamepadSlot.Slot9:
+                case GamepadSlot.Slot10:
+                case GamepadSlot.Slot11:
+                case GamepadSlot.Slot12:
+                    return s_GamepadConfigs[(int)gamepadSlot].triggerPresspoint;
+                case GamepadSlot.All:
+                    return s_GamepadConfigs[0].triggerPresspoint; // TODO what should it mean?
```
ekcoh: Agree, similar comment as previously, aggregating this one e.g. min/max or similar would yield a different result than applying it on each individual contributor.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -804,7 +847,12 @@ private static Vector2 NormalizeAxis(Vector2 axis, float deadzone)
         public static Vector2 GetAxis(Inputs left, Inputs right, Inputs up, Inputs down)
         {
             var axis = GetAxisRaw(left, right, up, down);
-            var deadzone = GetInputSystemDefaultDeadZone(); // TODO account for controls being on gamepads
+            var deadzone = (GetDeviceTypeForInput(left) == InputDeviceType.Gamepad
```
ekcoh: Does this makes sense? It seems like its intended to get dead zone from gamepad slot, but Inputs given above might come from e.g. 4 different gamepads? what dead zone should be used? Feels like deadzone should be passed to this function to move that decision higher up or have some restriction in place?

ekcoh (commented):
Providing some initial comments but will have another look when I can digest the full file (not only diffs from this iteration). I will come back to switch-cases and this Inputs enum which currently seem to me like its working against a straight forward implementation.

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -185,4 +185,41 @@ public void HighLevelAPI_CanQueryGetAxis()
...
+        gamepadState.leftTrigger = 0.3f;
+        InputSystem.QueueStateEvent(gamepad, gamepadState);
+        InputSystem.Update();
+        
+        Input.SetGamepadTriggerPressPoint(0.5f);
+        Assert.That(Input.IsControlPressed(Inputs.Gamepad_LeftTrigger), Is.False);
+
+        Input.SetGamepadTriggerPressPoint(0.3f);
+        Assert.That(Input.IsControlPressed(Inputs.Gamepad_LeftTrigger), Is.True);
+    }
```
jamesmcgill, line 210: I would add a couple of extra tests just to check that if you don't touch `Input.SetGamepadTriggerPressPoint()` again, but instead inject some more `InputSystem.QueueStateEvent()/InputSystem.Update();` call that the pressed state changes that way too.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -270,6 +270,37 @@ public static class Input
 
 #endif
 
+        private struct GamepadConfig
+        {
+            public float triggerPresspoint;
```
jamesmcgill: I feel this should be `PressPoint` with extra capital P. But that is debatable.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Review fix: shouldn't be any deadzones here

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Review fix: shouldn't be all, and passing Any should throw an exception

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -948,7 +1057,32 @@ public static bool IsGamepadButtonUp(GamepadButton input, GamepadSlot gamepadSlo
...
+                case GamepadSlot.Slot6:
+                case GamepadSlot.Slot7:
+                case GamepadSlot.Slot8:
+                case GamepadSlot.Slot9:
+                case GamepadSlot.Slot10:
+                case GamepadSlot.Slot11:
+                case GamepadSlot.Slot12:
+                    s_GamepadConfigs[(int)gamepadSlot].triggerPresspoint = pressPoint;
+                    break;
+                case GamepadSlot.All:
```
jimon, line 1044: Remove Any or All, flip a coin :)

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: I think doing enums is fine for now, let's check that on first customer API feedback and if int's are preferred then switch to int's

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Fixed: changed method to throw exception if All is passed

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Yes, some methods are waiting for stable arrays.

But this one is redundant now, so fixed.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Fixed similar to previous method - we throw an exception if all is passed

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Fixed by removing gamepad deadzone here

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: Removed .Any

andrew-oc (approved):

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
jimon: fixed

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
jimon: fixed

PR #1600, opened 2022-10-24
FIX: some actions not being triggered when other duplicate control bindings exist (ISXB-254).

jamesmcgill:
### Description

In v1.4 we improved shortcut support by having actions consuming their inputs. This meant, for example, that SHIFT+B would trigger it's action but any action bound to the 'B' key would be blocked.
However we have been receiving many reports that users are not getting the expected behaviour in their projects that may have unknowingly (or knowingly) multiple actions bound to the same controls.
E.g. the issues with WASD keys, with the UIInputModule Navigate action conflicting with Player controls, if both action maps were enabled and many other similar cases that were reported.

This PR restores the previous behaviour by default and makes the the new event consumption behaviour (shortcut support) opt-in now and accessible as a toggle via project settings. 

### Changes made

Added a toggle in project settings with the default value `false`.
Modified the `SetInternalFeatureFlag` and `IsFeatureEnabled` to use this new value when queried for `kDisableShortcutSupport`. It was deliberately done this way to preserve the current (non-public) API, as users are aware of it now. Rather than doing the natural thing of changing `kDisableShortcutSupport` to `kEnableShortcutSupport` or removing it entirely, which would have been cleaner.

### Notes
`InputSystem.settings.SetInternalFeatureFlag("DISABLE_SHORTCUT_SUPPORT", true);` is still available here as some users are aware of it now, but a deprecation warning is being emitted.
The setting a private field, but it does need to be saved in the asset.
I am preparing a v1.5 version of this with that removed and cleans up the settings code with proper api.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -14,6 +14,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed `ArgumentNullException` when opening the Prefab Overrides window and selecting a component with an `InputAction`.
 - Fixed `{fileID: 0}` getting appended to `ProjectSettings.asset` file when building a project ([case ISXB-296](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-296)).
 - Fixed `Type of instance in array does not match expected type` assertion when using PlayerInput in combination with Control Schemes and Interactions ([case ISXB-282](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-282)).
+- Reverted the `InputActions consume their inputs` behaviour introduced in v1.4. This shortcut support feature can still be enabled by is switched off by default for now ([case ISXB-254](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-254))).
 
 ## [1.4.3] - 2022-09-23
 
@@ -26,7 +27,6 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Improved performance of HID descriptor parsing by moving json parsing to a simple custom predicitve parser instead of relying on Unity's json parsing. This should improve domain reload times when there are many HID devices connected to a machine.
-- Disable shortcut support by default. This restores the behaviour from v1.3 and fixes the case of actions not triggering if multiple actions are enabled using the same controls ([case ISXB-254](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-254))).
 
 ## [1.4.2] - 2022-08-12
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -14,7 +14,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed `ArgumentNullException` when opening the Prefab Overrides window and selecting a component with an `InputAction`.
 - Fixed `{fileID: 0}` getting appended to `ProjectSettings.asset` file when building a project ([case ISXB-296](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-296)).
 - Fixed `Type of instance in array does not match expected type` assertion when using PlayerInput in combination with Control Schemes and Interactions ([case ISXB-282](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-282)).
-- Reverted the `InputActions consume their inputs` behaviour introduced in v1.4. This shortcut support feature can still be enabled by is switched off by default for now ([case ISXB-254](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-254))).
+- The `InputActions consume their inputs` behaviour for shortcut support introduced in v1.4 is opt-in now and can be enabled via the project settings ([case ISXB-254](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-254))).
 
 ## [1.4.3] - 2022-09-23
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -26,6 +26,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Improved performance of HID descriptor parsing by moving json parsing to a simple custom predicitve parser instead of relying on Unity's json parsing. This should improve domain reload times when there are many HID devices connected to a machine.
+- Disable shortcut support by default. This restores the behaviour from v1.3 and fixes the case of actions not triggering if multiple actions are enabled using the same controls ([case ISXB-254](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-254))).
 
 ## [1.4.2] - 2022-08-12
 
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Improved performance of HID descriptor parsing by moving json parsing to a simple custom predicitve parser instead of relying on Unity's json parsing. This should improve domain reload times when there are many HID devices connected to a machine.
+- Disable shortcut support by default. This restores the behaviour from v1.3 and fixes the case of actions not triggering if multiple actions are enabled using the same controls ([case ISXB-254](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-254))).
```
ekcoh: Do you think we should list this as a change? Or should we call it out as this having side-effects on behavior in prior 1.4.x versions and call out out as a regression? E.g. "Reverted shortcut support being enabled by default as introduced in v.1.4.0. This restores the behavior...."

ekcoh: (I think either is fine and its a minor thing in practise if called out as a bug or change, but it might be highlighted when reviewed by RM as a different thing)

ekcoh (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
jamesmcgill: I think that's fair enough. I've updated it.
I also realised it was in the wrong place as well (wrong version)

andrew-oc:
My take on this is that if we're making it opt-in now, we should make it part of the input system settings and add a checkbox to the project settings dialogue to allow switching it on globally. That also gives us a chance to put a little warning message in there about the repercussions should you have multiple action maps enabled at the same time with multiple actions using the same bindings. 

If we don't go this way, then I'd suggest changing the wording in the changelog from "Reverted" to "Changed to opt-in" or similar. When I see reverted I think the behaviour has been removed entirely.

jimon (approved):
LGTM. input settings logic feels a bit error prone though

jimon (approved):

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -675,6 +675,17 @@ public void SetInternalFeatureFlag(string featureName, bool enabled)
             if (string.IsNullOrEmpty(featureName))
                 throw new ArgumentNullException(nameof(featureName));
 
+            // REMOVE: this is a temporary crutch to disable shortcut support by default but while also preserving the
+            // existing flag name, as users are aware of that now.
+            if (featureName == InputFeatureNames.kDisableShortcutSupport)
```
andrew-oc, line 680: What would happen if we simply removed the internal feature flag? If users are using it, it's to turn off the feature, but now it's opt in, so if we remove it, their code will have the same behaviour.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -145,6 +145,17 @@ public override void OnGUI(string searchContext)
                 EditorGUILayout.Space();
                 EditorGUILayout.PropertyField(m_EditorInputBehaviorInPlayMode, m_EditorInputBehaviorInPlayModeContent);
 
+                EditorGUILayout.Space();
+                EditorGUILayout.LabelField("Shortcut support", EditorStyles.boldLabel);
```
andrew-oc: "Shortcut support" is referred to as "Improved Shortcut Support" below. Feel this should be consistent here. Also pascal case "support".

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -145,6 +145,17 @@ public override void OnGUI(string searchContext)
...
+                EditorGUILayout.Space();
+                EditorGUILayout.PropertyField(m_ImprovedShortcutSupportEnabled, m_ImprovedShortcutSupportContent);
```
andrew-oc: Should we hide the warning help box if the feature is not enabled?

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -145,6 +145,17 @@ public override void OnGUI(string searchContext)
...
+
+                EditorGUILayout.HelpBox("Please note that enabling Improved Shortcut Support will cause actions to consume and block any other actions which are enabled and sharing the same controls. "
+                    + "E.g. when pressing the 'Shift+B' keys, the associated action would trigger but any action bound to just the 'B' key would be prevented from triggering at the same time, which would be the expected result in this example. "
+                    + "However, in other cases this might not give the desired result. E.g. if the 'W', 'A', 'S' and 'D' keys are bound to different actions which are all enabled at the same time, then only one of those actions will trigger. "
+                    + "Therefore if you enable this feature, be sure to remove any of these duplicate control bindings so that controls are only ever bound to a single action. Alternatively, you can disable actions or action maps at runtime to remove these conflicts."
+                    , MessageType.None);
```
andrew-oc: Some suggestions:

- "Please note that enabling Improved Shortcut Support will cause actions _with composite bindings_ to consume _input_ and block..."
- Could point out that blocking occurs even across action maps. I don't think users are really often aware of where the actions live and just enable entire assets.
- I feel we can leave out the repeated instruction about Shift-B and B since that is more about the features intended purpose and is already in the tooltip.
- Would this benefit from a more detailed worked example describing two action maps, one for UI and one for Gameplay, where both use the up, down, left, and right keys (pretty common one on the forums) and maybe some other action that will be lower priority and only uses left and right keys for example.
- It would help users to know how priority is decided between actions. Should we include some instructions on that, i.e. number of composite parts determines priority but actions with the same number of parts are non-deterministic?
- Is the advice really to remove duplicate controls? I like the second bit about disabling actions but maybe we could expand it to something like "Since event consumption only occurs for enabled actions, you can resolve most unexpected issues by ensuring that only those action maps that are relevant to your game's current context are enabled. Enable or disable actions as your game or application moves between different contexts."

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
jamesmcgill: That's true. I was thinking that there would be no programmatic way to enable/disable this until we add the public property. But your right, only users who were disabling it before should know about the existence of this and it won't hurt them that it does nothing now.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
jamesmcgill: Good catch

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
jamesmcgill: That would be nicer, yes

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
jamesmcgill: I'm going to leave it in for v1.4.4. There are some tests for shortcuts that I need to enable this programmatically somehow.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
jamesmcgill: Tried to reword this incorporating your suggestions

PR #1602, opened 2022-10-31
NEW: Add new public setting to enable shortcut key consumption (ISXB-254)

jamesmcgill:
### Description

Removes the `InputSystem.settings.SetInternalFeatureFlag("DISABLE_SHORTCUT_SUPPORT")` method for enabling/disabling shortcut keys/composite input consumption and replace it with the public property: `InputSystem.settings.shortcutKeysConsumeInput`

### Changes made

Remove references to `"DISABLE_SHORTCUT_SUPPORT"` 
Added `shortcutKeysConsumeInput` property

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -72,7 +72,7 @@ public void Actions_WhenShortcutsDisabled_AllConflictingActionsTrigger()
     [TestCase(false)]
     public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
     {
-        InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kDisableShortcutSupport, false);
+        InputSystem.settings.shortcutKeysConsumeInput = true;
```
andrew-oc, line 75: I wonder if the name of the property would be better if it described the goal instead of the technical details, something like enableEnhancedShortcutKeySupport. The details about event consumption can go in a tooltip or help box.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -660,6 +660,31 @@ public bool disableRedundantEventsMerging
...
 
+        /// <summary>
+        /// Improves shortcut key support by making composite controls consume control input
+        /// </summary>
+        /// <remarks>
+        /// Actions are exclusively triggered and will consume/block other actions sharing the same input.
+        /// E.g. when pressing the 'Shift+B' keys, the associated action would trigger but any action bound to just the 'B' key would be prevented from triggering at the same time.
+        /// Please note that enabling this will cause actions with composite bindings to consume input and block any other actions which are enabled and sharing the same controls.
+        /// Input consumption is performed in priority order, with the action containing the greatest number of bindings checked first.
+        /// Therefore actions requiring less keypresses will not be triggered if an action using more keypresses is triggered and has overlapping controls.
```
andrew-oc: fewer?

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -660,6 +660,31 @@ public bool disableRedundantEventsMerging
...
+        /// This works for shortcut keys, however in other cases this might not give the desired result, especially where there are actions with the exact same number of composite controls, in which case it is non-deterministic which action will be triggered.
+        /// These conflicts may occur even between actions which belong to different Action Maps e.g. if using an UIInputModule with the Arrow Keys bound to the Navigate Action in the UI Action Map, this would interfere with other Action Maps using those keys.
+        /// </remarks>
```
andrew-oc, line 675: I just realized we forgot to point out in all these instructions the very important fact that event consumption only occurs within the same input action asset. That's sort of the trap door for people who really still want to use shortcuts but still have multiple actions run. Think we should add that.

Assets/Tests/InputSystem/CoreTests_Actions.cs
jamesmcgill: I actually originally had something like enableEnhancedShortcutKeys here and changed it for similar reasons  :D
My reasoning is that 'enhanced' doesn't really say anything to the user. And the thing that actually happens, even from the users perspective is that their shortcut keys will start to block/consume other conflicting actions. That would be the only reason for them to switch it on. So I'm not sure it's such an implementation detail but happy to reconsider if you feel strongly about it?

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
jamesmcgill: My english is the baddest.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
jamesmcgill: Yep, I'll add that in, thanks.

andrew-oc (approved):

PR #1592, opened 2022-10-06
Update to InputSystem package to distinguish pointer events for multidisplay

3iD0loN:
*Purpose of this PR*:

New input system has a displayIndex field for generic devices Mouse, Pen, and Touchscreen, which represents the display that the input of any of these devices came from. Currently, this property only being set in certain Player build targets that support multidisplay. Specifically, this feature is only implemented on Windows. There is also no implementation that properly passes this index over for any of the Editor or Android build targets.

*Release Notes*:

Improved @joel.barba:
Input System: Touch input is handled from USB secondary capacitive touch screens on Android

*Functional Testing status*:

Test project was created with a customized version of InputSystem packages that make displayIndex for Touchscreen and Mouse public and blitted fields. There is also a customized version of UGUI that was changed to recognize UI elements on different displays. The project has a sample scene with two cameras for display 0 and display 1, and a script that enables multi-display. The script also prints out mouse display index and the position of the mouse.

Tests in Android Player include:
* Plugging in secondary monitor while Unity app is running.
* Unplugging in secondary monitor while Unity app is running.
* Plugging in secondary monitor while Unity app is not running and running the app.
* Unplugging in secondary monitor while Unity app is not running and running the app.
* Alternating between plugging and unplugging 100 times while app is running to note memory consumption over time.
* Pressing hardware touchscreen 1 while hardware touchscreen 2 is plugged in.
* Alternating between touching touchscreen 1 and 2 while touchscreen 2 is plugged in and noting that touches have a complete begin/update/end-cancel cycle.
* Testing the regions of the screen with a cursor image, including ensuring that corners and edge are properly recognized and mapped.
* Testing that a modified UGUI package can properly handle touches from display index 0 and 1 on the proper screens that the UI is displayed on.
* Testing that UGUI and Input System packages that are unmodified will return the original display index value of zero.
Insuring that touch gestures tuch as tapping and dragging are recognized for both touchscreen 1 and 2.
* Insuring that calls to screen ray casting works on touchscreen 1 and 2 to return proper values.
* testing that New Input system Action mapping recognizes display index as a action mapping target.

Tests in the Editors include:
* Editor is laid out so that there are no GameViews.
* Editor is laid out so there is only one GameView set to Display 1 target in a docked area on hardware monitor 1.
* Editor is laid out so there is only one GameView set to Display 2 target in a docked area on hardware monitor  1.
* Editor is laid out so there is only one GameView set to Display 1 target in a docked area on hardware monitor  2.
[...]

PR #1603, opened 2022-11-01
Add support for distinguishing pointer events from multiple displays

EricFulton:
### Description

New input system has a displayIndex field for generic devices Mouse, Pen, and Touchscreen, which represents the display that the input of any of these devices came from. Currently, this property only being set in certain Player build targets that support multidisplay. Specifically, this feature is only implemented on Windows. There is also no implementation that properly passes this index over for any of the Editor or Android build targets.

### Notes

This pull request almost entirely copies https://github.com/Unity-Technologies/InputSystem/pull/1592 with two key differences:

1. This PR contains tests to explore all new code paths.
2. This PR comes from my account instead of Joel's because I don't have access to his.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1603) <br/>All committers have signed the CLA.

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -3761,6 +3761,264 @@ public IEnumerator UI_WhenCursorIsLockedToScreenCenter_PointerEnterAndExitEvents
...
+
+        // Touch Display 1
+        scene.ClearEvents();
+        BeginTouch(1, new Vector2(100, 100), queueEventOnly: true, displayIndex: targetDisplay);
+        yield return null;
+        EndTouch(1, new Vector2(100, 100), queueEventOnly: true, displayIndex: targetDisplay);
+        yield return null;
+
+        // Verify Clicks On Display 1 Are Ignored On Display 0
+        Assert.AreEqual(0, scene.parentReceiver.events.Count);
```
3iD0loN, line 3848: Do you think it would be worth adding tests that verify old behavior as well?

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -3761,6 +3761,264 @@ public IEnumerator UI_WhenCursorIsLockedToScreenCenter_PointerEnterAndExitEvents
         Assert.That(callbackReceiver.events.Any(e => e.type == EventType.PointerExit), Is.True);
     }
 
+    #region Multi Display Tests
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY
```
3iD0loN: We might also want to limit this by the platforms that are currently supported.

WinEditor, MacEditor, LinuxEditor and AndroidPlayer have this implemented.

MacPlayer, WinPlayer, LinuxPlayer, and iOSPlayer do not, and I am concerned that this might raise false red flags when running for those platforms.

3iD0loN (changes requested):
I think the platform changes are the only ones necessary.

Assets/Tests/InputSystem/Plugins/UITests.cs
EricFulton: @3iD0loN I'm not sure I understand.  Do you mean writing a test to make sure valid clicks are _not_ discarded?  That should be covered by these tests.

But I think I may not be understanding.

Assets/Tests/InputSystem/Plugins/UITests.cs
EricFulton: No problem, I'll add those in.

Assets/Tests/InputSystem/Plugins/UITests.cs
EricFulton: I spoke with Joel offline. He meant to add a test that verifies displayIndex is not set when UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY isn't set.

I'm not sure this is possible to test in any meaningful way because pointerData.displayIndex didn't exist before this review.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed `Given object is neither an InputAction nor an InputActionMap` when using `InputActionTrace` on input action from an input action asset ([case ISXB-29](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-29)).
 - Fixing devices not being removed if unplugged during domain reload (entering or exiting play mode) ([case ISXB-232](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-232)).
 
+### Changed
+- Exposed `displayIndex` property for `Pointer`, `Touchscreen`, `TouchControl`, `TouchState`, `Mouse`, `MouseState`; Enable multi-display input for all platforms that implement it;
```
jamesmcgill: This should go in the Unreleased section above

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1660,13 +1660,28 @@ private int GetPointerStateIndexFor(InputControl control, bool createIfNotExists
                 touchId = touchscreen.primaryTouch.touchId.ReadValue();
                 touchPosition = touchscreen.primaryTouch.position.ReadValue();
             }
+
+            int displayIndex = 0;
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
```
jamesmcgill: The PR description mentioned this was NOT available in editors and Android. Is the logic wrong here or have I misunderstood?

jamesmcgill (approved):
Awesome, thanks!

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
EricFulton: I just confirmed with Joel, this change does work in editor and android only. So this logic should be correct 👍

Packages/com.unity.inputsystem/CHANGELOG.md
EricFulton: Will do! 👍

Assets/Tests/InputSystem/Plugins/UITests.cs
ekcoh: If there is no behavior that is different and may not be evaluated when UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY is not defined it makes sense to exclude the test, if there is it makes sense to provide conditionally compiled outcomes/asserts that depend on the behavior of the Unity version, e.g. 2023.x, 2022.x, 2021.x, 2020.x.

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -3761,6 +3761,264 @@ public IEnumerator UI_WhenCursorIsLockedToScreenCenter_PointerEnterAndExitEvents
         Assert.That(callbackReceiver.events.Any(e => e.type == EventType.PointerExit), Is.True);
     }
 
+    #region Multi Display Tests
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
```
ekcoh: I have a comment regarding this but place it together with another similar comment further down.

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -4019,6 +4277,9 @@ private static ExtendedPointerEventData ClonePointerEventData(PointerEventData e
                 twist = eventData.twist,
                 radius = eventData.radius,
                 radiusVariance = eventData.radiusVariance,
+#endif
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
```
ekcoh: Same comment as for other conditional symbol evaluations.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -55,6 +55,11 @@
             "name": "com.unity.ugui",
             "expression": "1.0.0",
             "define": "UNITY_INPUT_SYSTEM_ENABLE_UI"
+        },
+        {
+            "name": "com.unity.ugui",
+            "expression": "1.1.0",
+            "define": "UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY"
```
ekcoh: Am I correct to assume that UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY here is intended to define whether the associated UGUI version is capable of multi display management in its raycasts? How about multi-display support via UITK, I wonder if that is missing or dependent on UGUI which would make this sufficient?

The problem I see with this is that this define alone won't tell if there is sufficient multi-display support. What if the user has com.unity ugui package 1.1.0 and a Unity version not supporting this? Would that be resolved indirectly by constraints from ugui 1.1.0 on required trunk version which would imply this define guarantees ugui capability and Unity capability?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1845,6 +1865,11 @@ private int AllocatePointer(int pointerId, int touchId, UIPointerType pointerTyp
                 eventData = new ExtendedPointerEventData(eventSystem);
 
             eventData.pointerId = pointerId;
+
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
```
ekcoh: Same comment as other defines

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1751,6 +1766,11 @@ private int GetPointerStateIndexFor(InputControl control, bool createIfNotExists
                     eventData.device = device;
                     eventData.pointerType = pointerType;
                     eventData.pointerId = pointerId;
+
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
```
ekcoh: Same comment as other defines

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1660,13 +1660,28 @@ private int GetPointerStateIndexFor(InputControl control, bool createIfNotExists
...
+#endif
+
             if (touchId != 0)
                 pointerId = ExtendedPointerEventData.MakePointerIdForTouch(pointerId, touchId);
 
             // Early out if it's the last used pointer.
             // NOTE: Can't just compare by device here because of touchscreens potentially having multiple associated pointers.
             if (m_CurrentPointerId == pointerId)
+            {
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
```
ekcoh: Same comment as other defines

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
ekcoh: Would it be possible to define a single symbol that may be used in source and tests that handle this combined expression so we can avoid having the full expression in every occurance it needs to be checked, e.g.

#if ...long expression
#define UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY
#endif

in a centralized place. If that is done, then instead the conditional ugui package version symbol define should maybe instead be called something like UNITY_INPUT_SYSTEM_UGUI_SUPPORTS_MULTIDISPLAY and be used in the long expression. Since I assume something similar would be needed for UITK? Or maybe that can be derived directly from Unity capability since distributed with Unity.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased]
 
+### Changed
+- Exposed `displayIndex` property for `Pointer`, `Touchscreen`, `TouchControl`, `TouchState`, `Mouse`, `MouseState`; Enable multi-display input for all platforms that implement it;
```
ekcoh: I know this concept of displayIndex exist in Unity and has been reused in this package. It is a bit misleading since display in industry typically means a physical display panel or monitor, while this actually refer to a "logical screen". I wonder whether we should clarify or this, e.g. "Exposed 'displayIndex' property .... which enables looking up the associated logical screen associated with the event available via https://docs.unity3d.com/ScriptReference/Display.html". This would hopefully avoid confusion that the event could be traced back to a physical display panel which is not currently the case. Such tracability could in the future be made available by also providing a reference to e.g. DisplayInfo-like information, but such information would only apply to certain devices such as e.g. TouchScreen, EyeTracker etc. It might be good to call out in the property documentation for displayIndex that the result, or associated coordinates may be incorrect if the display is projected on multiple physical monitors, especially if their resolution or aspect ratio is different.
Out of interest did you test any scenario with "duplicated desktop" where 2 physical displays map to a single "logical screen"?

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
```diff
@@ -180,6 +180,11 @@ public class Pointer : InputDevice, IInputStateCallbackReceiver
         /// </remarks>
         public ButtonControl press { get; protected set; }
 
+        /// <summary>
+        /// The index of the display that was pointed.
+        /// </summary>
+        public virtual IntegerControl displayIndex { get; protected set; }
```
ekcoh: Why does this need to be virtual? What is the use-case for this? I was under the impression this would be a plain set/get?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1845,6 +1865,11 @@ private int AllocatePointer(int pointerId, int touchId, UIPointerType pointerTyp
                 eventData = new ExtendedPointerEventData(eventSystem);
 
             eventData.pointerId = pointerId;
+
+#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
+            eventData.displayIndex = displayIndex;
```
ekcoh, line 1869: Isn't this range assignment a cast from int to byte? Maybe I am wrong but I noticed there is both byte and int representations of the displayIndex so it would be good to debug-assert any occurances of truncation with a debug-only range check to detect any incorrect out-of-range effects

ekcoh (changes requested):
Thanks for doing this work @3iD0loN @EricFulton, support for multiple display setups seems to be an area that benefits from improvements so happy to see some engagement in this area. Also good to see some test coverage being added.

As a general note, why would this functionality be dependent on ugui presence? I would like to understand this better and I would suspect it is related to outcome of raycasts when using key sets like {x, y, displayIndex} (as a supstitute for virtual desktop coordinate) to derive display from a coordinate, but I get the impression that events coming from native carries displayIndex data already? Would love to understand why ugui updates are needed and if they only affect ray-cast support for multi display or if there is more to it?

Assets/Tests/InputSystem/Plugins/UITests.cs
ekcoh: As I understand this UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY will be defined by Unity. Hence if the Unity version the tests are running against define this, a certain behavior is expected, but if this is not defined another ("legacy") behavior is expected. If the test in question verify a behavior previously not covered by existing tests, it may be valuable that the test exercises behaviors of when UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY is defined as well as when it is not defined since Input System package may be combined with any Unity version e.g. 2023.x 2022.x, 2021.x 2020.x.
Since UGUI changes comes with trunk at least there is no mismatch combination between UGUI and Input System package in addition to this as I understand.

EricFulton: The first clause of your first comment is correct.  There is no new functionality when UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY isn't set so we should be good.

In reference to your second comment: 
If I'm understanding correctly, it sounds like you agree with Joel: that there should be a test added for when UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY is not defined.  Is that correct?

If so, I'm not sure what that test would be testing. This PR adds a new field called "displayIndex" the compilation of which is gated by this flag UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY.

When UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY is not defined, there is no new displayIndex to test.  The most you could do is use reflection to verify that there is no property called displayIndex, but that to me doesn't seem like a meaningful test.

I could TOTALLY be misunderstanding here so please correct me if I've got it wrong.

EricFulton:
> Thanks for doing this work @3iD0loN @EricFulton, support for multiple display setups seems to be an area that benefits from improvements so happy to see some engagement in this area. Also good to see some test coverage being added.
> 
> As a general note, why would this functionality be dependent on ugui presence? I would like to understand this better and I would suspect it is related to outcome of raycasts when using key sets like {x, y, displayIndex} (as a supstitute for virtual desktop coordinate) to derive display from a coordinate, but I get the impression that events coming from native carries displayIndex data already? Would love to understand why ugui updates are needed and if they only affect ray-cast support for multi display or if there is more to it?

Sorry I should have added more context to this PR.  Here are the other changes that have been made in support of the multi-display feature:

- [Android support](https://ono.unity3d.com/unity/unity/pull-request/134020/_/ikin-ryz-android-presentation-input-changes-trunk)
- [WinEditor/MacEditor/LinEditor](https://ono.unity3d.com/unity/unity/pull-request/142411/_/ikin-ryz-input-system-display-index-binding-trunk)
- [UGUI](https://github.cds.internal.unity3d.com/unity/unity/pull/16207)
- And, of course, this InputSystem PR

This is the extent of the changes made to support the multi-display feature for IKIN (IKIN is one of our clients that works with holograms).  I think the code changes probably will explain what's been changed better than I can, but Joel Barba can probably give you more insight into these changes.

I'm very new to Unity so I don't have a firm grasp of what information is or isn't already added to these events.

Packages/com.unity.inputsystem/CHANGELOG.md
EricFulton: So all together, you'd like it to read:
```
- Exposed `displayIndex` property for `Pointer`, `Touchscreen`, `TouchControl`, `TouchState`, `Mouse`, `MouseState` which enables look up of the logical screen associated with a pointer event via (display documentation)[https://docs.unity3d.com/ScriptReference/Display.html]
```
Do I have that right?


Regarding the "duplicated desktop" question: I'll have to defer to @3iD0loN here. Joel did a lot of testing outside of the tests in this PR. @3iD0loN would you mind sharing if that's something you did?

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
EricFulton: @3iD0loN can you comment here too? I don't have context on why it's virtual.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
EricFulton: I don't know of any method (short of adding that #define to the top of every file) of doing that.  Do you know a way to define that symbol conditionally and globally?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
EricFulton: I don't have a good understanding for the flow of execution through this codebase.  I added:
```
Debug.Assert(m_CurrentDisplayIndices[pointerIndex] <= byte.MaxValue, "Display index was larger than expected");
```

To TouchSimluation based on a previous comment on this PR (which I can't find now).  It's TouchSimulation:294.  

I'm happy to add more of those, but I have no idea if they're superfluous or not.  Do you know?

EricFulton: Added :)

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
EricFulton: @3iD0loN can you comment on this?  I don't have the answers.

It might make sense to rename the variable just so that it's clearer what it means though since, you're right, even if you do have 1.1.0, if you're on XBOX, for instance, it won't work.

3iD0loN: Yes, you are correct in that assumption.

UGUI has support for multiple displays right now, aside from touch the controls that we have added. As for UI Toolkit, I have done some preliminary tests and UI Toolkit documents are drawn on all screens regardless of settings, so there is no proper support for that in UI Toolkit.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
3iD0loN: Yes, on the managed side the displayIndex is a byte and when it came to blitting from unmanaged to managed code, I did make an attempt to replace with a ByteControl instead of IntegerControl, but either one did not exist, or it broke the blit, I do not remember which

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
3iD0loN: Both Touchscreen and Mouse are Pointers in the hierarchy, so I wanted to present a way to access displayIndex through them. However, the Mouse and Touchscreen objects do not exactly align on how they internally store the displayIndex. Mouse contains many of its input controls directly, but since Touchscreen represents multiple pointers (multi-touch), it itself does not directly contain many of the input controls. Many of its controls are instead deferred to its `primaryTouch` member.

Packages/com.unity.inputsystem/CHANGELOG.md
3iD0loN: Actually, that's an interesting question, and interesting premise, I did not test if this would work on a mirrored display. Since there is no desktop implementation, I did not try, and for AndroidPlayer, there is a hard limit to 2 displays, so the only display that could be mirrored is Display 0. The only tests that could be conducted are for Editor, and those are sub-regions in the editor

Assets/Tests/InputSystem/Plugins/UITests.cs
ekcoh: Sorry if comment was fuzzy, but what I was after was covering previous and new behavior if feasible. Using reflection to check for a non-existent property is not a good approach so please do not do anything like that, which it also sounds like you already ruled out as not a great idea.

What I instead was after that I could have elaborated on in a better way was basically that I wanted to encourage to look into if there is indirect behavior to be observed given the provided stimuli. I.e. for the new behavior when source input event is associated with e.g. display 2, would the observable behavior be that it would still be ending up being emitted as an event for display 1. I guess you are saying that it doesn't make sense since there was no displayIndex that could be set for the touch prior to this update or possiblity to create a touch originating from something else that primary display (implicity). However, if that is the case wouldn't it be possible to have the display set to primary implicity and then instead associate the camera with display 2 and test both new and old behavior. In this case I assume new behavior should take this into account while for previous behavior it would incorrectly end up affecting the raycast of that camera?

In addition it would be great if possible to make the assert stronger here by also verifying the actual coordinate passed to scene.parentReceiver since it could in theory be off as long as within the scene here.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
ekcoh: Unfortunately no, so ignore my comment. I guess my C/C++ thinking took over and made me believe that this was trivial for a second. Checked through the docs for our Unity symbol support and it seems rather tightly designed for version expressions, e.g. https://docs.unity3d.com/Manual/PlatformDependentCompilation.html and sub page. As you say its not supported to have "pre-baked expression" from what I can see so ignore this comment. I guess we have to have the long expression everywhere unless used multiple times in a single file, then it might be worthwhile defining it at the top of the file and use an aliased symbol to reduce chance of mistakes and redundance within that file

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
ekcoh: OK, thanks, hence why we need the
`#if UNITY_INPUT_SYSTEM_ENABLE_MULTIDISPLAY && (UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)
`
as a combined expression per platform which relies on capabilities provided by UGUI package. 

I still struggle a bit with this and wonder if this is fragile since it would require every package utilizing or touching multi-display to provide similar #if checks which is, according to above, platform specific and there is a risk of symbol defines quickly getting outdated with what is actually provided supported by the platform. E.g. what if UNITY_IOS support is added to ugui package and/or unity engine in a couple of months? This would require Input package (and likely other packages intending to not be platform dependent) to be updated since the functionality isn't controlled by underlying dependency, i.e. symbol not being defined by imported assembly/code, hence I wonder if a Unity version define as part of manifest symbol defines is also needed depending on proper 2023 version or is this backported to all supported versions?. 

I just want to make sure we avoid diverging packages if possible. Ideally a multi-display capability define would have been exposed by a Unity assembly as I think of it, if this is needed. I also struggle to understand impact of Unity version on this or is this functionality backported against all supported Unity versions since the only thing checked at the moment is UGUI package version.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: OK, thanks for clarifying

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
ekcoh: Looking at the code I cannot see why IntegerControl displayIndex in Pointer cannot be made non-virtual as other controls. In addition to this e.g. map the displayIndex IntegerControl to displayIndex field in FinishSetup() for Mouse and map to underlying primaryTouch displayIndex for TouchScreen? (See examples of existing controls in FinishSetup). Did you try this and it didn't work? Since primaryTouch control always exist in TouchScreen I believe this should be feasible?

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
ekcoh: @jamesmcgill do you have any additional thoughts on this one?

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
ekcoh: @jamesmcgill do you have any additional thoughts on this one?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jamesmcgill: Sorry for back to my old question:
So then I should read the PR description about 'There is also no implementation that properly passes this index over for any of the Editor or Android build targets' as the behaviour _before_ this PR lands?

That implies that 'Currently, this property only being set in certain Player build targets that support multidisplay' is currently available today, before this PR lands.

If so, why is there a special path here for editors/android? And not just using whatever code path the windows player is using? It's probably a stupid question, but I think I'm missing something :)

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
jamesmcgill: I'm also a little uneasy about adding platform defines around the code.
I'm still not 100% on their use so I don't have any ideas on how to remove/reduce them just now, but I would love it if we can look into ways to do that.

I'd hope that these defines are a temporary stop-gap?  And could be removed once we can rely on the required UGUI version.
But with the platform defines, I'm not clear on when we could sunset these. It feels like something should be pushed as standard into the platform and that the package should be using editor version defines until versions catch up. I'm not sure our package should be dealing with likely long-term platform inconsistencies.

Assets/Tests/InputSystem/Plugins/UITests.cs
EricFulton: Good call on adding the parentReceiver.event.position check.  I just added that to the PR.

I think I understand what you're saying. I'll add a test that clicks on display 2 in version 2019 and make sure the click works.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
EricFulton: I'm not sure I understand that last part about "special code path".  I don't think there is a special code path for windows player unless I'm gravely mistaken. Can @3iD0loN or @ekcoh correct me if I'm wrong.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
EricFulton: So we ended up only able to make this change to UGUI trunk so Unity 2023 onward will have the displayIndex set.  The reason we need these defines is that InputSystem supports all the way back to 2019 where we can't backport our UGUI changes.

Once InputSystem only supports 2023 and up, you could sunset these #defines.  Otherwise, I don't really see a way around them.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
EricFulton: I'm having a little trouble following the explanation for why the virtual is need, but since I don't have a lot of exposure to either ecosystem, I'm going to have to rely on @ekcoh and @3iD0loN to steer me the right way here.

If you want the virtual gone, I'm happy to do it.  If it needs to stay, I'm happy to do that too.  @jamesmcgill, maybe you can break the tie here.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
jamesmcgill: Yeah, I would agree with @ekcoh . If it's possible to make it non-virtual using `FinishSetup`, then we should try with that first to keep it consistent.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jamesmcgill: Yeah, sorry I'm not clear here. I think I'm missing a bigger picture or not grasping something, hence why the probably stupid questions :)

I'm trying to understand if this is working in Windows Standalone Player already? (The PR description suggests it is).
And if so, why this code here is for Editors and Android only?
Couldn't at least Android work similarly to how the Windows Standalone Player works today? (whatever that is, I'm not sure).


Ideally editors would not have a special ifdef either, but I understand it is sometimes not possible and we do have some existing `#if UNITY_EDITOR' paths already. 
Having special code paths for handling android platform (or editors) differently is something I'd like to avoid or minimize as much as possible for future maintenance reasons. So I'd just like to understand more why it was needed like that.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
jamesmcgill: Would we be able to remove the full define after 2023?
Also the `(UNITY_EDITOR_WIN || UNITY_EDITOR_OSX || UNITY_EDITOR_LINUX || UNITY_ANDROID)` part?

Not sure about this suggestion: but would a UNITY_VERSION >= 2023 define work here instead then?

EricFulton: Yep, if InputSystem only supported 2023 onwards, we could remove the define. Also, if we backported this functionality we could remove the define and we'd like to do that if we can get permission.

This new feature only works in Editor and in Android.  I have to assume we'd like to see the functionality added to the other platforms at some point too, but not this time around.  Once the change is in the other platforms, we can get rid of the parenthetical condition you highlighted.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
EricFulton: Sure thing, I'll remove the virtual.

EricFulton: I was just about to remove virtual, but then I realized it's used in Touchscreen.cs:
```
        public override IntegerControl displayIndex
        {
            get
            {
                return primaryTouch.displayIndex;
            }

            protected set
            {
                primaryTouch.displayIndex = value;
            }
        }
```

I'm not sure how we could replicate that behavior in FinishSetup.  I might reach out on slack to better understand.

benoit-alain (approved):
I've nothing against the part of these changes that I understand, but I'm not enough of an expert to understand all of it. My main concern is with the trunk side of these changes, where I'm under the impression that PanelRaycaster will need to use this new `displayIndex` property as well, as it's UIToolkit way of interoperating with UGUI by piggybacking on UGUI's events. I'm not insisting that it be done right now, as this is already a step in the right direction and we can always come back for the rest when we feel it becomes important to address it.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
JoelBarba-Unity: There is no implementation for Players except AndroidPlayer. All Editors have this implemented in a way that the GameView when set to a display target can simulate this and populate the value.

Before the previous set of changes, the displayIndex value did not change on the managed side. That is because it did not change in the unmanaged side. So displayIndex was always 0 for all platforms. However, since we are now implementing it for AndroidPlayer, WinEditor, LinuxEditor, and MacEditor, as far as I'm aware we are now stuck with a discrepancy between platforms. Now there is an expectation on AndroidPlayer that display index will be 1 when second screen is touched, but for WindowsPlayer it will remain 0 regardless of multi-display being enabled. It is for this reason that asserting that displayIndex == 1 will fail on platforms that support multidisplay but do not implement the displayIndex population.

And so, the description of this PR is a bit incorrect and should be reworded to avoid further confusion.

As to why they are different values for different platforms, it because of how displayIndex is populated on the unmanaged side of code. The AndroidPlayer does this on the part where Android JNI Bridge interfaces with InputSystem, and for Editors, it retrieves the value from the GameView on the Editor's managed side and passes it towards the many platform-dependent structures. WindowsPlayer and others that are unimplemented would handle population of displayIndex by listening to changes to own native APIs. 
For deeper context down the avenue of platform dependent details, here are the changes that implement displayIndex population for Editors: https://ono.unity3d.com/unity/unity/pull-request/142411/_/ikin-ryz-input-system-display-index-binding-trunk
And for AndroidPlayer:
https://ono.unity3d.com/unity/unity/pull-request/134020/_/ikin-ryz-android-presentation-input-changes-trunk

Feel free to ask anything else as well.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
jamesmcgill: I just pushed a commit the removes the virtual keyword.
In doing so, it uncovered that Pointer.cs was missing a layout entry (because GetChildControl was not used before), so good catch @ekcoh .

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
```diff
@@ -180,6 +183,12 @@ public class Pointer : InputDevice, IInputStateCallbackReceiver
         /// </remarks>
         public ButtonControl press { get; protected set; }
 
+        /// <summary>
+        /// The index of the display that was pointed.
```
ekcoh: @jamesmcgill @EricFulton I wonder if we should call out here something like

"The index of the display associated with with this input event if supported by the platform, or the index of the primary display (zero) if not supported by the target platform. For more information see, <link to Unity or UGUI documentation or manual covering multi-display support for that platform>."

In order to avoid future bug reports on all platforms except Android until this is spread more widely.

ekcoh (approved):
This PR makes more sense now that conditional defines are targeting UNITY_2023_1_OR_NEWER. Hence, I do not think it would be wise to land this before trunk changes have been released or alternatively change those symbol defines to what was initially there in this PR e.g. UNITY_INPUT_SYSTEM_UGUI_DEPENDENCY_HAS_DISPLAY_INDEX or similar depending on UGUI version.
The displayIndex is unconditionally defined in UGUI (this is good), but will not be backed by all platforms. Hence it makes sense that input system do not have platform specific defines either.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
jamesmcgill: I've updated this comment to something similar to your suggestion.

JoelBarba-Unity (approved):
Lookin good

JoelBarba-Unity (approved):

3iD0loN (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
ekcoh: Thanks for elaborating on this. We decided to update this code to not have platform specific handling from Input System perspective, instead the displayIndex will be set to correct value when platform is supporting it and remain 0 when not supported. This drives a need to be very clear on what platforms are currently supporting this in user-communication / docs to not get a lot of bug reports which would basically be feature requests on platform capabilities.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
ekcoh: I reached out to UGUI team to conclude whether downgrading UGUI package is supported or not. If its not we can stick to current defines without platform specifics. @EricFulton based on the modifications we made we would need to update Input System if changes are made so this feature is supported in older Unity versions, so please include us if such backports are made.

We can still get rid of the platform specifics for Input System (we have added these changes) since from Input System perspective the displayIndex will be zero if platform doesn't provide any value, e.g. UGUI have this field and its not platform specific.

From our perspective feel free to resolve this conversation.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
ekcoh: @EricFulton Feel free to resolve this conversation if you are happy with outcome. We think its good now.

Packages/com.unity.inputsystem/InputSystem/Devices/Pointer.cs
ekcoh: Excellent, marking this resolved.

ekcoh:
@EricFulton @3iD0loN : @jamesmcgill and I believe we have addressed all review questions and done necessary edits to this point now if you feel satisfied with the outcome? However we would like to hold back on merging this until functionality is available in 23_1 so we can run CI on it and so that it makes sense to any user using this functionality, ok?

ekcoh:
Also noted CHANGELOG.md has a conflict that needs to be resolved.

jamesmcgill (changes requested):
Hi @EricFulton, could you remove the platform specific defines to make it consistent with the other changes without  it breaking? 
i.e. make it just `#if UNITY_2023_1_OR_NEWER`

jamesmcgill (approved):

jamesmcgill:
Hi @EricFulton could you see if it still works with the changes I just pushed?
I wanted to move the state modification out of the get index function as that seemed like a hidden side-effect in a getter.

ekcoh:
Changes reflected for CI verification in https://github.com/Unity-Technologies/InputSystem/pull/1623 through Git patching

jamesmcgill:
Closing this PR as this work has migrated to https://github.com/Unity-Technologies/InputSystem/pull/1623

PR #1604, opened 2022-11-01
Not setting control scheme on composite bindings during pasting

jimon:
### Description

Merging of #1024

Composite bindings by design shouldn't be assigned to control schemes, e.g. "groups" should be left empty.
But when pasting from clipboard we incorrectly assign an active control scheme as a group on composite bindings.
This doesn't produce any significant side effects.

### Changes made

Skipping composite bindings when assigning groups.
Also added tests as apparently this functionality was not tested at all.


jamesmcgill (approved):
lgtm

PR #1606, opened 2022-11-08
Hot path optimization for ReadValue/WriteValueIntoState for most used controls

jimon:
### Description

Current performance of `ReadValue` leaves a lot to be desired.

This PR is based on observation that for majority of controls:
- Memory layout is typical for the type, e.g. `AxisControl` is byte aligned float, `QuaternionControl` is byte aligned `Quaternion`, etc.
- No extra processing is applied.

### Changes made

- If memory layout is expected typical layout for the control type, we short circuit `ReadValue` to cast state pointer directly to expected type, bypassing `InputStateBlock` reading all together for selected controls. Same for `WriteValueIntoState`.
- Added control debugger visibility to check if control got hot path optimized or not.
- Require `ApplyParameterChanges` for application of control changes that change optimization status, like `AxisControl` processors parameters (e.g. `invert`, etc).

### Notes

We plan to land this feature as opt-in for now. And make it opt-out after some rigorous testing.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Controls.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Controls.cs
@@ -1378,4 +1378,45 @@ partial class CoreTests
         Assert.That(UnsafeUtility.SizeOf<TouchState>(), Is.EqualTo(TouchState.kSizeInBytes));
         Assert.That(touchscreen.touches[0].stateBlock.alignedSizeInBytes, Is.EqualTo(TouchState.kSizeInBytes));
     }
+
+    [Test]
+    [Category("Controls")]
+    public void Controls_OptimizedControls_TrivialControlsAreOptimized()
+    {
+        var mouse = InputSystem.AddDevice<Mouse>();
+
+        InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseOptimizedControls, false);
+        Assert.That(mouse.position.x.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.position.y.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.position.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.leftButton.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+
+        InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseOptimizedControls, true);
+        Assert.That(mouse.position.x.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatFloat));
+        Assert.That(mouse.position.y.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatFloat));
+        Assert.That(mouse.position.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatVector2));
+        Assert.That(mouse.leftButton.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+
+        InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseOptimizedControls, false);
+        Assert.That(mouse.position.x.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.position.y.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.position.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.leftButton.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+    }
+
+    [Test]
+    [Category("Controls")]
+    public void Controls_OptimizedControls_ParentChangesOptimization_IfChildIsNoLongerOptimized()
+    {
+        InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseOptimizedControls, true);
+
+        var mouse = InputSystem.AddDevice<Mouse>();
+
+        mouse.position.x.invert = true;
+        mouse.position.x.ApplyParameterChanges();
+
+        Assert.That(mouse.position.x.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.position.y.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatFloat));
+        Assert.That(mouse.position.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+    }
 }
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -13,6 +13,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Added
 - Added support for reading Tracking State in [TrackedPoseDriver](xref:UnityEngine.InputSystem.XR.TrackedPoseDriver) to constrain whether the input pose is applied to the Transform. This should be used when the device supports valid flags for the position and rotation values, which is the case for XR poses.
 - Added `InputSettings.shortcutKeysConsumeInput`. This allows programmatic access to opt-in to the enhanced shortcut key behaviour ([case ISXB-254](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-254))).
+- Significantly optimized cost of `ReadValue`/`ReadUnprocessedValueFromState`/`WriteValueIntoState` for some control types. Optimization is opt-in for now, please call `InputSystem.settings.SetInternalFeatureFlag("USE_OPTIMIZED_CONTROLS", true);` in your project to enable it. You can observe which controls are optimized by looking at new optimized column in device debugger. You will need to call a new `InputControl.ApplyParameterChanges()` method if the code is changing `AxisControl` fields after initial setup is done.
 
 ### Fixed
 - Fixed composite bindings incorrectly getting a control scheme assigned when pasting into input asset editor with a control scheme selected.
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
@@ -5,6 +5,8 @@ using UnityEngine.InputSystem.Utilities;
 
 ////REVIEW: change 'clampToConstant' to simply 'clampToMin'?
 
+////TODO: if AxisControl fields where properties, we wouldn't need ApplyParameterChanges, maybe it's ok breaking change? 
+
 namespace UnityEngine.InputSystem.Controls
 {
     /// <summary>
@@ -237,8 +239,6 @@ namespace UnityEngine.InputSystem.Controls
         /// <inheritdoc />
         public override unsafe float ReadUnprocessedValueFromState(void* statePtr)
         {
-            EnsureOptimizationTypeHasNotChanged();
-
             switch (m_OptimizedControlDataType)
             {
                 case InputStateBlock.kFormatFloat:
@@ -257,8 +257,6 @@ namespace UnityEngine.InputSystem.Controls
         /// <inheritdoc />
         public override unsafe void WriteValueIntoState(float value, void* statePtr)
         {
-            EnsureOptimizationTypeHasNotChanged();
-
             switch (m_OptimizedControlDataType)
             {
                 case InputStateBlock.kFormatFloat:
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
@@ -1,5 +1,7 @@
+using System.Runtime.CompilerServices;
 using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.InputSystem.Processors;
+using UnityEngine.InputSystem.Utilities;
 
 ////REVIEW: change 'clampToConstant' to simply 'clampToMin'?
 
@@ -177,6 +179,7 @@ namespace UnityEngine.InputSystem.Controls
         /// <seealso cref="normalize"/>
         /// <seealso cref="scale"/>
         /// <seealso cref="invert"/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         protected float Preprocess(float value)
         {
             if (scale)
@@ -234,16 +237,41 @@ namespace UnityEngine.InputSystem.Controls
         /// <inheritdoc />
         public override unsafe float ReadUnprocessedValueFromState(void* statePtr)
         {
-            var value = stateBlock.ReadFloat(statePtr);
-            ////REVIEW: this isn't very raw
-            return Preprocess(value);
+            EnsureOptimizationTypeHasNotChanged();
+
+            switch (m_OptimizedControlDataType)
+            {
+                case InputStateBlock.kFormatFloat:
+                    return *(float*)((byte*)statePtr + m_StateBlock.m_ByteOffset);
+                case InputStateBlock.kFormatByte:
+                    return *((byte*)statePtr + m_StateBlock.m_ByteOffset) != 0 ? 1.0f : 0.0f;
+                default:
+                {
+                    var value = stateBlock.ReadFloat(statePtr);
+                    ////REVIEW: this isn't very raw
+                    return Preprocess(value);
+                }
+            }
         }
 
         /// <inheritdoc />
         public override unsafe void WriteValueIntoState(float value, void* statePtr)
         {
-            value = Unpreprocess(value);
-            stateBlock.WriteFloat(statePtr, value);
+            EnsureOptimizationTypeHasNotChanged();
+
+            switch (m_OptimizedControlDataType)
+            {
+                case InputStateBlock.kFormatFloat:
+                    *(float*)((byte*)statePtr + m_StateBlock.m_ByteOffset) = value;
+                    break;
+                case InputStateBlock.kFormatByte:
+                    *((byte*)statePtr + m_StateBlock.m_ByteOffset) = (byte)(value >= 0.5f ? 1 : 0);
+                    break;
+                default:
+                    value = Unpreprocess(value);
+                    stateBlock.WriteFloat(statePtr, value);
+                    break;
+            }
         }
 
         /// <inheritdoc />
@@ -276,6 +304,29 @@ namespace UnityEngine.InputSystem.Controls
             }
 
             return NormalizeProcessor.Normalize(value, min, max, 0);
+        }
+
+        protected override FourCC CalculateOptimizedControlDataType()
+        {
+            var noProcessingNeeded =
+                clamp == Clamp.None &&
+                invert == false &&
+                normalize == false &&
+                scale == false;
+
+            if (noProcessingNeeded &&
+                m_StateBlock.format == InputStateBlock.FormatFloat &&
+                m_StateBlock.sizeInBits == 32 &&
+                m_StateBlock.bitOffset == 0)
+                return InputStateBlock.FormatFloat;
+            if (noProcessingNeeded &&
+                m_StateBlock.format == InputStateBlock.FormatBit &&
+                // has to be 8, otherwise we might be mapping to a state which only represents first bit in the byte, while other bits might map to some other controls
+                // like in the mouse where LMB/RMB map to the same byte, just LMB maps to first bit and RMB maps to second bit
+                m_StateBlock.sizeInBits == 8 &&
+                m_StateBlock.bitOffset == 0)
+                return InputStateBlock.FormatByte;
+            return InputStateBlock.FormatInvalid;
         }
     }
 }
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
@@ -5,7 +5,7 @@ using UnityEngine.InputSystem.Utilities;
 
 ////REVIEW: change 'clampToConstant' to simply 'clampToMin'?
 
-////TODO: if AxisControl fields where properties, we wouldn't need ApplyParameterChanges, maybe it's ok breaking change? 
+////TODO: if AxisControl fields where properties, we wouldn't need ApplyParameterChanges, maybe it's ok breaking change?
 
 namespace UnityEngine.InputSystem.Controls
 {
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
@@ -1,3 +1,4 @@
+using System.Runtime.CompilerServices;
 using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.Scripting;
 
@@ -75,6 +76,7 @@ namespace UnityEngine.InputSystem.Controls
         /// <returns>True if <paramref name="value"/> crosses the threshold to be considered pressed.</returns>
         /// <seealso cref="pressPoint"/>
         /// <seealso cref="InputSettings.defaultButtonPressPoint"/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public new bool IsValueConsideredPressed(float value)
         {
             return value >= pressPointOrDefault;
--- a/Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
@@ -1,7 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Runtime.CompilerServices;  
+using System.Runtime.CompilerServices;
 using UnityEngine.InputSystem.Controls;
 using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.InputSystem.Utilities;
@@ -869,7 +869,7 @@ namespace UnityEngine.InputSystem
         /// you can cast the Vector3Control state memory directly to Vector3 without calling ReadUnprocessedValueFromState on x/y/z axes.
         ///
         /// The value returned for any given control is computed automatically by the Input System, when the control's setup configuration changes. <see cref="InputControl.CalculateOptimizedControlDataType"/>
-        /// There are some parameter changes which don't trigger a configuration change (such as the clamp, invert, normalize, and scale parameters on AxisControl), 
+        /// There are some parameter changes which don't trigger a configuration change (such as the clamp, invert, normalize, and scale parameters on AxisControl),
         /// so if you modify these, the optimized data type is not automatically updated. In this situation, you should manually update it by calling <see cref="InputControl.ApplyParameterChanges"/>.
         /// </remarks>
         public FourCC optimizedControlDataType => m_OptimizedControlDataType;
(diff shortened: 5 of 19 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
```diff
@@ -190,8 +192,18 @@ public class PoseControl : InputControl<PoseState>
         /// </remarks>
         public PoseControl()
         {
-            m_StateBlock.format = new FourCC('P', 'o', 's', 'e');
+            m_StateBlock.format = PoseState.s_Format;
         }
+        
+        private void TryEnableFastPath()
+        {
+            m_FastPathReading = m_StateBlock.format == PoseState.s_Format;
```
jimon: Ok, after some thinking I think this is not sufficient, we also need to go through every subcontrol and check if there are no subprocessors on axis controls, etc.

jimon: Same applies for other composite types like vector3/quaternion/etc

jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
```diff
@@ -75,6 +76,7 @@ public ButtonControl()
         /// <returns>True if <paramref name="value"/> crosses the threshold to be considered pressed.</returns>
         /// <seealso cref="pressPoint"/>
         /// <seealso cref="InputSettings.defaultButtonPressPoint"/>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
andrew-oc, line 79: I'm a bit sceptical that all the aggressive inlining is going to actually speed anything up. Probably the opposite.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -853,6 +854,71 @@ protected internal uint stateOffsetRelativeToDeviceRoot
...
+
+            CalculateOptimizedControlDataType();
+        }
+
+        // This function exists to warn users to start using ApplyParameterChanges for edge cases that were previously not intentionally worked,
+        // where control properties suddenly change underneath us without us anticipating that.
+        // This is mainly to AxisControl fields being public and capable of changing at any time even if we don't thing it's an anticipated usage pattern.
+        // Also it's not clear if InputControl.stateBlock.format can potentially change at any time, likely not.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal void EnsureOptimizationTypeDoesntSuddenlyChange()
```
andrew-oc: Feel a more imperative-style name works better for this method. EnsureOptimizationTypeHasNotChanged

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -853,6 +854,71 @@ protected internal uint stateOffsetRelativeToDeviceRoot
...
+                    inputControl.CalculateOptimizedControlDataTypeRecursively();
+
+            CalculateOptimizedControlDataType();
+        }
+
+        // This function exists to warn users to start using ApplyParameterChanges for edge cases that were previously not intentionally worked,
+        // where control properties suddenly change underneath us without us anticipating that.
+        // This is mainly to AxisControl fields being public and capable of changing at any time even if we don't thing it's an anticipated usage pattern.
+        // Also it's not clear if InputControl.stateBlock.format can potentially change at any time, likely not.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
andrew-oc, line 936: Maybe mark this as [Conditional("DEBUG")] instead of using #IF DEBUG, so there's zero cost at runtime. Just a method call but still.

Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
```diff
@@ -277,5 +304,28 @@ public override unsafe float EvaluateMagnitude(void* statePtr)
...
+                m_StateBlock.bitOffset == 0)
+                return OptimizedControlDataType.Float;
+            if (noProcessingNeeded &&
+                     m_StateBlock.format == InputStateBlock.FormatBit &&
+                     // has to be 8, otherwise we might be mapping to a state which only represents first bit in the byte, while other bits might map to some other controls
+                     // like in the mouse where LMB/RMB map to the same byte, just LMB maps to first bit and RMB maps to second bit
+                     m_StateBlock.sizeInBits == 8 &&
+                     m_StateBlock.bitOffset == 0)
+                return OptimizedControlDataType.BoolAsByte;
+            return OptimizedControlDataType.NotOptimized;
```
andrew-oc: I think we might also need to check if the control uses state from another control (usesStateFromOtherControl) and make sure the other control also has non-default processing if so, potentially following a chain of state from other controls. This is getting messy

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
```diff
@@ -209,26 +211,66 @@ protected override void FinishSetup()
...
+                    velocity.WriteValueIntoState(value.velocity, statePtr);
+                    angularVelocity.WriteValueIntoState(value.angularVelocity, statePtr);
+                    break;
+            }
+        }
+        
+        public OptimizedControlDataType CalculateOptimizedControlDataType()
+        {
+            if (
+                m_StateBlock.sizeInBits == 60 * 8 &&
```
andrew-oc: Should be PoseState.kSizeInBytes instead of 60?

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
```diff
@@ -0,0 +1,34 @@
+﻿namespace UnityEngine.InputSystem.Controls
+{
+    public enum OptimizedControlDataType
```
andrew-oc: Not sure about this design. Users wouldn't be able to optimize their own controls with this. Can it not be an actual Type type?

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
```diff
@@ -0,0 +1,34 @@
...
+        Quaternion,
+
+        // PoseState
+        Pose
+    }
+    
+    internal interface IOptimizedControl
+    {
+        public OptimizedControlDataType CalculateOptimizedControlDataType();
+    }
```
andrew-oc: Could we get away with an isOptimized property on the base InputControl class instead of the interface?

andrew-oc (commented):
Really cool direction. Have some concerns about the implementation and I'm also strongly opposed to ever making it an opt-out behavior in any 1.x version, but I like it so far.

Packages/com.unity.inputsystem/InputSystem/Controls/AxisControl.cs
jimon: I though if we use state from other control we just do that .. use the state memory. As I don't understand how or where would processor stack from another control run.

Is there an example where one control uses state from another control and when reading first control we run processor stack from second control?

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
jimon: I want to easily switch case on the value, so enum would be better/faster.
It doesn't have to be public, we can make it internal. I just had struggle with internal interface public method, C# somehow wants this enum type being public, but we can wiggle around it I think.

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
jimon: No, we need to cache the value somehow, because for example for `Vector3Control` to decide that it can be optimized it needs to know if x/y/z controls are optimized and are of float type

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
jimon: Why? These checks are in the hot path, doing it inline _should_ be faster. But we don't have a good indicative test suit sigh

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
jimon: Not sure about users optimizing their own controls, this feature would be internal for a time being, if we get nudged we might consider opening it.

In general I'm not convinced that user defined types are a good idea, that greatly limits what we can do with regards to editor UX and data structures. If all controls types are known at our compile time we would be able to do much more effective architecture.

Users defining their own devices is different and should stay as-is.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
jimon: Will fix but it doesn't matter that much, as you can see down in the check we also check layout of controls, so if anything about `PoseState` changes we also need to change that code as well.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
jimon: tried with Type and the checks are much more expensive it seems https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmbjgKYB2ArlNgCqUTACidUp5Ja5/2AILYAvNgAMAGm4DsAIWl9ZBGdgC+AblWqADgCcAlgDcwwSkxbtO2M6w71RFu5wB0grUrL7jp84wCeOua2wAFBjsCBlAD2AGYAFAbUwACUHvyqWNgARtHRADbY9lD4ABaUAMYA1vEpxKr8cADsNpbFomLMzvQu+OnkauiZOLkFTFFllTV1vMq4LSFh5iJikUFx8bH50aZpqoNoakA===

jimon: After some discussion, we can switch to int (FourCC) the same way as `InputStateBlock` works

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
```diff
@@ -0,0 +1,34 @@
...
+        // Quaternion as 4 consecutive 32 bit floats, in order of x,y,z,w, byte aligned
+        Quaternion,
+
+        // PoseState
+        Pose
+    }
+    
+    internal interface IOptimizedControl
+    {
+        public OptimizedControlDataType CalculateOptimizedControlDataType();
```
jimon: Ok that's not gonna fly:
- Let's say we have a custom control that derives from `AxisControl` and does custom processing in `ReadUnprocessedValueFromState`
- We make `Vector3Control` of such controls
- Now `Vector3Control` has no way to check if the nested controls are optimized or not, because interface is already implemented on `AxisControl` level and can't be implemented again in derived controls.

Suggestion: use virtual method on `InputControl` instead

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/IOptimizedControl.cs
jimon: N/A, interface removed

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -853,6 +854,98 @@ protected internal uint stateOffsetRelativeToDeviceRoot
...
+
+        /// <summary>
+        /// Apply built-in parameters changes (e.g. <see cref="AxisControl.invert"/>, others) and recompute <see cref="InputControl.optimizedControlDataType"/> for impacted controls.
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
+        public void ApplyParameterChanges()
+        {
+            // First we go through all children of our own hierarchy
+            SetOptimizedControlDataTypeRecursively();
```
jimon, line 895: Or just computing status for current control is sufficient?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -853,6 +854,98 @@ protected internal uint stateOffsetRelativeToDeviceRoot
...
+        }
+
+        // This function exists to warn users to start using ApplyParameterChanges for edge cases that were previously not intentionally supported,
+        // where control properties suddenly change underneath us without us anticipating that.
+        // This is mainly to AxisControl fields being public and capable of changing at any time even if we were not anticipated such a usage pattern.
+        // Also it's not clear if InputControl.stateBlock.format can potentially change at any time, likely not.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        // Only do this check in debug in hope that it will be sufficient to catch any misuse during development.
+        [Conditional("DEBUG")]
+        internal void EnsureOptimizationTypeHasNotChanged()
```
jamesmcgill, line 939: Should this function go a bit further and ensure that `optimizeddControlDataType` is correct before continuing?

Assets/Tests/InputSystem/CoreTests_Controls.cs
```diff
@@ -1378,4 +1378,45 @@ public void Controls_TouchControlStateCorrespondsToTouchState()
...
+
+        var mouse = InputSystem.AddDevice<Mouse>();
+
+        mouse.position.x.invert = true;
+        mouse.position.x.ApplyParameterChanges();
+
+        Assert.That(mouse.position.x.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+        Assert.That(mouse.position.y.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatFloat));
+        Assert.That(mouse.position.optimizedControlDataType, Is.EqualTo(InputStateBlock.FormatInvalid));
+    }
```
jamesmcgill, line 1421: Maybe a quick test that the values coming out are correct (especially for quaternion would be nice), or hijack some existing tests by making the feature flag state conditional on TestCase?

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
jamesmcgill: Inclined to agree with Andrew, if we don't have good evidence otherwise, I would leave the compiler unconstrained to figure it out by itself.

jamesmcgill (approved):
Nice optimisation!

Assets/Tests/InputSystem/CoreTests_Controls.cs
ekcoh: +1 to what @jamesmcgill  said. Definately sounds best to parameterize existing tests since they should be invariant to optimization setting if functional tests right?

ekcoh: In addition related to testing, would it be possible to cover this performance targeted feature with performance tests, as in "Automation/Performance Testing" in internaldocs and benefit from that? This relates to discussion of having monitoring of performance KPIs and I understand if out of scope, but interested in understand if this would be feasable. It would give reporting and monitoring of any changes in this and a diff without doing manual profiling?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ however, it has to be formatted properly to pass verification tests.
 ### Actions
 - Extended input action code generator (`InputActionCodeGenerator.cs`) to support optional registration and unregistration of callbacks for multiple callback instances via `AddCallbacks(...)` and `RemoveCallbacks(...)` part of the generated code. Contribution by [Ramobo](https://github.com/Ramobo) in [#889](https://github.com/Unity-Technologies/InputSystem/pull/889).
 
+### Added
+
+- Significantly optimized cost of `ReadValue`/`ReadUnprocessedValueFromState`/`WriteValueIntoState` for some control types. Optimization is opt-in for now, please call `InputSystem.settings.SetInternalFeatureFlag("USE_OPTIMIZED_CONTROLS", true);` in your project to enable it. You can observe which controls are optimized by looking at new optimized column in device debugger. You might need to call a new `InputControl.ApplyParameterChanges()` method if the code is changing `AxisControl` fields after initial setup is done.
```
ekcoh: Regarding "You might need to call....", I assume this is required for the given condition and otherwise not? Maybe the correct approach is to only state that its required to be called for changes to have effect or what is the behavior if failing to do this?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -853,6 +854,98 @@ protected internal uint stateOffsetRelativeToDeviceRoot
...
+        {
+            return InputStateBlock.kFormatInvalid;
+        }
+
+        /// <summary>
+        /// Apply built-in parameters changes (e.g. <see cref="AxisControl.invert"/>, others) and recompute <see cref="InputControl.optimizedControlDataType"/> for impacted controls.
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
+        public void ApplyParameterChanges()
```
ekcoh, line 892: Looks like this is public API in public class right? Suggest to include @duckets on review when adding public API docs to get review feedback.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -853,6 +854,98 @@ protected internal uint stateOffsetRelativeToDeviceRoot
...
+            m_OptimizedControlDataType = InputSettings.optimizedControlsFeatureEnabled
+                ? CalculateOptimizedControlDataType()
+                : (FourCC)InputStateBlock.kFormatInvalid;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal void SetOptimizedControlDataTypeRecursively()
+        {
+            // Need to go depth-first because CalculateOptimizedControlDataType might depend on computed values of children
+            if (m_ChildCount > 0)
```
ekcoh, line 923: Would recommend adding braces to this if to avoid later mistakes on what is contained in it when doing multi-line statement

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -853,6 +854,98 @@ protected internal uint stateOffsetRelativeToDeviceRoot
...
+        // Also it's not clear if InputControl.stateBlock.format can potentially change at any time, likely not.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        // Only do this check in debug in hope that it will be sufficient to catch any misuse during development.
+        [Conditional("DEBUG")]
+        internal void EnsureOptimizationTypeHasNotChanged()
+        {
+            if (!InputSettings.optimizedControlsFeatureEnabled)
+                return;
+
+            if (CalculateOptimizedControlDataType() != optimizedControlDataType)
```
ekcoh: Good to see that error message contains info to track down the issue!

Packages/com.unity.inputsystem/InputSystem/Controls/QuaternionControl.cs
```diff
@@ -66,17 +67,61 @@ protected override void FinishSetup()
...
+            switch (m_OptimizedControlDataType)
+            {
+                case InputStateBlock.kFormatQuaternion:
+                    return *(Quaternion*)((byte*)statePtr + (int)m_StateBlock.byteOffset);
+                default:
+                    return new Quaternion(
+                        x.ReadValueFromState(statePtr),
+                        y.ReadValueFromState(statePtr),
+                        z.ReadValueFromState(statePtr),
+                        w.ReadUnprocessedValueFromState(statePtr));
```
ekcoh, line 79: If there was a conclusion why we have this odd patern of w not being processed, maybe add a comment explaining/motivating it if information was fresh?

Packages/com.unity.inputsystem/InputSystem/Controls/QuaternionControl.cs
```diff
@@ -66,17 +67,61 @@ protected override void FinishSetup()
...
+                    z.WriteValueIntoState(value.z, statePtr);
+                    w.WriteValueIntoState(value.w, statePtr);
+                    break;
+            }
+        }
+
+        protected override FourCC CalculateOptimizedControlDataType()
+        {
+            if (
+                m_StateBlock.sizeInBits == 128 &&
```
ekcoh: Why is this 128, get bit of magic number associations, would it be more clear with something like?
m_StateBlock.sizeInBits == sizeof(float) * 4 + sizeof(overhead....)
or similar?

Packages/com.unity.inputsystem/InputSystem/Controls/Vector3Control.cs
```diff
@@ -31,23 +32,57 @@ protected override void FinishSetup()
...
         public override unsafe float EvaluateMagnitude(void* statePtr)
         {
             ////REVIEW: this can go beyond 1; that okay?
             return ReadValueFromState(statePtr).magnitude;
         }
+
+        protected override FourCC CalculateOptimizedControlDataType()
+        {
+            if (
+                m_StateBlock.sizeInBits == 96 &&
```
ekcoh: Similar comment like on previous fixed size number

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputControlTreeView.cs
```diff
@@ -184,30 +189,18 @@ private ControlItem BuildControlTreeRecursive(InputControl control, int depth, r
...
-                offset = new GUIContent(offset.ToString()),
-                bit = new GUIContent(control.stateBlock.bitOffset.ToString()),
-                sizeInBits = new GUIContent(control.stateBlock.sizeInBits.ToString()),
-                type = new GUIContent(control.GetType().Name),
-                value = value,
-                values = values,
                 children = children
             };
 
+            ////TODO: come up with nice icons depicting different control types
```
ekcoh, line 199: Was this TODO added as part of this work? Some associated task that should be filed? Feels unrelated?

ekcoh (approved):
Nice work! Would be very beneficial to have performance metrics in place for this kind of work to see effect of introduced changes see comment in review.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputControlTreeView.cs
jimon: It's unrelated, just moved code around

Packages/com.unity.inputsystem/InputSystem/Controls/QuaternionControl.cs
jimon: Can fix, but dunno, most game devs know that vec4 is 128 bit, kinda a muscle memory I feel

Packages/com.unity.inputsystem/InputSystem/Controls/QuaternionControl.cs
jimon: We have no idea why is that, and we have no idea if anyone is relying on this. So I can't add a comment

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
jimon: I will keep this one because we have practical experience from dots mathematics where inlining helped, but will remove ones on ReadValue

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
jimon: Keep as is for now, probably safer that way as `ApplyParameterChanges` on root (device) will force-recompute all the states, so kinda an escape hatch of sorts

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
jimon: Now it sets correct value `optimizedControlDataType` after printing the error.
Fixed

Packages/com.unity.inputsystem/CHANGELOG.md
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/QuaternionControl.cs
jimon: Fixed

Packages/com.unity.inputsystem/InputSystem/Controls/Vector3Control.cs
jimon: Fixed

Assets/Tests/InputSystem/CoreTests_Controls.cs
jimon: We haven't found any good existing tests for this.
But I've added two performance tests.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
duckets: Is optimizedControlDataType and CalculateOptimizedControlDataType() intended to be used by end-users of the package? Or is this what "function shouldn't be called manually" refers to? (manually is confusing)

From the changelog it sounds like end users should only need to use SetInternalFeatureFlag and ApplyParameterChanges to actually use this feature, is this correct?

Also does this new feature have any corresponding documentation in the manual?

jimon: Depends, if a user only uses built-in control types (`Vector2Control`, `AxisControl`, etc) then they only need `SetInternalFeatureFlag` and `ApplyParameterChanges` to get started.

If user wants to enable this functionality for their own control types, they need to provide `CalculateOptimizedControlDataType` implementation.

duckets: I've updated the docs based on discussions with @jimon

duckets (approved):
Approved with docs amendments

PR #1609, opened 2022-11-21
FIX: Device disconnected events are missed when running in the background

andrew-oc:
### Description

Some platforms ignore the runInBackground flag and just always run. The Input System, however, respects that flag and when it is false, and that application doesn't have focus, it can early-out of the event processing loop and flush the event buffer. If this happens when a device disconnected event is in the buffer, that device won't be disconnected and the backend won't send another event, resulting in orphaned devices and exceptions when the an attempt is made to add the same device again.

### Changes made

Forced the runInBackground flag to true on PS5.

### Checklist

Before review:

- [x] Changelog entry added.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
@@ -227,7 +227,11 @@ namespace UnityEngine.InputSystem.LowLevel
         public double currentTimeOffsetToRealtimeSinceStartup => NativeInputSystem.currentTimeOffsetToRealtimeSinceStartup;
         public float unscaledGameTime => Time.unscaledTime;
 
-        public bool runInBackground => Application.runInBackground;
+        public bool runInBackground => Application.runInBackground ||
+        // certain platforms ignore the runInBackground flag and always run. Make sure we're
+        // not running on one of those.
+        // TODO: Add more platforms here as they're discovered.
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -227,7 +227,11 @@ public float pollingFrequency
         public double currentTimeOffsetToRealtimeSinceStartup => NativeInputSystem.currentTimeOffsetToRealtimeSinceStartup;
         public float unscaledGameTime => Time.unscaledTime;
 
-        public bool runInBackground => Application.runInBackground;
+        public bool runInBackground => Application.runInBackground ||
+        // certain platforms ignore the runInBackground flag and always run. Make sure we're
+        // not running on one of those.
+        // TODO: Add more platforms here as they're discovered.
+        Application.platform == RuntimePlatform.PS5;
```
ekcoh, line 234: I believe this change is ok to solve this problem, but if ignoring runInBackground is optional I think a more robust solution would have been to have a capability exposed by native stating if the platform respects/supports runInBackground settings and adapt to that.

ekcoh (approved):

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
andrew-oc: Agree, this is sort of a stop gap

PR #1611, opened 2022-11-22
Filtering out axis jitter on Switch Pro / DualShock 4 / DualSense gamepads connected via HID

jimon:
### Description

This is a continuation of #1585 with a much more robust way to compute if gamepads are significantly actuated or changed to mark them as current devices.

Most gamepads have noisy stick controls. And this makes them become `Gamepad.current` when user doesn't actuate any controls, which in result is problematic if two or more gamepads connected to the system.

### Changes made

- Added a terrible way to communicate back to InputManager if device _should not_ be set as current. This defies good programming practices, but unfortunately I see no other easy way to do that without breaking API (there are some hard ways to do it though).

### Notes

3 bits selected empirically myself observing live state from gamepads connected to my system. Value might not be optimal.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -3319,7 +3319,7 @@ namespace UnityEngine.InputSystem
 
         private bool m_ShouldMakeCurrentlyUpdatingDeviceCurrent;
 
-        // This is a dirty hot fix to expose entropy from device back to input manager to make choose if we want to make device current or not.
+        // This is a dirty hot fix to expose entropy from device back to input manager to make a choice if we want to make device current or not.
         // A proper fix would be to change IInputStateCallbackReceiver.OnStateEvent to return bool to make device current or not.
         internal void DontMakeCurrentlyUpdatingDeviceCurrent()
         {
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -3172,9 +3172,12 @@ namespace UnityEngine.InputSystem
                             var haveChangedStateOtherThanNoise = true;
                             if (deviceIsStateCallbackReceiver)
                             {
+                                m_ShouldMakeCurrentlyUpdatingDeviceCurrent = true;
                                 // NOTE: We leave it to the device to make sure the event has the right format. This allows the
                                 //       device to handle multiple different incoming formats.
                                 ((IInputStateCallbackReceiver)device).OnStateEvent(eventPtr);
+
+                                haveChangedStateOtherThanNoise = m_ShouldMakeCurrentlyUpdatingDeviceCurrent;
                             }
                             else
                             {
@@ -3312,6 +3315,15 @@ namespace UnityEngine.InputSystem
 
             DelegateHelpers.InvokeCallbacksSafe(ref m_AfterUpdateListeners,
                 "InputSystem.onAfterUpdate");
+        }
+
+        private bool m_ShouldMakeCurrentlyUpdatingDeviceCurrent;
+
+        // This is a dirty hot fix to expose entropy from device back to input manager to make choose if we want to make device current or not.
+        // A proper fix would be to change IInputStateCallbackReceiver.OnStateEvent to return bool to make device current or not.
+        internal void DontMakeCurrentlyUpdatingDeviceCurrent()
+        {
+            m_ShouldMakeCurrentlyUpdatingDeviceCurrent = false;
         }
 
         internal unsafe bool UpdateState(InputDevice device, InputEvent* eventPtr, InputUpdateType updateType)
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
@@ -557,8 +557,8 @@ namespace UnityEngine.InputSystem.DualShock
         }
 
         // filter out three lower bits as jitter noise
-        internal const byte AxisDeadZoneLow = 0b01111000;
-        internal const byte AxisDeadZoneHigh = 0b10000111;
+        internal const byte JitterMaskLow = 0b01111000;
+        internal const byte JitterMaskHigh = 0b10000111;
 
         public unsafe void OnStateEvent(InputEventPtr eventPtr)
         {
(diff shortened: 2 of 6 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3314,6 +3317,15 @@ private void InvokeAfterUpdateCallback(InputUpdateType updateType)
                 "InputSystem.onAfterUpdate");
         }
 
+        private bool m_ShouldMakeCurrentlyUpdatingDeviceCurrent;
+
+        // This is a dirty hot fix to expose entropy from device back to input manager to make choose if we want to make device current or not.
```
ekcoh: minor: typo "to make choose if we want to make..."?

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
```diff
@@ -516,6 +516,9 @@ unsafe bool IEventPreProcessor.PreProcessEvent(InputEventPtr eventPtr)
                 return eventPtr.type != DeltaStateEvent.Type; // only skip delta state events
 
             var stateEvent = StateEvent.FromUnchecked(eventPtr);
+            if (stateEvent->stateFormat == DualSenseHIDInputReport.Format)
+                return true; // if someone queued DSVS directly, just use as-is
```
ekcoh, line 520: Is "DSVS" an acronym that is well known? Is it one of the formats? I do not understand what it implies

ekcoh: Is it referring to: queued IOCTL command "DSVS"?

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
```diff
@@ -549,6 +552,50 @@ unsafe bool IEventPreProcessor.PreProcessEvent(InputEventPtr eventPtr)
                 return false; // skip unrecognized reportId
         }
 
+        public void OnNextUpdate()
+        {
+        }
+
+        // filter out three lower bits as jitter noise
+        internal const byte AxisDeadZoneLow = 0b01111000;
```
ekcoh: Do we need these repeated in every device-specific adaption or do you think we could use a common internal constant until they need to diverge?

ekcoh: It seems like this is based off complements? If this is the case maybe 
AxisDeadZoneHigh = ~AxisDeadZoneLow

...looking more carefully it seems approach is mid point based, hence maybe to make it clearer what's intended
internal const byte DeadZoneJitterRange = 7;
internal const float AxisOrigin = (float)Byte.MaxValue / 2..0f; // Non-actuated mid-point?
internal const byte AxisDeadZoneLow = (byte)Math.floor(AxisOrigin - DeadZoneJitterRange);
internal const byte AxisDeadZoneHigh = (byte)Math.ceil(AxisOrigin + DeadZoneJitterRange);

Or maybe clarify this with a comment if this doesn't evaluate to a constant in C#, likely not thats the reason.... What I am after is that the actual property here is distance from mid-point which affects the outcome.

andrew-oc: Should these values be user configurable? What if a user sets min and max deadzones to 0 and 1 at the control level? They'll still get input events correctly but the device won't be made current until they move outside the hard-coded deadzone. Is that ok?

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
```diff
@@ -549,6 +552,50 @@ unsafe bool IEventPreProcessor.PreProcessEvent(InputEventPtr eventPtr)
...
+                    || newState->rightStickX<AxisDeadZoneLow
+                                             || newState->rightStickX> AxisDeadZoneHigh
+                    || newState->rightStickY<AxisDeadZoneLow
+                                             || newState->rightStickY> AxisDeadZoneHigh
+                    // we need to make device current if triggers or buttons state change
+                    || newState->leftTrigger != currentState->leftTrigger
+                    || newState->rightTrigger != currentState->rightTrigger
+                    || newState->buttons0 != currentState->buttons0
+                    || newState->buttons1 != currentState->buttons1
+                    || newState->buttons2 != currentState->buttons2;
```
andrew-oc, line 585: Do we actually need to check all these non-jittery controls? Could we assume that because there's a state event we always want to make this device current except if the sticks have jittered? Or in other words, set actuated to true by default and only false if there's stick movement in the deadzones.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
jimon: Hard to say, probably shouldn't be user configurable. The value is specific to device and not controls, e.g. every gamepad is inheritably noisy due to resistor / ADC noise. So actual value is specific to device instance and will change over time as sticks degrade. Hence likely there is no optimal value out there, and this is just a good enough approximation.

A better way would be to do LPF on this, but that's a story for another time.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
jimon: We have to check every event because these gamepads send an event for every sensor update, which is around 100-300 Hz, but because we don't process sensors yet we get bunch of events where nothing changed.

andrew-oc: Ah of course. Missed that.

andrew-oc (approved):

ekcoh (approved):
This LGTM as a solution to mitigate the referenced problem. Long-term its likely better to have a separate low-pass filter with a large time-window evaluate that for this purpose instead to potentially allow for a slimmer deadzone filter but this seems like a good improvement and is straight-forward.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
jimon: It's a virtual state format that we convert incoming HID frames into, look for FourCC that states D S V S

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
ekcoh: Another way would be to do noise estimation to auto-adjust it, but that is also a story for another time.

jimon: It's just empirical observation of mine that 3 lower bits are jittery, I didn't want to mess with arithmetic to get to the result of filtering out 3 lower bits, and instead hardcoded them manually in a bitmask.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
```diff
@@ -549,6 +552,50 @@ unsafe bool IEventPreProcessor.PreProcessEvent(InputEventPtr eventPtr)
...
+        public unsafe void OnStateEvent(InputEventPtr eventPtr)
+        {
+            if (eventPtr.type == StateEvent.Type && eventPtr.stateFormat == DualSenseHIDInputReport.Format)
+            {
+                var currentState = (DualSenseHIDInputReport*)((byte*)currentStatePtr + m_StateBlock.byteOffset);
+                var newState = (DualSenseHIDInputReport*)StateEvent.FromUnchecked(eventPtr)->state;
+
+                var actuated =
+                    // we need to make device current if axes are outside of deadzone specifying hardware jitter of sticks around zero point
+                    newState->leftStickX<AxisDeadZoneLow
```
jamesmcgill: That looks a little weird to read at first glance.
I think because for the standard deadzone we also have low and high but they mean 'allowed' (inclusive)range and there I would have expected 'over' the Low threshold to be the actuated one (&& under High).

Here it is more the inverse, an exclusionary range (actuated if outside of range).
What if it were called something like the jitter zone/region, or something like that, so that it doesn't confuse with the other deadzone concept.

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepadHID.cs
jimon: Renamed to JitterMaskLow/High

PR #1618, opened 2022-12-13
Fix test framework define

jechter:
### Description

Unity.InputSystem.TestFramework was set up to depend on UNITY_INCLUDE_TESTS. This is a special define, which will cause the assembly to only be included if the containing package is set up as "testable" in the project manifest. (See code in  `IsCompatibleWith` - `Editor/Mono/Scripting/ScriptCompilation/CustomScriptAssembly.cs`). This means that another package cannot add tests using `InputTestFixture` without turning on all the InputSystem tests by adding the input system as a testable package. 

Instead, the assembly should use `UNITY_TESTS_FRAMEWORK` which is meant for this, and will cause the assembly to be built if testing is enabled, regardless of whether the input system package is set to testable.

This is analogous to how other test frameworks in packages are set up (ie: https://github.cds.internal.unity3d.com/unity/com.unity.test-framework.performance/blob/2e26d43ce4dd6fa6b3d19d11aed9ed582685a675/package/Editor/Unity.PerformanceTesting.Editor.asmdef

https://github.cds.internal.unity3d.com/unity/com.unity.testframework.graphics/blob/a27718d6e0271e0cb757a14c9c1928454adf58e0/com.unity.testframework.graphics/Runtime/UnityEngine.TestTools.Graphics.asmdef ).

### Changes made

See description.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


jamesmcgill (approved):
Thanks @jechter !

PR #1619, opened 2022-12-14
More gamepad related Basic API functionality

andrew-oc:
* Added initialization code and called it from the appropriate editor and player places.
* IsGamepadConnected
* DidGamepadConnectThisFrame
* DidGamepadDisconnectThisFrame
* IsControlDown/Up/Pressed gamepad overloads
* 'gamepads' collection of stable indexed devices
* Modified some code when querying gamepads using the Inputs enum that was allocating due to closure. Same needs to be done on other device types.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/API/Input.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/API/Input.cs
@@ -1,7 +1,10 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 using UnityEngine.InputSystem.Controls;
+using UnityEngine.InputSystem.Utilities;
+using UnityEngine.PlayerLoop;
 
 // TODO rename HighLevel to something that makes sense
 namespace UnityEngine.InputSystem.HighLevel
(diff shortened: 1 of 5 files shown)


Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -239,62 +268,115 @@ public static class Input
...
-        /// A collection of all keyboards currently connected to the system.
+        /// A collection of all joysticks currently connected to the system.
         /// </summary>
-        public static IReadOnlyList<Keyboard> keyboards { get; }
+        public static IReadOnlyList<Joystick> joysticks { get; }
+#endif
+
+        static Input()
+        {
+            s_GamepadSlotEnums = new ReadOnlyArray<GamepadSlot>(new[]
```
jimon, line 281: Is it a super hardcode version because `Enum.GetValues` allocates or something?

Packages/com.unity.inputsystem/InputSystem/Utilities/PlayerLoopSystemExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+                return true;
+            }
+
+            if (playerLoop.subSystemList == null)
+                return false;
+
+            for (var i = 0; i < playerLoop.subSystemList.Length; i++)
+            {
+                var system = playerLoop.subSystemList[i];
+                if (system.InsertSystemAsSubSystemOf<TNewSystem, TExistingSystem>(updateDelegate, insertFirst))
```
jimon, line 37: sneaky :)

jimon (approved):

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
andrew-oc: Yep. And because it's the simplest possible version of it. It's just nicer than filtering All (and Any was there at the time I wrote it too).

PR #1621, opened 2022-12-16
Minor OnScreenStick changes

andrew-oc:
### Description

Add some missing tooltips and inspector conveniences to OnScreenStick. Additionally, changing the range in dynamic origin mode now properly adjusts the size of the clickable region


jamesmcgill (approved):
Thank you

PR #1625, opened 2023-01-04
FIX: remove check that triggered NotImplementedException for Usage bindings (ISXB-373)

jamesmcgill:
### Description

If a Usage (e.g. Submit) was bound then an NotImplemented exception could occur because it hit a case that wasn't implemented yet. 

- The sample project I had was extremely simple to reproduce: just bind an action to Submit (then attempt to call `InputControlPath::ToHumanReadableString()` to trigger the failure).
- In that case the `control` will be `Key/Enter` and `control.device` will be `Keyboard`.
- Usages are stored at device level, not in the control subhierarchy AFAIK.

There is a `TODO: support scavenging a subhierarchy for usages` there, so my first instinct was to implement the missing case, however, at least in this case I couldn't see why we would want to search from `Key/Enter` downwards in the subhierarchy. And I couldn't come up with another use-case. I'm also unsure how this could be implemented if we wanted to (without moving usages out to be on per control basis which seems redundant?).

Perhaps I haven't understood the intention of the comment correctly, so happy to hear if someone knows better?

Even if we do have future plans for this, I think we should still remove this Exception today as the reproduction was extremely easy to trigger and I don't see any issue with looking at the root device for Usages in the meantime.

### Changes made

- Removed the check that triggered the exception and allow the code to fall through and use the device.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


lyndon-unity (approved):
I don't understand this area well enough to understand side effects but I feel that removing this exception is reasonable.

Perhaps a debug warning could be valuable but I'm comfortable clicking approve for the error case reported.

jamesmcgill:
Yeah, I'm also not 100% on what is intended missing feature here. Unless there is a rush to get this in, I think we can leave it open for the others to review incase I'm missing something.

I don't even think the warning is necessary, as my understanding at the moment is that this works as-is (i.e. continues to use the device when looking for the Usage in this case should be correct).

jimon (approved):
I couldn't came up with counter example. Given that nobody complained for years about this lets merge and see what breaks down the line.

PR #1626, opened 2023-01-04
FIX: Remove NotImplementedException when loading outdated binding overrides from json (ISXB-375)

jamesmcgill:
### Description

Another NotImplementedException. This one triggers if a binding ID that is saved in the override json file no longer exists at runtime (due to modifications in the inputasset) at the time it is reloaded.

Removed and replaced with a warning as the customer correctly argues for on the ticket: this crashes on startup for end-users if the developer pushed an update to them. Also any correct entries in the file will be lost due to one invalid entry terminating the loading early.

### Changes made

Replace exception with a warning.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


lyndon-unity (approved):
Looks good and given discussions in person it sounds like its valid approach so the code continues to execute past this.

PR #1629, opened 2023-01-19
Basic pointer and mouse APIs

andrew-oc:


Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -1171,11 +1171,83 @@ public static bool DidGamepadDisconnectThisFrame(GamepadSlot slot)
...
+        /// position of the first one. In window space coordinates, the bottom left corner of
+        /// the window is 0, 0.
+        /// pointerPosition will work for mouse, pen, and touchscreen devices.
+        /// Note that for Touchscreen devices, this will return the position of the primary touch.
+        /// </remarks>
+        public static Vector2 pointerPosition
+        {
+            get
+            {
+                foreach (var inputDevice in InputSystem.devices)
```
jimon: Wonder if `Pointer.current.position.ReadValue()` could be better here? Because if I have a mouse and a pen attached, then I move first mouse, and then pen, shouldn't I see updated position in both cases? As they move same physical cursor on the screen.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -1171,11 +1171,83 @@ public static bool DidGamepadDisconnectThisFrame(GamepadSlot slot)
...
+        public static Vector2 scrollDelta
+        {
+            get
+            {
+                foreach (var inputDevice in InputSystem.devices)
+                {
+                    if (!(inputDevice is Mouse mouse))
+                        continue;
+
+                    return mouse.scroll.ReadValue();
```
jimon: Same here as pointer position

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
andrew-oc: I was worried that the usual problems with current would come up and position would end up switching to whatever noisy pointer device last updated. Granted the committed code isn't ideal. You'd never be able to use a pen on a tablet because the touchscreen would probably always be registered first for example, but its a question of which behaviour is least surprising I think.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -1334,6 +1383,47 @@ internal static void Initialize()
                 () => s_TimeHasUpdatedThisFrame = false);
             UnityEngine.LowLevel.PlayerLoop.SetPlayerLoop(playerLoop);
 #endif
+            s_PointerPosition = Vector2.zero;
+            s_ScrollDelta = Vector2.zero;
+
+            s_PointerAction = new InputAction("Pointer", InputActionType.PassThrough, expectedControlType: "Vector2");
```
jimon, line 1389: And it's not value type to avoid max magnitude sorting?

jimon (approved):

PR #1632, opened 2023-02-01
Joystick API

andrew-oc:
Added support for joysticks added through the HID path that have multiple buttons.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/API/Input.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/API/Input.cs
@@ -24,6 +24,9 @@ namespace UnityEngine.InputSystem.HighLevel
     /// <seealso cref="Input.IsControlDown(Inputs)"/>
     /// <seealso cref="Input.IsControlUp(Inputs)"/>
     /// <seealso cref="Input.IsControlPressed(Inputs)"/>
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -772,25 +822,106 @@ private static ButtonControl GetGamepadButtonControl(Gamepad gamepad, Inputs inp
...
+                    break;
+                default:
+                    throw new InvalidEnumArgumentException(nameof(joystickButton), (int)joystickButton, typeof(JoystickButton));
+            }
+
+            // do a non-recursive search for the button based on name
+            for (var i = 0; i < joystick.allControls.Count; i++)
+            {
+                var control = joystick.allControls[i];
+                if (control.name == buttonName)
```
jimon, line 910: That ain't fast huh :)

jimon (approved):

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
andrew-oc: Oops. good catch. Fixed (or at least improved) in the next PR.

PR #1637, opened 2023-02-09
FIX: InputAction.bindings.count is not updated after calling Erase()

jfreire-unity:
### Description

This fixes `InputAction.bindings.count` not getting correctly updated after removing bindings with `Erase()`.

### Changes made

The [`Erase()`](https://github.com/Unity-Technologies/InputSystem/blob/3f0a6150fa13ee6d21a1220ebe68f5f21d666eef/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs#L1567-L1591) function didn't update `m_Action.m_BindingsCount` when a bindings were removed, and now it does.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -3609,6 +3609,40 @@ partial class CoreTests
 
     [Test]
     [Category("Actions")]
+    public void Actions_CanRemoveBindingsFromActions()
+    {
+        var gamepad = InputSystem.AddDevice<Gamepad>();
+        var action = new InputAction();
+
+        action.AddBinding("<Gamepad>/leftStick");
+        action.AddBinding("<Gamepad>/rightStick");
+
+        action.AddCompositeBinding("OneModifier")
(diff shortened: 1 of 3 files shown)


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1871,7 +1871,7 @@ public unsafe void Actions_CanReadValueFromAction_InCallback_WithoutKnowingValue
             }
         };
 
-        InputSystem.QueueStateEvent(gamepad, new GamepadState {leftStick = new Vector2(0.123f, 0.234f)});
```
jimon: Weird amount of formatting, is it our formatter got updated or something?

jimon (approved):

Assets/Tests/InputSystem/CoreTests_Actions.cs
jfreire-unity: I'm removing trailing whitespace in my editor and didn't notice this. But our formatter still allowed it.
I'll "put it back" as it was for consistency across the repo. Thanks for pointing it out 👍

PR #1638, opened 2023-02-10
Fixing trunk CI issues

jimon:
### Description

So com.unity.coding `0.1.0-preview.20` that we use no longer compiles on 2023.x.

### Changes made

I've updated it to `0.1.0-preview.24`, but in meanwhile the underlying scraping tool was also updated, so now the output of the tool no longer matched 1.0.0 reference API files we used for validation.

I've made a branch `dmytro/regenerate-scraped-file` on top of 1.0.0 tag to regenerate the .api files and updated them in this PR.

The rest of the changes is fixing exclusion lists because of all namespace changes.

### Notes

Good luck reviewing it! 🔥 :trollface: 
PS. I did my best to sanity check outputs, but .api file change is nearly impossible to review without some awk magic.

### DIFF
--- a/Tools/API/1.0.0/Unity.InputSystem.api
+++ b/Tools/API/1.0.0/Unity.InputSystem.api
@@ -17,7 +17,7 @@ namespace Unity.XR.GoogleVr
         [UnityEngine.InputSystem.Layouts.InputControl] [UnityEngine.Scripting.Preserve] public UnityEngine.InputSystem.Controls.ButtonControl volumeDown { get; }
         [UnityEngine.InputSystem.Layouts.InputControl] [UnityEngine.Scripting.Preserve] public UnityEngine.InputSystem.Controls.ButtonControl volumeUp { get; }
         public DaydreamController() {}
-        protected virtual void FinishSetup();
+        protected override void FinishSetup();
     }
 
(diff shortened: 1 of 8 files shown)


Tools/API/1.0.0/Unity.InputSystem.api
```diff
@@ -17,7 +17,7 @@ namespace Unity.XR.GoogleVr
         [UnityEngine.InputSystem.Layouts.InputControl] [UnityEngine.Scripting.Preserve] public UnityEngine.InputSystem.Controls.ButtonControl volumeDown { get; }
         [UnityEngine.InputSystem.Layouts.InputControl] [UnityEngine.Scripting.Preserve] public UnityEngine.InputSystem.Controls.ButtonControl volumeUp { get; }
         public DaydreamController() {}
-        protected virtual void FinishSetup();
```
jfreire-unity, line 20: Seems the pattern here is to change from `virtual` to `override`.  Is it just the way it is for the new format of  "com.unity.coding": "0.1.0-preview.24" or is there any other reason behind it? Anyway, this file was regenerated to meet the new format, correct?

jimon: It was .24 change, it's probably just to mark an overload with `override` instead of `virtual`

PR #1644, opened 2023-02-17
FIX: JSON Binding overrides are set to empty string when null

jfreire-unity:
### Description

When saving unsigned binding overrides (with `null` value) they were being set as empty strings "" when loading from JSON. The JsonUtility class seems to set `null` overrides to an empty string `""` when loading JSON data.

### Changes made

`BindingOverrideJson` now serializes `null` overrides with "null" strings when saving, and assignes `null` for "null" strings when loading.
An alternative would be to add `bool` values for all the override values to check if the values were set on not. However, it felt that we could use the existing fields to disambiguate the data without adding flags to the JSON structure. Happy to discuss if there are downsides to my approach.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
@@ -1463,15 +1463,25 @@ namespace UnityEngine.InputSystem
             public string interactions;
             public string processors;
 
-            public static BindingOverrideJson FromBinding(InputBinding binding)
+            public static BindingOverrideJson FromBinding(InputBinding binding, InputAction bindingAction = null)
             {
                 return new BindingOverrideJson
                 {
-                    action = binding.action,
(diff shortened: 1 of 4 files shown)


jimon (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -1463,15 +1463,25 @@ internal struct BindingOverrideJson
             public string interactions;
             public string processors;
 
-            public static BindingOverrideJson FromBinding(InputBinding binding)
+            public static BindingOverrideJson FromBinding(InputBinding binding, InputAction bindingAction = null)
```
andrew-oc, line 1466: We can't really do this unfortunately because it changes binary compatibility. It looks like a non-breaking change, but if someone is working against a precompiled dll of the input system, it breaks.

andrew-oc (changes requested):

PR #1646, opened 2023-02-17
Better shortcut key support

andrew-oc:
### Description

Existing shortcut key support falls a bit a short of being useful:
  * When shortcutKeysConsumeInput is off, there is basically no shortcut key support at all. Shift+B will trigger bindings to just B and the application has no reasonable way of determining that a different key combination was pressed
  * When all action maps are enabled (anecdotally the default for most users) event consumption causes confusion due to unintuitive behaviour. A Vector2 binding to WASD would consume the input of a binding to Ctrl+W when Ctrl+W is pressed.
  * The application has no way to override the default event consumption

### Changes made

* Composite bindings are now sorted by a priority system that comes from the bindings themselves, rather than as a count of the number of binding parts. In practice this means that for example a Vector2Composite binding does not have a higher priority than a one or two modifier composite binding i.e. pressing W to trigger a WASD action would not by default prevent a binding to Ctrl-W from performing.
* Added several APIs to allow application logic to control event handling/consumption.
    * CallbackContext.eventHandled can be used to check if a higher priority binding has already handled the current event
    * CallbackContext.HandleEvent() allows any 'performed' callback to set an input event as handled
    * InputAction.wasPerformedThisFrame(bool checkConflictingActions) allows polling code to determine if a higher priority binding has performed
    * InputBindingComposite.handleInputEvents allows applications to decide which composite bindings will handle events

### Notes

This PR does not change the default behaviour of code that has InputSettings.shortcutKeysConsumeInput turned off i.e. no bindings are blocked due to event consumption. However it is possible that due to the differences in how controls are sorted within the state monitor code that some bindings will now be triggered in a different order than before.

#### Future work
We can build two interesting features on top of this PR:
* It would be possible to allow users to set the priority of bindings on an instance by instance basis, and hence fully control the order that bindings perform in
* The action grouping that this PR comes with would allow us to detect conflicting bindings that we could display in the input action asset editor as warnings or even detect at runtime for players rebinding controls.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10230,68 +10232,200 @@ public void Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works
...
+    public void Actions_WithCompositeBindings_WithConsumeInputEventsSetToTrue_HandleInputEvents()
     {
-        InputSystem.settings.shortcutKeysConsumeInput = shortcutsEnabled;
+	    var keyboard = InputSystem.AddDevice<Keyboard>();
 
-        var keyboard = InputSystem.AddDevice<Keyboard>();
+	    var map = new InputActionMap("map1");
+		var action1 = map.AddAction(name: "action1");
+        // 1D axis composite doesn't by default consume input events, but by setting consumeInputEvent to true, it should
+		action1.AddCompositeBinding("1DAxis(handleInputEvents=true)")
```
jfreire-unity: In this comment do you mean `handleInputEvents=true` by referring to "setting consumeInputEvent to true" ? If so, could you fix the name in the comment to avoid causing a "slight confusion" to the reader? And maybe do the same to the name of the test?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10230,68 +10232,200 @@ public void Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works
...
+	    InputSystem.settings.shortcutKeysConsumeInput = true;
+
+	    var keyboard = InputSystem.AddDevice<Keyboard>();
+
+	    var map = new InputActionMap("map1");
+	    var action1 = map.AddAction(name: "action1");
+        action1.AddCompositeBinding("1DAxis")
+		    .With("Positive", "<Keyboard>/w")
+		    .With("Negative", "<Keyboard>/s");
+	    var action2 = map.AddAction("action2", binding: "<keyboard>/w");
```
jfreire-unity: Does this mean that the `action2.performed` will never be called? Only pressing "w+s" will call `action1.performed`?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1213,62 +1215,95 @@ public unsafe bool WasReleasedThisFrame()
...
+		/// This method can be made to check if a more specific binding has performed before this one in the current
+		/// frame. For example, when using shortcut key bindings, pressing Ctrl+C will trigger any actions that are
+		/// bound only to C, as well as any bound to Ctrl+C. Passing true for the 'checkConflictingActions' argument
+		/// will cause this method to return false if this action has been bound to the C key but Ctrl+C has been
+		/// pressed and another action is bound to that key combination.
+		///
+		/// Note that bindings can be made to consume input events, or in other words, prevent less specific bindings
+		/// from even performing in the first place. To enable this, set the <see cref="InputSettings.shortcutKeysConsumeInput"/>
+		/// property to true, and make sure that the 'Consume Input Events' property on the binding is also set to
+		/// true. See <see cref="InputBindingComposite.handleInputEvents"/> for more details.
```
jfreire-unity: Can we reference the right property you refer to the "Consume  Input Events" property? Or is it `InputBindingComposite.handleInputEvents` ?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1872,6 +1907,49 @@ public int valueSizeInBytes
                 }
             }
 
+            /// <summary>
+            /// Check if a more specific binding has already handled this event.
```
jfreire-unity: "...more specific binding" sounds a bit vague to me. Feels like it does something "special".
Could we replace it with something like "Check if a binding has already handled this event"? It might also be my lack of understanding of what it is a "more specific binding" in this context.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1997,5 +2075,32 @@ public override string ToString()
...
+	        for (var j = 0; j < groupCount; j++)
+	        {
+		        var groupIndex = state.m_ActionGroupIndirectionTable[actionState.actionGroupStartIndex + j + 1];
+		        var actionGroup = state.m_ActionGroups[groupIndex];
+                 
+		        if (actionGroup.lastEventHandledByAction == -1)
+			        continue;
+
+		        // if the last action in this group to handle an event didn't perform in the current step count,
+		        // then there is no way it can have run before this action, so early out
```
jfreire-unity: Does "early out" == "exit early"? Just trying to understand the comment

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -123,6 +124,14 @@ internal unsafe class InputActionState : IInputStateChangeMonitor, ICloneable, I
         private Action m_OnBeforeUpdateDelegate;
         private Action m_OnAfterUpdateDelegate;
 
+        // this is a sparse array that we index by control group index
+        internal NativeArray<ActionGroup> m_ActionGroups;
+        
+        // each action (TriggerState) has an index into this array. The element that the the TriggerState
+        // points at is a count of the number of groups that action appears in, and the next n elements
+        // are the indices into the m_ActionGroups where those groups live.
+        internal NativeArray<ushort> m_ActionGroupIndirectionTable;
```
jfreire-unity, line 134: Need to ask for more clarity on this comment offline. It's hard for me to understand it.
What is the purpose of `m_ActionGroupIndirectionTable` ?

jfreire-unity, line 134: What is the reason for it to be `ushort` ?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -150,47 +158,92 @@ private void ComputeControlGroupingIfNecessary()
...
 
-                    controlGroupingAndComplexity[i * 2] = (ushort)currentGroup;
+                    controlGroupingAndComplexity[i].group = (ushort)currentGroup;
 
                     ++currentGroup;
                 }
+
+				// add the action that the current binding belongs to to the relevant group, sorted by complexity
+                if (controlGroupingAndComplexity[i].group >= m_ActionGroups.Length)
+                    ArrayHelpers.Resize(ref m_ActionGroups, Mathf.Max(10, m_ActionGroups.Length + 10), Allocator.Persistent);
```
jfreire-unity: What's the reason for the resizing to the minimum of 10? Could we make this value a const so it's easier to refer to?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -2426,12 +2481,12 @@ private void ChangePhaseOfActionInternal(int actionIndex, TriggerState* actionSt
...
-            Debug.Assert(actionIndex >= mapIndices[trigger.mapIndex].actionStartIndex,
-                "actionIndex is below actionStartIndex for map that the action belongs to");
-            var action = map.m_Actions[actionIndex - mapIndices[trigger.mapIndex].actionStartIndex];
-            trigger.phase = newPhase;
+			// Let listeners know.
+			var map = maps[trigger.mapIndex];
+			Debug.Assert(actionIndex >= mapIndices[trigger.mapIndex].actionStartIndex,
+				"actionIndex is below actionStartIndex for map that the action belongs to");
+			var action = map.m_Actions[actionIndex - mapIndices[trigger.mapIndex].actionStartIndex];
+			trigger.phase = newPhase;
```
jfreire-unity: Formatting is weird here, too many "tabs".

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -2455,6 +2510,54 @@ private void ChangePhaseOfActionInternal(int actionIndex, TriggerState* actionSt
                     break;
                 }
             }
+
+            // now that the callbacks have run, we can let the composite decide whether to set the event as handled.
+            // we have to do this after the callbacks because it's expected that callbacks will check 
+            // CallbackContext.eventHandled to determine if another higher priority input action has already handled
+            // this input event and we don't want it to be true in the handler that is actually handling the event.
```
jfreire-unity: Nice one 👍

Packages/com.unity.inputsystem/InputSystem/Actions/InputBindingComposite.cs
```diff
@@ -31,7 +32,9 @@ namespace UnityEngine.InputSystem
     /// <seealso cref="InputBinding.isComposite"/>
     public abstract class InputBindingComposite
     {
-        /// <summary>
+	    private const int kDefaultCompositeBindingPriority = 2;
```
jfreire-unity: Why is 2 the default priority of a composite?

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -660,25 +660,32 @@ public bool disableRedundantEventsMerging
...
+		/// Improves shortcut key support by enabling composite bindings to consume input events.
+		/// </summary>
+		/// <remarks>
+		/// Input event consumption means that one binding can prevent input events from being processed
+		/// by another. For example, a binding to 'Shift+B' would prevent a binding to 'B' from performing
+		/// when 'Shift+B' is pressed. Note that input events can only be consumed by composite bindings
+		/// i.e. bindings that bind to more than one control. By default, only the
+		/// <see cref="Composites.OneModifierComposite"/> and <see cref="Composites.TwoModifiersComposite"/>
+		/// bindings consume events, but this can be switched on for any composite type by checking the
+		/// Consume Input Events checkbox for the binding in the Input Action Asset editor, or by setting
```
jfreire-unity: Thanks for the explanation. Now I understand what "consumption" means in this context.
Does this means consuming events "feature" will make only 1 of the events to be performed for the event with the highest priority?

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -660,25 +660,32 @@ public bool disableRedundantEventsMerging
...
+		/// the <see cref="InputBindingComposite.handleInputEvents"/> field to true.
+		///
+		/// Note that input event consumption is currently unreliable when used with multiple Input Action
+		/// Assets simultaneously, or with multiple standalone Input Actions created through code that do
+		/// not belong to the same Input Action Map. This is because input control groupings are created
+		/// in isolation for each Input Action Asset or individual Input Action Map, so group numbers can
+		/// overlap. This is a known issue and will be addressed in a later version.
```
jfreire-unity: Do we have this known issue tracked somewhere in our backlog?

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -660,25 +660,32 @@ public bool disableRedundantEventsMerging
...
+		/// </remarks>
+		/// <seealso cref="InputBindingComposite.handleInputEvents"/>
+		public bool shortcutKeysConsumeInput
+		{
+			get => m_ShortcutKeysConsumeInputs;
+			set
+			{
+				if (m_ShortcutKeysConsumeInputs == value)
                     return;
```
jfreire-unity: The formatting is weird here. Seems a closing bracket is missing?

jfreire-unity (changes requested):
Added questions for things that weren't clear when I look through it them. But I believe I need a walkthrough from you @andrew-oc to understand better the work involved.

I also noticed some formatting inconsistencies for you to look at. Don't know if it's the GitHub "diff" viewer.

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -927,6 +940,29 @@ Press(Keyboard.current.leftShiftKey);
 Press(keyboard.bKey);
 ```
 
+If an application needs even more control over when input events are consumed, it is possible to completely take over the handling of events by using the [`CallbackContext.HandleEvent`](api/UnityEngine.InputSystem.InputAction.CallbackContext.html#UnityEngine_InputSystem_InputAction_CallbackContext_HandleEvent) method. In the following example, event consumption is turned off for the `Shift+B` binding, and the performed handler only handles the input if the 'inventoryOpen' flag is true. The `B` binding then decides whether to perform some actions based on whether the input has been handled or not.
```
jfreire-unity: Nice example 🎉 !

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -927,6 +940,29 @@ Press(Keyboard.current.leftShiftKey);
...
+
+```CSharp
+bAction.AddBinding("<Keyboard>/b");
+shiftbAction.AddCompositeBinding("OneModifier(handleInputEvents=false)")
```
jfreire-unity, line 954: In this example "consume input events" is off, right?

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -927,6 +940,29 @@ Press(Keyboard.current.leftShiftKey);
...
+    .With("Modifier", "<Keyboard>/shift")
+    .With("Binding", "<Keyboard>/b");
+bool inventoryOpen;
+
+// Print something to the console when the actions are triggered.
+bAction.performed += ctx =>
+{
+    if(ctx.eventHandled == false)
+        Debug.Log("B action performed");
```
jfreire-unity, line 963: So, assuming consume events it's off by my previous comment, this means that there's still a way to know if "another" event for this key was called, right?

jfreire-unity (commented):
Overall, the work looks great and gives more control to developers. Great work 👍 

Regarding the PR, I would like to suggest that we have more granularity in the commits OR smaller PRs to have a better understanding of the changes added. I
It was a bit hard for me to follow everything added in just one commit, and some parts are still clueless to me. But you can take this with "a grain of salt" as I'm still very new to the codebase.

Assets/Tests/InputSystem/CoreTests_Actions.cs
jfreire-unity: Understood this after our sync yesterday. The line with `ctx.HandleEvent();` avoids event propagation to other actions.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
jfreire-unity: Need to look more into how `InputActionState` works on my own time but I got the general idea after our sync.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
jfreire-unity: From our sync, I understood that's a "random value" to avoid resizing multiple times with only 1 array. Still, making it a const makes sense to me, but it's not a blocking issue for this PR.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
jfreire-unity: Understood this better after our sync.

Assets/Tests/InputSystem/CoreTests_Actions.cs
andrew-oc: Done

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
andrew-oc: Done

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
andrew-oc: Changed "more specific" to "higher priority" and added links to the manual page that explains binding priorities.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
andrew-oc: It does, but that comment was a little out of date. Updated.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
andrew-oc: Just saving memory. The expectation is that there would never be greater than 65k action groups, so we don't need to store any indexes greater than that.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
andrew-oc: Made it a constant

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
andrew-oc: No but you reminded me I meant to put something in the documentation about it.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
andrew-oc: Huh, looks ok in the formatter

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
andrew-oc: Yep, it's off. I modified the example to make sure there's no question.

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
andrew-oc: Yep, consuming events and handling events are two separate concepts. Handled means the input caused some composite binding to trigger but it doesn't by itself stop lower priority events from also triggering. That's why callbacks need to check the handled flag to see if someone else has already triggered for the current event.

Consuming is what stops an event from triggering lower priority bindings. It works by checking the handled flag, but also the InputSettings.shortcutKeysConsumeInput.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10230,66 +10232,198 @@ public void Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works
...
+        Assert.That(wasHandled, Is.True);
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_ActionCallbacksCanSetEventsAsHandled()
+    {
         var keyboard = InputSystem.AddDevice<Keyboard>();
 
-        var map1 = new InputActionMap("map1");
```
lyndon-unity, line 10245: Why are we removeing this WASD test case that we've seen from forums (ie the UI / game use case)

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10230,66 +10232,198 @@ public void Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works
...
+    [Test]
+    [Category("Actions")]
+    public void Actions_WasPerformedThisFrameReturnsFalse_WhenHigherPriorityConflictingActionHasAlreadyPerformed()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+
+        var map = new InputActionMap();
+
+        // action with bindings to Ctrl+Shift+C and C should appear twice in the 'C' group
+        var action0 = map.AddAction("action1", binding: "<Keyboard>/c");
```
lyndon-unity: action0 variable and action 1 name is a bit confusing for me to read can we be consistent. E.g rename variable to match strings

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased]
 
+### Changed
+- Improved support for shortcut keys and input event consumption/handling in general.
+  * Composite bindings are now sorted by a priority system that comes from the bindings themselves, rather than as a count of the number of binding parts. In practice this means that for example a Vector2Composite binding does not have a higher priority than a one or two modifier composite binding i.e. pressing W to trigger a WASD action would not by default prevent a binding to Ctrl-W from performing.
+  * Added several APIs to allow application logic to control event handling/consumption.
+    * CallbackContext.eventHandled can be used to check if a higher priority binding has already handled the current event
+    * CallbackContext.HandleEvent() allows any 'performed' callback to set an input event as handled
+    * InputAction.wasPerformedThisFrame(bool checkConflictingActions) allows polling code to determine if a higher priority binding has performed
```
lyndon-unity: I'm not a fan of the bool here. I'd like to see that changing to an enum

InputAction.wasPerformedThisFrame(true) isn't very clear what that does

Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
```diff
@@ -2,6 +2,10 @@
 
 The following is a list of known limitations that the Input System currently has.
 
+## Actions
+
+* Input event consumption i.e. the ability to prevent a binding to "A" from triggering when a binding to "SHIFT+A" has already triggered, does not reliably work across multiple Input Action Assets, or multiple `InputActionMap` instances created through code. This is due to the way the runtime state for control grouping and priority is calculated in isolation for each Input Action Asset or standalone `InputActionMap`.
```
lyndon-unity, line 7: Is the limitation specific to multiple action maps created in code (i.e. multiple maps in an asset would work)

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1254,21 +1275,39 @@ public unsafe bool WasReleasedThisFrame()
...
+        /// is also set to true.
+        ///
+        /// For more in-depth reading on binding priorities, see the <a href="../manual/ActionBindings.html#conflicting-inputs">manual</a>.
         /// </remarks>
         /// <seealso cref="WasPressedThisFrame"/>
         /// <seealso cref="phase"/>
-        public unsafe bool WasPerformedThisFrame()
+        /// <seealso cref="InputSettings.shortcutKeysConsumeInput"/>
+        /// <seealso cref="InputBindingComposite.handleInputEvents"/>
+        public unsafe bool WasPerformedThisFrame(bool checkConflictingActions)
```
lyndon-unity: As mentioned earlier I'd prefer this to be an enum for the parameter to make it clearer in usage

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1997,5 +2082,32 @@ public override string ToString()
...
             }
         }
+
+        private unsafe bool HasHigherPriorityActionAlreadyPerformed()
+        {
+            var state = GetOrCreateActionMap().m_State;
+            ref var actionState = ref state.FetchActionState(this);
+
+            var groupCount = state.m_ActionGroupIndirectionTable[actionState.actionGroupStartIndex];
+            for (var j = 0; j < groupCount; j++)
```
lyndon-unity: Could remove the +1 below but making this loop specific 1 to groupCount
for (var j = 1; j <= groupCount; j++)

I'd prefer to see the variable name being the groupOffsetIndex or similar rather than the unclear 'j'

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -153,44 +162,93 @@ private void ComputeControlGroupingIfNecessary()
...
-                        if (!partBinding.isPartOfComposite || partBinding.compositeOrCompositeBindingIndex != compositeBindingIndex)
-                            break;
-                        ++complexity;
-                    }
+                    var compositeBinding = bindingStates[binding.compositeOrCompositeBindingIndex];
+                    var composite = composites[compositeBinding.compositeOrCompositeBindingIndex];
+                    complexity = composite.GetPriority();
                 }
-                controlGroupingAndComplexity[i * 2 + 1] = (ushort)complexity;
+                controlGroupingAndComplexity[i].complexity = (ushort)complexity;
```
lyndon-unity, line 168: Pleased to see this moved to a struct so the indexing is clearer

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -153,44 +162,93 @@ private void ComputeControlGroupingIfNecessary()
...
                     }
 
-                    controlGroupingAndComplexity[i * 2] = (ushort)currentGroup;
+                    controlGroupingAndComplexity[i].group = (ushort)currentGroup;
 
                     ++currentGroup;
                 }
+
+                // add the action that the current binding belongs to to the relevant group
+                if (controlGroupingAndComplexity[i].group >= m_ActionGroups.Length)
```
lyndon-unity: Wondering if we can move the resize outside the for loop so we only resize once

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -153,44 +162,93 @@ private void ComputeControlGroupingIfNecessary()
...
+                        Allocator.Persistent);
+
+                #if UNITY_2019_4
+                var grouping = m_ActionGroups[controlGroupingAndComplexity[i].group];
+                #else
+                ref var grouping = ref UnsafeUtility.ArrayElementAsRef<ActionGroup>(m_ActionGroups.GetUnsafePtr(),
+                    controlGroupingAndComplexity[i].group);
+                #endif
+                if (m_ActionGroups[controlGroupingAndComplexity[i].group].isCreated == false)
+                    grouping = new ActionGroup(5);
```
lyndon-unity: What's this 5 - perhaps we can make this a constant with a name

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
jfreire-unity: Yeah, locally looks ok. Damn you, Github viewer 🤷

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -3907,6 +4035,128 @@ public enum Flags
...
+
+            public bool isCreated { get; }
+
+            public int lastEventHandledByAction { get; set; }
+
+            private int m_Length;
+
+            public ActionGroup(int capacity)
+            {
+                actionIndicies = (int*)UnsafeUtility.Malloc(capacity, 8, Allocator.Persistent);
```
lyndon-unity: Could we add a named variable / static to make this 8 a bit clearer that its alignment (and why 8 chosen)

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -3907,6 +4035,128 @@ public enum Flags
...
+
+                lastEventHandledByAction = -1;
+                actionCount = 0;
+                isCreated = true;
+            }
+
+            public void AddActionIndex(int actionIndex)
+            {
+                // return immediately if the actionIndex already exists in this group
+                for (var i = 0; i < actionCount; i++)
```
lyndon-unity: Could call the contains function from below ?

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
jfreire-unity: Thank you for explaining. Now it's super clear to me 👍

lyndon-unity (changes requested):
There are a couple of suggested changes and I'd like to discuss with design/product before landing on develop as the preference would be to expand further to see what could be done to handle the common case [CTRL/SHIFT/ALT]+KEY overriding KEY (without impacting WASD in two places).

Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
andrew-oc: Yes. Multiple action maps in the same input action asset work as intended. This also extends to singleton input actions created in code, as internally they get their own InputActionMap instances.

Assets/Tests/InputSystem/CoreTests_Actions.cs
andrew-oc: It isn't really relevant anymore because the composites in that test don't by default handle input events, so they never get consumed. Those paths are tested by other tests though.

Assets/Tests/InputSystem/CoreTests_Actions.cs
andrew-oc: Good call

Packages/com.unity.inputsystem/CHANGELOG.md
andrew-oc: Done. Introduced InputAction.EventHandledBehaviour enum.

Packages/com.unity.inputsystem/Documentation~/KnownLimitations.md
andrew-oc: Multiples maps in an asset would work. The limitation is only for maps that don't belong to an asset.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
andrew-oc: Done

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
andrew-oc: Me too!

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
andrew-oc: Done. It means two loops over all the controls, but is probably still a win for larger input action assets, and the code is neater.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
andrew-oc: Done

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
andrew-oc: Went one better. 8 was in fact an incorrect alignment!

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
andrew-oc: Done, thanks

andrew-oc:
> There are a couple of suggested changes and I'd like to discuss with design/product before landing on develop as the preference would be to expand further to see what could be done to handle the common case [CTRL/SHIFT/ALT]+KEY overriding KEY (without impacting WASD in two places).

This PR changes things so that WASD *is* never impacted anywhere. Event consumption is still off by default, so no inputs will go missing, while still allowing users to check for shortcut keys, but even if event consumption is turned on, Vector2 controls (and 1DAxis) don't handle events by default, so they still won't be impacted. The only time you'll see a WASD getting blocked (out of the box) is when event consumption is on and a modifier is used with W A S or D keys.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -4155,6 +4156,11 @@ public void Dispose()
 
                 actionIndicies = null;
             }
+
+            private static void* Allocate(int length)
```
lyndon-unity, line 4160: This is a nice tidy up and using AlignOf

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -4119,16 +4123,13 @@ public ActionGroup(int capacity)
             public void AddActionIndex(int actionIndex)
             {
                 // return immediately if the actionIndex already exists in this group
-                for (var i = 0; i < actionCount; i++)
-                {
-                    if (actionIndicies[i] == actionIndex)
-                        return;
-                }
+                if (Contains(actionIndex))
```
lyndon-unity, line 4126: Thanks

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -246,9 +249,10 @@ private void ComputeControlGroupingIfNecessary()
```
lyndon-unity, line 249: Wondering if this Resize could be moved out of the nested for loop, e.g. by. spinning round in advance to calculate max size.

lyndon-unity: Also wondering if the actionCountInGroup index overflow the tempActionGroupIndexes above.
Perhaps splitting the loop could allow a sanity check (or perhaps a sanity check could be added without splitting the loop)

lyndon-unity (commented):
Thanks for the updates.

I still have a query about a resize call in a nested for loop and a bounds check.

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1646) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1646) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1646) it.</sub>

jfreire-unity:
Closing as it will not be merged.

PR #1650, opened 2023-02-22
Fixing GetBestMidPoint returning value outside of valid range

jimon:
### Description

Fixes ISXB-405.

### Changes made

Fixed range computation and added asserts in debug.


andrew-oc (approved):
Nice fix! Not sure the #if DEBUG statements are necessary since Debug.Asserts are conditionally compiled.

PR #1653, opened 2023-03-03
FIX: Allow setting background behavior UI regardless of Run In Background (UUM-21955)

jfreire-unity:
### Description

This makes sure that `IgnoreFocus` can be set to Background Behavior on the Input Package settings UI, without having to have "Run In Background" enabled in the Player settings.

It complements #1521 where we made sure `IgnoreFocus` could work around the fact that there is no way to enable "Run in Background" on Android's Player Settings. This has an impact on XR where the headset tracking freezes when the application goes out of focus (e.g., the application is paused).

### Changes made

Removed the guard to enable the selection of the Background Behavior dropdown menu.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -16,6 +16,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where connecting a gamepad in the editor with certain settings will cause memory and performance to degrade ([case UUM-19480](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-19480)).
 - Fixed issue leading to a stack overflow crash during device initialization in `InsertControlBitRangeNode` (case ISXB-405).
 - Fixed the issue where saving and loading override bindings to JSON would set unassigned overrides (that were `null`) to assigned overrides (as an empty string `""`).
+- Allow Input Package settings menu to select Background Behavior regardless of the setting "Run in Background". This is useful for XR where the "Run in Background" isn't available ([case UUM-21955](https://issuetracker.unity3d.com/issues/quest-head-tracking-freezes-when-pausing-the-app)).
 
 ## [1.5.0] - 2023-01-24
 
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
@@ -98,11 +98,7 @@ namespace UnityEngine.InputSystem.Editor
                 EditorGUI.BeginChangeCheck();
 
                 EditorGUILayout.PropertyField(m_UpdateMode, m_UpdateModeContent);
-                var runInBackground = Application.runInBackground;
-                using (new EditorGUI.DisabledScope(!runInBackground))
-                    EditorGUILayout.PropertyField(m_BackgroundBehavior, m_BackgroundBehaviorContent);
-                if (!runInBackground)
-                    EditorGUILayout.HelpBox("Focus change behavior can only be changed if 'Run In Background' is enabled in Player Settings.", MessageType.Info);
+                EditorGUILayout.PropertyField(m_BackgroundBehavior, m_BackgroundBehaviorContent);
 
                 EditorGUILayout.Space();
                 EditorGUILayout.PropertyField(m_CompensateForScreenOrientation, m_CompensateForScreenOrientationContent);
@@ -293,7 +289,7 @@ namespace UnityEngine.InputSystem.Editor
                 + "'Reset And Disable All Devices' soft-resets and disables *all* devices while the application does not have focus. No device will receive input while the application "
                 + "is running in the background.\n"
                 + "'Ignore Focus' leaves all devices untouched when application focus changes. While running in the background, all input that is received is processed as if "
-                + "running in the foreground.");
+                + "running in the foreground.\n\n");
             m_EditorInputBehaviorInPlayModeContent = new GUIContent("Play Mode Input Behavior", "When in play mode, determines how focus of the Game View is handled with respect to input.\n\n"
                 + "'Pointers And Keyboards Respect Game View Focus' requires Game View focus only for pointers (mice, touch, etc.) and keyboards. Other devices will feed input to the game regardless "
                 + "of whether the Game View is focused or not. Note that this means that input on these devices is not visible in other EditorWindows.\n"


jamesmcgill (commented):
The PR description should have the case number at the end for Jira to pick it up
e.g. "FIX: Allow setting background behavior UI regardless of Run In Background (UUM-21955)"

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where connecting a gamepad in the editor with certain settings will cause memory and performance to degrade ([case UUM-19480](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-19480)).
 - Fixed issue leading to a stack overflow crash during device initialization in `InsertControlBitRangeNode` (case ISXB-405).
 - Fixed the issue where saving and loading override bindings to JSON would set unassigned overrides (that were `null`) to assigned overrides (as an empty string `""`).
+- Allow Input Package settings menu to select Background Behavior regardless of the setting "Run in Background". This is useful for XR where the "Run in Background" isn't available.
```
jamesmcgill: Should add the case number (and issuetracker link if available) here
Also the PR description should have the case number at the end for Jira to pick it up
e.g. "FIX: Allow setting background behavior UI regardless of Run In Background (UUM-21955)"

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -98,11 +98,8 @@ public override void OnGUI(string searchContext)
                 EditorGUI.BeginChangeCheck();
 
                 EditorGUILayout.PropertyField(m_UpdateMode, m_UpdateModeContent);
-                var runInBackground = Application.runInBackground;
-                using (new EditorGUI.DisabledScope(!runInBackground))
-                    EditorGUILayout.PropertyField(m_BackgroundBehavior, m_BackgroundBehaviorContent);
-                if (!runInBackground)
-                    EditorGUILayout.HelpBox("Focus change behavior can only be changed if 'Run In Background' is enabled in Player Settings.", MessageType.Info);
+                EditorGUILayout.PropertyField(m_BackgroundBehavior, m_BackgroundBehaviorContent);
+                EditorGUILayout.HelpBox("Hover the mouse on this field to learn more about each selected setting.", MessageType.Info);
```
jamesmcgill: I would remove the Helpbox now as it no longer serves any meaningful purpose

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
```diff
@@ -293,7 +290,9 @@ private void InitializeWithCurrentSettings()
                 + "'Reset And Disable All Devices' soft-resets and disables *all* devices while the application does not have focus. No device will receive input while the application "
                 + "is running in the background.\n"
                 + "'Ignore Focus' leaves all devices untouched when application focus changes. While running in the background, all input that is received is processed as if "
-                + "running in the foreground.");
+                + "running in the foreground.\n\n"
+                + "If runBackground is false:\n"
+                + "Settings are ignored and devices and actions remain untouched while the application does not have focus. When focus is gained, all devices receive a sync request.");
```
jamesmcgill: I would restore the tooltip as it is still correct. The additional information about sync requests isn't really interesting enough to add more to this (already) huge wall of text.

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs
jfreire-unity: But the help box text: "Focus change behavior can only be changed if 'Run In Background' is enabled in Player Settings." no longer makes sense, right? Focus change behavior (also known as background behavior) can now be changed in the Input Settings settings even if "Run in Background" is not enabled in Player Settings. 

@jamesmcgill Would this make more sense?
"These settings are only in effect if `Run in Background` is enabled".

jfreire-unity: The point of the text above is to add it do the tooltip of your previous comment.
```
   EditorGUILayout.HelpBox("These settings are only in effect if Run in Background is enabled.", MessageType.Info);
   ```

Comment (jfreire-unity) [reply Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs]: Actually, you are right. I'll remove both additional tests and the help box isn't meaningful.

**Comment by jamesmcgill** [comment]
Sorry @jfreire-unity, I've been looking through the [docs ](https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/Documentation~/Devices.md#background-and-focus-change-behavior) trying to understand this better.

I'm now thinking that if we make this UI available it will be even more confusing for Users (perhaps this is what you were trying to explain to me before?).
1)  "On platforms such as iOS and Android, that do not support running Unity applications in the background, this is the only supported behavior."
2) Also on the matrix image there is no difference mentioned between the 3 modes for when `runInBackground=false`
Therefore why would Users pick anything in the UI when the result should be the same.


The change I made was a bit of a hack to unblock the headset issue until the main issue on the platform side gets solved. Not sure we should rework the design and docs for this. Is it strictly required to unlock the UI or can they not work programmatically?


Options could be:
1) Continue and unlock the UI: I'm concerned `ResetAndDisableNonBackgrounDevices` and `ResetAndDisableAllDevices` have no real differences and `IgnoreFocus` has only one small difference, so we present the User with a lot of options which don't do much and it will be confusing to explain (i.e. the tooltip). These are more design and documentation issues.

2) Try having all 3 modes uniform by having them all process input. Then the UI can remain locked and the documentation would be still correct and simple. Not sure the implications of that though. Would be a dangerous change as the common default case would have a behaviour change.

3) Leave the UI locked as-is.

File: Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsProvider.cs [hunk 4]
```diff
@@ -293,7 +289,7 @@ private void InitializeWithCurrentSettings()
                 + "'Reset And Disable All Devices' soft-resets and disables *all* devices while the application does not have focus. No device will receive input while the application "
                 + "is running in the background.\n"
                 + "'Ignore Focus' leaves all devices untouched when application focus changes. While running in the background, all input that is received is processed as if "
-                + "running in the foreground.");
+                + "running in the foreground.\n\n");
```
jamesmcgill, line 292: Some leftover newlines here

jamesmcgill:
I'm rambling a bit there. But to be clear, I'm fine to merge this PR if they need the option on the UI available to get unblocked. It's a just a minor design issue that I'm pointing out.

jfreire-unity:
Talked with James on Slack and agreed to ask the XR team if they could use the "hack" from #1521 programmatically.
The goal is to close this PR to avoid the risk of exposing confusing menus to users.

jfreire-unity:
Closing it for now.

PR #1648, opened 2023-02-17
Fix device disconnects being missed on console platforms while running in the background

andrew-oc:
### Description

Certain platforms ignore the runInBackground flag. If a device disconnect event occurs while the app is in the background on one of those platforms, the input system will discard the event, and the device instance will stay in memory.

### Changes made

Forced the input system to ignore the runInBackground flag in managed code on the troublesome platforms.

### Checklist

Before review:

- [x] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.

PR #1662, opened 2023-03-22
FIX: Allow InputSystem class to set runtime runInBackground field

jfreire-unity:
### Description

Some platforms' background behavior isn't reflected in `Application.runInBackground`. This PR exposes `InputSystem.runInBackground` to be called by those particular platforms to tell the Input System that the platform runs in the background.

It should be considered for "internal use" only even though it is part of the public API.

Context: This is the alternative for the rejected PR #1648. Specific platforms can call this API as they wish, otherwise, we rely on `Application.runInBackground` set in the Editor settings for platforms that allow being run in the background.

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 using UnityEngine.InputSystem.Haptics;
 using Unity.Collections.LowLevel.Unsafe;
 using UnityEngine.InputSystem.Controls;
@@ -74,10 +75,12 @@ namespace UnityEngine.InputSystem
     /// be called on the main thread. However, select APIs like <see cref="QueueEvent"/> can be
     /// called from threads. Where this is the case, it is stated in the documentation.
     /// </remarks>
+
     [SuppressMessage("Microsoft.Naming", "CA1724:TypeNamesShouldNotMatchNamespaces", Justification = "Options for namespaces are limited due to the legacy input class. Agreed on this as the least bad solution.")]
 #if UNITY_EDITOR
     [InitializeOnLoad]
 #endif
+
     public static partial class InputSystem
     {
         #region Layouts
@@ -3280,7 +3283,7 @@ namespace UnityEngine.InputSystem
         public static Version version => new Version(kAssemblyVersion);
 
         /// <summary>
-        /// Allows setting the player to run in the background.
+        /// Property for internal use that allows setting the player to run in the background.
         /// </summary>
         /// <remarks>
         /// Some platforms don't care about <see cref="Application.runInBackground"/> and for those we need to
--- a/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
@@ -3279,6 +3279,20 @@ namespace UnityEngine.InputSystem
         /// <value>Current version of the input system.</value>
         public static Version version => new Version(kAssemblyVersion);
 
+        /// <summary>
+        /// Allows setting the player to run in the background.
+        /// </summary>
+        /// <remarks>
+        /// Some platforms don't care about <see cref="Application.runInBackground"/> and for those we need to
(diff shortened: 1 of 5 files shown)


Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3192,6 +3192,25 @@ public static int ListEnabledActions(List<InputAction> actions)
...
+        /// Some platforms don't care about <see cref="Application.runInBackground"/> and for those we need to
+        /// enable them manually through this method.
+        /// </remarks>
+        /// <param name="value">The boolean value to set to <see cref="NativeInputRuntime.runInBackground"/></param>
+        public static void SetPlatformRunInBackgroundSetting(bool value)
+        {
+                s_Manager.m_Runtime.runInBackground = value;
+        }
+        
+        //TODO remove this, used for debugging only for now
```
lyndon-unity: Remove this if no longer needed. Or make 'internal' as you can setup so tests can get to that.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -227,11 +227,17 @@ public float pollingFrequency
...
-        // TODO: Add more platforms here as they're discovered.
-        Application.platform == RuntimePlatform.PS5;
+        public bool runInBackground
+        {
+            get =>
+                Application.runInBackground ||
+                // certain platforms ignore the runInBackground flag and always run. Make sure we're
+                // not running on one of those and set the values when running on specific platforms.
+                m_platformSpecificRunInBackground;
+            set => m_platformSpecificRunInBackground = value;
```
lyndon-unity: I'm wondering if we could do this in a way that only the platform specific input system packages could override the value and not general users. To prevent users interfering and causing issues. 

I'm also wondering if we can generalise it to a method for applying other settings that might be platform specific.

At the moment the platform specific input system packages seem to only use the public APIs from the input system such as InputSystem.RegisterLayout

lyndon-unity: That said I see InputSystem.s_Manager is internal so this is probably limited to the input system package. I see that the ios package references InputSystem.s_Manager.m_Runtime. However iOS plugin is the main package

https://github.com/Unity-Technologies/InputSystem/blob/6da074044b05034ebebb39c820aecd10c557fa31/Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSupport.cs

Can this definitely be used by the platform specific packages ?

lyndon-unity (commented):
Seems reasonable but I don't know all the connection between platform packages and the main package to be sure this is the best way to handle it.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
lyndon-unity: Oh I see now that SetPlatformRunInBackgroundSetting is the public API to access this field that both packages and users can set.

jfreire-unity: > I'm wondering if we could do this in a way that only the platform specific input system packages could override the value and not general users. To prevent users interfering and causing issues.

After some investigation, there's no easy way of doing with this without adding some more processes and proxy names/packages. I'll elaborate on Slack.

> Can this definitely be used by the platform-specific packages?

This was validated that it worked on Xbox.

jfreire-unity:
After going through all the options, I decided to go with the Public API direction for now. 
There's no "Plugin" concept in the InputSystem where a setting like this should live and it would involve a bigger effort at the moment. This should be the direction to go next.

martinh-unity (approved):

unity-cwells (approved):

MattWUnity3d (approved):

lyndon-unity (approved):

BraveSirAndrew:
@jfreire-unity Yo. Sorry to butt in but this is a breaking change for anyone who is implementing the IInputRuntime interface. Just saying. Also hi!

PR #1664, opened 2023-03-29
ISXB-449 fix for turkish letter i

ritamerkl:
### Description

Peter Pimley observes this bug, he suggested someone else of the team should look over it because there could be more similar cases where bugs like this appear. So it would be great if the reviewer could validate that this solution is appropriate
for all cases that may come up.

### Changes made

FIXED: Changed char comparison to use ToLower(CultureInfo) or ToLowerInvariant() to avoid problems with special characters like the turkish letter i. 
ADDED: Test for turkish culture 
[Link to the Bug](https://jira.unity3d.com/browse/ISXB-449)
[Link to the user Ticket](https://unity3d.atlassian.net/servicedesk/customer/portal/2/IN-36120)

### Notes

Due to the unfortunate case that James is sick the PR gets merged without his approval. 

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1664) <br/>All committers have signed the CLA.

peter-pimley-unity:
There are many places in the InputSystem source code where either `string.ToLower` or `char.ToLower` (or ToUpper) are called.  I guess that many (perhaps _all_) of them are done with the end goal of performing case-insensitive comparisons.  For example [here](https://github.com/Unity-Technologies/InputSystem/blob/99a451a44d6a3a2bb6d8cd7d6c79786a9b172045/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlLayout.cs#L1676-L1678) is one that attempts to do a case-insensitive match against the string `"ignore"`.  Will that suffer from the same problem?

Ideally we would do an audit of all occurances of ToLower or ToUpper on strings and chars in the source, and establish whether they should be culture-aware.  I would guess that nearly all of them should _not_ be.

There are some tips for string comparisons here: https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

jimon (approved):

jimon:
There are definitely a lot of places and likely just this fix is not sufficient, but I'm struggling to find a good strategy how to fix all of them, maybe just by using IDE "find all" tool?

peter-pimley-unity:
Yes I think simply a search in Visual Studio etc. for "ToUpper" and "ToLower".  The number of occurrences is high enough that it's more than just a 5 minute job but not so large as to be infeasible.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
```diff
@@ -600,7 +600,7 @@ internal string DetermineName()
                             return "hat";
                         var text = ((GenericDesktop)usage).ToString();
                         // Lower-case first letter.
-                        text = char.ToLowerInvariant(text[0]) + text.Substring(1);
+                        text = char.ToUpperInvariant(text[0]) + text.Substring(1).ToUpperInvariant();
```
jamesmcgill: I think based on the comment this part was probably correct before. (Just wanted the first character to be uppercase).

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -558,15 +558,15 @@ private static bool StringMatches(Substring str, InternedString matchTo)
                         return true; // Wildcard at end of string so rest is matched.
 
                     ++posInStr;
-                    nextChar = char.ToLower(str[posInStr]);
+                    nextChar = char.ToUpperInvariant(str[posInStr]);
 
-                    while (posInMatchTo < matchToLength && matchToLowerCase[posInMatchTo] != nextChar)
+                    while (posInMatchTo < matchToLength && char.ToUpperInvariant(matchToLowerCase[posInMatchTo]) != nextChar)
```
jamesmcgill: This `matchToLowerCase` looks suspicious here.
I think probably it needs renamed to `matchToUpperCase` and have it's contents changed to uppercase?
Then you don't need to uppercase it again here on this line.

Just need to be careful other code is relying on `matchToLowerCase` actually being lower case though.

jamesmcgill (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
ritamerkl: your right, there I was too zealously changing all ToLower

jimon (approved):

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -539,7 +539,7 @@ private static bool StringMatches(Substring str, InternedString matchTo)
             // Can't compare lengths here because str may contain wildcards and
             // thus be shorter than matchTo and still match.
 
-            var matchToLowerCase = matchTo.ToLower();
+            var matchToStringUpper = matchTo.ToString().ToUpperInvariant();
```
jamesmcgill: The `ToString()` will cost a GC allocation as it needs to create a new temporary string.

I looked at `matchTo` and why you couldn't just to `ToUpperInvariant` and I found it's because it's an `InternedString`. The docs say this is a special string class designed exactly for what you are doing here, case-insensitive comparisons.

However that class looks like it needs to be updated too as it might also be broken for turkish letters. Maybe it makes sense for it to store internally in uppercase instead and then you wouldn't need `ToString()` here either. I'm not sure how much `InternedString` is used, so it might be quite a bit of work.

jamesmcgill (dismissed):

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
ritamerkl: this is a lot of work. changing InternedString is a big thing because it entails changes at many spots

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -558,15 +558,15 @@ private static bool StringMatches(Substring str, InternedString matchTo)
                         return true; // Wildcard at end of string so rest is matched.
 
                     ++posInStr;
-                    nextChar = char.ToLower(str[posInStr]);
```
lyndon-unity, line 561: Could a fix be to convert to pass CultureInfo.InvariantCulture

E.g. 
nextChar =  char.ToLower(str[posInStr], CultureInfo.InvariantCulture);
or (if version allowed) :
nextChar =  char.ToLowerInvariant(str[posInStr]);

and a few lines down

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -558,15 +558,15 @@ private static bool StringMatches(Substring str, InternedString matchTo)
...
+                    nextChar = char.ToUpperInvariant(str[posInStr]);
 
-                    while (posInMatchTo < matchToLength && matchToLowerCase[posInMatchTo] != nextChar)
+                    while (posInMatchTo < matchToLength && matchToStringUpper[posInMatchTo] != nextChar)
                         ++posInMatchTo;
 
                     if (posInMatchTo == matchToLength)
                         return false; // Matched all the way to end of matchTo but there's more in str after the wildcard.
                 }
-                else if (char.ToLower(nextChar) != matchToLowerCase[posInMatchTo])
```
lyndon-unity, line 569: as above comment

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -1107,7 +1107,7 @@ private static bool MatchPathComponent(string component, string path, ref int in
                 }
 
                 var charInComponent = component[indexInComponent];
-                if (charInComponent == nextCharInPath || char.ToLower(charInComponent) == char.ToLower(nextCharInPath))
```
lyndon-unity, line 1110: and again as above

Packages/com.unity.inputsystem/InputSystem/Utilities/StringHelpers.cs
```diff
@@ -418,7 +418,7 @@ public static string Join<TValue>(IEnumerable<TValue> values, string separator)
                             var first = firstList[startIndexInFirst + i];
                             var second = secondList[startIndexInSecond + i];
 
-                            if (char.ToLower(first) != char.ToLower(second))
```
lyndon-unity, line 421: and one more as above

lyndon-unity (changes requested):
Added suggestions on comparison without moving to uppercase or altering InternedString

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
jamesmcgill: `ToUpperInvariant` is recommended here: https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings
I believe going in the other direction (lower) creates issues.

lyndon-unity: Thanks James. I missed that comment in the Microsoft documents. I mainly suggested this as it seemed we were missing the invariantCulture altogether but agree we may need the upper case for a more complete fix.

jamesmcgill: `ToUpperInvariant` is using Invariant Culture rules so that should be enough without needing to pass `CultureInfo`.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
lyndon-unity: I'm wondering why InternedString returns str.toString rather than str.m_StringOriginalCase
The later could avoid the GC allocation ?

https://sourcegraph.com/github.com/Unity-Technologies/InputSystem@30277637e5e9986c8cad3be939ad16185ba6114a/-/blob/Packages/com.unity.inputsystem/InputSystem/Utilities/InternedString.cs?L27:19-27:33

public static implicit operator string(InternedString str)

lyndon-unity: But agree switching to uppercase internals in InternedString seems the right thing to do

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
ritamerkl: @lyndon-unity @jamesmcgill I just updated whole branch to only use culture invariant char comparisons, this fixes the bug without changing to ToUpperInvariant(). I would suggest closing the ticket (as the bug is fixed) at this state and creating a new refactoring ticket for clean up the string ToLowerInvariant -> ToUpperInvariant situation. Maybe it even makes sense to do it in a bigger scale where we look at all our string storing and comparison situation

lyndon-unity (commented):
I'm happy with this as a fix for the immediate issue and then raising a separate task (not bug) on the backlog to refactor wider to use the upper case version which is a better long term fix but a bigger change which we don't have time for this sprint.

However there is a test failure outstanding

 IntegrationTests.Integration_CanSendAndReceiveEvents

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
lyndon-unity: I agree this is a good step and fixes the current but. We have captured the future step in a task here:
https://jira.unity3d.com/browse/ISX-1380

lyndon-unity (approved):
Overall I'm happy to proceed with this fix but ideally would like to see a new test case added to the input system test suite to make sure we do not regress with future changes.

ritamerkl:
> Overall I'm happy to proceed with this fix but ideally would like to see a new test case added to the input system test suite to make sure we do not regress with future changes.

After writing the test for this issue I came across the "real problem":
The previous ToLower() method converted the normal big I into a small turkish dotless ı. I found this info on the microsoft page: 

> "The casing operation that results from calling the ToLower() method takes the casing conventions of the current culture into account." See [here](https://learn.microsoft.com/en-us/dotnet/api/system.string.tolower?view=net-7.0)

So only by using the big 'I' the wrong conversion happens. A normal 'i' exists as well in the turkish language. The two big 'I's' do not differ from each other, that's where the problem arises. 
The ToLower() method with the CultureInfo property prevents this, so we are all good now.

lyndon-unity (approved):

PR #1665, opened 2023-03-29
ISXB-358  wildcards not working

ritamerkl:
### Description

Fixes the wildcard not working bug

### Changes made

Changed: 
BindingMasks get compared to Bindings with another method (StringMatches)
Fixed:
Wildcards working now
Added:
tests for wildcards

cases checked: 
* sign at the beginning
* sign at the end

### Notes

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
```diff
@@ -841,6 +843,18 @@ public bool Matches(InputBinding binding)
...
             return Matches(ref binding);
         }
+        
+        internal bool MatchesMask(ref InputBinding binding)
+        {
+            var appendix = path.Substring(path.Length-2,path.Length-(path.Length-2));
+            if (appendix == wildcard)
+            {
+                var before = path.Substring(0, path.Length - 2);
+                if (binding.path != null && binding.path.Contains(before))
```
lyndon-unity: Is Contains a bit too loose. Should this be StartsWith ?

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
```diff
@@ -841,6 +843,18 @@ public bool Matches(InputBinding binding)
         {
             return Matches(ref binding);
         }
+        
+        internal bool MatchesMask(ref InputBinding binding)
+        {
+            var appendix = path.Substring(path.Length-2,path.Length-(path.Length-2));
+            if (appendix == wildcard)
```
lyndon-unity: I'm concerned this is very specific to /* and not the wider wild card support listed in documentation here:

https://docs.unity3d.com/Packages/com.unity.inputsystem@1.0/manual/ActionBindings.html 
E.g <Gamepad>/button*

Is the wider case handled elsewhere?

lyndon-unity (changes requested):
Looks like various tests are failing.

I think at minimum MatchesMask needs a path!=null check at the start

[Error] NullReferenceException while resolving binding 'Move:<Gamepad>/leftStick[Gamepad]' in action map 'InGameHintsActions (UnityEngine.InputSystem.InputActionAsset):Gameplay''.

https://unity-ci.cds.internal.unity3d.com/job/22912980


I'd also like to see more detail in the PR description about cases checked.
It would be useful to also add a new test for the specific case we are fixing in this PR

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
ritamerkl: Your right, now that I look at the documentation I doubt that this is sufficient. I just saw that the StringMatches() Method in InputControlPath.cs does exactly that (Line 534).
But it is only used in some Tests if I don't miss something. Maybe it is worth to refactor that for this use? @lyndon-unity

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
ritamerkl: see comment below

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
ritamerkl: see my new commit - I also saw there are cases where there can be "**" signs to indicate a wildcard. I think this case is not covered yet, it is also not part of the documentaiton, but maybe to keep it in mind it makes sense to create another ticket for that?

lyndon-unity:
There are several test failures which seems to highlight problems with this change.

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
```diff
@@ -842,6 +842,13 @@ public bool Matches(InputBinding binding)
             return Matches(ref binding);
         }
 
+        internal bool MatchesMask(ref InputBinding binding)
+        {
+            if (path != null)
```
lyndon-unity, line 847: The matches function matches first on name, then path, then others like actions and groups.
It seems if path is non null we are skipping those other match checks.

Is this valid - is only one of path,name,action of group set or do we actually need the fallback cases?

I.e. perhaps this should be 

    if (path != null)
    {
        if (InputControlPath.BindingPathMatchesMask(path, binding.path))
            return true;
    }
    return Matches(ref binding, MatchOptions.EmptyGroupMatchesAny);

Assets/Tests/InputSystem/CoreTests_Controls.cs
```diff
@@ -1119,6 +1119,34 @@ public void Controls_CanFindControlsUsingWildcards_InMiddleOfNames()
...
+        Assert.That(InputControlPath.BindingPathMatchesMask("*/d", "<Keyboard>/e"), Is.False);
+        Assert.That(InputControlPath.BindingPathMatchesMask("*/d", "<Keyboard>/d"), Is.True);
+        Assert.That(InputControlPath.BindingPathMatchesMask("*/leftButton", "<Mouse>/leftButton"), Is.True);
+    }
+
+    [Test]
+    [Category("Controls")]
+    public void Controls_MatchingPathToMaskWithWildcardAtEnd()
+    {
+        Assert.That(InputControlPath.BindingPathMatchesMask("<Keyboard>/*", "<Keyboard>/e"), Is.True);
```
lyndon-unity, line 1274: Could we have a test that checks the higher level MatchesMask function too ?
I.e. covers the flow where name, action or group is non null

lyndon-unity (commented):
Looks reasonable for the specific cases but I've left a query about the non path cases.

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
ritamerkl: yes, true. I think till now the binding mask is only used with a path, but if it changes in future that will be a safer way

lyndon-unity: These test seem to suggest it works with groups
Actions_CanRecordActions_AndReadTheDataEvenIfBindingsHaveChanged
Actions_CanMaskOutBindingsByBindingGroup_OnAction_WhenEnabled

action.bindingMask = new InputBinding {groups = "B"};

ritamerkl: I never saw this tests before. You're right.

Assets/Tests/InputSystem/CoreTests_Controls.cs
```diff
@@ -1119,6 +1119,63 @@ public void Controls_CanFindControlsUsingWildcards_InMiddleOfNames()
...
+    public void Controls_MatchingBindingToMask()
+    {
+        var mask = new InputBinding("<Keyboard>/e");
+        var binding = new InputBinding("<Keyboard>/e");
+        Assert.That(mask.MatchesMask(ref binding), Is.True);
+        mask.path = "<Keyboard>/d";
+        Assert.That(mask.MatchesMask(ref binding), Is.False);
+        mask.path = "<Keyboard>/*";
+        Assert.That(mask.MatchesMask(ref binding), Is.True);
+        mask.path = null;
```
lyndon-unity: I'd be tempted to split some out into a new tests to improve readability

This is a good guide on test naming
https://internaldocs.unity.com/automation/Guidelines/naming/
UnitOfWork_WhenStateUnderTest_ProducesExpectedBehavior()

Perhaps some like:
Controls_BindingWithoutMask_Matches
Controls_BindingMaskWithSpecificKey_IgnoresOtherKey
Controls_BindingMaskWithWildcard_Matches
Controls_BindingMaskWithNullPath_Ignored
Controls_BindingMaskWithNullName_Ignored
Controls_BindingMaskWithMousePath_FiltersOutKeyboard

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
```diff
@@ -842,6 +842,14 @@ public bool Matches(InputBinding binding)
...
```
lyndon-unity, line 847: Minor point but with nested ifs I usually add braces around the code within the outer if, just to make it clear and avoid errors if someone adds to the inner if.

However in this case you could also collapse into a single if (path != null && InputControlPath....)

Minor point and not needing to hold up PR.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -481,6 +481,12 @@ private static string FindControlLayoutRecursive(ref PathParser parser, InputCon
             return currentResult;
         }
 
+        internal static bool BindingPathMatchesMask(string mask, string bindingPath)
+        {
+            var bindingPathString = new InternedString(bindingPath);
```
lyndon-unity, line 486: I'm a little worried about the InternedString creation here which is creating a lowercase variant of the bindingPath string which will be doing a GC allocation 

This BindingPathMatchesMask is called inside a loop over all bindings so this will be generating a lot of garbage.

Perhaps StringMatches could be altered to take a string variant. However I think it uses InternedString for the speed of the internal checks (as we've seen in the separate PR about turkish letter support)

I'm also seeing that StringMatches says the second 'matchTo' parameter may not contain wildcards and I think the bindingPath is allowed to contain wildcards.

lyndon-unity (changes requested):
So overall I'm actually now nervous about this change and think we need to do more work on performance and robustness.

I.e. we'd need to add some test cases where binding path contains wildcards.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
lyndon-unity: Yes binding path can contain wildcards
https://docs.unity3d.com/Packages/com.unity.inputsystem@1.5/manual/ActionBindings.html

> A Binding path can also contain wildcards, such as <Gamepad>/button*

PR #1667, opened 2023-04-05
Add dispatch of navigation events for InputForUI

jfreire-unity:
### Description

Dispatches `NavigationEvents` such as `Left`, `Right`, `Up`, `Down` for both Mouse and Keyboard, as they are part of the `_moveAction` setup by the default input actions asset.
`Next` and `Previous` currently are only setup for Keyboard, as there is not default action that has `tab` bindings. A new action was created for this.

### Notes

If this approach works and is good enough for now, I'll refactor the code a bit to avoid having less repetition for `NavigationEvents` in both `InputManagerProvider` (not part of this repository) and `InputSystemProvider`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
@@ -94,8 +94,9 @@ namespace InputSystem.Plugins.InputForUI
         {
             if (ctx.control.device is Keyboard)
             {
+                //TODO repeat rate
                 var keyboard = ctx.control.device as Keyboard;
-                EventProvider.Dispatch(Event.From(new NavigationEvent
+                DispatchFromCallback(Event.From(new NavigationEvent
                 {
                     type = NavigationEvent.Type.Move,
                     direction = keyboard.shiftKey.isPressed ? NavigationEvent.Direction.Previous : NavigationEvent.Direction.Next,
@@ -365,7 +366,6 @@ namespace InputSystem.Plugins.InputForUI
 
         private void OnSubmitPerformed(InputAction.CallbackContext ctx)
         {
-            // TODO repeat rate
             DispatchFromCallback(Event.From(new NavigationEvent
             {
                 type = NavigationEvent.Type.Submit,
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -62,30 +66,45 @@ private static void Bootstrap()
...
 
+        private void OnNextPreviousPerformed(InputAction.CallbackContext ctx)
+        {
+            if (ctx.control.device is Keyboard)
+            {
+                var keyboard = ctx.control.device as Keyboard;
+                EventProvider.Dispatch(Event.From(new NavigationEvent
+                {
+                    type = NavigationEvent.Type.Move,
+                    direction = keyboard.shiftKey.isPressed ? NavigationEvent.Direction.Previous : NavigationEvent.Direction.Next,
```
jimon, line 102: Hm, smart!

jimon: Should these have repeat rate as well?
Also don't use EventProvider.Dispatch here, I've added DispatchFromCallback so we can sort events later as order of callbacks is not really defined

jfreire-unity: Ah, I see. I will change it.

About repetition, I followed the IM provider's current implementation, actually.

@benoit-alain for which events do we need repetition handling? We only have for directional `NavigationEvents` => Left, Right, Up, Down, at the moment.

benoit-alain: For Next and Previous too it would make sense, since they're normally tied to KeyDown events which also repeat.

benoit-alain (approved):
Seems good, we can continue adding on this if something's missing (next/previous repeat?).

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
jfreire-unity: Thanks @benoit-alain , will implement it in a follow up PR.

PR #1670, opened 2023-04-12
InputForUI: Add repetition for Next/Previous navigation events

jfreire-unity:
### Description

This PR allows dispatching repeated Next/Previous navigation events while a control is still pressed.

### Changes made

It removes the callbacks when the `_nextPreviousAction` is performed. 
The logic is now in `DirectionNavigation()` method running in every `Update()` call.

### Notes

⚠️ Depends on the latest commits from `trunk`
At the moment, it only detects Shift or Shift+Tab as Next/Previous bindings (same as #1667). A configuration to set up these bindings for the Input System is necessary. The Input Manager provides configuration for this.


jfreire-unity:
@benoit-alain let me know if this works for you when using the Input System integration.

PR #1673, opened 2023-04-18
Expose the display index via the Touch object.

TonyAbouHaidar:
### Description

Currently, the display index can only be retrieved from the primary touch which doesn't reflect the correct display index for each touch in case touches are triggered by two different touch devices.  With this change, the display index of each individual touch can be accurately retrieved.

### Changes made

Added a new field to the Touch struct to expose the index of the display that was touched.

### Notes

Currently, there is no way to accurately retrieve the index of the touched display from the Touch object. The primary touch does expose the display index but since this is set once upon the first touch, touching two displays at the same time will end up storing the index of the first display touched in the primary touch.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


stefanunity:
landed via https://github.com/Unity-Technologies/InputSystem/pull/1677

PR #1672, opened 2023-04-14
Adding generalized support for controller usages with the binding path dropdown. Added optional info blurb on the binding paths covered

RogPodge:
### Description

This PR augments the binding path dropdown, allowing users to add control usages for specific devices.

This PR also adds a "Show Matching Paths" checkbox, which shows the users exactly which specific registered layout paths their current binding path will match with.

The goal of this PR is to address some grievances found in this thread: https://forum.unity.com/threads/openxr-primary2daxisclick-not-working.1120864/

https://user-images.githubusercontent.com/39840334/232162490-588f48e2-d521-4215-ab6a-3a8fa93b4554.mp4

https://user-images.githubusercontent.com/39840334/232162498-4089680c-bd06-4be6-a643-5c1bae0b7b29.mp4

### Changes made
Updated the InputControlPickerDropdown to allow for control usages to be selected and added to the paths in after specifying a desired device.

Augmented the UsageDropdown to be usable in more contexts.

Added unit tests for the new InputControlPath function.

Added new option to the InputBindingsPropertiesView.

### Notes

_Please write down any additional notes, remove the section if not applicable._

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #1676, opened 2023-04-27
Adding generalized support for controller usages with the binding path dropdown. Added optional info blurb on the binding paths covered

RogPodge:
Migrated from #1672 

### Description

This PR augments the binding path dropdown, allowing users to add control usages for specific devices.

This PR also adds a "Show Matching Paths" checkbox, which shows the users exactly which specific registered layout paths their current binding path will match with.

The goal of this PR is to address some grievances found in this thread: https://forum.unity.com/threads/openxr-primary2daxisclick-not-working.1120864/

https://user-images.githubusercontent.com/39840334/232162490-588f48e2-d521-4215-ab6a-3a8fa93b4554.mp4

https://user-images.githubusercontent.com/39840334/232162498-4089680c-bd06-4be6-a643-5c1bae0b7b29.mp4

### Changes made
Updated the InputControlPickerDropdown to allow for control usages to be selected and added to the paths in after specifying a desired device.

Augmented the UsageDropdown to be usable in more contexts.

Added unit tests for the new InputControlPath function.

Added new option to the InputBindingsPropertiesView.

### Notes

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Rene-Damm:
Using a foldout instead of a checkbox probably makes for a sightly nicer UI. Also, "Matching Controls" maybe slightly clearly naming than "Matching Paths".

RogPodge:
image

Tweaked the UI to the PR suggestions.

@reviewers this PR makes some changes to the editor display code, and potentially incurs some new performance overhead. Are there any tests or things we can run to see if the performance impacts are acceptable?

Billreyn:
Hi there, finally got time to look at this. 

So, as I understand it, the idea is to make it clear what the general purpose abstractions in the input system actually bind to, right? 

It's probably a good idea to add this - I thought about it before too. 

My idea for how to solve it was to do it an a more visual way though.

Question: Is the flyout open or closed by default? Probably I think it should be closed, otherwise it's quite overwhelming. 

And, is it displayed when you bind an action directly to an input control? Does it only show if you use an abstraction like GamePad? If not, it probably should do that I think - I mean only be visible when the user is binding to a device abstraction.

Billreyn:
So, I looked at this more. 

I suggest solving this problem in a slightly different way. These bindings are like 'derived bindings'. So I think it fits better in the binding list, closed by default and only visible when users pick an abstraction (usage or gamepad etc) 

Screenshot 2023-05-02 at 09 51 36

Billreyn:
Also, as you can see in the screenshot above, I am wondering if we really need to show the derived bindings different between Xbox Controller, Xbox Controller iOS, Xbox Controller Android? They always seem to be the same anyway, so it seems like redundant information?

lyndon-unity:
Note also that we are working on a UI Toolkit implementation of this UX so we need to add support into that implementation too. The current work is in develop and behind a flag. See https://docs.google.com/document/d/1ot1-Cr85TxpZGudm-VXvE4-C2YAATqiQcYkc_ww76eI/edit#heading=h.aasuoa63b4nd 

Specifically this feature flag needs to be set to true to see the (incomplete) UI TK version InputFeatureNames.kUseUIToolkitEditor

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -721,6 +721,39 @@ public static bool Matches(string expected, InputControl control)
             return MatchesRecursive(ref parser, control);
         }
 
+        internal static bool MatchControlComponent(ref ParsedPathComponent expectedControlComponent, ref InputControlLayout.ControlItem controlItem)
```
vrdave-unity: Since the expectedControlComponent is not modified, but used as an input reference, you can change from a `ref` to an `in` and make the intention more clear. 
```suggestion
        internal static bool MatchControlComponent(in ParsedPathComponent expectedControlComponent, ref InputControlLayout.ControlItem controlItem)
```

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs
```diff
@@ -102,11 +104,109 @@ protected override void DrawGeneralProperties()
...
+                    var rootLayout = EditorInputControlLayoutCache.allLayouts.FirstOrDefault(x => x.isDeviceLayout && !x.isOverride && !x.hideInUI && x.name == deviceLayout);
+                    if (rootLayout != null)
+                    {
+                        for (int i = 0; i < rootLayout.m_Controls.Length; i++)
+                        {
+                            if (InputControlPath.MatchControlComponent(ref parsedPath[1], ref rootLayout.m_Controls[i]))
+                            {
+                                EditorGUILayout.LabelField($"{rootLayout.displayName}/{rootLayout.m_Controls[i].displayName}");
+                                matchExists = true;
+                                continue;
```
vrdave-unity: Can this `continue` be removed? Or is this supposed to be a `break`?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs
```diff
@@ -102,11 +104,109 @@ protected override void DrawGeneralProperties()
...
+            {
+                foreach (var childLayout in matchedChildLayouts)
+                {
+                    for (int i = 0; i < childLayout.m_Controls.Length; i++)
+                    {
+                        if (InputControlPath.MatchControlComponent(ref pathControlComponent, ref childLayout.m_Controls[i]))
+                        {
+                            EditorGUILayout.LabelField($"{childLayout.displayName}/{childLayout.m_Controls[i].displayName}");
+                            matchExists = true;
+                            continue;
```
vrdave-unity: Same thing here.

vrdave-unity (dismissed):
Just a few questions and nitpicks.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs
RogPodge: I think it should be a break. For a controller layout, there shouldn't be instances where 2 buttons have the same usage right?

RogPodge:
Thanks for the suggestions @vrdave-unity. I incorporated all of them, though I have a question. On a given layout, 2 physical buttons should never have the same usage correct?

@Billreyn I really like the layout mockup idea you have there. It's much more elegant than an optional dropdown that you have to expand in the binding properties pane, and the nested structure really helps contain the bloat. Do you have a reference implementation, or can you point me to the code paths I need to touch to make this a reality?

> Also, as you can see in the screenshot above, I am wondering if we really need to show the derived bindings different between Xbox Controller, Xbox Controller iOS, Xbox Controller Android? They always seem to be the same anyway, so it seems like redundant information?

On this topic, I think it's still important to make this information accessible, since it helps expose all of the controller layouts that are currently registered (and in some cases, exposes the fact that you may have missed registering certain layouts!) This is especially useful in the XR space, since there's a bit of a manual process to configure your project to support the wide range of XR controllers available.

@lyndon-unity I'm not certain what additional work needs to be done here to account for the concurrent UI toolkit work. Can you give some more details?

ritamerkl:
Like Lyndon already mentioned we will need this feature to support the newer UIToolkit editor window as well.
Currently the UIToolkit InputActionsEditor is available on the develop branch but has to be enabled by using Unity 2022.1 or newer and modifying a line in InputActionsEditorWindow.cs (see this [document](https://docs.google.com/document/d/1ot1-Cr85TxpZGudm-VXvE4-C2YAATqiQcYkc_ww76eI/edit#heading=h.aasuoa63b4nd)) 
To support the UIToolkit for this feature additional files have to change. A starting point: You can find the binding properties view here: BindingPropertiesView.cs and the dropdown here: InputControlPathEditor.cs.

RogPodge:
Hey there guys,

I haven't gotten a chance to update the UIToolkit editor window, but I have updated the **Matching Controls** section to be a foldout.

image

I did try to incorporate the feedback to move this information directly into the bindings list, but encountered some issues with creating the UI elements there, since it's implemented as a `TreeView`. I also feel like it could make for a more confusing UI experience, especially when multiple bindings are mapped to the same action like here:

image

I left some comments in the code regarding some edge cases I encountered as I was debugging and implementing the foldouts, if anyone has any feedback, it'd be greatly appreciated!

https://github.com/Unity-Technologies/InputSystem/pull/1676/files#diff-6090a6d9adc85bc6f186a6206f2f0659bb1627a735d1474bf71e8153506ff84fR184-R189

Billreyn:
Hey so maybe I was not super clear before in the earlier mockup. Here's my attempt at clarifying: 

For each binding, we could show a disclosure triangle. By default they are all closed:
Screenshot 2023-05-30 at 11 33 59


Users can open each one to see what it maps to like so: 
Screenshot 2023-05-30 at 11 31 54

Make sense?

RogPodge:
I believe functionally, the code should be finalized. After discussion, we opted to keep with the current UI layout for now, as it would take more resources than we have available currently to migrate it into the Action's Pane rather than the Bindings Pane.

image

Some Notes:
The bindings displayed under Matched Controls use the `name` of the control, rather than the UI friendly `displayName`, as there are a number of layouts which don't have the latter property populated.

There is potential for duplicated information underneath the Matched Controls, such as the following (`Button East [Gamepad]` is the path, and we show that it matches the control `Gamepad > button east)`:
image

This is intentional, as there are scenarios where neglecting to show the "root" matched control can result in confusion.

1. The binding path provided specifies a Control Usage. In these cases, it isn't clear what "root" binding will be matched.
image

2. There is exactly 1 control which matches with the current binding path. By showing the user the Matched Controls, they can be sure that the path they provided is a valid one which corresponds to an existing layout. If we don't show this, it can create some ambiguity as to whether they specified an invalid/valid path. This could be remedied by utilizing other UI elements(i.e. displaying a check next to the Path field), but as a whole, it feels a bit odd given that we allow users to manually specify a path, rather than limiting them to the dropdown populated by all registered control layouts.

image

3. To counteract the cumbersome nature of having the "root" binding listed, it is expanded by default, as the intention of this new feature is to make it convenient for the user to verify that their path matches a broad range of controls, not just that of the "root" device they specify.

4. Performance:
image

While there is a noticeable performance spike when determining the matching control paths, this happens very infrequently, and should not impact the user's general experience when working with the Input Actions Editor

RogPodge:
@Billreyn @lyndon-unity this should be ready for another review.

lyndon-unity:
I'm seeing some duplicated information. 
E.g. if the path is 'Primary Action [Any]' (*/{PrimaryAction}), I see 'Mouse>LeftButton' twice in the list (along with others)

This was using the Assets/Samples/Visualizers/SimpleControls asset in the input system package (project at root level), on the 'fire' Action.

image

RogPodge:
Issue should be fixed @lyndon-unity 

image

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -721,6 +721,39 @@ public static bool Matches(string expected, InputControl control)
...
+                    for (var i = 0; i < usageCount; ++i)
+                    {
+                        if (StringMatches(usage, controlItem.usages[i]))
+                        {
+                            anyUsageMatches = true;
+                            break;
+                        }
+                    }
+
+                    if (!anyUsageMatches)
```
lyndon-unity: Could write if (i!=usageCount) here and avoid thy anyUsageMatches variable, but I think the existing code is easier to read.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -721,6 +721,39 @@ public static bool Matches(string expected, InputControl control)
...
+                        return false;
+                }
+            }
+
+            // Match name.
+            if (!expectedControlComponent.m_Name.isEmpty)
```
lyndon-unity: This seems faster than the for loop above, could this be moved above the for loop for a faster early out.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs
```diff
@@ -102,11 +103,172 @@ protected override void DrawGeneralProperties()
...
+                    matchedChildLayouts = EditorInputControlLayoutCache.allLayouts
+                        .Where(x => x.isDeviceLayout && !x.hideInUI && !x.isOverride && x.isGenericTypeOfDevice).OrderBy(x => x.displayName);
+                    break;
+                default:
+                    matchedChildLayouts = EditorInputControlLayoutCache.TryGetChildLayouts(deviceLayout.name);
+                    break;
+            }
+
+            // If this layout does not have a match, or is the top level root layout,
+            // skip over trying to draw any items for it, and immdiately try processing the child layouts
```
lyndon-unity: Minor spelling typo:

immdiately -> immediately

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlDropdownItem.cs
```diff
@@ -70,15 +70,31 @@ public OptionalControlDropdownItem(EditorInputControlLayoutCache.OptionalControl
...
 
-        public UsageDropdownItem(string usage)
+        public ControlUsageDropdownItem(string device, string usage, string controlUsage)
             : base(usage)
         {
-            m_Device = "*";
-            m_ControlPath = usage;
+            m_Device = string.IsNullOrEmpty(device) ? "*" : device;
+            m_Usage = usage;
+            m_ControlPath = $"{{{ controlUsage }}}";
```
lyndon-unity, line 96: With controlPath referencing this it seems we are exposing the braces around the control usage where we didn't used to have them.

I can't tell if that generates any unexpected side effects

lyndon-unity (approved):
Thanks for the update. My ui looks better now with the duplication fix.

Code seems reasonable but I don't know this code area well so I think it would be wise to get review from another input team member

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
RogPodge: I think I'll leave it as it for readability's sake.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlDropdownItem.cs
RogPodge: This shouldn't result in any issues. Previously, we added the braces when constructing the entire path at the end. With this approach, we add the braces "early" when we know we're adding a usage, in order to not have different path creation patterns during the construction step.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs
RogPodge: Done

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
RogPodge: Done

RogPodge:
@vrdave-unity can you reset your review/take another look?

lyndon-unity:
I've tried this binding which isn't showing up any matching control paths
<PSVR2Controller>{RightHand}/{Primary2DAxisClick}
Would you expect this to match a binding with the following setup?

I'm using 2022.2.15 + PS5 platform installed
Packages
- XR Plugin Management 
- XR interaction toolkit 
- OpenXR plugin 
- + Playstation VR2 (download - tarball install)

Ps5 platform active

I'm mentioning this as I'm looking at a bug report in this areas and was hoping this PR could help me check the bindings
https://jira.unity3d.com/browse/ISXB-315

lyndon-unity:
Ok I think I see why ps2vr isn't being picked up. This is only added in code on the psvr device in the native code

This native code references Primary2DAxisClick
[Native~/include/unity/IUnityXRInput.h#L236](https://github.cds.internal.unity3d.com/unity/xr.sdk.psvr2/blob/7828a2125fa7fe9e73d7c914b147f8087c3c0111/com.unity.xr.psvr2/Native~/include/unity/IUnityXRInput.h#L236) 
/// A binary feature representing a depression or click of the kUnityXRInputFeatureUsagePrimary2DAxis 2D axis.  If this is implemented, the device must implement kUnityXRInputFeatureUsagePrimary2DAxis.  If this is actuated, then kUnityXRInputFeatureUsagePrimary2DAxisTouch must be actuated as well, if it exists.
UnityXRInputFeatureUsage const kUnityXRInputFeatureUsagePrimary2DAxisClick = "Primary2DAxisClick";


And it seems we have a call to DeviceDefinition_AddFeatureWithUsage 
[Native~/src/psvr2_input.cpp#L639](https://github.cds.internal.unity3d.com/unity/xr.sdk.psvr2/blob/216ef283a9d86c9b3bf2811cfa2eb615de9eca9f/com.unity.xr.psvr2/Native~/src/psvr2_input.cpp#L639) 
featureMap[PSVR2ControllerFeature::ThumbstickClick] = s_Input->DeviceDefinition_AddFeatureWithUsage(definition, "Thumbstick - Click", kUnityXRInputFeatureTypeBinary, kUnityXRInputFeatureUsagePrimary2DAxisClick);

This does seem to be filled in here
[Native~/src/psvr2_input.cpp#L1149](https://github.cds.internal.unity3d.com/unity/xr.sdk.psvr2/blob/216ef283a9d86c9b3bf2811cfa2eb615de9eca9f/com.unity.xr.psvr2/Native~/src/psvr2_input.cpp#L1149) 
s_Input->DeviceState_SetBinaryValue(state, featureMap[PSVR2ControllerFeature::ThumbstickClick], (padData.buttons & SCE_PAD_VR_CONTROLLER_BUTTON_R3) == SCE_PAD_VR_CONTROLLER_BUTTON_R3);


So this looks like its going to be a limitation of the system as it stands at the moment :( 


I think ideally we'd add some stub reference here so the editor can pick it up too
https://github.cds.internal.unity3d.com/unity/xr.sdk.psvr2/blob/master/com.unity.xr.psvr2/Runtime/DeviceLayouts.cs#L212

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs
```diff
@@ -102,11 +103,172 @@ protected override void DrawGeneralProperties()
...
+                        {
+                            matchExists |= DrawMatchingControlPathsForLayout(genericDeviceLayout, in parsedPath);
+                        }
+                    }
+                }
+
+                // Otherwise, indicate that no layouts match the current path.
+                if (!matchExists)
+                {
+                    EditorGUILayout.LabelField("No registered control paths match this current binding");
```
lyndon-unity: Could be valuable to add a tooltip to this label as some binding control paths may only be registered at runtime so it may match at that time.

lyndon-unity:
<Gamepad>/a doesn't seem to match anything. This is an alias for <Gamepad>/buttonSouth and I would expect that to list the same set of controls.

image
image
image

lyndon-unity (changes requested):
Generally I approved, but want to check feedback around the aliases before landing.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs
RogPodge: Done

image

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
```diff
@@ -721,14 +721,33 @@ public static bool Matches(string expected, InputControl control)
...
+                    for (var i = 0; i < aliases.Count; i++)
+                    {
+                        if (StringMatches(expectedControlComponent.m_Name, aliases[i]))
+                        {
+                            hasMatchingAlias = true;
+                            break;
+                        }
+                    }
+
+                    if(!hasMatchingAlias)
```
lyndon-unity: I think this would prevent it matching usage versions (if there is an alias list but no matching aliases)

lyndon-unity (changes requested):
This seems to fail some cases where an alias is present but none match, then an usage is present that should match.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
lyndon-unity: `<Gamepad>/{Back}`

No longer matches with this update

lyndon-unity: I think we need more test cases added to the test to verify what does and doesn't match.

RogPodge: Added 2 new test cases to cover the Alias matching case and the Usages matching case. The regression from before should be fixed as well.

Billreyn:
Hi! Finally had time to test this. Overall it seems to be working. 

However, the way it works now does prompt another design issue to me. The way I see our bindings, is that some bindings are _direct_ bindings, to a specific button or control on a specific device. 

Then we have other bindings that are _abstractions_ - things like the abstracted GamePad or the Usages system..

The way I had seen this working is that the Matching Controls info would only appear when the user has bound an action to an abstracted binding. 

Currently you can bind something to, say, a keyboard button or mouse, in which case the Matching Controls info will just repeat that. To me that doesn't seem terribly useful?

Screenshot 2023-06-14 at 13 36 55


Screenshot 2023-06-14 at 13 36 15


Wouldn't it be nicer to just have this feature be shown when the user is binding to an abstraction?

RogPodge:
I was originally hesitant to get rid of the redundancy (https://github.com/Unity-Technologies/InputSystem/pull/1676#issuecomment-1574428285), but after digging into the cases (maybe for a bit too long), I think I have a satisfactory solution @Billreyn.

I'm still not quite satisfied with it, since we don't have additional UI to indicate that the user provided binding is itself a valid binding (doing so would probably best be done in another part of the codebase), but I think the dev time to make something completely comprehensive might be out of scope at the moment 😓

lyndon-unity:
Overall I like the way this PR is shaping up. I've asked out QA to do a pass over it.

Billreyn:
Hi Roger. Talked the naming though with Lyndon Homewood, and we landed on the term `Derived controls`.

If you just change that name I think we can accept and merge this PR

RogPodge:
Hey @lyndon-unity, I followed up with William on the naming, and after flip flopping a few times, feel that **Derived Bindings** would make more sense. The root for wanting to use bindings is so we don't conflate different concepts within the same UI element, as it can be confusing to see information about controls while editing a binding. 

In addition, in our ideal UI, the information would be placed directly under their parent in the Actions pane. In that scenario, they would be listed directly under a **binding**, so the items that appear under them would be interpreted as **Sub-Bindings**, or more formally, **Derived Bindings**

Let me know if that sounds good, and thanks everyone + QA for taking a pass on this!

Billreyn:
Hi Roger. Derived bindings is fine with us :)

lyndon-unity:
*/{PrimaryAction}
Doesn't seem to show any matches for me - this looks like a bug

image

lyndon-unity:
Also as you mention above

This doesn't show a match because its a single direct match
<XboxOneGamepadAndroid>/buttonEast

This does feel quite confusing to me. I thought it might be a invalid path, its not possible to tell the difference between invalid path and valid path (unless you've got one that has some derived types that it lists).
Ideally we'd add something like "Valid binding" for those that match something and "No matches" for those that have none.

>I'm still not quite satisfied with it, since we don't have additional UI to indicate that the user provided binding is itself a valid binding (doing so would probably best be done in another part of the codebase), but I think the dev time to make something completely comprehensive might be out of scope at the moment 😓

RogPodge:
> */{PrimaryAction} Doesn't seem to show any matches for me - this looks like a bug
> 
> image

You actually caught a bug that was introduced where we didn't process the 'match any device' case properly.

It should be fixed now.

Pauliusd01:
Found a bug, when adding 2 or more * at the front of the binding path the whole binding properties sidebar layout breaks. **Steps to repro:** Type in this path `*/{PrimaryAction}` -> add an extra * at the front
https://github.com/Unity-Technologies/InputSystem/assets/54306142/206099ac-3464-46fe-b76a-8bbbf96f72fe

Alternatively you can also break the UI by using multiple categories in the same path such as: `<Gamepad>/<Gamepad>`

RogPodge:
@Pauliusd01 bugs should be fixed!

lyndon-unity:
Formatting tests are failing - looks like it needs the formatting run on it


[19:05:56.456 Information] Starting process on subprocess runner
[19:06:04.263 Information] warning: Needs formatting: Packages/com.unity.inputsystem/InputSystem/Controls/InputControlPath.cs
[19:06:06.452 Information] warning: Needs formatting: Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputBindingPropertiesView.cs

lyndon-unity:
There is also a code analyzer failure.

Seem to be this error :
System.ExecutionEngineException: System.InvalidOperationException: Unexpected value '4' of type 'Microsoft.CodeAnalysis.RefKind' ---> System.InvalidOperationException: Unexpected value '4' of type 'Microsoft.CodeAnalysis.RefKind'

I can't quite see what's triggering it. Here's a link about that kind of error: https://github.com/dotnet/roslyn/issues/29371 

I see a few occurrences of this failure on your branch
https://unity-ci.cds.internal.unity3d.com/project/130/branch/inputbinding-controlusages-UI/jobDefinition/.yamato%2Fanalyze.yml%23code_analyser

I'm not seeing the failure on develop
https://unity-ci.cds.internal.unity3d.com/project/130/branch/develop/jobDefinition/.yamato%2Fanalyze.yml%23code_analyser?nav=jobDefinitions

RogPodge:
small ping after fixing the formatter errors @lyndon-unity @Pauliusd01

Pauliusd01 (approved):

Pauliusd01:
LGTM. Things checked:
Case sensitive scenarios
Normal usage such as (*/{PrimaryAction} / <Gamepad>/{Back})
Aliases such as (<Gamepad>/a/b/x/y/triangle/etc)
Direct control names such as (<Gamepad>/buttonSouth)
Checked what results are shown when using the binding menu itself

Note: One bug that I did notice was a fairly rare Editor freeze under the Application.Message.MouseMove call when selecting the binding path's text (only 4 times in the whole testing period) but it was ruled out as potentially caused by this PR by Roger so I'll try to nail it down and report separately

lyndon-unity (approved):
Happy to approve now that Paulius has QA'd it.

RogPodge:
Waiting for someone here to ultimately merge this PR

jamesmcgill (approved):
Thanks for getting this this over the line @RogPodge

PR #1686, opened 2023-06-08
[2022.3] Change #ifdef to allow for displayIndex (multi-display) in Unity 2022.3

reapazor:
### Description

Alters the #ifdef's that protect the displayIndex work to match supported Unity editor versions. It actually landed in late 2022.2 as well, but it was safer to use the UNITY_2022_3_OR_NEWER condition.

### Changes made

`#if UNITY_2023_1_OR_NEWER` is now `#if UNITY_2022_3_OR_NEWER` for displayIndex code


jamesmcgill (approved):
Thanks @reapazor. Looks good.

PR #1688, opened 2023-06-09
Fix for TrackedPoseDriver's BeforeRender update not being invoked

periodyctom:
### Description

This fixes the drift lag issue with `TrackedPoseDriver` that made it less usable than the deprecated `ARPoseDriver`

#### Issue Links
- https://github.com/Unity-Technologies/arfoundation-samples/issues/1063

### Changes made

`TrackedPoseDriver` now hooks into `Application.onBeforeRender` and invokes `InputSystem.Update(InputUpdateType.BeforeRender)`. `TrackedPoseDriver` expects that update type, yet it's not actually invoked from anywhere. Since `TrackedPoseDriver` requires OnBeforeRender input updates to be called to work properly, we simply have `TrackedPoseDriver` manage invoking the update.

### Notes

- Should we have the `OnBeforeRender` input update called from elsewhere, or is it acceptable for `TrackedPoseDriver` to manage it?
- XR `TrackedPoseDriver` Tests already covers BeforeRender update.
- No doc changes, as no API surface changes.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
@@ -422,6 +422,7 @@ namespace UnityEngine.InputSystem.XR
         /// </summary>
         protected void OnEnable()
         {
+            Application.onBeforeRender += PerformInputSystemUpdateBeforeRender;
             InputSystem.onAfterUpdate += UpdateCallback;
             BindActions();
 
@@ -437,6 +438,7 @@ namespace UnityEngine.InputSystem.XR
         {
             UnbindActions();
             InputSystem.onAfterUpdate -= UpdateCallback;
+            Application.onBeforeRender -= PerformInputSystemUpdateBeforeRender;
         }
 
         /// <summary>
@@ -451,6 +453,8 @@ namespace UnityEngine.InputSystem.XR
             }
 #endif
         }
+
+        void PerformInputSystemUpdateBeforeRender() => InputSystem.Update(InputUpdateType.BeforeRender);
 
         /// <summary>
         /// The callback method called after the Input System has completed an update and processed all pending events.
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -15,6 +15,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - Fixed UI clicks not registering when OS provides multiple input sources for the same event, e.g. on Samsung Dex (case ISX-1416, ISXB-342).
+- Fixed drift lag in `TrackedPoseDriver` due to `InputUpdateType.BeforeRender` not being invoked anywhere.
 
 ## [1.6.1] - 2023-05-26
 


vrdave-unity (approved):
Wow, that's it? We just were never hooking OnBeforeRender? 🤦🏻‍♂️ 
I also noticed that [in the LegacyInputHelper version](https://github.cds.internal.unity3d.com/unity/com.unity.xr.legacyinputhelpers/blob/master/Runtime/TrackedPoseDriver/TrackedPoseDriver.cs#L509) of this we set an attribute for the OnBeforeRender method:
```
[BeforeRenderOrder(-30000)]
protected virtual void OnBeforeRender()
{
    if (m_UpdateType == UpdateType.BeforeRender ||
        m_UpdateType == UpdateType.UpdateAndBeforeRender)
    {
        PerformUpdate();
    }
}
```

Should we also consider doing that [here](https://github.com/Unity-Technologies/InputSystem/blob/b4ae7b7821f0386538a275195192c71cfc183376/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs#LL559C48-L559C48)?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -452,6 +454,8 @@ protected virtual void OnDestroy()
 #endif
         }
 
+        void PerformInputSystemUpdateBeforeRender() => InputSystem.Update(InputUpdateType.BeforeRender);
```
chris-massie, line 457: This will cause the input system to step twice each frame. This isn't the way to fix the issue. The `TrackedPoseDriver.OnBeforeRender` method is correctly called when running a VR app, so this change would cause a double step and break any user that set the Edit > Project Settings > Input System Package > Update Mode to Manual.

What I think is happening is that the `HandheldARInputDevice` doesn't have `updateBeforeRender` set to true in the input control layout, so I'm guessing maybe the device doesn't have `UpdateBeforeRender` flag set. The main `InputManager.OnUpdate` method loop that is invoked by low level code has an early return if the `updateType` is not part of the `m_UpdateMask`. See https://github.com/Unity-Technologies/InputSystem/blob/c55cdd9f94422b4917045999a1c7b00f52e9bf1f/Packages/com.unity.inputsystem/InputSystem/InputManager.cs#L2857

The `m_UpdateMask` is set to add the `InputUpdateType.BeforeRender` ONLY if an added device has that set.
```csharp
if (device.updateBeforeRender)
    updateMask |= InputUpdateType.BeforeRender;
```
See https://github.com/Unity-Technologies/InputSystem/blob/c55cdd9f94422b4917045999a1c7b00f52e9bf1f/Packages/com.unity.inputsystem/InputSystem/InputManager.cs#LL1220C43-L1220C43

So my guess is that in the user's project, there aren't any added devices with that before render flag set, so then the callback in `TrackedPoseDriver` is never getting invoked when `InputState.currentUpdateType == InputUpdateType.BeforeRender`.

We have code in XRI where we have an `XRSimulatedHMD` where we set the attribute.
```csharp
[InputControlLayout(stateType = typeof(XRSimulatedHMDState), isGenericTypeOfDevice = false, displayName = "XR Simulated HMD", updateBeforeRender = true)]
[Preserve]
public class XRSimulatedHMD : XRHMD
```
See https://github.com/Unity-Technologies/com.unity.xr.interaction.toolkit/blob/80d057af71a66e9b32db863e9fc26f18e041b688/com.unity.xr.interaction.toolkit/Runtime/Inputs/Simulation/XRSimulatedHMD.cs#L12-L14

We might be able to just add that `updateBeforeRender = true` to `HandheldARInputDevice` in com.unity.xr.arsubsystems. Can we try that, or verify that the added device of that layout actually gets the flag set? OpenXR HMD devices get it added for instance. The screenshot is when I open Window > Analysis > Input Debugger, right-click the `HeadTrackingOpenXR` device and click Open Device Debug View. It shows that flag too with the XRI device simulator.
UpdateBeforeRender flags

chris-massie (changes requested):

periodyctom:
Actual Fix is in XR native module.

PR #1690, opened 2023-06-13
Apply the same scale factor as InputForUI InputManager provider for scroll delta

jfreire-unity:
### Description

The goal of this PR is to update the InputForUI provider with the latest changes that are about to be introduced on trunk that fix a high user-pain issue [(IN-24517)](https://issuetracker.unity3d.com/issues/windows-inputvalue-of-the-scroll-wheel-changes-with-the-step-of-plus-120-when-scrolling-using-input-system).

### Changes made

Applies the same scale factor as the Input Manager provider, that's consistent with UGUI values.

### Notes

⚠️  Should only be merged when there are changes in `trunk`.


jfreire-unity:
I still need to fix tests.

lyndon-unity (approved):
In discussion we aligned that this was good to land.
It relates to the new input system windows backend fix to make sure we take into account the WHEEL_DELTA (120) divisor applied in PlatformDependent/Win/Input/NewInput.cpp

See:
https://github.cds.internal.unity3d.com/unity/unity/pull/30853/files

If input system "input for ui" is used without the backend change the mouse wheel value will be a lot higher but a use can fix by upgrading unity version (Note "input for ui" is also only used in 2023.2+) or by using UGUI event system backend (which was the previous backend).

jfreire-unity:
Some unrelated tests to this PR we're failing and were disabled. Issue raised https://jira.unity3d.com/browse/ISX-1462

PR #1691, opened 2023-06-22
NEW: edit actions and bindings properties (ISX-1374, ISX-1126)

ritamerkl:
### Description

added the functionality to edit various fields for Actions, Bindings, Composites and Bindings in Composites. 
Includes adding, editing, reordering and deleting of Interactions and Processors. See the Ticket [here](https://jira.unity3d.com/browse/ISX-1374).

Another related ticket was done with this PR: https://jira.unity3d.com/browse/ISX-1126

### Changes made

added:
- editing properties of Actions, Bindings, Composites
- adding Interactions & Processors
- editing Interactions & Processors
- deleting Interactions & Processors
- reordering Interactions & Processors

### Checklist

Before review:

- [] Changelog entry added.
- [] Tests added/changed, if applicable.
    - Performance tests.
    - Integration tests.
- [] Docs for new/changed API's.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/InputParameterEditor.cs
```diff
@@ -249,12 +250,18 @@ public void OnDrawVisualElements(VisualElement root, Action onChangedCallback)
...
                 m_OpenInputSettingsButton = new Button(InputSettingsProvider.Open){text = m_OpenInputSettingsLabel.text};
                 m_OpenInputSettingsButton.AddToClassList("open-settings-button");
-                buttonContainer.Add(m_OpenInputSettingsButton);
 
                 settingsContainer.Add(m_FloatField);
                 settingsContainer.Add(m_DefaultToggle);
                 container.Add(settingsContainer);
-                container.Add(m_HelpBox);
+
+                //only add inputSettingsButton and help box if default is enabled
```
lyndon-unity: Not sure this comment adds much value.

I can see the default variable is controlling if the button and help box shows up, but I don't really understand what 'default is enabled' means. I'd suggest either remove the comment or expand to tell me more than the local code.

Packages/com.unity.inputsystem/InputSystem/Editor/InputParameterEditor.cs
```diff
@@ -278,18 +285,21 @@ private void ChangeSettingValue(ChangeEvent<float> evt)
                 }
             }
 
-            private void ToggleUseDefaultValue(ChangeEvent<bool> evt)
+            private void BlurFloatField(Action onChangedCallback)
```
lyndon-unity: Blur was a new event to me and therefore I also struggled with the naming on this. If this could be used in more than one place and isn't float specific could this perhaps be renamed to TriggerChangeCallback or something more action oriented?

Packages/com.unity.inputsystem/InputSystem/Editor/InputParameterEditor.cs
```diff
@@ -278,18 +285,21 @@ private void ChangeSettingValue(ChangeEvent<float> evt)
...
                 if (evt.newValue != m_UseDefaultValue)
                 {
                     m_SetValue(!evt.newValue ? m_GetDefaultValue() : m_DefaultInitializedValue);
+                    onChangedCallback.Invoke();
                 }
 
                 m_UseDefaultValue = evt.newValue;
-
                 m_FloatField?.SetEnabled(!m_UseDefaultValue);
-                m_HelpBox.visible = m_UseDefaultValue;
```
lyndon-unity, line 291: I like that these have moved to the setup phase and no longer needs to toggle visibility.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Resources/InputActionsEditorStyles.uss
```diff
@@ -136,12 +142,20 @@
     background-image: resource('Packages/com.unity.inputsystem/InputSystem/Editor/Icons/ChevronUp.png');
 }
 
+.name-and-parameters-list-view .upActive {
+    background-image: resource('Packages/com.unity.inputsystem/InputSystem/Editor/Icons/d_ChevronUp.png');
```
lyndon-unity, line 148: The icon seems to have a difference naming convention to the one above without the d_ 
Is there a reason these are inconsistent.

I seem to recall seeing some naming convention in a style guide somewhere but could only find these which didn't seem to help on this specific question
https://docs-style-guide.unity.com/media/images/image-type-guidelines/diagrams#icons
https://unityeditordesignsystem.unity.com/fundamentals/iconography#Overview

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionPropertiesView.cs
```diff
@@ -51,6 +52,7 @@ public override void RedrawUI((SerializedInputAction ? , List<string>) viewState
                 controlType.choices.Clear();
                 controlType.choices.AddRange(controlTypes.Select(ObjectNames.NicifyVariableName).ToList());
                 var controlTypeIndex = controlTypes.FindIndex(s => s == inputAction.expectedControlType);
+                controlTypeIndex = Math.Clamp(controlTypeIndex, 0, controlTypes.Count - 1);
```
lyndon-unity, line 56: Is this clamping to default to the zero's element if no match is found ?
Seems reasonable but I for a moment wondered why it was here. Perhaps a comment on the line would help

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -14,16 +15,74 @@ internal class NameAndParametersListView : ViewBase<InputActionsEditorState>
...
             CreateSelector(state => state);
         }
 
+        public void OnAddElement(string name)
+        {
+            if (m_ListProperty == null)
+                return;
+
+            var interactionsOrProcessorsList = NameAndParameters.ParseMultiple(m_ListProperty.stringValue).ToList();
+            var newElement = new NameAndParameters() { name = name};
```
lyndon-unity, line 37: Should this initialise the parameters too ?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -14,16 +15,74 @@ internal class NameAndParametersListView : ViewBase<InputActionsEditorState>
...
+            interactionsOrProcessorsList.Add(newElement);
+
+            m_ListProperty.stringValue = ToSerializableString(interactionsOrProcessorsList);
+            m_ListProperty.serializedObject.ApplyModifiedProperties();
+        }
+
+        private void MoveElement(int index, bool up)
+        {
+            var newIndex = index - 1;
+            if (!up)
```
lyndon-unity: Feels like you should collapse this into one line so you don't set the value twice if up is false. Trivial but if ever used in a loop could be more performant.

var newIndex = up ? index -1 : index +1;

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -14,16 +15,74 @@ internal class NameAndParametersListView : ViewBase<InputActionsEditorState>
...
+                newIndex = index + 1;
+            SwapElement(index, newIndex);
+        }
+
+        private void SwapElement(int oldIndex, int newIndex)
+        {
+            var interactionsOrProcessorsList = NameAndParameters.ParseMultiple(m_ListProperty.stringValue).ToArray();
```
lyndon-unity: This says ToArray but the variable is named List which is a bit confusing.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -14,16 +15,74 @@ internal class NameAndParametersListView : ViewBase<InputActionsEditorState>
...
+            newIndex = Math.Clamp(newIndex, 0, interactionsOrProcessorsList.Length - 1);
```
lyndon-unity: If the interactionsOrProcessorsList is empty this seems it would have an issue with clamping to range 0,-1 and the swap would fail below.

Perhaps this can never happen due to calling this code with always valid values, but an early out might be wise.
Also should we clamp in this case or just early out if the newIndex is out of range ?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -41,9 +100,19 @@ public override void RedrawUI(InputActionsEditorState state)
...
             m_ContentContainer.Clear();
-            foreach (var parameterListView in parameterListViews)
+            for (int i = 0; i < parameterListViews.Count; i++)
             {
-                new NameAndParametersListViewItem(m_ContentContainer, parameterListView);
+                var index = i;
+                var buttonProperties = new ButtonProperties()
+                {
+                    onClickDown = () => MoveElement(index, false),
+                    onClickUp = () => MoveElement(index, true),
```
lyndon-unity: perhaps consider an enum for the second parameter to Move element to make it clearer which is up and which is down. Or given how simple the function is, have a dedicated up and down variation.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -41,9 +100,19 @@ public override void RedrawUI(InputActionsEditorState state)
 
             m_Root.Q<Label>("no-parameters-added-label").style.display = new StyleEnum<DisplayStyle>(DisplayStyle.None);
             m_ContentContainer.Clear();
-            foreach (var parameterListView in parameterListViews)
+            for (int i = 0; i < parameterListViews.Count; i++)
             {
-                new NameAndParametersListViewItem(m_ContentContainer, parameterListView);
+                var index = i;
```
lyndon-unity, line 112: Why do we have a separate index and i - they seem to be the same.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/PropertiesView.cs
```diff
@@ -1,30 +1,121 @@
 #if UNITY_EDITOR && UNITY_INPUT_SYSTEM_UI_TK_ASSET_EDITOR
 using System;
+using System.Linq;
```
lyndon-unity, line 3: I always question if we need LINQ as it can be poor performance at times, but I think for the UI work it makes sense to use.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/PropertiesView.cs
```diff
@@ -1,30 +1,121 @@
...
+            toggle.Add(addProcessorButton);
+            return addProcessorButton;
         }
 
-        public override void RedrawUI(SelectionType selectionType)
+        private void CreateContextMenuProcessor(string expectedControlType, SerializedProperty serializedProperty)
+        {
+            var processors = InputProcessor.s_Processors;
+            Type expectedValueType = null;
+            if (!string.IsNullOrEmpty(expectedControlType))
```
lyndon-unity: When you have a value being set above and then a conditional to update it I always prefer folding these into one to set the value only once.

var data = conditional ? truevalue : falseValue;

lyndon-unity: So here

Type expectedValueType = string.IsNullOrEmpty(expectedControlType) ? null : EditorInputControlLayoutCache.GetValueType(expectedControlType);

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/PropertiesView.cs
```diff
@@ -1,30 +1,121 @@
...
+                expectedValueType = EditorInputControlLayoutCache.GetValueType(expectedControlType);
+
+            addProcessorButton.RegisterCallback<ContextualMenuPopulateEvent>(evt =>
+            {
+                evt.menu.ClearItems();
```
lyndon-unity, line 76: For bigger callback content like this I'm tempted to suggest snapping out to a function

ProjectSettings/MemorySettings.asset
```diff
@@ -0,0 +1,35 @@
+%YAML 1.1
```
lyndon-unity: Unsure why this file is a new addition - perhaps just never checked in before so probably makes sense.

lyndon-unity (changes requested):
Looks reasonable. I've made a couple of queries so will pause before approving. Answers might mean no changes needed though.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
ritamerkl: we need a local index to save a steady copy of the integer, otherwise we would reference the changing i variable of the loop

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Resources/InputActionsEditorStyles.uss
ritamerkl: After some research I found out that the d_ prefix is used for dark scheme of the editor. So I adjusted the code accordingly. Thank you for bringing that up.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -41,20 +41,25 @@ public void OnAddElement(string name)
...
         }
 
         private void SwapElement(int oldIndex, int newIndex)
         {
-            var interactionsOrProcessorsList = NameAndParameters.ParseMultiple(m_ListProperty.stringValue).ToArray();
-            newIndex = Math.Clamp(newIndex, 0, interactionsOrProcessorsList.Length - 1);
-            MemoryHelpers.Swap(ref interactionsOrProcessorsList[oldIndex], ref interactionsOrProcessorsList[newIndex]);
-            m_ListProperty.stringValue = ToSerializableString(interactionsOrProcessorsList);
+            var interactionsOrProcessors = NameAndParameters.ParseMultiple(m_ListProperty.stringValue).ToArray();
+            if (interactionsOrProcessors.Length == 0 || newIndex < 0 || newIndex >= interactionsOrProcessors.Length)
```
lyndon-unity: Looks good. I think we should probably check oldIndex too
If we never expect this to be out of range we should probably assert so they get flagged.

lyndon-unity (approved):
Good updates, minor comment added. Generally happy to approve

PR #1697, opened 2023-06-28
[ISXB-551] Fix InputControls for precompiled layouts

mtschoen-unity:
### Description

If you try to create a precompiled layout for a device which uses `QuaternionControl` (for example, `XRController`), the generated code fails to compile because it tries to set values to the child controls. It seems like the code generally assumes these properties will have setters, but there's no check for that.

### Changes made

Added a test that checks all `InputDevice` and `InputControl` classes to ensure their control properties have setters. Also fixed up any instances that fail the test.

### Notes

The test isn't completely exhaustive. I noticed that the `CustomDevice` in the repo just declares public properties without using the `InputControlAttribute` to decorate them. They are still picked up by the layout generator, but I'm not entirely sure why. If I do the check on all public properties, I get some false positives for properties like `Keyboard.spaceKey`. I added a comment to the test to mention this, but I'd love to know if there's a better way to check which properties are actually controls. As far as I can tell, we'd have to actually build the layouts and get their controls that way, which seems like more trouble than it's worth for this test.

This is an "added API" and will require a minor version bump. It cannot be backported according to semver rules.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Samples/CustomDevice/CustomDevice.cs
+++ b/Assets/Samples/CustomDevice/CustomDevice.cs
@@ -177,10 +177,10 @@ public class CustomDevice : InputDevice, IInputUpdateCallbackReceiver
     // for it a little bit. One thing we can do is expose the controls for our
     // device directly. While anyone can look up our controls using strings, exposing
     // the controls as properties makes it simpler to work with the device in script.
-    public ButtonControl firstButton { get; private set; }
-    public ButtonControl secondButton { get; private set; }
-    public ButtonControl thirdButton { get; private set; }
-    public StickControl stick { get; private set; }
+    public ButtonControl firstButton { get; protected set; }
+    public ButtonControl secondButton { get; protected set; }
+    public ButtonControl thirdButton { get; protected set; }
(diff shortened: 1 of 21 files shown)


RogPodge (approved):

jamesmcgill:
Hi @mtschoen-unity can you migrate this out of your personal fork and into a branch on the main internal repo so it can run CI?
I see your username already has write permission to this repo so it should be possible for you.

mtschoen-unity:
> can you migrate this out of your personal fork

Yep! I'm slammed on some other work right now but when I come up for air I'll push to the main repo and re-open the PRs.

mtschoen-unity:
Closing in favor of https://github.com/Unity-Technologies/InputSystem/pull/1701 which isn't from my fork.

PR #1699, opened 2023-07-03
Added a warning when a generated class is being finalized but an InputActionMap is left enabled.

peterhall-unity3d:
Added a warning when a generated class is being finalized but an InputActionMap is left enabled.

If it's left enabled that could lead to the StateChangeMonitorsForDevice.listeners growing. As it grows performance of input events will get worse.

I'm open to the idea that there might be a better way to deal with this - if you can think of a good alternative please let me know!

### Description

Working towards fixing ISX-1357 - https://jira.unity3d.com/browse/ISX-1357.  This change should warn users if they create an instance of a generated C# class, enable action maps, and the C# class is then finalized with the action map still enabled.

### Changes made

Added a finalizer to the generated class that asserts all the action maps are not enabled, with a message telling the user which action maps are left enabled.

### Notes

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


lyndon-unity (approved):
This looks like a useful warning, but I'm wondering why we don't call Disable here to do the required work for them ?

jamesmcgill (approved):
lgtm

Can you add an entry in the CHANGELOG?
TVOS jobs might be a known issue but the others like formatting should pass. There is a format script in the root of the repo.
Could possibly also consider adding a test to validate that the assert triggers.

peterhall-unity3d:
> This looks like a useful warning, but I'm wondering why we don't call Disable here to do the required work for them ?

So I did try doing this.  The problem I hit was that the finalization is running on another thread and the Disable call is not thread safe.  I considered trying to flag the item and auto-disabling it from elsewhere, though I don't think there's e.g. an Update call I can easily track that.  I guess I'm also not 100% certain that a user might somehow be relying on this behaviour and the automatic Disable call might break their project...

lyndon-unity:
Looks like an extra file needs committing to pass the tests
Assets/Tests/InputSystem/InputActionCodeGeneratorActions.cs

jamesmcgill:
Jobs are failing due to `error: Error building Player because scripts are compiling` issue
Needs some investigation from Input Team.

jamesmcgill:
Thanks for the PR @peterhall-unity3d . Merged now.

peterhall-unity3d:
> Thanks for the PR @peterhall-unity3d . Merged now.

Thank you for landing it @jamesmcgill !  🏆

PR #1701, opened 2023-07-07
[ISXB-551] Fix InputControls for precompiled layouts

mtschoen-unity:
### Description

Re-hash of #1697 -- I need to make a new PR to change the incoming branch

If you try to create a precompiled layout for a device which uses `QuaternionControl` (for example, `XRController`), the generated code fails to compile because it tries to set values to the child controls. It seems like the code generally assumes these properties will have setters, but there's no check for that.

### Changes made

Added a test that checks all `InputDevice` and `InputControl` classes to ensure their control properties have setters. Also fixed up any instances that fail the test.

### Notes

The test isn't completely exhaustive. I noticed that the `CustomDevice` in the repo just declares public properties without using the `InputControlAttribute` to decorate them. They are still picked up by the layout generator, but I'm not entirely sure why. If I do the check on all public properties, I get some false positives for properties like `Keyboard.spaceKey`. I added a comment to the test to mention this, but I'd love to know if there's a better way to check which properties are actually controls. As far as I can tell, we'd have to actually build the layouts and get their controls that way, which seems like more trouble than it's worth for this test.

This is an "added API" and will require a minor version bump. It cannot be backported according to semver rules.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


RogPodge (approved):

jamesmcgill:
Perhaps merging in the latest develop branch might help with the tvos failures (note the CHANGELOG.md might need fixing afterward, due to a recent release).
There is a script in the root for the repo for formatting which should fix that failure.

Packages/com.unity.inputsystem/InputSystem/Controls/QuaternionControl.cs
```diff
@@ -22,28 +22,28 @@ public class QuaternionControl : InputControl<Quaternion>
         /// </summary>
         /// <value>Control representing the X component.</value>
         [InputControl(displayName = "X")]
-        public AxisControl x { get; private set; }
+        public AxisControl x { get; set; }
```
jamesmcgill, line 25: Not protected like the others?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Controls/PoseControl.cs
```diff
@@ -132,7 +132,7 @@ public class PoseControl : InputControl<PoseState>
         /// </summary>
         /// <value>Control representing whether the pose is being fully tracked. Maps to the <see cref="PoseState.isTracked"/> value.</value>
         /// <seealso cref="PoseState.isTracked"/>
-        public ButtonControl isTracked { get; private set; }
+        public ButtonControl isTracked { get; set; }
```
jamesmcgill, line 135: Not protected?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/XRSupport.cs
```diff
@@ -283,11 +283,11 @@ public float rightEyeOpenAmount
     public class BoneControl : InputControl<Bone>
     {
         [InputControl(offset = 0, displayName = "parentBoneIndex")]
-        public IntegerControl parentBoneIndex { get; private set; }
+        public IntegerControl parentBoneIndex { get; set; }
```
jamesmcgill, line 286: protected?

jamesmcgill (approved):
Seems a reasonable change.
I'm wondering if a minor bump is enough to pass CI but we'll see what the API Validator (Dry Run) job says.

Once the other jobs are passing we can try doing a minor bump on it to see.
Please fix those. I suggest a merge with latest develop (likely requires a CHANGELOG fix) and running the formatter (format.ps1 script in the root of the repo).

Packages/com.unity.inputsystem/InputSystem/Controls/QuaternionControl.cs
mtschoen-unity: We can't use protected on `Control` classes, because the precompiled layout is an `InputDevice`. It needs to set these properties on controls, but is not itself an extension of the control class.

jamesmcgill:
@mtschoen-unity Could you take a look at the test failures?

mtschoen-unity:
I took a look at the test failures and fixed what I can. The remaining issue seems to be from API validation. I'm a little confused, because it looks like this package does its own API scraping, rather than using the validation package. I thought maybe I needed to update the `.api` file in `Tools`, but it appears to be a snapshot of the 1.0.0 version? Is there some magic script I need to run after bumping the version number?

jamesmcgill:
Looking at the APIVerification tests I found this line which looks very similar to what changed now:
https://github.com/Unity-Technologies/InputSystem/blob/af1009867a0df6861e6a106aa831411016efe759/Assets/Tests/InputSystem/APIVerificationTests.cs#L277

Perhaps creating another list for 1.7.0 changes will fix it, although that might be tedious as there are so many that TestRunner is truncating the list when I run that test.


PVS also runs in the "[Yamato] Publish to Internal Registry (Dry Run)" job. That is passing now and previously wasn't before the v1.7.0 bump.

jamesmcgill:
Thanks for the PR @mtschoen-unity!

PR #1709, opened 2023-07-13
Add UNITY_VISIONOS ifdefs where appropriate

vvuk:
### Description

Add visionOS ifdefs to necessary places to support input devices on visionOS

### Changes made

Added visionOS to some ifdefs alonside IOS. Explicitly did _not_ add them to the CoreMotion ones; CoreMotion is available, but pedometer is not. If we need CoreMotion bits without pedometer we can do that in a future PR.

### Checklist

Before review:

- [X] Changelog entry added.
- [x] No tests added.
- [x] No docs added.


jamesmcgill (approved):
Looks good

PR #1715, opened 2023-07-19
TrackedPoseDriver: Use SetlocalPositionAndRotation instead of 2 seperate calls to improve performance

smitdylan2001:
### DONT MERGE YET
This PR is open for feedback and to get this noticeable improvement to the team.

### Description

XR setups can involve a lot of game objects, especially when doing hand animation etc. Right now the tracked pose driver has 2 extern calls, which both calculate the position and rotation of all children. Combining these will help with performance, without changing any behavior.

### Changes made

Changed Track Pose Driver's SetLocalTransform() method to integrate SetLocalPositionAndRotation when tracking type is to rotation and position.

### Notes

This is supported in [2021.3.11+](https://unity.com/releases/editor/whats-new/2021.3.11) an 2022.3+
It is recommended to add conditionals to manually compile if the feature is supported.
Since you cannot use conditional compilation with the bugfix version in Unity using _OR_NEWER, the asmdef needs to add support.
This can easily be done with the following statements, or by implementing the conditionals for UNITY_2021_3_11_OR_NEWER
```
        {
            "name": "Unity",
            "expression": "[2021.3.11,2022.1)",
            "define": "HAS_POSITION_AND_ROTATION"
        },
        {
            "name": "Unity",
            "expression": "2022.3",
            "define": "HAS_POSITION_AND_ROTATION"
        }
```

```cs
#if HAS_POSITION_AND_ROTATION
//New code
#else
//Old code
#endif
```

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
@@ -581,18 +581,23 @@ namespace UnityEngine.InputSystem.XR
             var positionValid = m_IgnoreTrackingState || (m_CurrentTrackingState & TrackingStates.Position) != 0;
             var rotationValid = m_IgnoreTrackingState || (m_CurrentTrackingState & TrackingStates.Rotation) != 0;
 
-            if (rotationValid &&
-                (m_TrackingType == TrackingType.RotationAndPosition ||
-                 m_TrackingType == TrackingType.RotationOnly))
+            if (rotationValid && positionValid && m_TrackingType == TrackingType.RotationAndPosition)
             {
-                transform.localRotation = newRotation;
+                transform.SetLocalPositionAndRotation(newPosition, newRotation);
             }
-
-            if (positionValid &&
-                (m_TrackingType == TrackingType.RotationAndPosition ||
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/TrackedPoseDriver.cs
```diff
@@ -581,18 +581,23 @@ protected virtual void SetLocalTransform(Vector3 newPosition, Quaternion newRota
...
             }
-
-            if (positionValid &&
-                (m_TrackingType == TrackingType.RotationAndPosition ||
-                 m_TrackingType == TrackingType.PositionOnly))
+            else
             {
-                transform.localPosition = newPosition;
+                if (rotationValid &&
+                m_TrackingType == TrackingType.RotationOnly)
```
chris-massie, line 591: This won't be exactly the correct behavior we want if rotation is valid but position is not when tracking type is RotationAndPosition. I believe I have the correct code we want in another PR https://github.com/Unity-Technologies/InputSystem/pull/1716

I think we can close this PR in favor of that one if you would like.

Thanks again for the improvement suggestion (and for suggesting `Transform.SetLocalPositionAndRotation` itself).

smitdylan2001:
Closed in favor of the better implementation at #1716

PR #1717, opened 2023-07-24
Listen button fix for older unity versions (ISXB-535)

ritamerkl:
### Description

added conditional compilation for line to avoid older unity versions to throw a warning. 

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/Internal/AdvancedDropdown/AdvancedDropdownGUI.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/Internal/AdvancedDropdown/AdvancedDropdownGUI.cs
@@ -10,7 +10,7 @@ namespace UnityEngine.InputSystem.Editor
     {
         private static class Styles
         {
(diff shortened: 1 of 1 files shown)


lyndon-unity (approved):
Seems like a reasonable solution to the issue. Thank you.

sttz:
The condition only works for the specific patch versions checked (2021.3.28 and 2022.3.1), not for any newer patch versions.

I'm running 2022.3.7 and the listen button is broken again, since the `#if` does not pass and `ToolbarSeachTextField` is used again.

The proper fix would to use `UNITY_2021_3_OR_NEWER` and then exclude all versions without the style name change, for 2022 e.g. `!UNITY_2022_1 && !UNITY_2022_2 && !UNITY_2022_3_0`. For 2021 this would have to check all 28 previous patch versions, so maybe a dynamic check would be more sensible?

Issue #1562, opened 2022-07-21
[API Proposal] Global Input Actions and wrapper classes


andrew-oc:
### Background and Motivation

This API proposal adds a simplification layer around Input Actions and introduces the concept of Global Actions, an input actions asset that comes pre-built with the package and sits in the InputManager asset in Project Settings (bit of an implementation detail but that seems the best place to put it). Global actions are enabled by default and remove the need for the user to a) remember to call Enable and b) create a new Input Action Asset before they can even get started with input (along with the requirement to understand all the concepts around Input Actions that go with that).

The API favours the polling approach to input by default and funnels users towards that type of usage. While the original Input Actions API and all their events are always accessible, they are another layer down, and none of the types in this proposal directly expose any events.

It also tries to make a distinction between the concept of Input Actions as an abstraction over a set of controls, and the interaction model of started -> performed -> cancelled, because this often seems to be a source of confusion.

Finally, this API removes the need for users to know what action type an action has. In the existing API, users must call the ReadValue<T> method to retrieve the value of the action, and if the wrong type is provided, a runtime exception is thrown. This is an easy way for uncaught errors to sneak into a product, as the type of an action can be changed in the asset at any time. Using source generators, this API proposes building strongly-typed wrappers around actions.

### Proposed API

```csharp
namespace UnityEngine.InputSystem
{
  public static partial class Input
  {
    public static bool IsPressed(string actionName, string actionMapName = "");
    public static bool WasPressedThisFrame(string actionName, string actionMapName = "");
    public static bool WasReleasedThisFrame(string actionName, string actionMapName = "");
  }

  public class Input<TActionType> where TActionType : struct
  {
    public InputAction action { get; }
    public bool isPressed { get; }
    public bool wasPressedThisFrame { get; }
    public bool wasReleasedThisFrame { get; }
    public TActionType value { get; }

    public Input(InputAction action);

    public bool WasPerformedThisFrame<TInteraction>() where TInteraction:IInputInteraction;
    public bool WasStartedThisFrame<TInteraction>() where TInteraction : IInputInteraction;
    public bool WasEndedThisFrame<TInteraction>() where TInteraction : IInputInteraction;

    public bool SetInteractionParameter<TInteraction, TParameter>(Expression<Func<TInteraction, TParameter>> expr, TParameter value);
    public bool GetInteractionParameter<TInteraction, TParameter>(Expression<Func<TInteraction, TParameter>> expr, out TParameter value);
[...]

PR #1720, opened 2023-08-02
NEW: Project-wide Actions and Device API

jamesmcgill:
### Description

This is a WIP branch that includes:
- Global Actions (Now called Project-wide Actions)
- Device API
- Project-wide Actions API

As it is WIP it remains incomplete of functionality and is untested.
In particular the source generator will be removed in the first phase.

==========================

## Project-wide Actions
In-built set of Actions which is installed (currently inside InputManager.asset) and is available via `InputSystem.actions`
New UITK based editor installed in Project Settings provides a UI for editing these Actions

## Device API
Based on this [proposal](https://github.com/Unity-Technologies/InputSystem/issues/1559)

New API layer with simple functions similar to what InputManager provided
e.g. `Input.IsPressed(Inputs.Key_LeftCtrl)`

## Project-wide Actions API
Based on this [proposal](https://github.com/Unity-Technologies/InputSystem/issues/1562)

New API wrapper with convenience functions around the Project-wide Actions.
e.g. `Input.IsPressed("Fire")`

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/Samples/ProjectWideActionsTest/ProjectWideActionsTest.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        // Project-Wide Actions Direct (InputSystem.actions)
+        if (sprint.isPressed)
+        {
+            cube.GetComponent<Renderer>().material.color = Color.blue;
+        }
+
+        // Project-Wide Actions via API
+        if (Input.IsPressed("Crouch", "Player"))
```
lyndon-unity: If we are removing this API we should change this example and show a pressed example in the source generated type safe api set
E.g. InputActions.player.crouch.isPressed

Assets/TestGlobalActions.cs
```diff
@@ -0,0 +1,81 @@
+#if UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API
```
lyndon-unity, line 1: This whole file looks like WIP code and I think should probably be completely removed

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10353,6 +10376,42 @@ public void Actions_ImprovedShortcutSupport_ConsumesWASD(bool shortcutsEnabled)
         }
     }
 
+    [Test]
+    [Category("Actions")]
+    public void Actions_ActiveBindingIndex_PointsAtRelativeIndexInActionBindingsArray()
```
lyndon-unity, line 10381: Good to see new test but I'm not clear its related to new actions/device API code

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -6172,6 +6172,29 @@ public void Actions_CanRemoveExistingBindingOnAction()
...
+    [Test]
+    [Category("Actions")]
+    public void Actions_CanRemoveExistingInteractionOnBinding()
+    {
+        var action = new InputAction();
+        action.AddBinding("<Gamepad>/buttonSouth", interactions: "hold,tap(duration=0.5),hold(duration=0.5)");
+
+        action.Enable();
+
+        action.ChangeBinding(0).RemoveInteraction<HoldInteraction>();
```
lyndon-unity, line 6184: Seems to be new test for new RemoveInteraction API, so good to see

Assets/Tests/InputSystem/CoreTests_Actions_Rebinding.cs
```diff
@@ -1701,6 +1701,7 @@ public void Actions_CanApplyParameterOverrides_WithOverridesNotGettingLostOnReRe
         map.Enable();
 
         action.ApplyParameterOverride("scale:factor", 0.25f);
+        action.ApplyParameterOverride("hold:duration", 1.5f);
```
lyndon-unity, line 1704: Not sure why this is added but I don't see a problem either

Assets/Tests/InputSystem/CoreTests_GlobalActions.cs
```diff
@@ -0,0 +1,500 @@
...
+        var testInputManager = ScriptableObject.CreateInstance<TestInputManager>();
+        AssetDatabase.CreateAsset(testInputManager, "Assets/TestInputManager.asset");
+
+        // create a template input action asset from which the input action asset stuffed inside the InputManager will be created
+        var globalActions = ScriptableObject.CreateInstance<InputActionAsset>();
+        globalActions.AddActionMap("ActionMapOne").AddAction("ActionOne");
+        m_TemplateAssetPath = Path.Combine(Environment.CurrentDirectory, "Assets/TestGlobalActions.inputactions");
+        File.WriteAllText(m_TemplateAssetPath, globalActions.ToJson());
+
+        InputSystem.SetGlobalActionAssetPaths(m_TemplateAssetPath, "Assets/TestInputManager.asset");
```
lyndon-unity, line 36: This seems to be overriding what the global asset points at.
I'm wondering why we can't just use/test the global asset default bindings

lyndon-unity: I see test below forcing in additional bindings and then triggering specific control activations.
We could potentially change these to expected default binding control activations and remove the custom override.

However this isn't critical for 0.8.0-pre and might require platform specific binding checks

Assets/Tests/InputSystem/CoreTests_GlobalActions.cs
```diff
@@ -0,0 +1,500 @@
...
+    [Category(TestCategory)]
+    public void GlobalActions_CanQueryActionsByStringName()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+
+        InputSystem.actions.FindAction("ActionOne").AddBinding("<keyboard>/w");
+
+        Set(keyboard.wKey, 1);
+
+        Assert.That(Input.WasPressedThisFrame("ActionOne"), Is.True);
```
lyndon-unity: As we are proposing removing these Input.IsPressed/WasPRessetThisFrame/WasReleasedThisFrame functions then this test should be removed too.

Assets/Tests/InputSystem/CoreTests_GlobalActions.cs
```diff
@@ -0,0 +1,500 @@
...
+        Press(keyboard.bKey, currentTime);
+        currentTime += 0.5;
+        InputSystem.Update();
+
+        Assert.That(holdPerformed, Is.False);
+    }
+
+    [Test]
+    [Category(TestCategory)]
+    public void GlobalActions_CanPollInteractionPhaseChanges()
```
lyndon-unity, line 201: I wonder if we should split 'interaction' tests into a separate file. The test themselves seem reasonable

Assets/Tests/InputSystem/CoreTests_GlobalActions.cs
```diff
@@ -0,0 +1,500 @@
...
+            .With("modifier", "<keyboard>/leftShift")
+            .With("binding", "<keyboard>/a");
+        action.AddBinding("<gamepad>/leftStick");
+        action.AddCompositeBinding("2DVector")
+            .With("up", "<keyboard>/w")
+            .With("down", "<keyboard>/s")
+            .With("left", "<keyboard>/a")
+            .With("right", "<keyboard>/d");
+
+        var collection = new InputBindingsCollection(action, enumerationBehaviour);
```
lyndon-unity, line 386: Seems to be testing newly added concept which is good to see.

The indexing below 0,1,2,5,6 isn't obvious to me though.

0 seems to be w
1 seems to be a
2,3,4 are composites (OneModifier, leftstick, a)
5 seems to be gamepad left stick
I would expect 6,7,8,9,10 to be composites  (2dvector, w,s,a,d)

so I am puzzled with the index 6 in the 'skipcompositeparts' list

Assets/Tests/InputSystem/CoreTests_GlobalActions.cs
```diff
@@ -0,0 +1,500 @@
...
+        var action = new InputAction(binding: "<keyboard>/w");
+        action.AddBinding("<keyboard>/a").WithGroup("Gamepad");
+        action.AddCompositeBinding("OneModifier")
+            .With("modifier", "<keyboard>/leftShift")
+            .With("binding", "<keyboard>/a");
+        action.AddBinding("<gamepad>/leftStick").WithGroup("Gamepad");
+
+        var bindingCollection = new InputBindingsCollection(action, InputBindingsCollection.EnumerationBehaviour.SkipCompositeParts);
+        Assert.That(bindingCollection.WithControlScheme("Gamepad").Select(b => b.binding),
+            Is.EqualTo(new[] { action.bindings[1], action.bindings[5] }));
```
lyndon-unity, line 413: Again puzzled why 0 is not in the list

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
+#if UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API
```
lyndon-unity, line 1: Would suggest we renamed the file to CoreTests_DeviceAPI

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
...
+using UnityEngine.TestTools.Constraints;
+using UnityEngine.TestTools.Utils;
+using GamepadButton = UnityEngine.InputSystem.LowLevel.GamepadButton;
+using Input = UnityEngine.InputSystem.Input;
+using Is = NUnit.Framework.Is;
+using Random = UnityEngine.Random;
+
+internal partial class CoreTests
+{
+    // TODO rename HighLevelAPI to something that makes sense
```
lyndon-unity, line 28: Delete comment as we have done that at namespace level

Although perhaps we should change the category to DeviceAPI below

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
...
+    // one perspective is to see this API as an evolution of input manager API, aka input manager 2.0
+    [Test]
+    [Category("HighLevelAPI")]
+    public void HighLevelAPI_CanQueryControl()
```
lyndon-unity, line 32: This whole test can be deleted if we remove Input.IsPressed/WasPressedThisFrame/WasReleasedThisFrame

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
...
+        [FieldOffset(5)] public int buttons;
+
+        public FourCC format => new FourCC('H', 'I', 'D');
+    }
+
+    [Test]
+    [Category("HighLevelAPI")]
+    [TestCase(InputSlot.Slot1)]
+    [TestCase(InputSlot.Slot2)]
+    public void HighLevelAPI_CanQueryGamepadControl(InputSlot slot)
```
lyndon-unity, line 180: As above this can be removed

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
...
+        {
+            Assert.That(Input.WasPressedThisFrame(gamepadButton, gamepadSlot), Is.EqualTo(controlDown));
+            Assert.That(Input.IsPressed(gamepadButton, gamepadSlot), Is.EqualTo(controlPressed));
+            Assert.That(Input.WasReleasedThisFrame(gamepadButton, gamepadSlot), Is.EqualTo(controlUp));
+        }
+    }
+
+    [Test]
+    [Category("HighLevelAPI")]
+    public void HighLevelAPI_CanQuerySpecificJoystick()
```
lyndon-unity, line 246: As above this can be removed

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
...
+        Assert.That(Input.IsPressed(JoystickButton.Button2, InputSlot.Slot2), Is.False);
+        Assert.That(Input.WasReleasedThisFrame(JoystickButton.Button2, InputSlot.Slot2), Is.True);
+
+        InputSystem.Update();
+
+        Assert.That(Input.WasReleasedThisFrame(JoystickButton.Button2, InputSlot.Slot2), Is.False);
+    }
+
+    [Test]
+    [Category("HighLevelAPI")]
```
lyndon-unity, line 272: This is actually DeviceAPI now so should be renamed in category and the prefix

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
...
+        Assert.That(Input.GetAxis(Inputs.Gamepad_West, Inputs.Gamepad_East, Inputs.Gamepad_North, Inputs.Gamepad_East),
+            Is.EqualTo(new Vector2(-0.71f, 0.71f)).Using(new Vector2EqualityComparer(0.01f)));
+
+        // sticks go via different path
+        Assert.That(Input.GetAxis(GamepadAxis.LeftStick), Is.EqualTo(new Vector2(-0.71f, -0.71f)).Using(new Vector2EqualityComparer(0.01f)));
+        Assert.That(Input.GetAxis(GamepadAxis.RightStick), Is.EqualTo(new Vector2(0.71f, 0.71f)).Using(new Vector2EqualityComparer(0.01f)));
+    }
+
+    [Test]
+    [Category("HighLevelAPI")]
```
lyndon-unity, line 323: Renamed to DeviceAPI here and below

Assets/Tests/InputSystem/CoreTests_HighLevelAPI.cs
```diff
@@ -0,0 +1,753 @@
...
+        Input.SetGamepadStickDeadzone(0.3f);
+        Assert.That(Input.GetAxis(GamepadAxis.LeftStick), Is.EqualTo(new Vector2(-0.13f, 0.13f)).Using(new Vector2EqualityComparer(0.01f)));
+        Assert.That(Input.GetAxis(GamepadAxis.RightStick), Is.EqualTo(new Vector2(0.70f, -0.70f)).Using(new Vector2EqualityComparer(0.01f)));
+    }
+
+    [Test]
+    [TestCase("Mouse")]
+    [TestCase("Pen")]
+    [TestCase("Touchscreen")]
+    [Category("HighLevelAPI")]
```
lyndon-unity, line 509: Still DeviceAPI though to here and below so continue renaming

Assets/Tests/InputSystem/Unity.InputSystem.Tests.asmdef
```diff
@@ -40,6 +40,11 @@
             "name": "Unity",
             "expression": "2022.1",
             "define": "TEMP_DISABLE_STANDALONE_OSX_XINPUT_TEST"
+        },
+        {
+            "name": "Unity",
+            "expression": "2020.2",
```
lyndon-unity, line 46: Do we want to enable this support all the way back to 2020.2? 
Is that valid? Does the UI TK feature set for the editor side work that far back?

Packages/com.unity.inputsystem/InputSystem/API/BindingIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+        // Constructor exists so we can create the None BindingIndex
+        private BindingIndex(int index)
+        {
+            m_InputAction = null;
+            m_Value = index;
+            type = IndexType.SkipCompositeParts;
+        }
+
+        /// <summary>
+        /// TODO
```
lyndon-unity, line 71: Need to fill this in

Packages/com.unity.inputsystem/InputSystem/API/BindingIndex.cs
```diff
@@ -0,0 +1,241 @@
...
+
+            if (m_InputAction.GetOrCreateActionMap().bindingsAreContiguous)
+                return m_InputAction.m_BindingsStartIndex + bindingIndex;
+
+            // do a slower lookup by name and id
+            return m_InputAction.BindingIndexOnActionToBindingIndexOnMap(bindingIndex);
+        }
+
+        /// <summary>
+        /// TODO
```
lyndon-unity, line 47: Need to fill this in

Packages/com.unity.inputsystem/InputSystem/API/GlobalInputActions.inputactions
```diff
@@ -0,0 +1,995 @@
...
+                    "name": "Look",
+                    "type": "Value",
+                    "id": "6b444451-8a00-4d00-a97e-f47457f736a8",
+                    "expectedControlType": "Vector2",
+                    "processors": "",
+                    "interactions": "",
+                    "initialStateCheck": true
+                },
+                {
+                    "name": "Attack",
```
lyndon-unity, line 27: Ben gave some feedback about a policy to avoid 'violent' names, proposing to change this to 'Use'
(and we could then drop 'Interact')

Packages/com.unity.inputsystem/InputSystem/API/GlobalInputActions.inputactions
```diff
@@ -0,0 +1,995 @@
...
+                    "name": "Point",
+                    "type": "PassThrough",
+                    "id": "32b35790-4ed0-4e9a-aa41-69ac6d629449",
+                    "expectedControlType": "Vector2",
+                    "processors": "",
+                    "interactions": "",
+                    "initialStateCheck": true
+                },
+                {
+                    "name": "PrimaryAction",
```
lyndon-unity, line 497: Need to check with Marika but UI team gave feedback on the naming that was less obvious that the older 'Click', 'SecondaryClick' etc.

Packages/com.unity.inputsystem/InputSystem/API/GlobalInputActions.inputactions
```diff
@@ -0,0 +1,995 @@
...
+                    "interactions": "",
+                    "processors": "",
+                    "groups": ";Keyboard&Mouse",
+                    "action": "SecondaryAction",
+                    "isComposite": false,
+                    "isPartOfComposite": false
+                }
+            ]
+        }
+    ],
```
lyndon-unity, line 932: Please could you add a screenshot of the default bindings to the PR itself (I find it easier to view)

Packages/com.unity.inputsystem/InputSystem/API/GlobalInputActions.inputactions
```diff
@@ -0,0 +1,995 @@
...
+                    "processors": "",
+                    "groups": "Touch",
+                    "action": "PrimaryAction",
+                    "isComposite": false,
+                    "isPartOfComposite": false
+                },
+                {
+                    "name": "",
+                    "id": "bb9e6b34-44bf-4381-ac63-5aa15d19f677",
+                    "path": "<XRController>/trigger",
```
lyndon-unity, line 889: So we do have some XR Controller bindings and could apply some to the player actions too.
Worth discussing with Marika

lyndon-unity (changes requested):
First review down to 'Input.actions.cs' - submitting to split it up into more manageable feedback chunks.

Mostly looks reasonable to far.
Needs some renaming from HighLevelAPI to DeviceAPI
Removal of some Input.IsPressed/WasPressedThisFrame/WasReleasedThisFrame 'string based' API functions/tests
Some missing documentation in TODO sections
A couple of the binding test indices look odd in a couple of tests

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        internal const string kGlobalActionsAssetName = "GlobalInputActions";
+        internal const string kGlobalActionsAssetConfigKey = "com.unity.inputsystem.globalactionsasset";
+
+        /// <summary>
+        /// True if the specified action is currently pressed.
+        /// </summary>
+        /// <param name="actionMapName"></param>
+        /// <param name="actionName"></param>
+        /// <returns></returns>
+        public static bool IsPressed(string actionName, string actionMapName = "")
```
lyndon-unity: As discussed we are looking to remove these next 3 string based API functions due to 3 key factors
* performance concerns and that the
* type safe API is a cleaner alternative when working
* string based is still viable via the action asset interface for any dynamic needs

lyndon-unity: as actionName can be a just action (E.g. "Move") or full (E.g. "Player/Move") name I really don't see the point of the actionMapName parameter either

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        {
+#if UNITY_EDITOR
+            if (!EditorApplication.isPlayingOrWillChangePlaymode)
+                return;
+
+            InputSystem.actions = GlobalActionsAsset.GetOrCreateGlobalActionsAsset(assetPath, defaultAssetPath);
+#else
+            // find the source generated global actions runtime file in the users assembly and use it to load the
+            // runtime actions
+            var runtimeAssetLoaderType = AppDomain.CurrentDomain.GetAssemblies().SelectMany(a => a.GetExportedTypes())
```
lyndon-unity: I wonder how CoreCLR will impact this code, not a problem to solve for 0.8.0-pre

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+                    // thrown when we try to read from those controls.
+                    return m_Action.ReadValue<TActionType>();
+                }
+                catch (InvalidOperationException ex)
+                {
+                    Debug.LogWarning(ex.Message);
+                }
+
+                return m_Action.ReadDefaultValue<TActionType>(
+                    new BindingIndex(m_Action, m_Action.activeBindingIndex, BindingIndex.IndexType.IncludeCompositeParts)
```
lyndon-unity, line 94: Is this creating any Garbage?
If so perhaps we could create a bindingIndex up front and update the value with the active binding index if needed/changes.

Not critical for 0.8.0-pre

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        /// </summary>
+        /// <param name="action"></param>
+        /// <remarks>
+        /// </remarks>
+        public Input(InputAction action)
+        {
+            Debug.Assert(action != null);
+
+            m_Action = action ?? throw new ArgumentNullException(nameof(action));
+            m_Action.Enable();
```
lyndon-unity, line 120: Ah - referencing the action this way enables it.

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+        public bool WasPerformedThisFrame<TInteraction>()
```
lyndon-unity, line 123: Needs API docs

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        /// for purposes of indexing.
+        /// </summary>
+        /// <remarks>
+        ///
+        /// </remarks>
+        public InputBindingsCollection bindings =>
+            new InputBindingsCollection(m_Action, InputBindingsCollection.EnumerationBehaviour.SkipCompositeParts);
+
+        /// <summary>
+        ///
```
lyndon-unity, line 110: Add simple doc about 'Constructor' ?

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+        public bool WasCanceledThisFrame<TInteraction>()
+            where TInteraction : IInputInteraction
+        {
+            return WasPhaseChangedThisFrame<TInteraction>(BindingIndex.None, state => state.lastCanceledInUpdate);
+        }
+
+        /// <summary>
+        ///
```
lyndon-unity, line 142: Missing documentation

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        /// this method can be made to consider the interaction from a specific binding by passing the 'bindingIndex'
+        /// argument. This index is relative to the bindings of just the Input Action. Most lower-level APIs are relative
+        /// to the Input Action Map in which the binding is contained.</remarks>
+        public bool WasPerformedThisFrame<TInteraction>(BindingIndex bindingIndex)
+            where TInteraction : IInputInteraction
+        {
+            return WasPhaseChangedThisFrame<TInteraction>(bindingIndex, state => state.lastPerformedInUpdate);
+        }
+
+        public bool WasStartedThisFrame<TInteraction>(BindingIndex bindingIndex)
```
lyndon-unity, line 157: Missing docs as the following is too

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+                .RemoveInteraction<TInteraction>();
+        }
+
+        /// <summary>
+        /// Returns the value of the indicated parameter from the first interaction of type TInteraction that
+        /// exists on the bindings of this action.
+        /// </summary>
+        /// <typeparam name="TInteraction"></typeparam>
+        /// <typeparam name="TParameter"></typeparam>
+        /// <param name="expr"></param>
```
lyndon-unity, line 231: This field in particular could do with some explanation

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+        public static implicit operator InputAction(Input<TActionType> input)
+        {
+            return input.action;
+        }
+
+        private unsafe bool WasPhaseChangedThisFrame<TInteraction>(BindingIndex bindingIndex,
+            Func<InputActionState.InteractionState, uint> lastChangedInUpdateGetter)
+            where TInteraction : IInputInteraction
+        {
+            var map = m_Action.GetOrCreateActionMap();
```
lyndon-unity, line 287: Slightly puzzled why 'OrCreateActionMap'
I would expect it to fail/assert if none attached.

Packages/com.unity.inputsystem/InputSystem/API/Input.actions.cs
```diff
@@ -0,0 +1,534 @@
...
+            var bindingsCount = bindings.Count;
+            var bindingStartIndexInMap = m_Action.m_BindingsStartIndex;
+
+            if (bindingIndex != BindingIndex.None)
+            {
+                bindingStartIndexInMap = bindingIndex.ToMapIndex();
+                bindingsCount = 1;
+            }
+
+            for (var i = 0; i < bindingsCount; i++)
```
lyndon-unity, line 299: Doesn't seem the most efficient solution given spinning these lists on every query call to this function

It does lead me to question the performance of the 'simplified' interaction API aspect of this part of the work.

It also makes me wonder if the interactions need to be exposed in this level of the API.

lyndon-unity (commented):
Reviewed further up to Input.cs but pausing again as out of time.

High level summary 
I am wondering if the interactions API additions add enough value for the complexity added.

As this is not a critical baseline for 0.8.0-pre my preference would be we consider shipping without it, but I appreciate this adds time to strip it out.

IIRC some tests were disabled in this area as it wasn't fully working. If thats true then it would add further concerns on releasing this part of the API.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        A = South,
+        B = East,
+        Cross = South,
+        Square = West,
+        Triangle = North,
+        Circle = East
+    }
+
+    /// <summary>
+    /// An enum for addressing gamepad slots.
```
lyndon-unity, line 195: Should say or joysticks

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+            switch (input)
+            {
+                case Inputs.Gamepad_DpadUp: return gamepad.dpad.up;
+                case Inputs.Gamepad_DpadDown: return gamepad.dpad.down;
+                case Inputs.Gamepad_DpadLeft: return gamepad.dpad.left;
+                case Inputs.Gamepad_DpadRight: return gamepad.dpad.right;
+                case Inputs.Gamepad_North: return gamepad.buttonNorth;
+                case Inputs.Gamepad_East: return gamepad.buttonEast;
+                case Inputs.Gamepad_South: return gamepad.buttonSouth;
+                case Inputs.Gamepad_West: return gamepad.buttonWest;
```
lyndon-unity, line 724: Why the different indenting styles - ideally should be consistent.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                    break;
+                case Inputs.Joystick_Button8:
+                    buttonName = "button8";
+                    break;
+                default:
+                    throw new InvalidEnumArgumentException(nameof(input), (int)input, typeof(JoystickButton));
+            }
+
+            // do a non-recursive search for the button based on name
+            for (var i = 0; i < joystick.allControls.Count; i++)
```
lyndon-unity: I'd prefer this to be an array lookup rather than a search for performance if possible, but I expect it would be if viable here.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+
+        private static ButtonControl GetJoystickButtonControl(Joystick joystick, JoystickButton joystickButton)
+        {
+            if (joystick == null)
+                return null;
+
+            string buttonName;
+            switch (joystickButton)
+            {
+                case JoystickButton.Trigger:
```
lyndon-unity: If JoystickButton enum ids matched Input.Joysticks* values in the Input enum we could potentially use the same code as above and delete this function

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                        if (inputDevice is Keyboard keyboard &&
+                            GetKeyboardButtonControl(keyboard, input).isPressed)
+                            return true;
+                    }
+
+                    return false;
+                case InputDeviceType.Mouse:
+                    foreach (var inputDevice in InputSystem.devices)
+                    {
+                        if (inputDevice is Mouse mouse &&
```
lyndon-unity, line 836: Seems the mouse loop and keyboard loops could both use the slot code to allow selection of which device in a consistent way.

I.e. add : 

if (slot != InputSlot.All && (int)slot != i) continue;

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                        if (gamepad == null) continue;
+
+                        if (GetGamepadButtonControl(gamepad, input).ReadValue() >=
+                            GetGamepadTriggerPressPoint((InputSlot)i))
+                            return true;
+                    }
+
+                    return false;
+                case InputDeviceType.Joystick:
+                    for (var i = 0; i < (int)InputSlot.Joystick_Max; i++)
```
lyndon-unity, line 857: I think this should use < s_Joysticks.Length

(+ I'm not sure why we need a Joystick_Max that might not match)

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                case InputDeviceType.Gamepad:
+                    for (var i = 0; i < s_Gamepads.Length; i++)
+                    {
+                        if (slot != InputSlot.All && (int)slot != i) continue;
+
+                        var gamepad = s_Gamepads[i];
+                        if (gamepad == null) continue;
+
+                        if (GetGamepadButtonControl(gamepad, input).ReadValue() >=
+                            GetGamepadTriggerPressPoint((InputSlot)i))
```
lyndon-unity, line 851: I do wonder if we really need GetGamepadTriggerPressPoint per gamepad.
I feel there is an opportunity to gave a generic TriggerPressPoint and use it across all gamepads and potentially all Joysticks rather than the one hardcoded value below

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        /// Is the indicated gamepad button currently pressed.
+        /// </summary>
+        /// <param name="button">A control from the InputGamepadButton enum.</param>
+        /// <param name="slot">Which gamepad to check for input. Default is 'Any'.</param>
+        /// <returns>True if the input is currently held down, false if the input is not pressed,
+        /// or if 'slot' is specified and no gamepad exists in that slot.</returns>
+        public static bool IsPressed(InputGamepadButton button, InputSlot slot = InputSlot.All)
+        {
+            if (slot != InputSlot.All)
+                return s_Gamepads[(int)slot] != null &&
```
lyndon-unity: I'd suggest splitting into 2 lines like the joystick variant below to make this slightly more readable.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        public static bool WasPressedThisFrame(Inputs input, InputSlot slot = InputSlot.All)
+        {
+            var deviceType = GetDeviceTypeForInput(input);
+
+            switch (deviceType)
+            {
+                case InputDeviceType.Keyboard:
+                    foreach (var inputDevice in InputSystem.devices)
+                    {
+                        if (inputDevice is Keyboard keyboard &&
```
lyndon-unity, line 899: As above add slot check on keyboard and mouse

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                var control = GetJoystickButtonControl(joystick, button);
+                return control != null && control.isPressed;
+            }
+
+            for (var i = 0; i < (int)InputSlot.Joystick_Max; i++)
+            {
+                if (s_Joysticks[i] == null) continue;
+
+                var control = GetJoystickButtonControl(s_Joysticks[i], button);
+                if (control != null && control.isPressed)
```
lyndon-unity: Lacks the press point logic seen above in the variant that takes "Inputs" as 1st parameter. Should make this threshold consistent.

I.e. this bit:
                        if (button != null && button.ReadValue() >= InputSystem.settings.defaultButtonPressPoint)
                            return true;

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                    }
+
+                    return false;
+                case InputDeviceType.Joystick:
+                    for (var i = 0; i < (int)InputSlot.Joystick_Max; i++)
+                    {
+                        if (slot != InputSlot.All && (int)slot != i) continue;
+
+                        var button = GetJoystickButtonControl(s_Joysticks[i], input);
+                        if (button != null && button.wasPressedThisFrame)
```
lyndon-unity, line 937: Lacks the press point logic seen in the IsPressed variant

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        /// when 'slot' is specified and no joystick is connected to that slot.</returns>
+        public static bool WasPressedThisFrame(JoystickButton button, InputSlot slot = InputSlot.All)
+        {
+            if (slot != InputSlot.All)
+            {
+                var joystick = s_Joysticks[(int)slot];
+                if (joystick == null)
+                    return false;
+
+                var control = GetJoystickButtonControl(joystick, button);
```
lyndon-unity: Again lacks the press point logic see in is pressed

var button = GetJoystickButtonControl(s_Joysticks[i], input);
                        if (button != null && button.ReadValue() >= InputSystem.settings.defaultButtonPressPoint)
                            return true;

Potentially could save this till after 0.8.0-pre as the deadzone/trigger aspects needs refinement.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        public static bool WasReleasedThisFrame(Inputs input, InputSlot slot = InputSlot.All)
+        {
+            var deviceType = GetDeviceTypeForInput(input);
+
+            switch (deviceType)
+            {
+                case InputDeviceType.Keyboard:
+                    foreach (var inputDevice in InputSystem.devices)
+                    {
+                        if (inputDevice is Keyboard keyboard &&
```
lyndon-unity, line 974: Add the slot id check as above

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                    }
+
+                    return false;
+                case InputDeviceType.Joystick:
+                    for (var i = 0; i < (int)InputSlot.Joystick_Max; i++)
+                    {
+                        if (slot != InputSlot.All && (int)slot != i) continue;
+
+                        var button = GetJoystickButtonControl(s_Joysticks[i], input);
+                        if (button != null && button.wasReleasedThisFrame)
```
lyndon-unity, line 1012: Missing trigger threshold seen in the IsPressed variant.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        /// True in the frame that the button was released.
+        /// </summary>
+        /// <param name="button">A control from the GamepadButton enum.</param>
+        /// <param name="slot">Which gamepad to check for input. Default is 'Any'.</param>
+        /// <returns>True if the input was released in the current frame, otherwise false.
+        /// Also returns false if 'slot' is specified and no gamepad is connected to that slot.</returns>
+        public static bool WasReleasedThisFrame(InputGamepadButton button, InputSlot slot = InputSlot.All)
+        {
+            if (slot != InputSlot.All)
+                return s_Gamepads[(int)slot] != null &&
```
lyndon-unity: Suggest splitting into 2 lines for readability (like joystick function below)

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        {
+            var deviceType = GetDeviceTypeForInput(input);
+            var maxValue = 0.0f;
+
+            switch (deviceType)
+            {
+                case InputDeviceType.Keyboard:
+                    foreach (var device in InputSystem.devices)
+                    {
+                        if (!(device is Keyboard keyboard)) continue;
```
lyndon-unity, line 1046: Add slot id check like gamepad (and do to mouse version too)

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+        /// </summary>
+        /// <param name="left"></param>
+        /// <param name="right"></param>
+        /// <param name="up"></param>
+        /// <param name="down"></param>
+        /// <returns></returns>
+        public static Vector2 GetAxis(Inputs left, Inputs right, Inputs up, Inputs down)
+        {
+            var axis = GetAxisRaw(left, right, up, down);
+            return NormalizeAxis(axis, InputSystem.settings.defaultDeadzoneMin);
```
lyndon-unity: Suggest we pass optional deadzone into this function and use the default if non passed.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+            foreach (var frame in s_GamepadsDisconnectedFrames)
+            {
+                if (frame != Time.frameCount)
+                    return false;
+            }
+
+            return true;
+        }
+
+        private static float GetGamepadTriggerPressPoint(InputSlot gamepadSlot)
```
lyndon-unity, line 1317: Do we really need this per slot ?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+
+        /// <summary>
+        /// Set the value at which gamepad triggers need to be actuated before they will be considered pressed.
+        /// </summary>
+        /// <param name="pressPoint"></param>
+        /// <param name="gamepadSlot"></param>
+        /// <remarks>
+        /// If this is set and a gamepad subsequently connects, the set values should also apply to that gamepad.
+        /// </remarks>
+        public static void SetGamepadTriggerPressPoint(float pressPoint, InputSlot gamepadSlot = InputSlot.All)
```
lyndon-unity, line 1349: Again do we really need per slot?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                case InputSlot.All:
+                    for (var i = 0; i < s_GamepadConfigs.Length; ++i)
+                        s_GamepadConfigs[i].TriggerPressPoint = pressPoint;
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(gamepadSlot), gamepadSlot, null);
+            }
+        }
+
+        private static float GetGamepadStickDeadZone(InputSlot gamepadSlot)
```
lyndon-unity, line 1376: And again does this really make sense per slot?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                    throw new ArgumentOutOfRangeException(nameof(gamepadSlot), gamepadSlot, null);
+            }
+        }
+
+        /// <summary>
+        /// Set the stick deadzone for both gamepad sticks.
+        /// </summary>
+        /// <param name="deadzone"></param>
+        /// <param name="gamepadSlot"></param>
+        public static void SetGamepadStickDeadzone(float deadzone, InputSlot gamepadSlot = InputSlot.All)
```
lyndon-unity, line 1405: Same as above - do we need per slot?

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
lyndon-unity: ^ ignore this - the array is initialised to Joystick_Max , so arguably it could just be s_Joysticks.Length

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+#endif
+            s_PointerPosition = Vector2.zero;
+            s_ScrollDelta = Vector2.zero;
+
+            s_PointerAction = new InputAction("Pointer", InputActionType.PassThrough, expectedControlType: "Vector2");
+            s_PointerAction.AddBinding("<Pointer>/position");
+            s_PointerAction.performed += OnPointerMoved;
+            s_PointerAction.Enable();
+
+            s_ScrollAction = new InputAction("Scroll", InputActionType.PassThrough, expectedControlType: "Vector2");
```
lyndon-unity, line 1506: I'm wondering if this forces mouse processing even if we never use it.
Something to check on performance side later.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+                if (s_Gamepads[i] != device) continue;
+
+                s_Gamepads[i] = null;
+
+                var frameCount = Time.frameCount;
+#if UNITY_EDITOR
+                if (!s_TimeHasUpdatedThisFrame)
+                    frameCount += 1;
+#endif
+                s_GamepadsDisconnectedFrames[i] = frameCount;
```
lyndon-unity, line 1591: Should we be clearing the s_GamepadsConnectedFrames[i[ value too.
I'm wondering if we would get a bogus connect on frame wrapping the max counter, although that would take along time for an int value so unlikely to occur in practice.

(likewise above clearing s_GamepadsDisconnectedFrames on a connection.

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
```diff
@@ -0,0 +1,1862 @@
...
+
+        private static void AddJoystickToFirstFreeSlot(InputDevice device)
+        {
+            if (!(device is Joystick joystick)) return;
+
+            for (var i = 0; i < (int)InputSlot.Joystick_Max; i++)
+            {
+                if (s_Joysticks[i] != null) continue;
+
+                s_Joysticks[i] = joystick;
```
lyndon-unity, line 1604: This basic add/remove to free slot feels like it only preserves the joystick index for simple single device add/remove and would fail to preserve index if I remove one device, add a different one, then add my old device back.

Packages/com.unity.inputsystem/InputSystem/API/InputBindingsCollection.cs
```diff
@@ -0,0 +1,203 @@
...
+                    throw new ArgumentOutOfRangeException(nameof(index));
+
+                if (m_EnumerationBehaviour == EnumerationBehaviour.SkipCompositeParts)
+                {
+                    var currentIndex = 0;
+                    var i = 0;
+                    while (i < m_Bindings.Count)
+                    {
+                        if (currentIndex == index)
+                            return new BindingWithIndex(m_Bindings[currentIndex],
```
lyndon-unity, line 63: This line is tricky to parse, perhaps it can be made simpler to read, at minimum currentIndex could be replaced with index for less text as they are equal here.

Packages/com.unity.inputsystem/InputSystem/API/InputBindingsCollection.cs
```diff
@@ -0,0 +1,203 @@
...
+                // been added or deleted.
+                if (m_InitialBindingsCount != m_InputAction.bindings.Count)
+                    throw new InvalidOperationException("The collection was modified after the enumerator was created.");
+
+                if (m_CurrentWithCompositeParts >= m_Bindings.Count)
+                    return false;
+
+                if (m_EnumerationBehaviour == EnumerationBehaviour.SkipCompositeParts)
+                {
+                    // ++m_Current;
```
lyndon-unity, line 129: Remove commented out lines

Packages/com.unity.inputsystem/InputSystem/API/InputBindingsCollection.cs
```diff
@@ -0,0 +1,203 @@
...
+                    while (++m_CurrentWithCompositeParts < m_Bindings.Count)
+                    {
+                        if (m_Bindings[m_CurrentWithCompositeParts].isPartOfComposite)
+                            continue;
+
+                        ++m_Current;
+                        return true;
+                    }
+
+                    // if (m_CurrentWithCompositeParts == m_Bindings.Count && m_Bindings[m_CurrentWithCompositeParts - 1].isPartOfComposite)
```
lyndon-unity, line 140: Delete old code

Packages/com.unity.inputsystem/InputSystem/API/InputBindingsCollection.cs
```diff
@@ -0,0 +1,203 @@
...
+                return m_CurrentWithCompositeParts != m_Bindings.Count;
+            }
+
+            public void Reset()
+            {
+                m_Current = -1;
+            }
+
+            public BindingWithIndex Current =>
+                new BindingWithIndex(
```
lyndon-unity, line 154: Quite a long line but it does make sense. Would be nice if it could be easier to read

Packages/com.unity.inputsystem/InputSystem/API/InputBindingsCollection.cs
```diff
@@ -0,0 +1,203 @@
...
+        public InputBinding binding { get; }
+        public BindingIndex index { get; }
+    }
+
+    public static class InputBindingsCollectionExtensions
+    {
+        public static IEnumerable<BindingWithIndex> WithControlScheme(
+            this InputBindingsCollection collection, string controlScheme)
+        {
+            return collection.Where(b =>
```
lyndon-unity, line 198: Consider simplifying to remove LINQ usage

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -608,6 +608,39 @@ public unsafe InputControl activeControl
             }
         }
 
+        /// <summary>
+        /// Get the index of the binding within this input action that last caused it to fire.
+        /// </summary>
+        /// <remarks>
+        /// Note that this index is relative to the Input Action bindings.
+        ///
+        /// TODO: Describe relative indexing and ignoring of composite part bindings, and also suggest using callbacks
```
lyndon-unity, line 617: This needs updating

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1614,7 +1662,7 @@ internal int BindingIndexOnMapToBindingIndexOnAction(int indexOfBindingOnMap)
             {
                 ref var binding = ref bindingsInMap[i];
 
-                if (string.Compare(binding.action, actionName, StringComparison.InvariantCultureIgnoreCase) == 0 ||
+                if (string.Compare(binding.action, actionName, StringComparison.OrdinalIgnoreCase) == 0 ||
```
lyndon-unity, line 1669: Unclear why this change is made?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
```diff
@@ -1245,9 +1291,9 @@ public BindingSyntax WithInteraction<TInteraction>()
                 if (!valid)
                     throw new InvalidOperationException("Accessor is not valid");
 
-                var interactionName = InputProcessor.s_Processors.FindNameForType(typeof(TInteraction));
+                var interactionName = InputInteraction.s_Interactions.FindNameForType(typeof(TInteraction));
```
lyndon-unity, line 1294: This looks like an important fix to used interactions list rather than process or list. Looks like this never worked before.

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
```diff
@@ -751,7 +751,7 @@ public string ToDisplayString(DisplayStringOptions options = default, InputContr
         internal bool TriggersAction(InputAction action)
         {
             // Match both name and ID on binding.
-            return string.Compare(action.name, this.action, StringComparison.InvariantCultureIgnoreCase) == 0
+            return string.Compare(action.name, this.action, StringComparison.OrdinalIgnoreCase) == 0
```
lyndon-unity, line 754: Another change that isn't clear why

Packages/com.unity.inputsystem/InputSystem/Controls/Paths/FastPathParser.cs
```diff
@@ -0,0 +1,106 @@
+using UnityEngine.InputSystem.Utilities;
+
+namespace UnityEngine.InputSystem
+{
+#if UNITY_2020_1_OR_NEWER
+    // NOTE: Must not allocate!
```
lyndon-unity, line 6: Question : Do we have any tests that check this is true

Packages/com.unity.inputsystem/InputSystem/Controls/Paths/FastPathParser.cs
```diff
@@ -0,0 +1,106 @@
...
+                if (leftIndexInPath == length)
+                    return false;
+            }
+
+            // Parse <...> layout part, if present.
+            var layout = new Substring();
+            if (rightIndexInPath < length && path[rightIndexInPath] == '<')
+                layout = ParseComponentPart('>');
+
+            ////FIXME: with multiple usages, this will allocate
```
lyndon-unity, line 54: This comment is concerning

lyndon-unity (commented):
Reviewed up to GlobalActionsAsset.cs

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -660,6 +660,52 @@ public bool disableRedundantEventsMerging
...
+                m_DisableHighLevelApi = value;
+                OnChange();
+            }
+        }
+
+        /// <summary>
+        /// The global set of actions.
+        /// </summary>
+        /// <remarks>
+        /// TODO
```
lyndon-unity: Just remove the TODO at a minimum

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3513,6 +3536,13 @@ internal static void OnPlayModeChange(PlayModeStateChange change)
                     // Nuke all InputActionMapStates. Releases their unmanaged memory.
                     InputActionState.DestroyAllActionMapStates();
 
+#if UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API
+                    if (!settings.disableHighLevelAPI)
```
lyndon-unity: Should this potentially always call shutdown in case someone has toggled this off after initialise?

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
```diff
@@ -0,0 +1,387 @@
...
+
+// Code formatter does not like raw literals
+//*begin-nonstandard-formatting*
+        public void Execute(GeneratorExecutionContext context)
+        {
+            // only run at build time
+            if (!IsBuildTime)
+                return;
+
+            // TODO(JIM): Disabling this for now as it creates a chicken and egg scenario. Types are not available if not used,
```
lyndon-unity: Consider just deleting this code.

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
```diff
@@ -90,6 +90,12 @@ public virtual void Setup()
 
                 runtime = new InputTestRuntime();
 
+#if UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API
+                // we want to run most tests without the high level API being enabled because it interferes with
```
lyndon-unity, line 94: [CRITICAL] This 'interfering with global states' is concerning!

This seems a risk to landing it / shipping

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Unity.InputSystem.SourceGenerators.dll
lyndon-unity, line 1: Should this be committed or generated in a build step before shipping ?
In 0.8.0-pre I believe we would not merge this file so we don't get the type safe API

lyndon-unity (changes requested):
Finished review - although source generator not well reviewed.

Concerned about high level API needing to be disabled for tests to pass.
I believe that is the following task in the sprint
[ISX-1455](https://jira.unity3d.com/browse/ISX-1455)

I see it as a risk to ship with that needing to be disabled as it shows the high level API impacts outside its own API.

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1720) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1720) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1720) it.</sub>

Packages/com.unity.inputsystem/InputSystem/API/Input.cs
BraveSirAndrew: I was thinking about local multiplayer games where each user might want different sensitivities. Also hi:)

jamesmcgill:
Project-wide actions work moved to: https://github.com/Unity-Technologies/InputSystem/pull/1735

PR #1722, opened 2023-08-04
FIX: exception on renaming control scheme (ISX-1544)

ritamerkl:
### Description

This PR fixes the exception captured in this [ticket](https://jira.unity3d.com/browse/ISX-1544).

### Changes made

provided old name to compare control scheme with serialized properties

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
@@ -31,7 +31,7 @@ namespace UnityEngine.InputSystem.Editor
                     state.selectedControlScheme.deviceRequirements.Where((r, i) => i != selectedDeviceIndex)));
             };
         }
-
+        
         public static Command SaveControlScheme(string oldName = "", bool updateExisting = false)
         {
             return (in InputActionsEditorState state) =>
@@ -39,6 +39,7 @@ namespace UnityEngine.InputSystem.Editor
                 var controlSchemeName = state.selectedControlScheme.name;
 
                 var controlSchemesArray = state.serializedObject.FindProperty(nameof(InputActionAsset.m_ControlSchemes));
+                //use the old name to find the control scheme in the array
                 var controlScheme = controlSchemesArray
                     .FirstOrDefault(sp => sp.FindPropertyRelative(nameof(InputControlScheme.m_Name)).stringValue == (string.IsNullOrEmpty(oldName) ? controlSchemeName : oldName));
 
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
@@ -32,7 +32,7 @@ namespace UnityEngine.InputSystem.Editor
             };
         }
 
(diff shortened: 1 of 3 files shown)


jamesmcgill (approved):
Functionally it looks correct but it's not really clear here what is going on or why.

I don't understand what `m_OldName` is doing. Is it the `original` name or `initial` name of the control scheme?
Why would it be empty during save that it needs patched up?

ritamerkl:
added some explanatory comments @jamesmcgill

jamesmcgill:
Things to consider:
- maybe it could be called `OriginalName` as that is how the comment clarifies it?
- maybe you don't need to to have conditional logic `IsNullOrEmpty` and just always just find by `OriginalName`. It should always have a name at the start. (This would remove the need to empty the variable as that might be fragile).
- or maybe there is some way to avoid the new temp variable completely (ensure the renaming got saved so the lookup always works or storing the original name in the state somewhere). Not sure that's possible.

ritamerkl:
I refactored a bit. I changed it to storing the newName to make it simpler to not overwrite the old name if save is not pressed. Also I moved the Close method to the ControlSchemesView as the only usages of it are there. 
Unfortunately I will have to empty the variable OnClose because there I see no other (uncomplicated, clean) way to restore the oldName.
I tried one other way to avoid using the variable: that was to compare control schemes by Id's instead of their names (as I think this is also cleaner), it turned out to be completely over-engineered for such a small problem with a lot of overhead. I also did not want to bloat up the state with various variables to keep it clear. 
If you have another concrete idea I am open for it :) @jamesmcgill

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
```diff
@@ -32,13 +32,14 @@ public static Command RemoveDeviceRequirement(int selectedDeviceIndex)
...
 
-        public static Command SaveControlScheme(bool updateExisting = false)
+        public static Command SaveControlScheme(string newName = "", bool updateExisting = false)
         {
             return (in InputActionsEditorState state) =>
             {
                 var controlSchemeName = state.selectedControlScheme.name;
 
                 var controlSchemesArray = state.serializedObject.FindProperty(nameof(InputActionAsset.m_ControlSchemes));
+                //use the old name to find the control scheme in the array
```
jamesmcgill: Unneeded now

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
```diff
@@ -72,6 +73,8 @@ public static Command SaveControlScheme(bool updateExisting = false)
 
                 state.serializedObject.ApplyModifiedProperties();
 
+                if (!string.IsNullOrEmpty(newName))
+                    return state.With(selectedControlScheme: new InputControlScheme(controlScheme));
```
jamesmcgill: Should `newName` be used here, if it was tested in the line above?

jamesmcgill (approved):
Looks better now I think.
Still don't like having to remember to set to "" on close, but if there is no alternative then so be it.

PR #1683, opened 2023-05-26
RELEASE: 1.6.1

jamesmcgill:
### Description

Release v1.6.1 with fix for compiler errors in 2022.1 editors and with XRI toolkit

### Changes made

Bump to v1.6.1
Hide Experimental UITK code completely behind compiler define

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #1723, opened 2023-08-04
FIX: Remove errors when using UITK Asset Editor (ISX-1531)

jfreire-unity:
### Description

[ISX-1531](https://jira.unity3d.com/browse/ISX-1531): This PR fixes errors for currently supported Unity versions that do not support UITK Asset Editor features we are using.

### Changes made

- Implemented a stub for the default implementation of `InputParameter<T>.OnDrawVisualElements` to avoid breaking the code of packages that have classes that derive from `InputParameter<T>`. Otherwise we would have a breaking change.

### Notes

Questions: 
- ~This doesn't fix problems for releasing. There are still issues for with XR Interaction Tool Kit. More comments in [ISX-1531](https://jira.unity3d.com/browse/ISX-1531) for the options we have. cc @lyndon-unity @jamesmcgill @ritamerkl~ It fixes it now :)
- Should UITK be enabled as well by default? Right now seems to be behind a [feature flag](https://github.com/Unity-Technologies/InputSystem/blob/132deb5acd63fa149a356c6f1199aa06a59a5f54/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs#L20) disabled by default.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/InputParameterEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/InputParameterEditor.cs
@@ -182,6 +182,17 @@ namespace UnityEngine.InputSystem.Editor
             OnEnable();
         }
 
+#if UNITY_INPUT_SYSTEM_UI_TK_ASSET_EDITOR
+        /// <summary>
+        /// Default stub implementation of <see cref="InputParameterEditor.OnDrawVisualElements"/>.
+        /// Should be overridden to create the desired UI.
+        /// </summary>
+        public override void OnDrawVisualElements(VisualElement root, Action onChangedCallback)
+        {
+        }
+
+#endif
+
         /// <summary>
         /// Helper for parameters that have defaults (usually from <see cref="InputSettings"/>).
         /// </summary>


ritamerkl (approved):
LGTM

jamesmcgill:
Questions
1) This PR is essentially reverting https://github.com/Unity-Technologies/InputSystem/pull/1683

But we needed that to get the 1.6 release out. So I we can't merge this into develop just now as it would cause the same issues when we want to do a release (and v1.7 is in flight soon I believe).

The issues with the release and XR toolkit would need to be investigated and fixed first.

2) It should be enabled by default, yes. However develop should not have any functional changes visible to users. 
Not until the global actions feature lands, and we move to 1.8-prerelease then we can make those changes in develop.

That work (and other compiler flag work) should be done on https://github.com/Unity-Technologies/InputSystem/tree/api/global-actions-source-generators-wip

jfreire-unity:
> Questions
> 
> 1. This PR is essentially reverting [RELEASE: 1.6.1 #1683](https://github.com/Unity-Technologies/InputSystem/pull/1683)
> 
> But we needed that to get the 1.6 release out. So I we can't merge this into developing just now as it would cause the same issues when we want to do a release (and v1.7 is in flight soon I believe).
> 
> The issues with the release and XR toolkit would need to be investigated and fixed first.

I see. Even if we still have `UNITY_INPUT_SYSTEM_UI_TK_ASSET_EDITOR` enabled in a .asmdef in our package, it will still break XR toolkit when releasing. So meanwhile we can only land this kind of work in a branch with work for 1.8-pre release / or not merge in `develop` for now (my preference).

> 2. It should be enabled by default, yes. However develop should not have any functional changes visible to users.
>    Not until the global actions feature lands, and we move to 1.8-pre release then we can make those changes in develop.
> That work (and other compiler flag work) should be done on https://github.com/Unity-Technologies/InputSystem/tree/api/global-actions-source-generators-wip

Follows my previous comment. This kind of work can only land when we are close to making the 1.8-pre release. Otherwise, we're blocked from releasing anything from `develop` until the XR Toolkit issues are fixed.

jfreire-unity:
@jamesmcgill pointed out we had a potential breaking change due to the requirement of the derived class to override the abstract method `InputParameterEditor.OnDrawVisualElements`. I added a default implementation in the generic class to overcome this problem and re-introduced the define `UNITY_INPUT_SYSTEM_UI_TK_ASSET_EDITOR`

lyndon-unity:
I'm glad to see the default implementation so code overriding this no longer breaks

"Should UITK be enabled as well by default? Right now seems to be behind a feature flag"
Agree with James. This will need to be enabled by default on 1.8.0-pre, but NOT on develop yet as we need a 1.7.0 release first where its not active.
Also we should enable UI TK editor for the project-wide settings but NOT input assets in 1.8.0-pre so we can get some wider user testing before fixing the input asset specific features.
Overall this should mean this PR is valid for develop now. (Assuming XR ITK works with it)

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -70,6 +71,11 @@
             "name": "Unity",
             "expression": "2022.2",
             "define": "HAS_SET_LOCAL_POSITION_AND_ROTATION"
+        },
+        {
+            "name": "Unity",
+            "expression": "2022.3",
+            "define": "UNITY_INPUT_SYSTEM_UI_TK_ASSET_EDITOR"
```
jamesmcgill: Let remove this for now. We don't want to enable any behaviour on `develop`.

However on the Project-wide actions branch it should be enabled by default there.
And just directly with `#if UNITY_VERSION_**` rather than the extra symbol in the asmdef.

Packages/com.unity.inputsystem/InputSystem/Editor/InputParameterEditor.cs
```diff
@@ -182,6 +182,17 @@ internal override void SetTarget(object target)
             OnEnable();
         }
 
+#if UNITY_INPUT_SYSTEM_UI_TK_ASSET_EDITOR
+        /// <summary>
+        /// Default stub implementation of <see cref="InputParameterEditor.OnDrawVisualElements"/>.
+        /// Should be overridden to create the desired UI.
+        /// </summary>
+        public override void OnDrawVisualElements(VisualElement root, Action onChangedCallback)
```
jamesmcgill, line 190: Great that you figured out a solution to the XR toolkit issue we had with the v1.6.0 release

jamesmcgill (changes requested):
Let's put the fix for XR toolkit in develop, but after Paulius does the 1.7.0 release.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
jfreire-unity: Agreed, I'll remove it for now in `develop` 👍

jamesmcgill (approved):
Great! Please hold off merging until Paulius is finished with the release.

PR #1726, opened 2023-08-11
Compact InputActionProperty drawer

chris-massie:
### Description

The property drawer for `InputActionProperty` takes up a lot of room, using 3 lines when using a reference doesn't make optimal use of the Inspector window when there are a lot of serialized properties. This adds a new default method of drawing them to make them much shorter, shrinking down to a single line when using a reference.

For a motivating example, look at the before and after screenshots of a component in XR Interaction Toolkit:
Before
After

### Changes made

Added a new setting to the **Edit** > **Project Settings** > **Input System Package** window to choose between three modes: Compact, Multiline Effective, Multiline Both. Compact is the new default as shown above. Multiline Effective is the original. Multiline Both is a way to show both the underlying values, which is helpful when comparing prefab overrides or if you want to view them without having to toggle the value, dirtying the scene.

Compact adds a `⋮` button that shows a dropdown menu to choose between Use Reference and Use Action instead of having a dedicated line for that sub property. This style matches the format used for datum drawers in XRI/XR Core Utilities.
Popup menu

### Notes

For comparison between the 3 modes, an example Tracked Pose Driver.
**Compact**, showing all references and all embedded actions:
TPD Compact - References
TPD Compact - Actions

**Multiline Effective** (old method):
TPD MultilineEffective - References
TPD MultilineEffective - Actions

**Multiline Both**:
TPD MultilineBoth

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


chris-massie:
I'm waiting to add a changelog entry until after 1.7.0 is finished releasing.

lyndon-unity:
Added William for UX input.
Note we are looking at switching over to a UI Toolkit implementation soon so would prefer to land UI changes in the newer UI backend.

Billreyn:
Interesting. Here are my initial thoughts:

  - Why is there a need for the Use Reference checkboxes?
  - I assume the need to pick action references could go away with Global Actions?
  - What does the Use Action / Use Reference options do?
  - Why would we need 3 layout options for this stuff?

Overall, my thought is to simplify, streamline & uncomplicate.

chris-massie:
> Interesting. Here are my initial thoughts:
> 
>     * Why is there a need for the Use Reference checkboxes?
> 
>     * I assume the need to pick action references could go away with Global Actions?
> 
>     * What does the Use Action / Use Reference options do?
> 
>     * Why would we need 3 layout options for this stuff?
> 
> 
> Overall, my thought is to simplify, streamline & uncomplicate.

**Why is there a need for the Use Reference checkboxes?**
**What does the Use Action / Use Reference options do?**
The `InputActionProperty` allows a MonoBehaviour to either reference some input action in an Input Action Asset (the Use Reference option), or directly serialize an input action on the component itself (the Use Action option). "Use Reference" and having your input actions in an asset is usually the preferred way for users to read an input value. But we've had users say they like serializing an input action directly on the component for quick prototyping.

The reason we serialize with type `InputActionProperty` in many behaviors is to give users the flexibility of either method, giving them the choice of reference (instead of only serializing `InputActionReference` in a behavior) or action (instead of only serializing `InputAction` in a behavior).

By convention in many XR behaviors that have `InputActionProperty` values, if it's a Use Action (directly serialized) type, we automatically enable and disable the input action with the enable and disable of the behavior itself, which simplifies prototyping. For Use Reference, we leave that up to another script to manage the lifecycle of enabling/disabling actions in an Input Action Asset (and note that Input System has them all disabled by default). But as a project scales up in complexity, and for the future of OpenXR input, it's usually better to have the actions all in one or more Input Action Assets so you can see and configure them in a central location and enable/disable actions in mass with maps or with control schemes.

**I assume the need to pick action references could go away with Global Actions?**
I would assume a behavior would still need to create a stable reference to an input action even if there's one global list. If you had to get it by a string name, that would be more fragile than having something that serializes the GUID of the action you want to read from.

**Why would we need 3 layout options for this stuff?**
We don't need to have 3, Compact is probably sufficient for most people. But I added the option to allow people to revert to how it was drawn before with `Multiline Effective` (where it only shows the active one that would be returned by the [`InputActionProperty.action`](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.6/api/UnityEngine.InputSystem.InputActionProperty.html#UnityEngine_InputSystem_InputActionProperty_action) property). And a 3rd option `Multiline Both` if you want to compare prefab overrides easier in the inspector or the Overrides dropdown in the Unity Editor if you have a mix of ones that override Use Reference so you can see all the changes side-by-side. It also makes it easier to see the value of the serialized input action (Use Action) on a bunch of them if they are all set to Use Reference without having to click one-by-one to change them, and then revert the scene override as you modify that serialized setting just so you can see both values.

I have found myself at least switching between Compact and Multline Both depending on what I'm doing.

chris-massie:
> Note we are looking at switching over to a UI Toolkit implementation soon so would prefer to land UI changes in the newer UI backend.

According to [PropertyDrawer](https://docs.unity3d.com/ScriptReference/PropertyDrawer.html) in the docs:
>You cannot have UI Toolkit running inside IMGUI, which means that if your custom PropertyDrawer only has a UI Toolkit implementation, it will not work inside an IMGUI custom Inspector or a parent IMGUI custom PropertyDrawer. In Unity 2022.2 and above, the default Inspector uses UI Toolkit exclusively in custom PropertyDrawers. Prior to 2022.2, it is recommended that you either implement both IMGUI and UI Toolkit versions of each PropertyDrawer, or make sure they are exclusively used inside custom UI Toolkit inspectors.

So if we want to allow it to use UI Toolkit, we would need to implement that method in addition to IMGUI.

lyndon-unity:
> Added William for UX input. Note we are looking at switching over to a UI Toolkit implementation soon so would prefer to land UI changes in the newer UI backend.

Given that this is code in the normal inspector window my comment is redundant here. Thanks for the extra information your side Chris.

Billreyn:
OK so I finally had time to look at this again...

I don't want to block progress too much here. However, I guess I am not fully convinced that this is going to feel simpler to users in aggregate, as long as there is now an extra setting to choose between various list layouts. It's worsened by the fact that this option lives somewhere different from the UI itself. 

So my thought is that we could move ahead but preferably in a way that avoids the need for that setting to be there. 

Separately to this PR, what we really would like, is to work much more closely with you all on the XR/input stuff. We feel like there are huge opportunities to vastly simplify and really make using input with XR simple and straight forward.

chris-massie:
> OK so I finally had time to look at this again...
> 
> I don't want to block progress too much here. However, I guess I am not fully convinced that this is going to feel simpler to users in aggregate, as long as there is now an extra setting to choose between various list layouts. It's worsened by the fact that this option lives somewhere different from the UI itself.
> 
> So my thought is that we could move ahead but preferably in a way that avoids the need for that setting to be there.
I would imagine almost all users won't bother to change the setting from the default. We could remove the Multiline Effective option since it doesn't add any advantage over the Compact mode, simplifying the setting to just two options, but I kept it for any users that prefer or need that old behavior for any reason.

To address your reservations about adding a setting, we could hide it from the Edit > Project Settings window and just keep it as an API option or available to users who right-click the Inspector window tab of the `.inputsettings` asset and change from Normal to Debug. Are you fine with that?

> Separately to this PR, what we really would like, is to work much more closely with you all on the XR/input stuff. We feel like there are huge opportunities to vastly simplify and really make using input with XR simple and straight forward.

We have our monthly Input Working Group Meeting with many on the XR team that meets next week. @lyndon-unity is the organizer, should @Billreyn be added to that? We can also talk about this PR then if more discussion needs to be had and also talk about some work in progress with how we read input in XRI.

ekcoh (approved):
Code looks good to me and see value in the mentioned presentation modes. However I agree that Project Settings as the place to change the view representation of the inspector might be a far reach. If any alternative approach requires substantial work I assume it could be reworked or improved at a later point if required since the current "..." menu may indicate a menu but it feels like the ability to present the currently selected variant type is not taken opportunity of. 
In general this improves readability of a configuration since screen space is more efficiently used.

ekcoh:
@chris-massie @lyndon-unity @Billreyn I propose we add this as one topic for the Input Working Group to sync. However I approved code changes since these LGTM. One improvement that could be considered would be to add some test coverage on InputActionProperty and related editor via unit tests.

ekcoh:
@chris-massie I followed up with UX and PM from Input team and lets do what we discussed yesterday and remove this as a visible option from the Input Settings editor and only allow switching via Input Settings asset debug mode to avoid introducing additional options when use case is mainly internal. We could easily make it non-internal later if there is any pull/requests for it. Could you do these updates and then we can merge it?

chris-massie:
> @chris-massie I followed up with UX and PM from Input team and lets do what we discussed yesterday and remove this as a visible option from the Input Settings editor and only allow switching via Input Settings asset debug mode to avoid introducing additional options when use case is mainly internal. We could easily make it non-internal later if there is any pull/requests for it. Could you do these updates and then we can merge it?

I removed the option from the **Edit** > **Project Settings** window with https://github.com/Unity-Technologies/InputSystem/pull/1726/commits/0d87b83d33eeca8b34e5d6876bef81038dacae46 and tweaked the changelog/docs for that change. It should be ready to merge now.

ekcoh:
Thanks for updating @chris-massie . Looks good and I also tested it. Will make sure this is merged as soon as CI is green on the branch. Just restarted the build since there seems to have been partial dependency failures.

ekcoh:
@chris-massie I will merge this as soon as current CI issues relating to performance test package has been resolved. develop is currently red due to this environment issue.

ekcoh:
@chris-massie FYI I merged latest develop into this branch to resolve CI issue

ekcoh:
@chris-massie This have now been merged.

PR #1727, opened 2023-08-11
FIX: save and reset input action asset (ISX-1525, ISX-1483, ISX-1559)

ritamerkl:
### Description

Three tickets around saving/auto-saving and reset the input action asset including a dialog (showing up if no auto-save is enabled and the editor window is closed) are solved in this PR. You can find them here: [ISX-1525](https://jira.unity3d.com/browse/ISX-1525), [ISX-1483](https://jira.unity3d.com/browse/ISX-1483), [ISX-1559](https://jira.unity3d.com/browse/ISX-1559).

### Changes made

fixed:
- reset dialog properly if changes got discarded

added:
- dialog that shows up on closing the editor window and auto-save is disabled 
- dirty asset editor window if auto-save disabled and asset has unsaved changes

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -74,8 +78,12 @@ private static InputActionsEditorWindow GetOrCreateWindow(int id, out bool isAlr
 
         private void SetAsset(InputActionAsset asset)
         {
-            var serializedAsset = new SerializedObject(asset);
+            m_AssetPath = AssetDatabase.GetAssetPath(asset);
+            var clone = Instantiate(asset);
```
lyndon-unity: Does this just create a local copy of the asset rather than cloning it in the database?
If it clones it in the database it would be problematic.

lyndon-unity: Do we at least need to delete this after making the serialized asset?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -95,6 +103,8 @@ private void CreateGUI()
                 if (m_State.serializedObject == null)
                 {
                     var asset = GetAssetFromDatabase();
+                    m_AssetPath = AssetDatabase.GetAssetPath(asset);
+                    m_AssetJson = File.ReadAllText(m_AssetPath);
                     var serializedAsset = new SerializedObject(asset);
```
lyndon-unity, line 106: Doesn't this flow also need to create the clone/copy?
Would be worth extracting this code out into a supporting function to remove dupliation

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -119,10 +129,58 @@ private void BuildUI()
...
         {
+            DirtyInputActionsEditorWindow(newState);
             if (InputEditorUserSettings.autoSaveInputActionAssets)
                 SaveAsset(m_State.serializedObject);
         }
 
+        private void DirtyInputActionsEditorWindow(InputActionsEditorState newState)
+        {
+            m_IsDirty = !InputEditorUserSettings.autoSaveInputActionAssets && HasAssetChanged(newState.serializedObject);
+            titleContent = m_IsDirty ? new GUIContent("(*) Input Actions Editor") : new GUIContent("Input Actions Editor");
```
lyndon-unity, line 141: Suggest we only change the title if the dirty state has changed to avoid the GUIContent calls when no differences to dirty state.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -119,10 +129,58 @@ private void BuildUI()
...
+                    break;
+            }
+        }
+
+        private void ReshowEditorWindowWithUnsavedChanges()
+        {
+            var window = CreateWindow<InputActionsEditorWindow>();
+            window.m_AssetId = m_AssetId;
+            window.m_State = m_State;
+            window.BuildUI();
```
lyndon-unity, line 181: Have you tested this flow. Does it work?
I ask as I'm wondering if you need to call SetAsset too ?

lyndon-unity (changes requested):
Some questions and suggestions above

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ritamerkl: resolved as I found a better solution: 
now the old asset is restored from json after discarding changes- this is the cleanest solution without allocating the database

lyndon-unity (approved):
Saving and auto saving seems to work as expected for me.
This update looks good

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
lyndon-unity: Seems to work for me.

ritamerkl: It works now, (latest commit) - no need to SetAsset() as long as the properties get transferred

PR #1729, opened 2023-08-14
ISX-1551 Make Project-Wide Actions and UITK editor available from 2022.3+

jfreire-unity:
### Description

This PR makes the Project-Wide Actions and UITK available from 2022.3+. More in the notes below.
**🔔  It targets the current WIP project wide actions branch #1720**

### Changes made

It removes `UNITY_INPUT_SYSTEM_UI_TK_ASSET_EDITOR` and replaces it with `UNITY_2022_3_OR_NEWER` since UITK functionality is only available from 2022.3 forward. 
- The main reason is that we use the element `TreeView` which is only available from 2022.2 but we still get some errors due to the features we're using that are only stable in 2022.3. So it's more reasonable to enable it for 2022.3.

For `UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API` I did not do the same. Rather I updated the `.asmdef` to make Project Wide Actions available from 2022.3 to match UI TK Asset Editor.  
- However, It might be possible to remove the definition altogether and not restrict it to only 2022.3 or any other version. For example, the device-only API doesn't have to depend on UI TK. 
- But before doing that, I'd like to have a stable branch so that I can run automated testing against it. At the moment, a lot of tests are still failing in `api/global-actions-source-generators`.

I also didn't remove the feature flags yet. We could do that once we do a stable `1.8.0`  when both these features available have been tested and used more and are stable enough.

### Notes

Also, a decision will need to be made about having Project Wide Actions without UI TK support, since it is only available for 2022.3. 
There's no point in having Project Wide Actions if the users cannot see them in the Project Settings. And at the moment there is no easy way to just use the IMGUI implementation of the Asset Editor and embed it in the Project Settings editor AFAIK.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Samples/ProjectWideActionsTest/ProjectWideActionsTest.cs
+++ b/Assets/Samples/ProjectWideActionsTest/ProjectWideActionsTest.cs
@@ -1,4 +1,4 @@
-#if UNITY_2020_2_OR_NEWER
+#if UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API
 using System;
 using UnityEngine;
 using UnityEngine.InputSystem;
--- a/Assets/Samples/ProjectWideActionsTest/Unity.InputSystem.ProjectWideActionsTest.asmdef
+++ b/Assets/Samples/ProjectWideActionsTest/Unity.InputSystem.ProjectWideActionsTest.asmdef
@@ -1,0 +1,22 @@
+{
+    "name": "InputSystem.Samples",
+    "rootNamespace": "",
+    "references": [
+        "GUID:75469ad4d38634e559750d17036d5f7c"
+    ],
+    "includePlatforms": [],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": false,
+    "precompiledReferences": [],
+    "autoReferenced": true,
+    "defineConstraints": [],
+    "versionDefines": [
+        {
+            "name": "Unity",
+            "expression": "2022.3",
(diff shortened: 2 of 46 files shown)


lyndon-unity (approved):
Whats the reason to replace UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API with UNITY_2020_2_OR_NEWER rather than changing the asmdef

I only ask as the other define UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API is using the asm def rather than UNITY_2022_3_OR_NEWER 

Why not use the same consistent approach ?

It sounds like you wanted to do this in stages and would prefer to replace the UNITY_INPUT_SYSTEM_ENABLE_GLOBAL_ACTIONS_API in a later commit.

I'm comfortable with this change


Regarding the note:
>a decision will need to be made about having Project Wide Actions without UI TK support, since it is only available for 2022.3.

I think we've agreed that on slack. Its fine for project wide actions to only be supported in 2022.3 as its a new feature.

ritamerkl (approved):

jfreire-unity:
No need to merge this, #1735 deals with this already.

PR #1743, opened 2023-08-30
NEW: Focus on selected action when opening UITK Asset Editor (ISX-1131)

jfreire-unity:
### Description

This PR adds functionality to focus on selected action when editor is opened and an input action from the Asset folder is selected: https://jira.unity3d.com/browse/ISX-1131

### Changes made

- Change the view state to select the action and actionMap  selected in the Editor window
- Focus on the tree view item that corresponds to the action selected. Scrolling to the element ID on the tree view is needed because the action might not be visible in case there are a lot of tree view elements (a lot of actions and bindings for example).

### Notes

📣  **Should only be merged after the release process of `1.8.0-pre.1` is done.**

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
@@ -27,6 +27,9 @@ namespace UnityEngine.InputSystem.Editor
 
     internal class InputActionsEditorWindow : EditorWindow
     {
+        [SerializeField] private InputActionsEditorState m_State;
+        [SerializeField] private string m_AssetGUID;
+
         private static readonly string k_FileExtension = "." + InputActionAsset.Extension;
         private int m_AssetId;
         private string m_AssetPath;
@@ -49,9 +52,26 @@ namespace UnityEngine.InputSystem.Editor
             //       without forcing a checkout.
             var obj = EditorUtility.InstanceIDToObject(instanceId);
             var asset = obj as InputActionAsset;
-            if (asset == null)
-                return false;
 
+            string actionMapToSelect = null;
+            string actionToSelect = null;
+
+            // Means we're dealing with an InputActionReference, e.g. when expanding the an .input action asset
+            // on the Asset window and selecting an Action.
+            if (asset == null)
+            {
+                var actionReference = obj as InputActionReference;
+                if (actionReference != null)
+                {
+                    asset = actionReference.asset;
+                    actionMapToSelect = actionReference.action.actionMap.name;
+                    actionToSelect = actionReference.action.name;
+                }
+                else
+                {
+                    return false;
+                }
+            }
 
             var window = GetOrCreateWindow(instanceId, out var isAlreadyOpened);
             if (isAlreadyOpened)
@@ -62,6 +82,7 @@ namespace UnityEngine.InputSystem.Editor
             window.m_IsDirty = false;
             window.m_AssetId = instanceId;
             window.titleContent = new GUIContent("Input Actions Editor");
+            window.SetAsset(asset, actionToSelect, actionMapToSelect);
             window.minSize = k_MinWindowSize;
             window.SetAsset(asset);
             window.Show();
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -49,9 +52,26 @@ public static bool OpenAsset(int instanceId, int line)
...
+            string actionToSelect = null;
+
+            // Means we're dealing with an InputActionReference, e.g. when expanding the an .input action asset
+            // on the Asset window and selecting an Action.
+            if (asset == null)
+            {
+                var actionReference = obj as InputActionReference;
+                if (actionReference != null)
+                {
+                    asset = actionReference.asset;
```
jamesmcgill, line 66: `actionReference.asset` and `actionReference.action` (+possibly `actionMap`) can be `null`

You probably want to early out if `asset` is still null.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -62,6 +82,7 @@ public static bool OpenAsset(int instanceId, int line)
             window.m_IsDirty = false;
             window.m_AssetId = instanceId;
             window.titleContent = new GUIContent("Input Actions Editor");
+            window.SetAsset(asset, actionToSelect, actionMapToSelect);
```
jamesmcgill: There is duplicate `window.SetAsset()` call a couple of lines below, I assume this one should be moved down to replace the other one, so it's after `minSize` is set?

jamesmcgill (approved):
I didn't try it out but looks good.
Couple of small notes to look at.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
jfreire-unity: Ah, good catch. It was a leftover from a merge conflict fix. Done ✅

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
jfreire-unity: ✅

jfreire-unity:
@Pauliusd01 or @stefanunity could try this out and see if you find any issues? Thanks 🙏

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -61,7 +61,7 @@ public static bool OpenAsset(int instanceId, int line)
             if (asset == null)
             {
                 var actionReference = obj as InputActionReference;
-                if (actionReference != null)
+                if (actionReference != null && actionReference.asset != null)
                 {
                     asset = actionReference.asset;
                     actionMapToSelect = actionReference.action.actionMap.name;
```
jamesmcgill: `actionReference.action` and `actionReference.action.actionMap' could be `null` too.

jfreire-unity: Could you give me an example of when those fields can be `null` in this context? Maybe some workflow in the Editor I'm not familiar with? The IMGUI code had no checks for this.
But I now added null conditionals for those fields anyway.

Pauliusd01:
I will take a look, Stefan is going to be busy with the release process

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -49,9 +52,26 @@ public static bool OpenAsset(int instanceId, int line)
...
+
+            // Means we're dealing with an InputActionReference, e.g. when expanding the an .input action asset
+            // on the Asset window and selecting an Action.
+            if (asset == null)
+            {
+                var actionReference = obj as InputActionReference;
+                if (actionReference != null && actionReference.asset != null)
+                {
+                    asset = actionReference.asset;
+                    actionMapToSelect = actionReference.action.actionMap?.name;
```
lyndon-unity, line 67: Should this be checking if action is null too 
actionReference.action?.actionMap?.name;

Pauliusd01 (approved):
LGTM, here's what I tried:

- Made sure the focus is correct when opening actions while searching in the Project view
- Checked what would happen with multiple assets selected -> multiple windows are correctly opened with the correct item focused
- Tried opening public action references in the inspector and verified that they also focus correctly

PR #1746, opened 2023-09-04
bump version to 1.8.0-pre.2

stefanunity:
### Description

Version bump, assembly files and C# generation not part of this, since these don't handle "-pre.x" version changes

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/ValidationExceptions.json
+++ b/Packages/com.unity.inputsystem/ValidationExceptions.json
@@ -1,4 +1,9 @@
 {
-    "ErrorExceptions": [],
+    "ErrorExceptions": [
+		{
+             "ValidationTest": "API Updater Configuration Validation",
+             "PackageVersion": "1.8.0-pre.2"
+        }
+	],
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/ValidationExceptions.json
```diff
@@ -1,4 +1,9 @@
 {
-    "ErrorExceptions": [],
+    "ErrorExceptions": [
```
jfreire-unity: Any reason to have this error exception?

jfreire-unity (commented):
Looking good, just a question below 👇

Packages/com.unity.inputsystem/ValidationExceptions.json
stefanunity: removed this again since we didn't have this in the last few releases either

PR #1747, opened 2023-09-06
Save / Auto save now disabled only for project wide settings, but enabled for input assets (ISX-1556)

lyndon-unity:
### Description

Save / Auto save now disabled only for project wide settings, but enabled for input assets ([ISX-1556](https://jira.unity3d.com/browse/ISX-1556))

image
image

### Changes made

Added explicit code to find save container and hide it (rather than being in the USS style sheet file) This is in the settings window code and not in the asset dialog code, so only effects project wide input actions.

UI TK editor now enabled for input assets as well as project wide input actions

Also removed some duplicate code for accessing stylesheets
And fixed some tests:

- a missing teardown call in InputUI tests
- a warning on an obsolete API (from 2023.2 onward)
- a camera override warning

### Checklist

Before review:

- Changelog entry not added as this is not a user facing change from what they already see
- Tests changed, if applicable.
    - Functional tests - those listed above
- Docs for new/changed API's. - no API's changed

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
@@ -67,11 +67,23 @@ namespace UnityEngine.InputSystem.Editor
                 }
             }
 
+            OpenWindow(asset, actionMapToSelect, actionToSelect);
+            return true;
+        }
+
+        private static InputActionsEditorWindow OpenWindow(InputActionAsset asset, string actionMapToSelect = null, string actionToSelect = null)
+        {
+            int instanceId = asset.GetInstanceID();
+
+            ////REVIEW: It'd be great if the window got docked by default but the public EditorWindow API doesn't allow that
+            ////        to be done for windows that aren't singletons (GetWindow<T>() will only create one window and it's the
+            ////        only way to get programmatic docking with the current API).
+            // See if we have an existing editor window that has the asset open.
             var window = GetOrCreateWindow(instanceId, out var isAlreadyOpened);
             if (isAlreadyOpened)
             {
                 window.Focus();
-                return true;
+                return window;
             }
             window.m_IsDirty = false;
             window.m_AssetId = instanceId;
@@ -80,7 +92,18 @@ namespace UnityEngine.InputSystem.Editor
             window.SetAsset(asset, actionToSelect, actionMapToSelect);
             window.Show();
 
-            return true;
+            return window;
+        }
+
+        /// <summary>
+        /// Open the specified <paramref name="asset"/> in an editor window. Used when someone hits the "Edit Asset" button in the
+        /// importer inspector.
+        /// </summary>
+        /// <param name="asset">The InputActionAsset to open.</param>
+        /// <returns>The editor window.</returns>
+        public static InputActionsEditorWindow OpenEditor(InputActionAsset asset)
+        {
+            return OpenWindow(asset, null, null);
         }
 
         private static InputActionsEditorWindow GetOrCreateWindow(int id, out bool isAlreadyOpened)
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
@@ -122,13 +122,17 @@ namespace UnityEngine.InputSystem.Editor
             var stateContainer = new StateContainer(rootVisualElement, m_State);
             stateContainer.StateChanged += OnStateChanged;
 
-            var theme = EditorGUIUtility.isProSkin
-                ? AssetDatabase.LoadAssetAtPath<StyleSheet>(InputActionsEditorConstants.PackagePath + InputActionsEditorConstants.ResourcesPath + "/InputAssetEditorDark.uss")
-                : AssetDatabase.LoadAssetAtPath<StyleSheet>(InputActionsEditorConstants.PackagePath + InputActionsEditorConstants.ResourcesPath + "/InputAssetEditorLight.uss");
-
-            rootVisualElement.styleSheets.Add(theme);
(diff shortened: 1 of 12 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -21,7 +21,7 @@ static EnableUITKEditor()
             // At the moment, the UITK Asset Editor doesn't have feature parity with the IMGUI version.
             // This is set to false to show the IMGUI version of the InputActionAsset Editor instead.
             // UITK Editor is always be used for the Project Settings Editor regardless of this setting.
-            InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseUIToolkitEditorForAllAssets, false);
+            InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseUIToolkitEditorForAllAssets, true);
```
jamesmcgill: Should we remove the feature flag now (and it it's associated code) to tidy that up?

Keeping the feature flag will make it easier to rollback if we are not 100% about it right now. But at some point this should be removed.

jamesmcgill (approved):
Nice one.

This will bee a user observable change though (going from pre1 to pre2) so it probably does warrant a changelog entry.

Run `format.ps1` to fix the formatting CI job.
The others failures are the known bee buildsystem issues.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
Pauliusd01: Can a new flag be added to instead switch to the IMGUI version? Let's say if there's an UITK issue in the future that's blocking some person it would be nice to offer that as a temporary workaround. We could remove it once UITK version has been battle tested properly with the users

jamesmcgill: Yeah, that would be the proper way of keeping the option available (the reversed logic).

We really should not have `SetInternalFeatureFlag` anywhere in our code. That is an oversight that it was left in. The flag should be unset (`false`) for whatever the default state is. And then the flag should be set by external user code. By having it in our code it will possibly override what the user was trying to set this to.

Pauliusd01 (changes requested):
It appears that the save button doesn't actually save the asset properly (can't compare with current develop though since It's missing there)
https://github.com/Unity-Technologies/InputSystem/assets/54306142/593a57c3-3955-4765-90e5-a3d403f0b97c

jamesmcgill (changes requested):
Hiding the auto-save button visually is not enough, as in this case auto-save needs to be logically enabled too.
Currently `InputEditorUserSettings.autoSaveInputActionAssets == false` and without the AutoSave button there is no way to set it to `true`.

Therefore `InputActionsEditorWindowUtils::SaveAsset()` never gets called now.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
lyndon-unity: Good idea - flag renamed in latest comment, logic reversed to be opt in for emergency short term usage and SetInternalFeatureFlag no longer used in package

lyndon-unity:
> Nice one.
> 
> This will bee a user observable change though (going from pre1 to pre2) so it probably does warrant a changelog entry.
> 
> Run `format.ps1` to fix the formatting CI job. The others failures are the known bee buildsystem issues.

Done the format part, will update changelog too

lyndon-unity:
> It appears that the save button doesn't actually save the asset properly (can't compare with current develop though since It's missing there) https://github.com/Unity-Technologies/InputSystem/assets/54306142/593a57c3-3955-4765-90e5-a3d403f0b97c

I've noticed it does save, but it doesn't correctly update the dirty state.

jamesmcgill:
> InputActionsEditorWindowUtils::SaveAsset()

I didn't see `InputActionsEditorWindowUtils::SaveAsset()` getting called at all.
I see the file being changed when I close the editor, so I'm assuming AssetDatabase is flushing it then. But we cannot rely on this behaviour once we stop using `InputManager.asset` it will need to be manually serialized by that function.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
lyndon-unity: I've also added a changelog

lyndon-unity:
> > InputActionsEditorWindowUtils::SaveAsset()
> 
> I didn't see `InputActionsEditorWindowUtils::SaveAsset()` getting called at all. I see the file being changed when I close the editor, so I'm assuming AssetDatabase is flushing it then. But we cannot rely on this behaviour once we stop using `InputManager.asset` it will need to be manually serialized by that function.

I think this is in reference to the project wide input asset when that asset level 'auto save' flag is unchecked.
This is an asset only flag and should not be checked in the project wide input asset flow. 
I've removed that in the latest commit.
(InputActionsEditorSettingsProvider.cs onStateChanged is where the save gets triggered of project wide input actions)

For asset save I've fixed the dirty state (and reset the cached asset data) when save occurs.
(InputActionsEditorWindow.cs onStateChanged is where auto save get triggered for assets)

jamesmcgill (approved):

lyndon-unity:
The latest commit makes the Edit asset button open the UI TK window too
image

Pauliusd01 (approved):
Looks good to me. Here's what I checked:

- Saving in UITK/IMGUI versions manually and with auto save (When renaming, deleting, adding things, etc)
- Undo functionality unchanged with both versions
- Project Wide Actions saving and undos also checked

PR #1751, opened 2023-09-08
NEW: Typesafe API for Project Wide Actions

jamesmcgill:
### Description

Use a Roslyn Source Generator to automatically create a wrapper API around the new Project Wide Actions.
This enables things like: 
` if (InputActions.player.attack.wasPressedThisFrame)`
and
 `if (InputActions.player.move.value.x < 0.0f)`

Ported the Source Generator and necessary support code from Andrew's branch: https://github.com/Unity-Technologies/InputSystem/pull/1720

### Changes made

- New SourceGenerator project for building the source generator dll
- Added a working dll binary to the package 
- New wrapper APIs added to support this (e.g. `Input<TActionType>`)
- Changed the Project Wide Actions Sample to use the generated API

### Notes

WIP with many TODOs including:
~~- generating from the user's actual ProjectWide actions asset (currently it reads from the static template actions asset file) [ISX-1530](https://jira.unity3d.com/browse/ISX-1530)~~
~~- fix Intellisense [ISX-1476](https://jira.unity3d.com/browse/ISX-1476)~~
~~- ensure typesafe api is regenerated after editing asset [ISX-1623](https://jira.unity3d.com/browse/ISX-1623)~~
~~- fix type conflict warnings [ISX-1619](https://jira.unity3d.com/browse/ISX-1619)~~
~~- fix for entering safe mode on first opening of the project [ISX-1624](https://jira.unity3d.com/browse/ISX-1624)~~
~~- wrapper APIs in the `TypeSafeAPI` directory should somehow be hidden from users  [ISX-1620](https://jira.unity3d.com/browse/ISX-1620)~~
~~- CoreTests.Actions_ActiveBindingIndex_PointsAtRelativeIndexInActionBindingsArray test failures [ISX-1621](https://jira.unity3d.com/browse/ISX-1621)~~
- Also needs automated tests created for the typesafe API itself [ISX-1622](https://jira.unity3d.com/browse/ISX-1622)

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


lyndon-unity:
I see some test failures such as this - this seems to be one of the new tests

playmode Test Failure: CoreTests.Actions_ActiveBindingIndex_PointsAtRelativeIndexInActionBindingsArray
  Expected: ActionTwo:OneModifier
  But was:  ActionTwo:<Keyboard>/2

Assets/Samples/ProjectWideActionsTest/ProjectWideActionsTest.cs
```diff
@@ -1,63 +1,42 @@
...
-    InputAction interact;
-    InputAction next;
-    InputAction previous;
-    InputAction sprint;
-    InputAction crouch;
-
     // Start is called before the first frame update
     void Start()
     {
-        // Project-Wide Actions
```
lyndon-unity, line 23: I'm thinking It would be good to keep one example of the explicit InputSystem.actions FindAction route for any user wanting to lookup by string for some reason of dynamic access, but this is the more advanced option

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -10434,6 +10457,42 @@ public void Actions_ImprovedShortcutSupport_ConsumesWASD(bool shortcutsEnabled)
...
+            .With("modifier", "<Keyboard>/leftShift")
+            .With("binding", "<Keyboard>/1");
+        actionTwo.AddCompositeBinding("OneModifier")
+            .With("modifier", "<Keyboard>/leftShift")
+            .With("binding", "<Keyboard>/2");
+
+        Press(keyboard.leftShiftKey);
+        PressAndRelease(keyboard.digit2Key);
+
+        Assert.That(actionTwo.bindings[actionTwo.activeBindingIndex], Is.EqualTo(actionTwo.bindings[4]));
```
lyndon-unity: This seems to be failing in test run

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
```diff
@@ -1232,6 +1234,49 @@ public BindingSyntax WithInteractions(string interactions)
...
+                    return this;
+
+                var nameAndParameters = NameAndParameters.ParseMultiple(interactionsString);
+                var interactions = nameAndParameters as List<NameAndParameters> ?? nameAndParameters.ToList();
+                var lastIndexOfInteraction = -1;
+                for (var i = interactions.Count - 1; i >= 0; i--)
+                {
+                    if (interactions[i].name == interactionName)
+                    {
+                        lastIndexOfInteraction = i;
```
lyndon-unity: Do we allow the interaction to be present multiple times?
If its present multiple times should we remove all?

(+ Should we prevent multiple interactions of the same name in the add code?
 I guess we don't know what impact that would have for some odd user custom interactions. )

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
```diff
@@ -1245,9 +1290,9 @@ public BindingSyntax WithInteraction<TInteraction>()
                 if (!valid)
                     throw new InvalidOperationException("Accessor is not valid");
 
-                var interactionName = InputProcessor.s_Processors.FindNameForType(typeof(TInteraction));
+                var interactionName = InputInteraction.s_Interactions.FindNameForType(typeof(TInteraction));
```
lyndon-unity, line 1248: Probably should call this bug fix out in the change log

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
```diff
@@ -0,0 +1,372 @@
...
+using System.Text.Json.Serialization;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Unity.InputSystem.SourceGenerators
+{
+    [Generator]
+    public class GlobalInputActionsSourceGenerator : ISourceGenerator
+    {
+        //@TODO: This needs a JSON file
```
lyndon-unity: Why do we have this TODO. The code below seems to handle extracting the incoming data out into a JSON file so shouldn't we just make this a statement.

lyndon-unity: Hmm the more critical aspect seems to be this should not reference the template. It needs to access the InputManager.asset

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
```diff
@@ -0,0 +1,372 @@
...
+
+            var projectPath = additionalFile.GetText().ToString();
+            var globalActionsAssetPath = Path.Combine(projectPath, GlobalActionsAssetPath);
+            if (!File.Exists(globalActionsAssetPath))
+            {
+                Console.Error.Write($"Global input actions asset file not found at '{globalActionsAssetPath}'.");
+                return;
+            }
+
+            using var fileStream = new FileStream(globalActionsAssetPath, FileMode.Open);
```
lyndon-unity: This seems to just be processing the template file and note the real data in project wide input actions ?

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
```diff
@@ -0,0 +1,372 @@
...
+                        break;
+                    hash2 = ((hash2 << 5) + hash2) ^ str[i + 1];
+                }
+
+                return hash1 + (hash2 * 1566083941);
+            }
+        }
+    }
+
+    public struct InputActionAsset
```
lyndon-unity: Can we share any of these definitions with the main package code to avoid duplication and risk of going out of sync.
Or at least comment both versions with a reference of where to update to its clear they need to be kept in sync.

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/InputActionsSyntaxReceiver.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Unity.InputSystem.SourceGenerators;
+
+public class InputActionsSyntaxReceiver : ISyntaxReceiver
```
lyndon-unity: What does this do - can we add a small comment to explain this class

Packages/com.unity.inputsystem/InputSystem/TypeSafeAPI/Input.actions.cs
```diff
@@ -0,0 +1,90 @@
...
+                    // add incompatible bindings, and the best we can do then is to catch the exceptions
+                    // thrown when we try to read from those controls.
+                    return m_Action.ReadValue<TActionType>();
+                }
+                catch (InvalidOperationException ex)
+                {
+                    Debug.LogWarning(ex.Message);
+                }
+
+                return m_Action.ReadDefaultValue<TActionType>(
```
lyndon-unity: Could we add a test that checks this flow. Less critical as its something we shouldn't hit in most cases

lyndon-unity (changes requested):
Failing test is a concern.
Shouldn't land until we resolve some of the issues mentioned in the main PR body.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,19 +4,21 @@
...
         internal const string kDefaultAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.inputactions";
-        internal const string kAssetPath = "ProjectSettings/InputManager.asset";
+        internal const string kInputManagerAssetPath = "ProjectSettings/InputManager.asset";
+        internal const string kAssetPath = "ProjectSettings/InputSystemActions.inputactions";
         internal const string kAssetName = InputSystem.kProjectWideActionsAssetName;
 
         static string s_DefaultAssetPath = kDefaultAssetPath;
         static string s_AssetPath = kAssetPath;
 
+        public static string assetPath => s_AssetPath;
```
lyndon-unity: Is this writable for tests. 
Seems in most cases we want this read only.

lyndon-unity: As we use SetAssetPaths to write to the backing variable for tests I'm thinking this could be marked read only.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -38,69 +40,77 @@ internal static void InstallProjectWideActions()
...
-            var json = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath));
-
-            var asset = ScriptableObject.CreateInstance<InputActionAsset>();
-            asset.LoadFromJson(json);
-            asset.name = kAssetName;
+            InputActionAsset asset;
 
-            AssetDatabase.AddObjectToAsset(asset, s_AssetPath);
+            asset = LoadFromProjectSettings();
+            if (asset != null) return asset;
```
lyndon-unity: I'm thinking we want to move this below the LoadFromInputManagerSettings so we use the newer location by default, and it we were to save (or copy) a version in the newer location (ProjectSettings/InputSystemActions.inputactions) it would pick that up in preference.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -38,69 +40,77 @@ internal static void InstallProjectWideActions()
...
 
-            AssetDatabase.SaveAssets();
+            // Create a new one if we couldn't find any existing one to load.
+            return CreateNewActionAsset();
+        }
+
+        private static InputActionAsset CreateNewActionAsset()
+        {
+            var templateAssetPath = Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath);
+            var projectAssetPath = Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath);
```
lyndon-unity: Looks like both paths are the same - shouldn't the second be created from s_AssetPath

lyndon-unity (changes requested):
I think the CreateNewActionAsset function needs a fix

lyndon-unity:
Paulius please can you do some preliminary testing of this branch with regards to upgrades from older versions.
Specifically testing upgrades:

- from 1.7.0 where project wide input actions not present
-- Do the new settings get setup correctly and the new file appear on disk (ProjectSettings/InputSystemActions.inputactions)
- from 1.8.0.pre-1 where project wide input actions are present
-- Do the settings come over correctly in the UI
-- Do further updates correctly get applied and appear in the UI (after project reload) (in the new file on disk 
-- Does the Input Manager UI remain working, or fix itself from 1.8.0.pre-1 after upgrading to this branch.
ProjectSettings/InputSystemActions.inputactions)
- new project

Does the Input Manager UI remain working in all the above cases.

Pauliusd01:
image
Are these tests supposed to be broken or do I need to enable something besides switching to this branch?

Pauliusd01:
On that same note the Project Actions tab is empty with "argument cannot be null" errors in console. Gonna clear the repo and repull in case of any leftover files
image

jamesmcgill:
This is a WIP branch, so it's not ready for any sort of testing. Specifically there are auto-save related issues with the editor that just landed in `develop` and now need integrated into this branch to fix the save flow.

I will fix this up just now so that you can test some things and will let you know.

However the first image is something I see and it basically means the source generator didn't generate the API (yet). Entering into and out of playmode may help. It's known and I wouldn't bother with that.


The second one may be related around the asset importing which I believe Lyndon wants you to test.
I.e. the flow should be:
For new projects (or projects that haven't run on v1.8pre yet) -> new file `ProjectSettings/InputSystemActions.inputactions` should be created (with new template actions) and the editor UI should show the contents of that (and save to that file)

For projects that already run on v1.8pre -> new file `ProjectSettings/InputSystemActions.inputactions` should be created with existing project actions copied into it. Editor UI should show from and save to the contents of the new file.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jamesmcgill: If I understand correctly it should be working as you describe.
`LoadFromProjectSettings()`  (new json file) is loaded first as preference
`LoadFromInputManagerSettings()'  (pre1 file) is attempted as a backup

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jamesmcgill: This was fixed, but I'll need to check it didn't get commited somehow

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jamesmcgill: Yep, it can be readonly

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/InputActionsSyntaxReceiver.cs
jamesmcgill: This is removed now, but it is a Roslyn interface for visiting syntax nodes.

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
jamesmcgill: They can't be shared ASAIK as this is not a Unity project but a MSBuild project (maybe in new future projects we will change this story).
But yeah, agree it needs to be documented as there needs to be parity between the asset and the generator code.

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
jamesmcgill: Yes, this was noted in the PR description but now fixed.

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
jamesmcgill: Noted in PR description against [ISX-1530](https://jira.unity3d.com/browse/ISX-1530) and now resolved.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jamesmcgill: Perhaps you are reviewing commits rather than the PR because that fix was pushed?

jamesmcgill:
@Pauliusd01 You should be able to create/load/save asset now for testing.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
chris-massie: XR Interaction Toolkit ships an input actions asset with a binding with multiple interactions of the same type (but with different properties set).

Packages/com.unity.inputsystem/InputSystem/TypeSafeAPI/Input.actions.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <summary>
+        /// True when the action is pressed.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <remarks>
+        /// </remarks>
+        public static implicit operator bool(Input<TActionType> input)
+        {
+            return input.m_Action.IsPressed();
+        }
```
chris-massie: I don't know if we should add this implicit operator, especially since it is using `IsPressed` instead of something like `phase == InputActionPhase.Performed`. If you have a Button input action with a Hold interaction on the binding, `IsPressed` will be `true` the moment you press the button, but it won't be performed until 0.4 seconds later in the default. Using `IsPressed` here seems bad for the default.

Since there's subtle differences between `IsPressed`, `WasPerformedThisFrame`, and `phase == InputActionPhase.Performed`, I think it should remain explicit for what the user wants, and I think this operator should be deleted.

jamesmcgill: Thanks for looking this over @chris-massie. Good points, I think this should be okay to remove entirely.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
jamesmcgill: I think `RemoveInteraction` is not actually needed by this PR, so I suggest we remove it entirely here and only add it if we have a clear need to have it and design that resolves these issues.

ekcoh (commented):
@jamesmcgill I noticed we have added a precompiled InputSystemActionsAPIGenerator.dll into SourceGenerators folder. Is this binary something we have compiled or is it a renamed external source generator lib related to Roslyn? I noticed there is only a .dll and no .so etc so how would this work on macOS and Linux editors?

jamesmcgill:
> @jamesmcgill I noticed we have added a precompiled InputSystemActionsAPIGenerator.dll into SourceGenerators folder. Is this binary something we have compiled or is it a renamed external source generator lib related to Roslyn? I noticed there is only a .dll and no .so etc so how would this work on macOS and Linux editors?

The dll _is_ the source generator. It's presence (and it's .meta file) are the only files needed to make that work.
It is compiled (by us), but it's still .net based so it works on mac/rider too. 

If you build `Unity.InputSystem.SourceGenerators.sln`, this file will automatically be updated (no copying needed)

Pauliusd01:
@lyndon-unity 
From Lyndon's comment: 
   
    "from 1.7.0 where project wide input actions not present
    Do the new settings get setup correctly and the new file appear on disk (ProjectSettings/InputSystemActions.inputactions)"
     - If by new setting you mean project wide actions then yes, upgraded stable 1.7.0 test project (e29989176f3a8473ba30854e4bfae953083eba28) to use this branch package. Although I do not see a new asset in the path you listed, maybe It's supposed to be a different path?
    
    "from 1.8.0.pre-1 where project wide input actions are present
    Do the settings come over correctly in the UI"
    - If you mean the Project Wide actions then yes, if It's something else I need to look at please let me know.
   
    "Do further updates correctly get applied and appear in the UI (after project reload) (in the new file on disk"
    - Same thing again, no new asset in your provided path. Could it be that you're referring to "InputSystem\Packages\com.unity.inputsystem\InputSystem\Editor\ProjectWideActions\ProjectWideActionsTemplate.inputactions" instead? Althought that one did not get a new date modified stamp after I edited the Project Actions
    
    "Does the Input Manager UI remain working, or fix itself from 1.8.0.pre-1 after upgrading to this branch.
    ProjectSettings/InputSystemActions.inputactions)"
    - No it does not, but all the user has to do is restart the Editor.
    
    "Does the Input Manager UI remain working in all the above cases. "
    - Yes but it still requires a restart if upgrading.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/TemplateActions.InputSystemActionsAPIGenerator.additionalfile
lyndon-unity, line 1: I'm confused about this file TemplateActions.InputSystemActionsAPIGenerator.additionalfile

Why do we have this file?
* Is it generating the output prior to use creating a .additionalfile in the users folder?

What happens once a we do create the .additionalfile in the users folder?
* Don't both try to create the project wide input actions code? One from the template and one from the updated asset in project settings?

lyndon-unity: Also I don't see any code to write the .additionalfile in this branch. So isn't this always creating the same data and not connected to the asset data stored in project settings ?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -102,6 +107,42 @@ private static InputActionAsset CreateNewActionAsset()
...
+        // the Roslyn source generator will be able to read. This file is automatically generated
+        // and is created in the root of the User's Assets directory.
+        internal static void CreateRoslynAdditionalFile()
+        {
+            try
+            {
+                if (!Directory.Exists("Assets"))
+                    Directory.CreateDirectory("Assets");
+
+                var assetJson = InputSystem.actions.ToJson();
```
lyndon-unity, line 129: Can we search for existing file in the AssetDB at any path, use that location if it exists and overwrite, then only create at the fixed path if not present elsewhere in the Assets folder.

This would allow people to move it.

lyndon-unity: If we do this the delete would have to also search for all files of that name which does add complexity.

lyndon-unity: For pre-2 this could could be sufficient so I wouldn't block merging for it.
Perhaps capture this additional ask in a task if you agree.

Packages/com.unity.inputsystem/InputSystem/SourceGenerators/Source~/SourceGenerators/GlobalInputActionsSourceGenerator.cs
```diff
@@ -0,0 +1,520 @@
...
+    } // class _Input
+} // namespace UnityEngine.InputSystem.TypeSafeAPIInternals
+
+namespace UnityEngine.InputSystem
+{
+    /// <summary>
+    /// A container class to hold the project-wide actions typesafe api.
+    /// </summary>
+    [CompilerGenerated]
+    public static partial class ProjectActions
```
lyndon-unity, line 283: I want to check with PM and design that all are comfortable with this naming.
I can see why renamed from Input to avoid clash, but we refer to as ProjectWideActions elsewhere so I want to make sure we are aligned on the naming here. I'll come back to this once the others give me feedback on the naming.

Assets/Samples/ProjectWideActionsSample/ProjectWideActionsSample.cs
```diff
@@ -0,0 +1,46 @@
...
+            if (ProjectActions.player.attack.wasPressedThisFrame)
+            {
+                cube.GetComponent<Renderer>().material.color = Color.red;
+            }
+            else if (ProjectActions.player.attack.wasReleasedThisFrame)
+            {
+                cube.GetComponent<Renderer>().material.color = Color.green;
+            }
+
+            if (ProjectActions.player.move.value.x < 0.0f)
```
ekcoh, line 27: Not high priority to address on this branch and not really introduced here, but as I commented on a previous PR this can be replaced by a one or two liner to simplify sample:

`cube.transform.Translate(ProjectActions.player.move.value * Time.deltaTime);` 

Maybe we should also add a comment after the translate as such:
// Multiplying with Time.deltaTime since assuming action binding values are absolute and not relative

ekcoh: (Such a change could have been outside TypeSafe API branch and added to develop since not relating to that functionality?)

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -14,43 +12,27 @@
...
     [SetUp]
     public override void Setup()
     {
-        // @TODO: Currently we can only inject the TestActionsAsset in PlayMode tests.
-        // It would be nice to be able to inject it as a Preloaded asset into the Player tests so
-        // we don't need different tests for the player.
-        // This also means these tests are dependant on the content of InputManager.asset not being changed.
-#if UNITY_EDITOR
-        // This asset takes the place of ProjectSettings/InputManager.asset for the sake of testing, as we don't
+        // This asset takes the place of the project wide actions asset for the sake of testing, as we don't
```
ekcoh: Changes here do not relate to Typesafe API and could have been done separately against develop on its own PR?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -8,28 +8,29 @@
 
 namespace UnityEngine.InputSystem.Editor
 {
+    class ProjectWideActionsAssetPostprocessor : AssetPostprocessor
+    {
+        static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths, bool didDomainReload)
+        {
+            if (ArrayHelpers.Contains(importedAssets, ProjectWideActionsAsset.kAssetPath))
```
ekcoh, line 15: Does this imply that the .additionalfile has a fixed path? Wouldn't it be possible to instead scan for it or have its location tracked by this and persisted as part of InputManager.asset scriptable object or as an optional serialised property of the InputActionAsset updated via this if an importer cannot be registered?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -8,28 +8,29 @@
...
+                ProjectWideActionsAsset.CreateRoslynAdditionalFile();
+                AssetDatabase.AllowAutoRefresh();
+            }
+        }
+    }
+
     internal static class ProjectWideActionsAsset
     {
-        internal const string kDefaultAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.json";
+        internal const string kTemplateAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.json";
```
ekcoh, line 24: None of the changes of this class relates to Type-safe API right? (Could have been a separate PR against develop)

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -8,28 +8,29 @@
...
-            s_AssetPath = assetPath;
-        }
-
-        internal static void Reset()
-        {
-            s_DefaultAssetPath = kDefaultAssetPath;
-            s_AssetPath = kAssetPath;
-        }
-
+        internal static InputActionAsset testAsset  { get; set; }
```
ekcoh: I am afraid just returning another InputActionAsset here might just be a partial workaround for tests depending on what they are covering. There is no code to replace / swap out object in AssetDatabase added via AddObjectToAsset (references and InputActionAsset) so any attempt to access via non-instance or doing edits would cause a mixed perspective of reality. Maybe this is ok for now, but just want to highlight that if we really want to properly substitute the InputActionAsset we might have do more work. Possibly I guess we could remove this property and modify/restore the AssetDatabase from Test fixture instead which would eliminate this property and if statement below. Maybe its as simple as deleting the asset object from the path via ADB (but that could remove local changes) or extract instance, modify ADB and restore?

In test TearDown:
Restore InputActionAsset by invoking AssetDatabase.AddObjectToPath(kAssetPath);

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetDrawer.cs
```diff
@@ -94,7 +94,7 @@ static bool IsAssetProjectWideActions(SerializedProperty property)
             if (property.objectReferenceValue != null)
             {
                 var asset = (InputActionAsset)property.objectReferenceValue;
-                isAssetProjectWideActions = asset?.name == ProjectWideActionsAsset.kAssetName;
+                isAssetProjectWideActions = asset?.name == InputSystem.kProjectWideActionsAssetName;
```
ekcoh, line 97: I would recommend we delegate this comparison to ProjectWideActions class as a static function, i.e.

var isAssetProjectWideActions = InputSystem.IsProjectWideActions(property.objectReferenceValue);

Instead of having such comparisons scattered around in the code, then its more maintainable.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
```diff
@@ -14,12 +14,15 @@ internal class InputActionsEditorWindowUtils
         public static void SaveAsset(SerializedObject serializedAsset)
         {
             var asset = (InputActionAsset)serializedAsset.targetObject;
+
             // for the global actions asset: save differently (as it is a yaml file and not a json)
             if (asset.name == ProjectWideActionsAsset.kAssetName)
```
ekcoh, line 19: Same comment as for drawer above, delegate check to InputSystem.IsProjectWideActionAsset(asset) or similar

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3028,31 +3028,31 @@ public static InputActionAsset actions
...
                     throw new ArgumentNullException(nameof(value));
 
-                if (s_projectWideActions == value)
+                if (s_ProjectWideActions == value)
                     return;
 
-                s_projectWideActions?.Disable();
-                s_projectWideActions = value;
-                s_projectWideActions.Enable();
+                s_ProjectWideActions?.Disable();
```
ekcoh, line 3053: I noticed there is a warning on these lines saying it disables Unity object lifetime checks to use ?. operator here. Consider maybe an if-statement instead?

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3746,11 +3746,11 @@ private static void Reset(bool enableRemoting = false, IInputRuntime runtime = n
...
 #if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
             // Avoid touching the `actions` property directly here, to prevent unwanted initialisation.
-            if (s_projectWideActions)
+            if (s_ProjectWideActions)
             {
-                s_projectWideActions.Disable();
-                s_projectWideActions?.OnSetupChanged();  // Cleanup ActionState (remove unused controls after disabling)
-                s_projectWideActions = null;
+                s_ProjectWideActions.Disable();
+                s_ProjectWideActions?.OnSetupChanged();  // Cleanup ActionState (remove unused controls after disabling)
```
ekcoh, line 3752: This ?. check looks redundant?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
```diff
@@ -14,12 +14,15 @@ internal class InputActionsEditorWindowUtils
...
             {
                 AssetDatabase.SaveAssets();
+                ProjectWideActionsAsset.CreateRoslynAdditionalFile();
```
ekcoh, line 22: If we tracked the path to .additionalfile via asset or via InputActionAsset serialised property I assume we could have e.g. 

```
if (asset.m_RoslynCodeGenerationAssetPath != null)
     ProjectWideActionsAsset.CreateRoslynAdditionalFile();
```
     
     or similarly via another separate serialised object stored within InputManager.asset we could determine the same thing and/or also use it as a setting?

jamesmcgill:
Closing as it has been decided to discontinue work on this feature.

PR #1753, opened 2023-09-12
CHANGE: Add property drawer for InputActionsAsset to use Project-Wide Actions (ISX-1321)

jfreire-unity:
### Description

Implements https://jira.unity3d.com/browse/ISX-1321 for InputActionAsset. Follow-up PRs will update InputActionReference with project-wide actions.

### Changes made

Kapture 2023-09-15 at 11 28 32

This PR introduces a `CustomPropertyDrawer` for InputActionAsset. Whenever an InputActionAsset field is shown in the editor, it will show a dropdown menu to allow the selection for an actions asset or project-wide actions.
A "Open" button is also added when the user selects project wide actions.

 `InputSystemUIInputModule` is also updated so that the custom inspector editor shows the appropriate actions when using project-wide actions.

### Notes
- This is only available for 2022.3+, as it is related to project-wide actions.
- Fixes some bugs on the 'Input System UI Input Module' when working null asset selected, or missing actions in a valid asset.
- TrackedDevicePosition and TrackedDeviceOrientation added back into the project-wide actions in the UI action map so the default 'Input System UI Input Module' mappings function.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetPropertyDrawer.cs
```diff
@@ -0,0 +1,58 @@
+using System.Linq;
+using UnityEditor;
+using UnityEditor.SearchService;
+using UnityEngine.Search;
+using UnityEngine.UIElements;
+
+namespace UnityEngine.InputSystem.Editor
+{
+    // [CustomPropertyDrawer(typeof(InputActionAsset))]
+    [CustomPropertyDrawer(typeof(InputActionAssetProperty))]
```
ekcoh: It would be highly desirable to have this as [CustomPropertyDrawer(typeof(InputActionAsset))] instead. Then any third-party or other package code would just work, while this approach may enable assigning project-wide or asset it has the drawback of requiring anyone interested in referencing project wide to update/adapt their code. 

I did a quick test with CustomPropertyDrawer(typeof(InputActionAsset))] and that enables a drawer but leaves you with the problem of figuring out what is really selected to be able to set the value of the first high-level drop-down, which leads me to my next question:
How do we currently reference the project-wide asset? Is it available as a static instance somewhere? If it is, and I interpret InputActionAsset as basically being a reference (project wide is serialised independently by the settings?), why can't we leave PlayerInput and PlayerINputEditor unmodified and instead provide a new property on the InputActionAsset (evaluated by the drawer) which is bool IsProjectWide().
That getter could in turn base the result, ideally on reference comparison with the project wide instance, or on detecting a tag from json parsing (InputActionAsset.FromJson) not serialised for any other asset. What I am after is that any solution not introducing a new type here would be desirable in my perspective.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetProperty.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using UnityEngine.Serialization;
+
+namespace UnityEngine.InputSystem.Editor
+{
+    [Serializable]
+    public struct InputActionAssetProperty
```
ekcoh: Summary of problems I could see with this new type:
- This type is another type than InputActionAsset which implies we do not benefit from automatic property drawers on any existing code where InputActionAsset is a property, e.g. components in other packages or projects. Still I understand that you need a way to tell whether a certaain InputActionAsset is the global one or not, hence the bool.
- Given this solution what would happen if multiple InputActionAssetProperties exist which have the bool set to true but pointing to different assets? This type would allowing setting the bool from user code which may lead to  unexpected effects.

ekcoh (commented):
Only commenting for now since WIP

ekcoh:
Excellent with a GIF animation by the way quickly providing a good comprehension of what I am reviewing.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetPropertyDrawer.cs
jfreire-unity: My initial idea was to do a custom drawer for `InputActionAsset` but after not being able to make it work properly I decided with a different strategy, and found that this pattern would do what we wanted.
That idea sounds interesting and I'll try it out, thanks. 👍

lyndon-unity:
> Excellent with a GIF animation by the way quickly providing a good comprehension of what I am reviewing.

Agreed although the default scheme, auto switch and default map doesn't seem present for the input action asset type if I'm interpretting the gif correctly. I'd expect them to be present for project wide input actions and non null input action assets.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetPropertyDrawer.cs
lyndon-unity: Good proposal and seems reasonable if this works.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetProperty.cs
jfreire-unity: Good points, thanks.
From my view, point 2 is the more critical from an API perspective.
We could probably leave with point 1 as we would only update our known components, leaving users to do what they wanted. Not desirable, but reasonable enough if there.
Based on your comments, I'll try again implementing this to `InputActionAsset` instead of creating a new type.

jfreire-unity:
> > Excellent with a GIF animation by the way quickly providing a good comprehension of what I am reviewing.
> 
> Agreed although the default scheme, auto switch and default map doesn't seem present for the input action asset type if I'm interpretting the gif correctly. I'd expect them to be present for project wide input actions and non null input action assets.

That only happens if the actions asset has control schemes created. The one I selected in the GIF didn't have, but for instance, this one has:
image

This was the behavior that was already implemented by the way. I didn't change anything with regard to this area.

jfreire-unity:
@ekcoh thanks for the idea of checking if the asset selected is a project-wide action or not. For some reason the first time I tried it did not work. I must have tried it in a rush and failed to get a simpler solution.

jfreire-unity:
@ekcoh thanks for the idea of checking if the asset selected is a project-wide action or not. It works! For some reason the first time I tried it did not work. I must have tried it in a rush and failed to get a simpler solution.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetPropertyDrawer.cs
jfreire-unity: Addressed in newer commits. Thanks @ekcoh for the suggestion

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetDrawer.cs
```diff
@@ -0,0 +1,71 @@
...
+    internal class InputActionAssetDrawer : PropertyDrawer
+    {
+        static readonly string[] k_ActionsTypeOptions = new[] { "Project-Wide Actions", "Actions Asset" };
+
+        public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+        {
+            EditorGUI.BeginProperty(position, label, property);
+
+            var isAssetProjectWideActions = IsAssetProjectWideActions(property);
+            var selectedAssetOptionIndex = isAssetProjectWideActions ? 0 : 1;
```
lyndon-unity: Would be tidier if this was an enum rather than 0/1 int value

You can use EditorGUILayout.EnumPopup below to assist this

lyndon-unity (commented):
This looks like a simpler approach that is quite robust and would work in more places. Overall I'd approve but I think a change to use enums rather than int's for section would be easier to read.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionAssetDrawer.cs
jfreire-unity: Added those changes now

lyndon-unity (approved):
The enum changes make that easier to read - thank you

lyndon-unity:
Can you indent the asset when the "Actions Asset" is selected?

Both in the player input component and my own component its not obvious its related to the drop down when the next field is at the same indenting depth

image

In the player input component its a bit confusing that both fields are called 'Actions'. I think the name with the drop down should perhaps be taken from the field name and then the actual asset selector just called 'Asset'

(in the above example my field in the 'Digital and analog script' is called "Input Action Asset" so I'd expect to see that even when the selection is on "Project-Wide Actions")

lyndon-unity:
On a related note the InputActionReference I have is not listing the project wide input actions when it pops up its selector. We need to find a way to inject it here too, but that could be a separate PR

image

Note changes in this area might clash with this PR
https://github.com/Unity-Technologies/InputSystem/pull/1726

jfreire-unity:
> Can you indent the asset when the "Actions Asset" is selected?
> 
> Both in the player input component and my own component its not obvious its related to the drop down when the next field is at the same indenting depth
> 
> image
> 
> In the player input component its a bit confusing that both fields are called 'Actions'. I think the name with the drop down should perhaps be taken from the field name and then the actual asset selector just called 'Asset'
> 
> (in the above example my field in the 'Digital and analog script' is called "Input Action Asset" so I'd expect to see that even when the selection is on "Project-Wide Actions")

Ah yeah, good catch. I had a different name before but that was lost during refactoring. Will add "Actions Asset" label to that field.

jfreire-unity:
Adding QA to see if there are any issues we didn't catch in our testing cc @stefanunity or @Pauliusd01

Pauliusd01:
Would changing the naming of one of the dropdown options from "Actions Asset" to "Custom Actions Asset" be decent idea? I feel like we hear the words Project Wide Actions a lot and we know what it means specifically so we can tell the 2 options apart easily, but to the user they both might sound too similar? Just a thought I had 😅

Pauliusd01:
https://github.com/Unity-Technologies/InputSystem/assets/54306142/8fe75706-37df-47f5-9465-3d774c5895a6

Got 2 questions/potential issues:

- If simply switching between Project Wide Actions and Actions Assets should the already set Actions asset be saved in the Actions Asset dropdown? (To avoid users losing their progress, however small that may be when using the dropdown)
- In the video above I switch between the dropdown and then undo, my character no longer works. Does that mean there's an issue with the undos and the dropdown?

Pauliusd01:
> Unity_2023-09-15_14-12-58.mp4
> 
> Got 2 questions/potential issues:
> 
>     * If simply switching between Project Wide Actions and Actions Assets should the already set Actions asset be saved in the Actions Asset dropdown? (To avoid users losing their progress, however small that may be when using the dropdown)
> 
>     * In the video above I switch between the dropdown and then undo, my character no longer works. Does that mean there's an issue with the undos and the dropdown?

Disregard the undo issue I guess, I reproduced it without your changes as well by just switching actions assets on Develop

jfreire-unity:
> > Unity_2023-09-15_14-12-58.mp4
> > Got 2 questions/potential issues:
> > ```
> > * If simply switching between Project Wide Actions and Actions Assets should the already set Actions asset be saved in the Actions Asset dropdown? (To avoid users losing their progress, however small that may be when using the dropdown)
> > 
> > * In the video above I switch between the dropdown and then undo, my character no longer works. Does that mean there's an issue with the undos and the dropdown?
> > ```
> 
> Disregard the undo issue I guess, I reproduced it without your changes as well by just switching actions assets on Develop

Yeah, I was just checking it. Maybe we can open a bug for it @Pauliusd01 ?

Pauliusd01 (approved):
Looks good to me, I only have the 2 not super consequential nitpicks I listed previously. Here's what I checked:

1. Checking existing, previously set up objects with play input and made sure that they still work the same.
2. Set up new controls with player input and Project Wide Actions, also made sure my modifications to the Project Wide Actions are still picked up by my player input object I set up. Tested it with multiple control schemes set.
3. Checked undoing when switching between Project Wide Actions and custom ones (One of the nitpicks I have mentioned)

jfreire-unity:
Merging with un-related failures for 2023.1. Applying what #1750 did, doesn't solve the issue so we will need to wait for the `trunk` changes to be propagated.

PR #1754, opened 2023-09-26
Fixed UI Submit / Cancel usages when using Switch Pro controller

Secticide:
### Description

The Switch Pro controller previously wasn't usable with UI since the "Submit" and "Cancel" usages were not assigned to the relevant controls. This PR rectifies this issue.

### Changes made

- Added "Submit" and "Cancel" usages to the relevant controls on the Switch Pro controller


jfreire-unity (approved):
Looks good to me. Maybe add an entry to the changelog as well? 

I'm wondering if there is any way to test this, to validate that the "new behavior" works for these controllers. Were you able to test it?

Secticide:
> Looks good to me. Maybe add an entry to the changelog as well?
> 
> I'm wondering if there is any way to test this, to validate that the "new behavior" works for these controllers. Were you able to test it?

I tested it locally using the bug-provided test project. The only way I can think of to test this sort of thing would be to write up a validator for our controller layouts which loops over all of the defined layouts to ensure a "Submit" and "Cancel" usage is present. That way we're defended against this in the future.

To be a bit more specific on the local test; there was a UI setup where you were able to select a few different buttons using the controller and "Submit" via a button press which would lead to a log showing. This would be very difficult to setup as a test project.

Secticide:
@jfreire-unity Added a changelog entry, thanks for pointing it out!

jfreire-unity:
> > Looks good to me. Maybe add an entry to the changelog as well?
> > I'm wondering if there is any way to test this, to validate that the "new behavior" works for these controllers. Were you able to test it?
> 
> I tested it locally using the bug-provided test project. The only way I can think of to test this sort of thing would be to write up a validator for our controller layouts which loops over all of the defined layouts to ensure a "Submit" and "Cancel" usage is present. That way we're defended against this in the future.
> 
> To be a bit more specific on the local test; there was a UI setup where you were able to select a few different buttons using the controller and "Submit" via a button press which would lead to a log showing. This would be very difficult to setup as a test project.

I see, thanks for explaining. That sounds like a good idea of something to have in the future, thanks! I'll capture that to see if we could do it eventually.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - Partially fixed case ISX-1357 (Investigate performance regressing over time).  A sample showed that leaving an InputActionMap enabled could lead to an internal list of listeners growing.  This leads to slow-down, so we now warn if we think this is happening.
+- Fixed case ISXB-580 (UI Submit / Cancel not working with Switch Pro controller) by adding "Submit" & "Cancel" usages to the Switch Pro controller input controls.
```
ekcoh: Suggest providing ISXB-580 as a hyperlink to the issue tracker since this is what we have done previously. This provides a convenient way to check the corresponding issue that the fix-claim references. In this case e.g. https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-580

ekcoh (changes requested):
In general LGTM, please adda hyperlink to the related issue-tracker in the changelog before merging.

A side note to usages is that I agree it could be beneficial with unit testing layouts for the presence of controls with Cancel and Submit. However, this would only apply to Layouts that would correspond to a certain "standard model" and couldn't be used as a general requirement on a Layout as I see it.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Also noted that CHANGELOG.md has a conflict that needs to be resolved.

Secticide: Resolved conflicts and added link to issue tracker in changelog in 827232f.

ekcoh (approved):

PR #1755, opened 2023-09-27
Modified `onIMECompositionChange` documentation to bring clarity to intended usage

Secticide:
### Description

When investigating a user's issue with the `onIMECompositionChange` event with a specific IME on Windows 10 (Korean), we noticed that there are cases where the blank event is not fired to signify the submission or resetting of the composition. The more consistent way to use this events is to use the IME Composition Change event to display the text that is currently being edited and rely on the `onTextInput` event for submission (resetting can be handled by listening for a new composition event and updating the displayed text accordingly).

The specific line that likely caused the user to *expect* (and therefore rely on) the empty string event is as follows:
> Fired once for every change, sends the entire string to date, and sends a blank string whenever a composition is submitted or reset.

### Changes made

- Modified the `onIMECompositionChange` event documentation to better point users to intended usage

### Notes

This is a work in progress PR with the intention of iterating on the documentation wording.


ekcoh (commented):
Documentation update LGTM. @duckets any feedback from your side?

duckets (approved):

ekcoh (approved):
Approving this one, lets merge it as soon as CI is working again. Doc checks might pick something up but looks good so likely not.

ekcoh:
@Secticide I believe you just need to resolve conflict in CHANGELOG.md now and lets CI rerun this branch. Then you may merge it to develop.

PR #1758, opened 2023-09-28
FIX: Exceptions in UITK action asset editor (ISX-1647, ISX-1648)

ekcoh:
### Description

Provides fixes for:
- Exception thrown when deselecting an action in UITK Input action asset editor. (ISX-1647)
- Exception printed when deselecting an action map in UITK Input action asset editor. (ISX-1648).

**Steps to reproduce first issue:**
1. Open UITK editor from Project Settings > Input actions to editor actions.
2. Select the first Action in actions list. 
3. Press ESC key.

An exception is thrown when accessing a IEnumerable<int> in editor code assuming the enumerable is never an empty sequence. 

**Steps to reproduce second issue:**

1. Open Project Settings and Input Actions editor
2. Select any Action Map
3. Press ESC to unselect

An NullReferenceException is thrown but is suppressed and instead a Debug.Log is generated in the editor.

This fix addresses the previously incorrect assumption that a selection change always results in a selection index sequence with count > 0. 

### Changes made

Updated code to handle both the case when selected indices is a sequence containing elements as well as an empty sequence. 
Second fix was to handle trying to lookup a null action map.

Update 2023-09-29: Added DeselectionHelper which currently helps preventing deselection in Action Map list and Action tree when there are maps and/or actions available. Left a boolean const to be able to quickly switch between allowing deselection or preventing it during development to identify issues with either mode. Considered this required since when bugs where eliminated with deselection this opened up for deselecting items which results in confusing UI representation.

### Notes

Open issues observed to follow up as part of different PR/tasks:
- Action tree not displaying "Tree is empty" or "List is empty" to be consistent with Action Map list when there are no actions.
- Action Properties not displaying "No action selected" when no action is selected or no actions exist.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ekcoh:
Realize now that there might be some additional behavioural work (lets see based on feedback) since it seems another issue is Binding properties aren't updated if action selection goes to null. Another approach would be to prevent null-selection at all in both views.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -108,7 +108,7 @@ public ActionsTreeView(VisualElement root, StateContainer stateContainer)
 
             m_ActionsTreeView.selectedIndicesChanged += indicies =>
             {
-                var index = indicies.First();
+                var index = indicies.FirstOrFallback(-1);
```
jamesmcgill: What about
`var index = indicies.Any() ? indicies.First() : -1;`
instead of making extension functions?

Also, not part of this PR but `indicies` should be `indices`.

jfreire-unity: Agree with @jamesmcgill here

ekcoh: Using Any() makes a lot of sense, however it is a little bit unreliable and will generate a warning since e.g. `var index = indicies.Any() ? indicies.First() : -1;` could possibly result in multiple-enumeration which isn't something guaranteed by IEnumerable<T> as a generic interface. In this particular situation its still safe since the underlying implementation is a List enumerator, but it becomes implementation dependent if UITK at any point would replace it by an enumerator not allowing multiple-enumeration. Still want me to change? 
This is the whole reason why I added an extension not having that problem.

ekcoh: I will begin updating the typo though until we just settle on this detail.

ekcoh:
Had to update this PR to allow preventing deselection since it leads to undesirable UX. Branch still supports both modes by flipping a bool at the moment.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/DeselectionHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        private const bool kAllowDeselection = true;
+
+        private int m_SelectedIndex = -1;
+
+        public bool Select(BaseVerticalCollectionView view, int selectedIndex)
+        {
+            if (!kAllowDeselection)
+            {
+                // This is a workaround to prevent deselection in the list or tree-view, e.g. pressing ESC key.
```
ekcoh: Following up on this statement to verify that there is no built-in support or guidance for best-practises with UITK team. For now this workaround seem to do the job.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
jamesmcgill: I didn't see a warning but of course, if you are getting one then it's not good and you should go with your solution in that case.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -132,6 +134,8 @@ private void OnKeyDownEventForDelete()
             item.DeleteItem();
         }
 
+        //private int m_PrevSelectedIndex = -1;
```
jamesmcgill: Some leftover comment

ekcoh: This has now been removed

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -106,13 +107,21 @@ public ActionsTreeView(VisualElement root, StateContainer stateContainer)
...
 
-            m_ActionsTreeView.selectedIndicesChanged += indicies =>
+            m_ActionsTreeView.selectedIndicesChanged += indices =>
             {
-                var index = indicies.First();
-                if (index == -1)
-                    return;
-                var item = m_ActionsTreeView.GetItemDataForIndex<ActionOrBindingData>(index);
-                Dispatch(item.isAction ? Commands.SelectAction(item.name) : Commands.SelectBinding(item.bindingIndex));
+                if (!m_DeselectionHelper.Select(m_ActionsTreeView, indices))
```
jfreire-unity: Nitpick: Would it make more sense to have this as something like `m_SelectionHelper.Deselect(...)` , `m_SelectionHelper.Disable(..)` or `m_SelectionHelper.DisableSelection(..)`? Just because it feels we're using a lot "select" wording all around so it would be nice if we could be consistent. Having "deselection" kind of seems counterintuitive.
But maybe it's really needed to have this "deselection" wording, I didn't do a full deep dive on this so I might lack context anyway.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -101,11 +108,6 @@ private void ChangeActionMapName(int index, string newName)
             Dispatch(Commands.ChangeActionMapName(index, newName));
         }
 
-        private void SelectActionMap()
-        {
-            Dispatch(Commands.SelectActionMap((string)m_ListView.selectedItem));
-        }
```
ritamerkl, line 107: A Coding-guideline question: I think outsourcing of code snippets can increase the readability. But that is a style decision.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/DeselectionHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Currently left as a constant to allow turning deselection on/off to be able to remove issues
+        /// related to not having selections more quickly. Also it might be desirable to either support or
+        /// not support deselection (E.g. pressing ESC while having a current selection).
+        /// </summary>
+        private const bool kAllowDeselection = false;
+
+        private int m_SelectedIndex = -1;
+
+        public bool Select(BaseVerticalCollectionView view, int selectedIndex)
```
ritamerkl: I agree with João. I think this method name is a bit confusing. Maybe something like IsSelectionValid()? As it's not really selecting or deselecting items in the method itself.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ekcoh: I think that's a valid comment but I also noticed this was the only place I could see where that function wasn't inlined with a lambda (comparing other view classes). Since this function was only called in a single place I considered it to be more readable when put in context with the triggering event callback (instead of having to lookup who is actually calling SelectActionMap) and also avoids anyone to call the function (removed from scope). I can move it back into a function as `void SelectActionMap(IEnumerable<int>) ` if you'd like?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ekcoh: I do not like the name DeselectionHelper either but failed to come up with a better name at the time of writing. The reason I named it Select is rather that it may be transparently returning a selection index or acting upon the current index or deselecting. Will try to come up with a better name..... would e.g. `m_SelectionFilter.Apply(...)` be better since it actually filters the selection change and modifies it if needed ? Typically for other UI frameworks the behaviour I am preventing here is typically called "de-selection"

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/DeselectionHelper.cs
ekcoh: I get that, I would like to avoid IsSelectionValid since that sounds like a read-only getter or evaluation method while this has side-effects for some cases and assigns selection to prevent deselection. Maybe ViewSelectionManager or similar is better? The function could be just be called Apply or be function operator overload if the class name makes sense since its basically a stateful functor.

ekcoh: I have updated the code based on your reviews and generalised the selection filter so it would be ready to also handle multi-selection whenever/if we would like to support that (not in current development scope). Also renamed in an attempt to make class names and identifiers make more sense

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ritamerkl: Good point, I didn't consider the aspect of the extended scope, I think it is good to avoid unwanted calls. Also if it is consistent to the other view classes you are right to keep it that way. I might have had a wrong perception of it. _Personally_ I like it more to split things up and give meaningful method names to make it readable. Now I am convinced to leave it as it is due to the reasons you mentioned.

ritamerkl (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/DeselectionHelper.cs
ritamerkl: It's hard to find a simple lucid name here. I think your new changes are appropriate!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ekcoh: This is no longer relevant since FirstOrFallback was removed from this PR so marking this resolved.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/DeselectionHelper.cs
ekcoh: Haven't received any feedback on this one so marking it resolved given workaround seem to do the job and will follow up later if needed.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/DeselectionHelper.cs
ekcoh: Lets rename it later (its an internal) if a better name is found.

Pauliusd01 (approved):
LGTM. Tried ESC with the bug steps in project wide actions and custom actions window, tried it with empty views/no selections and made sure things like rename fields are still cancelled out properly with the ESC key (All checked on both UITK and IMGUI versions)

PR #1761, opened 2023-09-29
FIX: Avoid making new created assets dirty when opened

jfreire-unity:
### Description

https://jira.unity3d.com/browse/ISX-1591

### Changes made

Assets without any action maps and control schemes are not set as "dirty" anymore.

### Notes

@Pauliusd01 since you reported this, please try it out and check if it solves the reported issue.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ritamerkl:
LGTM, only one thing: what happens if you have an existing asset and delete all action maps and control schemes? should still be dirty then, right? (it's a edge case, but still worth trying I think)

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -210,6 +210,11 @@ private void DirtyInputActionsEditorWindow(InputActionsEditorState newState)
         private bool HasAssetChanged(SerializedObject serializedAsset)
         {
             var asset = (InputActionAsset)serializedAsset.targetObject;
+
+            // Checks if the new asset is empty. If it is, there's nothing to save.
+            if (asset.actionMaps.Count == 0 && asset.controlSchemes.Count == 0)
+                return false;
+
             var newAssetJson = asset.ToJson();
             return newAssetJson != m_AssetJson;
```
ekcoh, line 235: I would recommend determining (if not really checked) why the this evaluation evaluates to true instead of false (?) instead of introducing additional conditions for some corner case that shouldn't exist if the underlying asset json produces different or similar result. Isn't the asset json for the newly created asset "empty" similar to what asset.ToJson() would produce or is there any difference there to look into and resolve?

Pauliusd01:
> LGTM, only one thing: what happens if you have an existing asset and delete all action maps and control schemes? should still be dirty then, right? (it's a edge case, but still worth trying I think)

Good catch, this indeed does not trigger the dirty state even though it should. @jfreire-unity

Pauliusd01 (changes requested):
Updating status to waiting for fixes

jfreire-unity:
@Pauliusd01 thanks! you can try again

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
jfreire-unity: Currently, the asset on disk has a specific layout.
Ideally, the layout shouldn't be `"{}"`. It should be a JSON serialized version of an empty asset. 
However, the `name` field is the problem. If you create and rename an input asset, its `name` field inside the asset is not updated at that stage. When opening the asset, there are actually changes to be made (the asset is flagged dirty) since the filename was changed.
The reason why this seems to happen is because on importing an asset, [we're actually setting `asset.name` based on the filename.](https://github.com/Unity-Technologies/InputSystem/blob/b0bcfbfbd0aa0561069447e1ec76d2ba7ec042fa/Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs#L85-L87)

I spent some time on this and am feeling it can become a rabbit hole and introduce regressions. But I think we should revisit this workflow.

jfreire-unity:
> LGTM, only one thing: what happens if you have an existing asset and delete all action maps and control schemes? should still be dirty then, right? (it's a edge case, but still worth trying I think)

Yep, quite on point use case. Thanks! I fixed it now

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -210,6 +210,14 @@ private void DirtyInputActionsEditorWindow(InputActionsEditorState newState)
         private bool HasAssetChanged(SerializedObject serializedAsset)
         {
             var asset = (InputActionAsset)serializedAsset.targetObject;
+
+            // Checks if the asset being edited is a new asset that was never saved before.
+            // If it is, there's nothing to save.
+            // At the moment, an asset only has the default asset layout content on disk when it is first created.
+            // So in this case we cannot go through the normal path and compare what's on disk with what has been serialized.
+            if (InputActionAsset.HasDefaultJsonLayout(m_AssetJson) && asset.IsEmpty())
```
ekcoh: Isn't this specific to this editor? I would recommend moving the logic to the associated utils class or making it an editor-only extension for the asset?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -83,6 +83,8 @@ public class InputActionAsset : ScriptableObject, IInputActionCollection2
         /// InputActionAssets.
         /// </remarks>
         public const string Extension = "inputactions";
+        ////REVIEW: actually pre-populate with some stuff?
```
ekcoh, line 86: What is the content after a save of an empty asset? Not sure why do not initialise it like that when its created? Then we wouldn't need these checks at all and the equality check would solve that for us

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
jfreire-unity: Thanks. Remove the method and just have it inline as I don't know if it needs to be used elsewhere at the moment.
Checking if the asset is empty is not editor dependent.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
jfreire-unity: Check https://github.com/Unity-Technologies/InputSystem/pull/1761#discussion_r1345711499

ritamerkl (approved):

Pauliusd01 (approved):
LGTM, did not notice anymore unusual scenarios where the asset gets dirty when it shouldn't (when undoing changes, redoing, creating new ones, cancelling already made ones and seeing if they are still dirty, etc)

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: Sounds odd we are setting name on import and not directly after constructing the asset, e.g. create, import, set name, write if needed, but I think you know best what is the best way to resolve this.

ekcoh (approved):

PR #1763, opened 2023-10-02
Added `EvaluateMagnitude` override to `Vector3Composite` to fix issues with callbacks

Secticide:
### Description

Issue tracker: [ISXB-251](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-251).

When an action is set to `Vector3Composite` using keyboard keys would result in started and performed callbacks firing but a cancel callback would only fire when losing focus. This doesn't map to what is explained in the documentation.

The issue was caused by [InputActionState.IsActuated](https://github.com/Unity-Technologies/InputSystem/blob/39f7cf63a8fd6700ca10af87735ec3b81d826c20/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs#L2690C8-L2690C8) returning true consistently for `Vector3Composite` actions. The result when calling `trigger.magnitude` was always -1 due to `Vector3Composite` not overriding `EvaluateMagnitude` (this was already resolved in `Vector2Composite`).

### Changes made

- Added `EvaluateMagnitude` override to `Vector3Composite`.
- Added new tests for cancel callback for both `Vector3Composite` & `Vector2Composite`
- Modified both `Vector3Composite` & `Vector2Composite` to be more consistent

### Notes

While changes have been made to `Vector3Composite` & `Vector2Composite` for consistency, they are entirely cosmetic.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -8059,6 +8059,39 @@ partial class CoreTests
 
     [Test]
     [Category("Actions")]
+    public void Actions_Vector2Composite_WithKeyboardKeys_CancelOnRelease()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+
+        // Set up classic WASD control.
+        var action = new InputAction();
+        action.AddCompositeBinding("Dpad")
+            .With("Up", "<Keyboard>/w")
+            .With("Down", "<Keyboard>/s")
+            .With("Left", "<Keyboard>/a")
+            .With("Right", "<Keyboard>/d");
+        action.Enable();
+
+        bool wasCanceled = false;
+        action.canceled += ctx => { wasCanceled = true; };
+
+        // Test all directions to ensure they are correctly canceled
+        var keys = new Key[] { Key.W, Key.A, Key.S, Key.D };
+        foreach (var key in keys)
+        {
+            wasCanceled = false;
+            InputSystem.QueueStateEvent(keyboard, new KeyboardState(key));
+            InputSystem.Update();
+
+            InputSystem.QueueStateEvent(keyboard, new KeyboardState());
+            InputSystem.Update();
+
+            Assert.That(wasCanceled, Is.EqualTo(true));
+        }
+    }
+
+    [Test]
+    [Category("Actions")]
     public void Actions_CanCreateComposite_WithPartsBeingOutOfOrder()
     {
         var gamepad = InputSystem.AddDevice<Gamepad>();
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector2Composite.cs
```diff
@@ -124,7 +124,7 @@ public class Vector2Composite : InputBindingComposite<Vector2>
         /// </code>
         /// </example>
         /// </remarks>
-        public Mode mode;
+        public Mode mode = Mode.DigitalNormalized;
```
ekcoh, line 127: Is this a API/behaviorial change (may not be a valid patch) or is it just a case of explicit initialisation? (fine)

ekcoh: I answer my own question, it seems like its not a change since Mode.DigitalNormalized was assigned value zero which would be equivalent to default(Mode), hence this is better in my opinion since explicit.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -8057,6 +8057,39 @@ public void Actions_Vector2Composite_RespectsButtonPressurePoint()
         Assert.That(value.Value, Is.EqualTo(Vector2.left));
     }
 
+    [Test]
+    [Category("Actions")]
+    public void Actions_Vector2Composite_WithKeyboardKeys_CancelOnRelease()
```
ekcoh, line 8062: Excellent to see you've added tests to cover this functionality which seem to have been missing.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector3Composite.cs
```diff
@@ -46,7 +46,7 @@ public class Vector3Composite : InputBindingComposite<Vector3>
         /// </remarks>
         // ReSharper disable once MemberCanBePrivate.Global
         // ReSharper disable once FieldCanBeMadeReadOnly.Global
-        [InputControl(layout = "Axis")] public int up;
+        [InputControl(layout = "Axis")] public int up = 0;
```
ekcoh: Not sure these need explicit initialisation since default(int) is known to be zero in C#?

ekcoh (approved):
Approved, I would consider removing explicit initialization of ints since known to default to zero implicitly in C# but keep the enum initialisation for clarity? Updating CHANGELOG.md to resolve conflict.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector3Composite.cs
Secticide: That makes sense, in that case I'll change the Vector2Composite to remove the explicit initialisation - just noticed some discrepancies in consistency and wanted to clean them up!

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector2Composite.cs
Secticide: I must admit, I'm a little sad that `Vector2Composite` and `Vector3Composite` both have a `Mode` enum with the same values (although they have been re-ordered for Vector2) and that the mode defaults are different.

But yes, this is just a cosmetic change for explicitness and consistency.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector3Composite.cs
Secticide: Resolved in f15b4feb5bbb9acf439cb77e17d2396ef6f69e04.

ekcoh: Ah, you decide what is best, consistency wins, not a big deal either way.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector2Composite.cs
ekcoh: Yes that inconsistency is indeed sad, something to address when a breaking change can be done.

Pauliusd01 (approved):
LGTM, I've basically just checked that the bug is fixed with the user repro project.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/Vector3Composite.cs
jamesmcgill: This change broke the test API_MinorVersionsHaveNoBreakingChanges on my branch.
It took me a while to trace it here because CI on develop and on this branch both seemed to have passed... 😸

PR #1764, opened 2023-10-03
FIX: Undo single action map deletion should restore actions

tspiller:
### Description

https://jira.unity3d.com/browse/ISX-1500 
When deleting the last remaining action map then undoing the deletion the action map is being restored without actions. 
This is a result of the selected action map index becoming -1 after deletion, which the undo system does not have the information to update back, and that index will return null when retrieving the related actions.

### Changes made

Updated the function that finds the selected action map after deletion so it matches the default index of 0 when there are no more action maps.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ekcoh (approved):
Small change :) Looks good to me. I think it relates to pending fix https://github.com/Unity-Technologies/InputSystem/pull/1758 so it will be interesting to test these in conjunction.

ekcoh:
Looks like branch is outdated with respect to develop so recommending updating branch and rerunning CI @tspiller

ekcoh:
Added @Pauliusd01 to get QA perspective on this, especially since not covered by new automatic tests

Pauliusd01:
https://github.com/Unity-Technologies/InputSystem/assets/54306142/e86dced2-8eb8-4713-af30-3562340cf9a2

Getting InvalidOperationException:/ObjectDisposedException:  errors when deleting something after an undo. 
Steps to repro are: Create a new Input Actions asset and fill it with at least 1 map and 1 actions binding -> Save and reopen just to be sure -> Delete the binding and undo -> now delete the map -> observe the errors

Pauliusd01:
By the way should the names of what you're undoing be a bit more readable? 
image

tspiller:
> By the way should the names of what you're undoing be a bit more readable?
> image

@Pauliusd01 These are a result of using default undo naming from serialised object changes. Not the most readable I know.
In order to make them more readable it is probably worth raising a separate issue for that because it would involve actively providing a more correct undo name with each relevant action which is outside the scope of this bug.

Pauliusd01 (approved):
LGTM, undoing and reodoing seems to properly return bindings/actions to how they were in both custom Input Actions window and Project Wide actions one. I did find odd that I am able to undo things even when the window is not open but that is in develop as well so I will report it separately.

PR #1765, opened 2023-10-04
Change inputactions asset auto-save behavior to trigger on editor focus lost

ekcoh:
### Description

Modifies the auto-save behaviour of all Input Action Asset editors as long as UNITY_INPUT_SYSTEM_INPUT_ACTIONS_EDITOR_AUTO_SAVE_ON_FOCUS_LOST is defined and otherwise keeps existing behaviour. 

The purpose with the change is to reduce unnecessary processing and generation during input action asset editing and trigger such auto-saves (if enabled or implicit) when the editor window looses focus.

### Changes made

- Modified `InputActionsEditorWindow` so that asset is only saved if dirty and window looses focus. It is assumed that this reflects the user switching focus to work on something else where it could be vital to have changes reflected. The goal with this behaviour is to not disrupt asset modifications with any domain reloads or other processing while doing edits.
- Modified `InputActionsEditorWindow` dirty flag and title so that its a direct reflection of asset being dirty when UNITY_INPUT_SYSTEM_INPUT_ACTIONS_EDITOR_AUTO_SAVE_ON_FOCUS_LOST is defined.
- Added UNITY_INPUT_SYSTEM_INPUT_ACTIONS_EDITOR_AUTO_SAVE_ON_FOCUS_LOST to input system asmdef.
- Added logic to `InputActionsEditorSettingsProvider` to derive a "edit focus" state which is not equivalent to Window focus as in `InputActionsEditorWindow` since it will not trigger on window focus alone, but instead is tied to any UI elements of the UITK editor having focus. Hence its a more restricted version of focus. When UNITY_INPUT_SYSTEM_INPUT_ACTIONS_EDITOR_AUTO_SAVE_ON_FOCUS_LOST is defined, the asset is now saved in the embedded project settings editor when focus is lost similar to the other editor. If not defined, asset is saved on every change.  

### Notes

The feature flag for the change in auto-save behaviour may be deleted later if agreed this is the desired behaviour. 

Intended squash merge commit message would be similar to CHANGELOG.md entry.

Would be good to get feedback in review in case I missed something, might have missed use-case of having UITK editor in standalone (editor window) mode?

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ekcoh:
If you are confused about commit history here its because I derived this form analytics branch commit so hence all the cleanups. It will be removed when squash committed.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
```diff
@@ -25,12 +26,66 @@ public override void OnActivate(string searchContext, VisualElement rootElement)
...
+            {
+                m_HasEditFocus = true;
+            }
+        }
+
+        private void OnEditFocusLost(FocusOutEvent @event)
+        {
+            // This can be used to detect focus lost events of container elements, but will not detect window focus
+            var element = (VisualElement)@event?.relatedTarget;
+            if (element == null && m_HasEditFocus)
```
jfreire-unity, line 74: I'd like to add something like this comment here as checking for `null` is not a common pattern :
```
            // `event.relatedTarget` contains the element that gains focus, which is null if we select
            // elements outside of project settings Editor Window.
 ```

**Review by jfreire-unity — APPROVED** [review]

File: Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs [hunk 2]
```diff
@@ -214,6 +227,15 @@ private bool HasAssetChanged(SerializedObject serializedAsset)
             return newAssetJson != m_AssetJson;
         }
 
+        private void OnLostFocus()
+        {
+            // Auto-save triggers on focus-lost instead of on every change
+            #if UNITY_INPUT_SYSTEM_INPUT_ACTIONS_EDITOR_AUTO_SAVE_ON_FOCUS_LOST
```
lyndon-unity, line 233: I think we should be checking if dirty before calling save here.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
ekcoh: Absolutely, will add, maybe I accidentally dropped that part from your investigation on this

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: You're absolutely right, will fix

ekcoh:
Thanks for reviews @jfreire-unity and @lyndon-unity, realize this PR lacks an adaption of the old editor used when UITK is not available as well so will re-add the work in progress label to this and address tomorrow morning. I will follow-up on the review comments as part of the next commit.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: This has now been addressed.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
ekcoh: Added a comment as suggested.

Pauliusd01 (approved):
Genuinely don't know if my review is useful here as I saw no difference with "UNITY_INPUT_SYSTEM_INPUT_ACTIONS_EDITOR_AUTO_SAVE_ON_FOCUS_LOST" added to scripting defines. The project wide actions autosave still worked immediately (visually), I couldn't get it to lose any data by abrubtly closing the window without losing focus (good) and the custom Actions asset window worked just as well with auto save on. So either all's good or I've set something up wrong with the define :D

PR #1768, opened 2023-10-05
NEW: Object picker for InputActionReference combining results from Project-wide Input Actions Asset and Project Input Actions Assets

ekcoh:
### Description

This PR provides a new object picker for InputActionReference that allows selecting Input Actions from regular .inputaction assets combined with matches from Project-Wide actions. The intention with this PR is to support the workflow with Project Settings based input actions and map. 

### Changes made

Implemented an object picker based on SearchProvider with the support and guidance of @lochrist that allows using a single object picker (Search window) to select `InputActionReference` assets from Project Settings folder asset and from regular .inputactions assets governed by the AssetDatabase. The functionality is wrapped into `InputReferencePropertyDrawer` which allows including it as an object picker in the Inspector. The component relies on UITK for property drawer.  

Note that all property drawers above are conditionally included based on the feature flag for project wide action functionality. If this feature is not enabled, behaviour should be similar to prior this PR and default to existing or default object prickers and property drawers.

Extracted code from `InputActionAssetImported` relating to icons into new class `InputActionAssetIconProvider` to be able to reference this functionality without caring about implementation details from importer and also from new search/property drawers.

### Notes

**Impact on existing property drawers:**
- PR modifies InputActionReference property drawer to rely on a picker capable of selecting from Project Settings asset actions as well as regular .inputaction assets without requiring the user to make any decision outside of the picker.
image

InputActionProperty doesn't require any modifications since it exposes an editor for a InputActionReference when it is configured to use a reference. When configured to represent an InputAction there is no implication on the property drawer by Project Wide actions. 
image

- Previous modifications to InputActionAsset could still make sense since None may be confusing. Probably needs discussion if this PR would be merged, but likely should remain as-is.
image

InputActionMap property drawer remains unaffected since its similar to InputAction drawer and allows creating and editing a map owned by the component and not providing support for indirect references.

There currently is no InputActionMapReference support in Input System but if there was, such a picker could be implemented the same way as the InputActionReferencePropertyDrawer part of this PR with little effort.

**Open issues:**
- Currently the asset loaded contains InputActionReference duplicates with different GUIDs. Everything points to this problem being unrelated to this PR or the editor functionality. **Update:** this have been resolved now by not attempting to load the InputActionReference assets directly, however it is unclear if the load is additive in some way leading to duplicates. If someone could help me understand as part of reviewing it would be helpful.
- It seems that InputActionReferences referencing Project-wide actions are detached from Property drawer monitoring. This is visible by referencing an action and then delete the action via the editor. Alternatively, rename the referenced action. The inspector will not realize this nor update compared to the same inspector property drawer referencing a .inputactions asset action where it will show "Missing (InputActionReference)" for the same scenario. Not sure if something is broken with the InputManager.asset, if need custom monitoring/scanning of InputActionReferences in some way or how we could resolve this. It seems unrelated to the editor in this PR but might not be unrelated to how InputActionReferences are retrieved. It could be related to creating new references I assume, when picking using regular ADB I assume we fetch InputActionReferences directly without creating new ones but not sure how this could be achieved for InputManager.asset.
- Minimum zoom level obfuscates items to only display description. Currently a fallback solution with a non-standard description is used that contains path and map/action name to allow the user to understand what asset it refers to at the minimum zoom level. For all other zoom levels this seem to not be an issue.
- ~~There seems to be some exceptions entering playmode where a `null` serialised property is reported in UI code. Not sure what is wrong here at the moment. This seems to be related to IMGUI since the UITK variant of this doesn't seem to suffer from the same issue. Needs more investigation.~~ It seems this must have been related to something in /Library while working on this. Haven't seen it since I cleaned up the /Library folder.

**Open issues (likely bugs) found in SearchProvider functionality - currently being investigated**
(PENDING) Issue with description - scenarios:
1. It seems that if SearchProvider.CreateItem(...) IS NOT given a description, SearchProvider IS GIVEN a fetchDescription then...
...the given description is derived via fetchDescription at all times regardless of Window zoom-level.
2. It seems that if SearchProvider.CreateItem(...) IS GIVEN a description and NOT a fetchDescription then...
...the given description to CreateItem is used as description for all zoom-levels of the Window.
3. It seems that if SearchProvider.CreateItem(...) IS GIVEN a description and IS GIVEN a fetchDescription then...
...the given description to CreateItem is used as description for all zoom-levels of the Window EXCEPT minimum zoom-level where fetchDescription is used instead.
Based on this behaviour (3) should be used to provide desired behaviour but it is needed to verify that this undocumented behaviour is not a bug to be reliable.

(PENDING) Incorrect documentation.
SearchProvider.CreateItem has incorrect documentation for `data` argument which says its a thumbnail when it really is a "user-provided reference passed to function callbacks". This should be reported and correct.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/InputActionReferenceContainer.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.InputSystem;
+
+public class InputActionContainer: MonoBehaviour
```
lyndon-unity: I assume this is just for testing and potentially shouldn't be committed ?
(Although as we don't ship this folder in the package I guess its less important)

Assets/InputActionReferenceContainer.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+using System.Collections;
```
lyndon-unity: 3 duplicated lines here

lyndon-unity: using System.Collections;
using UnityEngine;

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
```diff
@@ -0,0 +1,89 @@
...
+        private static IEnumerable<Object> GetInputActionReferenceAssets()
+        {
+            // Extract actions from actions maps from ProjectWideActionAsset.
+            // Note that we cannot do something like:
+            //  return AssetDatabase.LoadAllAssetsAtPath(ProjectWideActionsAsset.kAssetPath)
+            //      .Where((asset) => (asset is InputActionReference));
+            // Since that would return an outdated asset that is out of sync with ProjectWideActionsAsset.
+            // TODO Understand why this is a problem and update this comment
+            var asset = ProjectWideActionsAsset.GetOrCreate();
+            return (from actionMap in asset.actionMaps from action in actionMap.actions select InputActionReference.Create(action)).ToList();
```
lyndon-unity: Seems we need to merge both the assets and the project wide setting info here.
ATM this is just project wide info

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        private static IEnumerable<Object> GetInputActionReferenceAssets()
+        {
+            // Extract actions from actions maps from ProjectWideActionAsset.
+            // Note that we cannot do something like:
+            //  return AssetDatabase.LoadAllAssetsAtPath(ProjectWideActionsAsset.kAssetPath)
+            //      .Where((asset) => (asset is InputActionReference));
+            // Since that would return an outdated asset that is out of sync with ProjectWideActionsAsset.
+            // TODO Understand why this is a problem and update this comment
```
lyndon-unity: I think we simply need to rename the template file to just .json to exclude it from being a real item in the asset database

InputSystem\Packages\com.unity.inputsystem\InputSystem\Editor\ProjectWideActions\ProjectWideActionsTemplate.inputactions

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                var label = asset.name;
+                if (!label.Contains(context.searchText, System.StringComparison.InvariantCultureIgnoreCase))
+                    continue; // Ignore: not matching search text filter
+
+                yield return provider.CreateItem(context,
+                    asset.GetInstanceID().ToString(),
+                    label,
+                    $"{AssetDatabase.GetAssetPath(asset)} ({label})",
+                    (thumbnail == null) ? AssetDatabase.GetCachedIcon(ProjectWideActionsAsset.kAssetPath) as Texture2D : thumbnail,
```
lyndon-unity: I like that we have a new icon to indicate the project wide ones

These icons seem to update at I'm viewing the UI which is a bit distracting. 
(Would be nice if it was at least blank before updating to reduce distraction.)

I think this is potentially a UI issue outside the scope of this PR though.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
lyndon-unity: Seems in the UI it combines both - I'm not fully clear how that works, but the end result looks good.

lyndon-unity (approved):
Overall this is what I was expecting from the original task on this area
(https://jira.unity3d.com/browse/ISX-1321)

So looks good IMO.

I would generally approved the PR with a small comment on some duplicate code.

We do need to agree where backend data is stored asset vs project settings as we still have other issues with this project settings storage
* the bug with the UI when storing data in project settings (https://jira.unity3d.com/browse/ISX-1600)
* the issue with code generation requiring a file in the assets folder (in part https://jira.unity3d.com/browse/ISX-1635)
* potential standardisation for users with storing data as an asset.
* Issue with template file showing up in asset selectors (https://jira.unity3d.com/browse/ISX-1588)

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
lyndon-unity: See https://jira.unity3d.com/browse/ISX-1588

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
lyndon-unity: Icons I'm seeing - opportunity to revise the icon for project wide input actions (currently seems to be a magnifying glass)

image

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: Is it just this one or all package assets?

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: Thing is we do not have any new icons, I just used the one registered with asset database. The icon (blue+orange) is the .inputactions icon registered by InputActionImporter. This should really be the way to fetch it (the default value handling I have in the code). However the "Flash icon" is what is I currently set for the Project-wide in this search. That is currently used with .inputactions sub-assets (InputActions), but it could be debated what to use. I see the choice of icon as minor for this PR and could be addressed separately. However I propose here to differentiate assets and project-wide for increased visual cue.

ekcoh: The magnifying glass is a default icon from the SearchProvider

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: Not sure how it relates to this PR either, is it something relating to the search results you are after? It can be affected with SearchFlags and/or explicit filtering here as well.

Assets/InputActionReferenceContainer.cs
ekcoh: Yes, this file should be removed from the PR

Assets/SceneWithInputContainer.unity
ekcoh, line 1: This file should be removed from PR

Assets/SceneWithInputContainer.unity.meta
ekcoh, line 1: This file should be removed from PR

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -29,8 +29,8 @@ internal class InputActionImporter : ScriptedImporter
     {
         private const int kVersion = 13;
 
-        private const string kActionIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/InputAction.png";
-        private const string kAssetIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/InputActionAsset.png";
+        internal const string kActionIcon = "Packages/com.unity.inputsystem/InputSystem/Editor/Icons/InputAction.png";
```
ekcoh: Thinking about it, it likely makes sense to either have internal access on importer to load/fetch Texture2D instead or move these or such logic to its own class.

ekcoh: Even better is relying on AssetDatabase.GetCachedIcon but this doesn't seem to do the job for sub-assets. Might need more investigation.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
lyndon-unity: Its specific to this one file that we don't want showing up. We can ignore for the PR specifically, but I'm just indicating that this other issue could resolve the problem that the template data shows up. We should just fix that other issue ASAP to avoid this confusion

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: Not sure I follow? This SearchProvider only handles Project-wide, the "ADB" default provider handles the assets without custom code. The framework / standard solutions handles the combining for us.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: @lyndon-unity Do you see Input Action Icon (Flash) in Project-Wide Input Actions search group?

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
lyndon-unity: Yep I missed that in the code, but running the UI I see it. Ignore this.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace UnityEngine.InputSystem.Editor
+{
+    internal static class SearchConstants
+    {
+        // SearchFlags : these flags are used to customize how search is performed and how search
+        // results are displayed.
+        internal static readonly SearchFlags SearchFlags = SearchFlags.Sorted |
+            SearchFlags.OpenPicker |
+            SearchFlags.Packages;
```
lyndon-unity: I'd suggest it shouldn't have packages on by default. 
Turning this off makes it smaller and more manageable and more like the previous version

lyndon-unity (changes requested):
Requesting a change of whats listed

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
lyndon-unity: (This also excludes the unwanted project wide input actions 'template' until its renamed)

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: > I like that we have a new icon to indicate the project wide ones
> 
> These icons seem to update at I'm viewing the UI which is a bit distracting. (Would be nice if it was at least blank before updating to reduce distraction.)
> 
> I think this is potentially a UI issue outside the scope of this PR though.

Not sure was this was referring to? I need to understand this to address it I believe

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -10,18 +10,18 @@ namespace UnityEngine.InputSystem.Editor
 {
     internal static class ProjectWideActionsAsset
```
ekcoh, line 11: This refactoring doesn't provide value for the PR other than that it reduces use of internals and eliminates redundant similar code.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: I can update this since it wouldn't make a difference anyway. I assume you are referring to results produced by ADB which includes packages since otherwise a user couldn't refer to action imported via third-party or other packages, e.g. XRI.

lyndon-unity (approved):
Updates looks good

Assets/InputActionReferenceContainer.cs
ekcoh: This have been removed, resolving

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AssetSearchProviders.cs
ekcoh: Resolving this since updated according to request

ekcoh:
Going to comment on this myself, this branch needs auto-tests to verify behaviour.

ekcoh:
Transitioning this back to work-in-progress again unfortunately since I have detected issues with UI behavior that needs follow-up. Seems to be possible to work around but needs verification that we are not relying on side-effects from Search-related bugs.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -9550,6 +9549,36 @@ public void Actions_CanResolveActionReference()
...
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_CanImplicitlyConvertReferenceToAction_WhenAssigningActionFromReference()
+    {
+        var reference = ScriptableObject.CreateInstance<InputActionReference>();
+        InputAction action = reference; // implicit conversion
+        Assert.That(reference.action, Is.Null);
+    }
```
jfreire-unity, line 9581: It's unclear to me why these tests are needed in this PR as they seem to be unrelated to the object picker. Maybe I'm missing something or this is just being used for testing.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputActionReferenceValidator.cs
```diff
@@ -0,0 +1,185 @@
...
+    //      notice that inspector changes to "Missing (Input Action Reference)" as object is destroyed.
+    //      Then if user undo the operation there are two issues:
+    //      - Editor is restored to an object not referencing an existing editor object.
+    //      - The deletion in editor and inspector re-resolve is considered two different undo steps, they need
+    //        to be an atomic step.
+    //
+    // TODO Instead we likely need to integrate all reference management into each step of the editor.
+    internal class InputActionReferenceValidator
+    {
+        public static void ValidateReferences(InputActionAsset asset)
```
ritamerkl, line 77: Can we split this method up a bit to make it more readable? 
Maybe just outsource some code to methods with meaningful names like "HandleAddedOrRemovedActions()"?

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputActionReferenceValidator.cs
```diff
@@ -0,0 +1,185 @@
...
+                {
+                    Debug.Log("Removing invalid or dangling InputActionReference: " + references[i]);
+                    RemoveReferenceAtIndex(references, ref count, i);
+                }
+                else // Reference is associated with an action of this asset as expected
+                {
+                    // Look for first duplicate reference referencing the same action and eliminate this reference
+                    // if duplicates exist (Should not happen, basically corrupt asset). Additional duplicates are
+                    // covered since this is evaluated for each element.
+                    for (var j = i - 1; j >= 0; --j)
```
ritamerkl, line 120: for example here: calling a method like "EleminateDuplicateReferences" could help explain a lot without reading a comment

ekcoh:
Majority of this PR was picked up in another PR attempting to resolve this issue. given https://github.com/Unity-Technologies/InputSystem/pull/1834 these changes are no longer relevant so will close this PR.

ekcoh:
Majority of this PR was picked up in another PR attempting to resolve this issue. given https://github.com/Unity-Technologies/InputSystem/pull/1834 these changes are no longer relevant so will close this PR.

PR #1769, opened 2023-10-05
NEW: Add an IME input sample scene

Secticide:
### Description

We have had a select few bugs around IME input recently and I thought it would be a good idea to have a sample scene which doubles up as an example of how to handle IME input.

It isn't absolutely perfect since it doesn't handle everything perfectly, but it is a barebones scene that can be built on top of. Will be useful for us for testing IME related issues as well as customers to have a starting point when implementing IME input.

### Changes made

- Added IME input sample scene.


Assets/Samples/IMEInput/IMEInputHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+            m_CompositionField.text = string.Empty;
+            m_TextField.text = string.Empty;
+            m_CombinedField.text = string.Empty;
+
+            m_CurrentInput = string.Empty;
+            m_ComposingViaIME = false;
+        }
+
+        // Ensures all fields are correctly referenced
+        private void AssertInvariance()
```
lyndon-unity: Strange naming but I like the check

lyndon-unity (approved):
Overall I think this is a good addition.
Would suggest getting feedback from another member of the input team though.

Assets/Samples/IMEInput/IMEInputHandler.cs
Secticide: I agree actually, maybe `EnsureReferencesNotNull` or something like that would be clearer. Will get this switched out today.

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1769) <br/>All committers have signed the CLA.

jfreire-unity (approved):

PR #1773, opened 2023-10-09
CHANGE: Move project-wide actions into separate file

jamesmcgill:
### Description

Move Project-wide actions from `ProjectSettings/InputManager.asset` into it's own file `ProjectSettings/InputSystemActions.inputactions`.
This is to skirt around an issue where the ProjectSettings UI for InputManager would become corrupted when we first appended the `InputActionAsset` and required restarting the editor to clear.

### Changes made
Store ProjectWide Actions in `ProjectSettings/InputSystemActions.inputactions`.
Renamed ProjectWideActionsTest to ProjectWideActionsSample
Removed need to test ProjectWide Actions differently when running in the player by injecting a test asset.
Fix for BindingSyntax WithInteraction() which was using processors

### Notes

This is all the work from https://github.com/Unity-Technologies/InputSystem/tree/project-wide-actions-phase2-project-settings with the typesafe work removed.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,156 @@
...
-using UnityEngine.InputSystem.Utilities;
 
 namespace UnityEngine.InputSystem.Editor
 {
     internal static class ProjectWideActionsAsset
     {
-        internal const string kDefaultAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.inputactions";
-        internal const string kAssetPath = "ProjectSettings/InputManager.asset";
-        internal const string kAssetName = InputSystem.kProjectWideActionsAssetName;
+        internal const string kTemplateAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.inputactions";
```
lyndon-unity: May clash with Joao's PR but simple to resolve

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,156 @@
...
+        internal const string kInputManagerAssetPath = "ProjectSettings/InputManager.asset";
+        internal const string kAssetPath = "ProjectSettings/InputSystemActions.inputactions";
```
lyndon-unity, line 14: Is there any benefit to the .inputactions name rather than .json extension here?
(Minor question)

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,156 @@
...
         [InitializeOnLoadMethod]
         internal static void InstallProjectWideActions()
         {
             GetOrCreate();
         }
 
-        internal static InputActionAsset GetOrCreate()
+        // Copies the inputactionasset content from the Project Setting asset into a file that
+        // the Roslyn source generator will be able to read.
+        internal static void CreateRoslynAdditionalFileForAsset()
```
lyndon-unity: I don't see this being called yet so I assume this is just landing the prep work that has no effect yet

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,156 @@
...
+        {
+            try
+            {
+                if (File.Exists(kAssetPath))
+                {
+                    File.Copy(kAssetPath, kAdditionalFilePath, true);
```
lyndon-unity: Seems this should potentially only copy/input of the new data would be different to whats on disk, but those checks may happen prior to calling this function

lyndon-unity: Seems we are stuck with a fixed path here too, would be nice if we could still allow a user to move this. Perhaps querying the asset DB for a file of the specific name prior to saving so we overwrite a moved version rather than just recreating

lyndon-unity (approved):
Some more work needed on CreateRoslynAdditionalFileForAsset
Potentially not needed to be in this PR at all, but not a blocker as its not called AFAICT

Would like to confirm this change still keep project wide input actions working in both play mode and player, but looks reasonable from code.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jamesmcgill: No, that also should be removed

Assets/Samples/ProjectWideActionsSample/ProjectWideActionsSample.cs
```diff
@@ -0,0 +1,68 @@
...
+            if (attack.WasPressedThisFrame())
+            {
+                cube.GetComponent<Renderer>().material.color = Color.red;
+            }
+            else if (attack.WasReleasedThisFrame())
+            {
+                cube.GetComponent<Renderer>().material.color = Color.green;
+            }
+
+            var moveVal = move.ReadValue<Vector2>();
```
ekcoh, line 48: (Minor) I guess you only renamed this file without git move so looks like its not related but why not simplify the translation part since its not the main thing about this sample anyway?

Basically, this

```
        var moveVal = move.ReadValue<Vector2>();
        if (moveVal.x < 0.0f)
        {
            cube.transform.Translate(new Vector3(-10 * Time.deltaTime, 0, 0));
        }
        else if (moveVal.x > 0.0f)
        {
            cube.transform.Translate(new Vector3(10 * Time.deltaTime, 0, 0));
        }
        if (moveVal.y < 0.0f)
        {
            cube.transform.Translate(new Vector3(0, -10 * Time.deltaTime, 0));
        }
        else if (moveVal.y > 0.0f)
        {
            cube.transform.Translate(new Vector3(0, 10 * Time.deltaTime, 0));
        } 
```

Could be replaced by (Skipping the zero-check if for simplicity and relying on implicit cast from Vector2 to Vector3):

```
var moveVal = move.ReadValue<Vector2>();
cube.transform.Translate(moveVal * (10.0f * Time.deltaTime));
```

ekcoh: It could be a one liner but maybe good to keep the ReadValue outside of the Translate for clarity

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -14,43 +12,27 @@
...
+
+        var map = testAsset.AddActionMap("InitialActionMapOne");
         map.AddAction("InitialActionOne");
         map.AddAction("InitialActionTwo");
 
-        m_TemplateAssetPath = Path.Combine(Environment.CurrentDirectory, "Assets/ProjectWideActionsTemplate.inputactions");
-        File.WriteAllText(m_TemplateAssetPath, templateActions.ToJson());
-
-        ProjectWideActionsAsset.SetAssetPaths(m_TemplateAssetPath, TestAssetPath);
+#if UNITY_EDITOR
```
ekcoh, line 32: Good this was simplified

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -14,43 +12,27 @@
...
         // This is used as a backing for `InputSystem.actions` in PlayMode tests.
-        var testAsset = ScriptableObject.CreateInstance<TestActionsAsset>();
-        AssetDatabase.CreateAsset(testAsset, TestAssetPath);
-
-        // Create a template `InputActionAsset` containing some test actions.
-        // This will then be used to populate the initially empty `TestActionsAsset` when it is first acessed.
-        var templateActions = ScriptableObject.CreateInstance<InputActionAsset>();
-        templateActions.name = "TestAsset";
-        var map = templateActions.AddActionMap("InitialActionMapOne");
+        var testAsset = ScriptableObject.CreateInstance<InputActionAsset>();
```
ekcoh, line 25: (Minor) Since this is constructed in SetUp, maybe it would be good to keep the reference and call ScriptableObject.Destroy(testAsset) in TearDown() instead of leaking it and only relying on it being destroyed with the Scene?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ however, it has to be formatted properly to pass verification tests.
 - Static analysis warnings regarding flag enums have been suppressed in order to avoid compile-time warnings or errors.
 - Action Map and Action Tree views of the UI Toolkit based Input Action Editor now prevents deselection in both views when Escape key is pressed.
 - Input Action Asset editors Auto-save feature has been modified to trigger on focus-lost when activated instead of triggering on every modification to the asset in order to reduce impact of processing required to handle modified assets.
+- Store ProjectWide Actions in `ProjectSettings/InputSystemActions.inputactions`.
```
ekcoh, line 20: Minor: Since this is a change log maybe
"- Store ProjectWide Actions in `ProjectSettings/InputSystemActions.inputactions` instead of `ProjectSettings/InputManager.asset`"

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ however, it has to be formatted properly to pass verification tests.
...
+- Renamed ProjectWideActionsTest to ProjectWideActionsSample
```
ekcoh, line 21: Minor: Missing full stop at the end.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionSetupExtensions.cs
```diff
@@ -1245,9 +1245,9 @@ public BindingSyntax WithInteraction<TInteraction>()
                 if (!valid)
                     throw new InvalidOperationException("Accessor is not valid");
 
-                var interactionName = InputProcessor.s_Processors.FindNameForType(typeof(TInteraction));
+                var interactionName = InputInteraction.s_Interactions.FindNameForType(typeof(TInteraction));
                 if (interactionName.IsEmpty())
-                    throw new NotSupportedException($"Type '{typeof(TInteraction)}' has not been registered as a processor");
+                    throw new NotSupportedException($"Type '{typeof(TInteraction)}' has not been registered as a interaction");
```
ekcoh, line 1250: I am not native English so might be wrong but should it be ".... registered as **an** interaction"?

ekcoh: Additionally, if this is due to the user of this forgetting to register it maybe the message should guide to the class/method to use?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: It might relate to `InputActionImporter` being triggered on this or not since its registered to that type? If it is that would have major impact on how this works and what we need to do. Basically it could solve issues I am currently investigating for picker.

ekcoh: InputActionImporter is the logic that picks up such files and makes AssetDatabase capable of managing them. Not sure this would be the case though since outside Assets folder, but maybe this relates to how its registered in native (trunk)?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,132 @@
...
-        internal static void SetAssetPaths(string defaultAssetPath, string assetPath)
+        internal static void SetTestAsset(InputActionAsset testAsset)
         {
-            s_DefaultAssetPath = defaultAssetPath;
-            s_AssetPath = assetPath;
+            s_TestAsset = testAsset;
         }
 
-        internal static void Reset()
+        internal static void ResetTestAsset()
```
ekcoh, line 24: Do we need this Reset function when we could just call `SetTestAsset(null)`?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,132 @@
...
-        internal const string kDefaultAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.json";
-        internal const string kAssetPath = "ProjectSettings/InputManager.asset";
-        internal const string kAssetName = InputSystem.kProjectWideActionsAssetName;
+        internal const string kTemplateAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.json";
+        internal const string kInputManagerAssetPath = "ProjectSettings/InputManager.asset";
+        internal const string kAssetPath = "ProjectSettings/InputSystemActions.inputactions";
 
-        static string s_DefaultAssetPath = kDefaultAssetPath;
-        static string s_AssetPath = kAssetPath;
+        static InputActionAsset s_TestAsset = null;
```
ekcoh, line 16: This field is only used in tests and should be inside #if UNITY_INCLUDE_TESTS guard

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,132 @@
...
+            return asset;
         }
 
-        private static InputActionAsset CreateNewActionAsset()
+        internal static InputActionAsset LoadFromInputManagerSettings()
         {
-            var json = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath));
+            var objects = AssetDatabase.LoadAllAssetsAtPath(kInputManagerAssetPath);
+            if (objects != null)
+                return objects.FirstOrDefault(o => o != null && o.name == InputSystem.kProjectWideActionsAssetName) as InputActionAsset;
```
ekcoh, line 69: I am not fond of this `FirstOrDefault`.... basically by using this code we are saying that its OK to have multiple assets with name `kProjectWideActionsAssetName` at path `kInputManagerAssetPath`. 
Its fine to select it using approach above with `FirstOrDefault` if we prior to calling that would check that there is only ONE or ZERO such InputActionAsset otherwise I think we should at least write an error message to the console or possibly throw an exception. 
This comment could be ignored if we create a separate PR/ticket on addressing this.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -4,103 +4,132 @@
...
-                foreach (var action in map.actions)
-                {
-                    var actionId = action.m_Id;
-                    if (string.IsNullOrEmpty(actionId) || asset.actionMaps.Sum(m => m.actions.Count(a => a.m_Id == actionId)) > 1)
-                        action.GenerateId();
-                }
+            // v1.8.0-pre1 stored the Actions in ProjectSettings/InputManager.asset.
+            // Check if we have a project that was saved on that version and migrate it
+            // to save in the default location used from the pre2 release.
+            asset = LoadFromInputManagerSettings();
```
ekcoh, line 87: I cannot see that we actually remove InputActionAsset from InputManager.asset here so are we really migrating and not just making a copy? I think we need to remove the old one from InputManager.asset, otherwise it would be enumerated and possible to load if loading InputManager.asset.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3028,31 +3028,31 @@ public static InputActionAsset actions
...
-                    return s_projectWideActions;
+                if (s_ProjectWideActions != null)
+                    return s_ProjectWideActions;
 
                 #if UNITY_EDITOR
-                s_projectWideActions = Editor.ProjectWideActionsAsset.GetOrCreate();
+                s_ProjectWideActions = ProjectWideActionsAsset.GetOrCreate();
                 #else
-                s_projectWideActions = Resources.FindObjectsOfTypeAll<InputActionAsset>().FirstOrDefault(o => o != null && o.name == kProjectWideActionsAssetName);
+                s_ProjectWideActions = Resources.FindObjectsOfTypeAll<InputActionAsset>().FirstOrDefault(o => o != null && o.name == kProjectWideActionsAssetName);
```
ekcoh, line 3037: Similar as previous comment on FirstOrDefault, should we make sure there are not multiple fulfilling this?

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3028,31 +3028,31 @@ public static InputActionAsset actions
...
+                return s_ProjectWideActions;
             }
 
             set
             {
                 if (value == null)
                     throw new ArgumentNullException(nameof(value));
 
-                if (s_projectWideActions == value)
+                if (s_ProjectWideActions == value)
```
ekcoh, line 3050: If set is allowed, shouldn't make sure its has the name kProjectWideActionsAssetName or enforce it? Otherwise those other checks would fail?

ekcoh (changes requested):
Overall looks good. Added a bunch of minor comments but some more major that needs follow-up.

jamesmcgill:
The majority of this commit history will be in https://github.com/Unity-Technologies/InputSystem/pull/1751
If this needs resurrected in future, cherry pick the single commit out of this that actually does the file move.

PR #1775, opened 2023-10-18
FIX: TreeView tracking of collapse/expanded states (ISX-1626)

ritamerkl:
### Description

On deleting or duplicating bindings in composites the TreeView was collapsing all items sometimes. Another behavior (not tracked in the ticket) was that the collapsed/expanded states of the actions and bindings were often not correct after adding/duplicating and deleting actions and bindings on the TreeView. This behavior got worse when having more items in the TreeView. The main cause was that the id's of the visual elements did not match the items anymore and the auto save mode of expansion states was not turned on on the TreeView. The ticket to the issue can be found [here](https://jira.unity3d.com/browse/ISX-1626).

### Changes made

fixed:
- changed index of VisualElements to be constant through all Rebuilds of the UI to make the saving of their expand state possible
- added viewDataKey to start the automatic saving of expanded states in TreeView

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem.Editor/SelectorsTests.cs
+++ b/Assets/Tests/InputSystem.Editor/SelectorsTests.cs
@@ -1,6 +1,8 @@
 // UITK TreeView is not supported in earlier versions
 // Therefore the UITK version of the InputActionAsset Editor is not available on earlier Editor versions either.
 #if UNITY_EDITOR && UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
+using System;
+using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
 using UnityEngine;
@@ -22,7 +24,7 @@ internal class SelectorsTests
         var actionTwo = actionMap.AddAction("Action2", binding: "<Keyboard>/d");
 
 
-        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate());
+        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate(), new Dictionary<Guid, int>());
 
 
         Assert.That(treeViewData.Count, Is.EqualTo(2));
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -33,6 +33,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed case [ISXB-580](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-580) (UI Submit / Cancel not working with Switch Pro controller) by adding "Submit" & "Cancel" usages to the Switch Pro controller input controls.
 - Fixed an issue where undoing deletion of Action Maps did not restore Actions correctly.
 - Fixed case [ISXB-628](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-628) (OnIMECompositionChange does not return an empty string on accept when using Microsoft IME) by clarifying expectations and intended usage for the IME composition change event.
+- Fixed case [ISX-1626](https://jira.unity3d.com/browse/ISX-1626) where the expanded/collapsed state of items in the input action editor was not properly saved between rebuilds of the UI.
 
 ## [1.8.0-pre.1] - 2023-09-04
 
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/SerializedInputAction.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/SerializedInputAction.cs
@@ -12,6 +12,7 @@ namespace UnityEngine.InputSystem.Editor
             // serialized fields and make sure they're present?
             wrappedProperty = serializedProperty ?? throw new ArgumentNullException(nameof(serializedProperty));
 
+            id = serializedProperty.FindPropertyRelative(nameof(InputAction.m_Id)).stringValue;
             name = serializedProperty.FindPropertyRelative(nameof(InputAction.m_Name)).stringValue;
             expectedControlType = ReadExpectedControlType(serializedProperty);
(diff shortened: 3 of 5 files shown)


ritamerkl:
@Pauliusd01 could you test and confirm it's fixed?

Pauliusd01:
Found this one weird edge case, I don't know exactly why It's happening but I've got 100% repro at least. Do you think this PR should have fixed it too? (It's reproable here as well as in develop)

https://github.com/Unity-Technologies/InputSystem/assets/54306142/784daf45-7d7a-42bb-80de-826316c24c53

Pauliusd01:
Other than the one minor edge case mentioned I think this would be safe to merge, what I checked was: duplicating and deleting different bindings in Project Settings and Custom Input Actions Window, then undoing redoing, duplicating and deleting -> changing folds -> undoing, etc and looking at the the folds to make sure they still make sense. 

This part below could be ignored but I'm just mentioning it in case you also saw something like this and or you think your PR could have contributed to these:
```
Two things that I also found but could not pin down properly were: some rare scenarios where I would delete a binding, fold the parent action and undo -> my expectation would be for the fold to unfold but with your PR it wouldn't (develop would do so) 
AND
It would sometimes ignore my right clicks right after duplicating and deleting but it is hard to pin down and I don't see why your PR would even be at fault here.
```

ritamerkl:
> Found this one weird edge case, I don't know exactly why It's happening but I've got 100% repro at least. Do you think this PR should have fixed it too? (It's reproable here as well as in develop)
> 
>  Unity_2023-10-20_14-15-00.mp4

Looking at the video it's unclear to me what you are doing. Do you press any keys to change the selection or does it happen automatically?

ritamerkl:
About your other two findings: 

> some rare scenarios where I would delete a binding, fold the parent action and undo -> my expectation would be for the fold to unfold but with your PR it wouldn't (develop would do so) 

This is a good point. What is the expected behavior here?

> It would sometimes ignore my right clicks right after duplicating and deleting but it is hard to pin down and I don't see why your PR would even be at fault here.

Interesting - does this happen in develop as well?

Pauliusd01:
> > Found this one weird edge case, I don't know exactly why It's happening but I've got 100% repro at least. Do you think this PR should have fixed it too? (It's reproable here as well as in develop)
> > Unity_2023-10-20_14-15-00.mp4
> 
> Looking at the video it's unclear to me what you are doing. Do you press any keys to change the selection or does it happen automatically?

Sorry, I rushed that comment out and forgot to write the steps. What I'm doing there is selecting the UI map -> Right Click -> Right Button binding and deleting it via the shortcut -> My selection jumps to a different folded binding which is then unfolded. And what I wanted to show in the video is that it happens on both this PR and Develop so it was more of a "Is this the same issue and should it have been fixed here?" thing.

Pauliusd01:
> About your other two findings:
> > some rare scenarios where I would delete a binding, fold the parent action and undo -> my expectation would be for the fold to unfold but with your PR it wouldn't (develop would do so)
> This is a good point. What is the expected behavior here?

My expectation is for the undo to select the object that was undone and thus also unfold its parent if it is folded. But I would not spend too much time on this, it was just a minor observation.

> > It would sometimes ignore my right clicks right after duplicating and deleting but it is hard to pin down and I don't see why your PR would even be at fault here.
> Interesting - does this happen in develop as well?

This is super hard to tell It's more of a "feel" thing so again I included this in the "ignore if you didn't notice it yourself" section. It just "seems" like it ignores my clicks on your PR but not on develop but I, over long time of trying to pin something decent for you to try out yourself wasn't able to. Feel free to ignore it if you didn't get the same issue

ritamerkl:
> > > Found this one weird edge case, I don't know exactly why It's happening but I've got 100% repro at least. Do you think this PR should have fixed it too? (It's reproable here as well as in develop)
> > > Unity_2023-10-20_14-15-00.mp4
> > 
> > 
> > Looking at the video it's unclear to me what you are doing. Do you press any keys to change the selection or does it happen automatically?
> 
> Sorry, I rushed that comment out and forgot to write the steps. What I'm doing there is selecting the UI map -> Right Click -> Right Button binding and deleting it via the shortcut -> My selection jumps to a different folded binding which is then unfolded. And what I wanted to show in the video is that it happens on both this PR and Develop so it was more of a "Is this the same issue and should it have been fixed here?" thing.

Ok- thank you for clarifying. Do we have a ticket for this? 
To answer your initial question: no, this PR is not a fix for this issue and also not related. My guess is that this issue happens because the selected Index is set wrongly/not updated after deleting items from the list. I was able reproduce the same behavior after undoing a duplication.

ritamerkl:
> > > some rare scenarios where I would delete a binding, fold the parent action and undo -> my expectation would be for the fold to unfold but with your PR it wouldn't (develop would do so)
> > > This is a good point. What is the expected behavior here?
> 
> My expectation is for the undo to select the object that was undone and thus also unfold its parent if it is folded. But I would not spend too much time on this, it was just a minor observation.

This is interesting either way. I think it is worth putting into a ticket with low priority so we don't forget about it. But I agree, it is a minor issue.

> 
> > > It would sometimes ignore my right clicks right after duplicating and deleting but it is hard to pin down and I don't see why your PR would even be at fault here.
> > > Interesting - does this happen in develop as well?
> 
> This is super hard to tell It's more of a "feel" thing so again I included this in the "ignore if you didn't notice it yourself" section. It just "seems" like it ignores my clicks on your PR but not on develop but I, over long time of trying to pin something decent for you to try out yourself wasn't able to. Feel free to ignore it if you didn't get the same issue

I never experienced that, but I am working on Mac, and if I remember correctly you are on Windows, right? Could be a platform dependent thing.

Pauliusd01 (approved):

ekcoh:
@ritamerkl @Pauliusd01 did you file that additional inconvenience in behavior in a separate ticket? If so I believe we can overlook that in the context of this PR.

ekcoh (approved):

PR #1776, opened 2023-10-18
Fixed GetHapticCapabilitiesCommand always failing to execute

chris-massie:
### Description

Fixed `GetHapticCapabilitiesCommand` always failing to execute.

### Changes made

Added new fields to the command to make the payload size in bytes match the expected size.

~~I modified the names of the command fields to match the names in Unity core and the XR module, and created deprecated properties that map to the new fields. The naming makes it more clear they only apply for buffered haptics. Though this is technically a breaking change, since the command never worked at all, I think this is acceptable.~~ EDIT: I kept the old names to avoid API breaking.

~~I also renamed the old properties in `HapticCapabilities` to match XR module with auto migration. This is not a breaking change.~~ EDIT: Reverted since the `API_MinorVersionsHaveNoBreakingChanges` test isn't sophisticated enough to recognize this is allowed.

### Notes

Starting with Unity 2018.3, the `UnityXRHapticCapabilities` struct was modified to add new properties, which adjusted the size of the struct. The `XRInputSubsystem` verifies that the size of the buffer passed to the command matches. It was previously always failing because 12 bytes != 20 bytes.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
```diff
@@ -1,31 +1,104 @@
...
+        /// <summary>
+        /// Initializes and returns an instance of <see cref="HapticCapabilities"/>.
+        /// </summary>
+        /// <param name="numChannels">The number of haptic channels available on this device.</param>
+        /// <param name="supportsImpulse">This device supports sending a haptic impulse.</param>
+        /// <param name="supportsBuffer">This device supports sending a haptic buffer.</param>
+        /// <param name="bufferFrequencyHz">The buffer frequency the device operates at in Hertz.</param>
+        /// <param name="bufferMaxSize">The max amount of buffer data that can be stored by the device.</param>
+        /// <param name="bufferOptimalSize">The optimal size of a device's buffer, taking into account frequency and latency.</param>
+        public HapticCapabilities(uint numChannels, bool supportsImpulse, bool supportsBuffer, uint bufferFrequencyHz, uint bufferMaxSize, uint bufferOptimalSize)
```
vrdave-unity, line 24: Is this not a breaking change to remove the old constructor and add a new one with different parameters?

vrdave-unity (approved):
Good find.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
```diff
@@ -1,31 +1,104 @@
...
 
 namespace UnityEngine.InputSystem.XR.Haptics
 {
-    public struct HapticCapabilities
+    /// <summary>
+    /// Describes the haptic capabilities of a specific device.
+    /// </summary>
+    public readonly struct HapticCapabilities
     {
-        public HapticCapabilities(uint numChannels, uint frequencyHz, uint maxBufferSize)
```
jfreire-unity, line 11: Thanks for this PR, Chris!

This seems to be a breaking change for the package and we would need to maintain this method. A bit annoying to have to leave it as we won't do a 2.0 soon. But maybe there's a way to fix the issue you found without removing this constructor?

jfreire-unity (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
```diff
@@ -1,31 +1,104 @@
...
+        /// <seealso cref="SendBufferedHapticCommand"/>
+        public bool supportsBuffer { get; }
+
+        /// <summary>
+        /// The buffer frequency the device operates at in Hertz. This impacts how fast the device consumes buffered haptic data.
+        /// </summary>
+        /// <remarks>
+        /// This value is greater than 0 if <see cref="supportsBuffer"/> is <see langword="true"/>, and 0 otherwise.
+        /// </remarks>
+        public uint bufferFrequencyHz { get; }
```
jfreire-unity: I assume this naming logic comes from some XR package naming, right?
Is this sample frequency of buffer "of samples" frequency? 
Is this buffer a PCM buffer by the way?
Curiousity: Is this part of OpenXr spec?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
chris-massie: The constructor isn't removed, I just moved it under the new one. It has the same signature and parameter names, so it should be backwards compatible.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
chris-massie: It comes from [`HapticCapabilities`](https://docs.unity3d.com/ScriptReference/XR.HapticCapabilities.html) from the `UnityEngine.XR` namespace. It also comes from the [struct](https://github.cds.internal.unity3d.com/unity/unity/blob/trunk/Modules/XR/Subsystems/Input/ProviderInterface/IUnityXRInput.h#L429-L443) in the Unity interface that XR providers use.

`supportsBuffer` is currently always false with OpenXR, so I don't really know the details of these `uint` properties. But I verified with a temp modification of OpenXR package to set dummy data and they came through fine.

chris-massie:
Like @jamesmcgill I was getting failures of API_MinorVersionsHaveNoBreakingChanges when running locally on my computer, see https://github.com/Unity-Technologies/InputSystem/pull/1763#discussion_r1350477430. I reverted the change of adding a redundant explicit field initialization to `Vector2Composite` to make the test pass.

I also updated the branch to revert the obsolete and renaming I did previously. The names won't match the UnityEngine.XR namespace, but it seems we can't really do so without doing a breaking change.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
jfreire-unity: Ah, my bad. Should have looked better in the diff. Thanks and great to know 👍

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
```diff
@@ -6,26 +6,85 @@
...
+        public uint optimalBufferSize { get; }
     }
 
     [StructLayout(LayoutKind.Explicit, Size = kSize)]
     public struct GetHapticCapabilitiesCommand : IInputDeviceCommandInfo
     {
         static FourCC Type => new FourCC('X', 'H', 'C', '0');
 
-        const int kSize = InputDeviceCommand.kBaseCommandSize + sizeof(uint) * 3;
+        // 20 bytes of data from uint(4) + bool(1) + bool(1) + padding + uint(4) + uint(4) + uint(4)
```
jfreire-unity, line 89: 👍

jfreire-unity (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
```diff
@@ -35,13 +94,22 @@ public struct GetHapticCapabilitiesCommand : IInputDeviceCommandInfo
...
 
-        [FieldOffset(InputDeviceCommand.kBaseCommandSize + (sizeof(uint) * 2))]
+        [FieldOffset(InputDeviceCommand.kBaseCommandSize + 12)]
         public uint maxBufferSize;
 
-        public HapticCapabilities capabilities => new HapticCapabilities(numChannels, frequencyHz, maxBufferSize);
+        [FieldOffset(InputDeviceCommand.kBaseCommandSize + 16)]
+        public uint optimalBufferSize;
+
+        public HapticCapabilities capabilities => new HapticCapabilities(numChannels, supportsImpulse, supportsBuffer, frequencyHz, maxBufferSize, optimalBufferSize);
```
jamesmcgill, line 142: Maybe an xmldoc for the new property? (I know the rest of the struct is already missing docs)

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Haptics/GetHapticCapabilitiesCommand.cs
chris-massie: I added docs to everything in cde310ab503e6b3ed3f146d9969388f6c3891a69.

jamesmcgill:
@chris-massie There is another PR about to land, so it's going to need another update soon I'm afraid. 😄 
I can do that and make sure it's merged next, if you are okay with that?

ekcoh (approved):

PR #1779, opened 2023-10-25
NEW: Allow selection of InputActionReferences from project-wide actions

jfreire-unity:
### Description

https://jira.unity3d.com/browse/ISX-1660
https://jira.unity3d.com/browse/ISX-1664

### Changes made

This PR allows users to select InputActionReferences from project-wide actions.

It is based on most of the POC work done in #1768 but with some refactoring:
- It adds a new UI advanced picker for the InputActionReference property
- Deals with reference management issues caused by having an InputActionAsset outside of the Assets folder. Project-wide actions are part of "ProjectSettings/InputManager.asset" which abides by a different workflow as `.inputaction` assets. More specifically, project-wide actions don't have InputActionReferences as sub-assets, which means they are not updated once the `ScriptedImporter.OnImportAsset runs`. This means we need to "manually" update the InputActionReferences to match the existing input actions in the project-wide actions asset. Essentially, making sure that input actions in the asset have references, and also making sure that the references point to the correct input action.

### Notes

Feel free to capture issues regarding the Undo system so that we can deal with them in a separate PR.
I believe if users can use action references they added, renamed, or deleted in project-wide actions, it is still valuable to pre.2, even if there's some stabilization to be done regarding Undo actions.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -212,12 +210,51 @@ public override void OnImportAsset(AssetImportContext ctx)
             InputActionEditorWindow.RefreshAllOnAssetReimport();
         }
 
+        internal static IEnumerable<InputActionReference> LoadInputActionReferencesFromAsset(InputActionAsset asset)
+        {
+            //Get all InputActionReferences are stored at the same asset path as InputActionAsset
+            return AssetDatabase.LoadAllAssetsAtPath(AssetDatabase.GetAssetPath(asset)).Where(
+                o => o is InputActionReference && o.name != "InputManager").Cast<InputActionReference>();
```
lyndon-unity, line 217: Do we need this part?
 && o.name != "InputManager

InputManager data shouldn't be an InputActionReference should it ?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -86,8 +86,16 @@ private static InputActionAsset CreateNewActionAsset()
...
+            CreateInputActionReferences(asset);
+
+            AssetDatabase.SaveAssets();
+
+            return asset;
+        }
+
+        private static void CreateInputActionReferences(InputActionAsset asset)
+        {
+            var maps = asset.actionMaps;
```
lyndon-unity, line 98: Do we need a check here to make sure the asset doesn't already have the InputActionReferences before they are created?

var actionReference = ScriptableObject.CreateInstance<InputActionReference>();

lyndon-unity: I guess not as its only called from CreateNewActionAsset

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -97,10 +105,51 @@ private static InputActionAsset CreateNewActionAsset()
...
+                {
+                    actionReference.Set(null);
+                    AssetDatabase.RemoveObjectFromAsset(actionReference);
+                }
+            }
+
+            // Check if all actions have a reference
+            foreach (var action in asset)
+            {
+                var actionReference = existingReferences.FirstOrDefault(r => r.m_ActionId == action.id.ToString());
```
lyndon-unity, line 134: Feels like we could do this faster, e.g. by generating a hash in the above foreach look and then checking if its in the hash here - rather than splining the eixtingReferences list on every action in the asset.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -97,10 +105,51 @@ private static InputActionAsset CreateNewActionAsset()
...
+                    AssetDatabase.AddObjectToAsset(actionReferenceNew, asset);
+                }
+                else
+                {
+                    // Update the name of the reference if it doesn't match the action name.
+                    if (actionReference.name != InputActionReference.GetDisplayName(action))
+                    {
+                        AssetDatabase.RemoveObjectFromAsset(actionReference);
+                        actionReference.name = InputActionReference.GetDisplayName(action);
+                        AssetDatabase.AddObjectToAsset(actionReference, asset);
```
lyndon-unity, line 149: This is making me think that InputActionReference's displayed in other parts of the UI won't be renamed as they have their own InputActionReference (with asset / action id and **cached name**)

(But perhaps when those are displayed the name is refreshed there too?)

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            if (property?.objectReferenceValue is InputActionReference reference)
+            {
+                var action = reference?.asset?.FindAction(reference.action.id);
+                if (action is null)
+                {
+                    property.objectReferenceValue = null;
+                    property.serializedObject.ApplyModifiedProperties();
+                }
+            }
```
lyndon-unity, line 46: Perhaps here we also need to refresh the InputActionReference reference name in case its changed in the action.

jfreire-unity: No need.
When renaming an input action, we need to manually recreate the action reference in the asset ([code here](https://github.com/Unity-Technologies/InputSystem/blob/f23f58720ff7dd5e7678ffb3469e999a61325f5a/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs#L145-L150)). Otherwise, the action reference serialized object in the .asset file has the name of an action that no longer exists.
At the same time, the change in name is reflected in the inspector UI of this property drawer.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: Thanks. We probably can. I'll leave it for the end in case the direction in this PR is accepted.
Are you suggesting to create some kind of dictionary to associate the action ID hash of the reference with the reference itself ? So that we can then use the `action.id` here as a key ?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
jfreire-unity: You're right, in theory it shouldn't. 
But when removing an input action reference from the asset, it seems to show InputManage in the picker window. This was a just a quick fix. Something to improve, depending on the time we have to understand what's happening in InputManager.asset.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: > (But perhaps when those are displayed the name is refreshed there too?)

From my testing, this is what currently happens.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionAssetIconLoader.cs
```diff
@@ -0,0 +1,35 @@
+#if UNITY_EDITOR
+using UnityEditor;
+
+namespace UnityEngine.InputSystem.Editor
+{
+    /// <summary>
+    /// Provides access to icons associated with <code>InputActionAsset</code>.
```
jamesmcgill: Could be `<see cref="InputActionAsset"/>`
There are several other places in the PR where references could be used, I won't list them all.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -97,10 +105,51 @@ private static InputActionAsset CreateNewActionAsset()
...
+        internal static void UpdateInputActionReferences()
+        {
+            var asset = GetOrCreate();
+            var existingReferences = InputActionImporter.LoadInputActionReferencesFromAsset(asset).ToList();
 
-            return asset;
+            // Check if referenced input action exists in the asset and remove the reference if it doesn't.
+            foreach (var actionReference in existingReferences)
+            {
+                var action = asset.FindAction(actionReference.action.id);
```
ekcoh, line 123: When testing the branch I have a NullReferenceException here in PlayMode after some edits and entering PlayMode (Not sure about exact steps to reproduce, will update his comment if I can reproduce again):

```
NullReferenceException: Object reference not set to an instance of an object
UnityEngine.InputSystem.Editor.ProjectWideActionsAsset.UpdateInputActionReferences () (at ./Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs:123)
UnityEngine.InputSystem.Editor.InputActionsEditorWindowUtils.SaveAsset (UnityEditor.SerializedObject serializedAsset) (at ./Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs:21)
UnityEngine.InputSystem.Editor.InputActionsEditorSettingsProvider.OnEditFocusLost (UnityEngine.UIElements.FocusOutEvent event) (at ./Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs:79)
UnityEngine.InputSystem.Editor.InputActionsEditorSettingsProvider.OnDeactivate () (at ./Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs:54)
UnityEditor.SettingsWindow.OnDisable () (at /Users/bokken/build/output/unity/unity/Editor/Mono/Settings/SettingsWindow.cs:130)
UnityEditor.ContainerWindow:InternalCloseWindow() (at /Users/bokken/build/output/unity/unity/Editor/Mono/ContainerWindow.cs:360)
```

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh, line 134: Is it guaranteed that ADB will never contain more than a single reference referring to an action? (Since using FirstOrDefault here).

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            InputActionReferenceSearchProviders.CreateInputActionReferenceSearchProviderForAssets(),
+            InputActionReferenceSearchProviders.CreateInputActionReferenceSearchProviderForProjectWideActions(),
+        }, string.Empty, SearchConstants.PickerSearchFlags);
+
+
+        public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+        {
+            // Sets the property to null if the action is not found in the asset.
+            ValidatePropertyWithDanglingInputActionReferences(property);
```
ekcoh, line 26: Why do we need to do this? Note that this also introduces a different behavior than for normal .inputaction assets which resolve to "Missing (Input Action Reference)" when the referenced ScriptableObject is Destroyed and marked available for GC. I assume this will resolve to null if its a dangling instance from broken Undo and maybe its better than having a "dangling reference" but wouldn't this also prevent recovery via Undo?

ekcoh: It seems like this will trigger a full iteration of all action maps and all actions on every redraw right?

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferenceSearchProviders.cs
```diff
@@ -0,0 +1,98 @@
...
+            | Search.SearchViewFlags.DisableSavedSearchQuery;
+    }
+
+    internal static class InputActionReferenceSearchProviders
+    {
+        const string k_AssetFolderSearchProviderId = "AssetsInputActionReferenceSearchProvider";
+        const string k_ProjectWideActionsSearchProviderId = "ProjectWideInputActionReferenceSearchProvider";
+
+        // Search provider for InputActionReferences for all assets in the project, without project-wide actions.
+        internal static SearchProvider CreateInputActionReferenceSearchProviderForAssets()
```
ekcoh, line 30: I'm a little bit torn about overriding the default ADB implementation for this one since it would provide another experience/result than Edit > SearchAll with filter "t:InputActionReference", I guess its fine for pre2 scope but I think we need to ask ourselves if we really want to override actual ADB search for regular assets which are already covered by ADB search. This is the reason I didn't do it on the original branch. Also note that the original branch referenced here didn't use the package search flag if that was a driving reason for doing a custom searcher for assets?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
```diff
@@ -35,7 +35,7 @@ public override void OnActivate(string searchContext, VisualElement rootElement)
             // Note that focused element will be set if we are navigating back to
             // an existing instance when switching setting in the left project settings panel since
             // this doesn't recreate the editor.
-            if (m_RootVisualElement.focusController.focusedElement != null)
+            if (m_RootVisualElement?.focusController?.focusedElement != null)
```
ekcoh, line 38: I noticed you picked up this diff from the original branch, just wanted to clarify this is a general bug fix relating to a post around potential focus controller bug and isn't really related to the subject of the PR, I think it is fine to add anyway but might require its own FIX line in the CHANGELOG / commit message.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
```diff
@@ -14,10 +15,13 @@ internal class InputActionsEditorWindowUtils
...
         {
             var asset = (InputActionAsset)serializedAsset.targetObject;
-            // for the global actions asset: save differently (as it is a yaml file and not a json)
+            // For project-wide actions asset save works differently. The asset is in YAML format, not JSON.
             if (asset.name == ProjectWideActionsAsset.kAssetName)
             {
+                ProjectWideActionsAsset.UpdateInputActionReferences();
                 AssetDatabase.SaveAssets();
+
+                var existingReferences = InputActionImporter.LoadInputActionReferencesFromAsset(ProjectWideActionsAsset.GetOrCreate()).ToList();
```
ekcoh: I cannot see this `existingReferences` and Load is used for anything? Remove?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -212,12 +210,51 @@ public override void OnImportAsset(AssetImportContext ctx)
...
+            // Then the whole asset is loaded and all the InputActionReferences are extracted from it.
+            // Also, the action references are duplicated to have backwards compatibility with the 1.0.0-preview.7. That
+            // is why we look for references withouth the `HideFlags.HideInHierarchy` flag.
+            var inputActionReferencesList = new List<InputActionReference>();
+            foreach (var guid in inputActionReferenceGUIDs)
+            {
+                var assetName = AssetDatabase.GUIDToAssetPath(guid);
+                var assetInputActionReferenceList = AssetDatabase.LoadAllAssetsAtPath(assetName).Where(
+                    o => o is InputActionReference &&
+                    !((InputActionReference)o).hideFlags.HasFlag(HideFlags.HideInHierarchy))
```
ekcoh, line 244: I think its good you have added the flag evaluation here even though I believe its never set (it could be)?

ekcoh (changes requested):
I believe this could be a valuable change to enable testing of pre2 but I do not think its a solution for InputActionReference (nor is the branch it is based on). I still believe that a way forward could/would be either integrating InputActionReference into UITK editor workflow or use OnValidate() on related assets to fix InputActionReferences. Likely a combination is needed and have a serialised array in the InputActionAsset instead of Load to track them in order to allow that array to follow the same Undo operations and the ScriptableObjects themselves.

I added some questions/comments below relating to the implementation. My main comments for feedback with Request changes are:
- This PR (as the referenced PoC PR) needs unit tests for the added functionality to at least have some test coverage as part of CI.

ekcoh:
Did some additional testing of this branch but failed to trigger the NullReference exception again. In general it seems that this is more reliable than behavior on the PoC branch being referenced and hence this is likely a better choice for pre2.
I noticed however in one case I managed to get into a situation where I could no longer assign any Project-wide action to via the picker and the search result from the picker and the UITK editor window was out of sync. Unfortunately I haven't been able to reproduce but may open an issue if I manage to do this again.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferenceSearchProviders.cs
jfreire-unity: As explained in Slack, this was implemented since using the solution from the PoC branch (which uses the "ADB search" you mentioned) was not working in 2022.3.0f. In the latest 2022 LTS (at least from 2022.3.9f on), the PoC branch solution works but it doesn't show the InputActionReference icon for those references. So far I have not found an immediate solution to change the asset icon while using ADB search.

However, for a final release, I'd also prefer to use the ADB search while customizing the asset icon. It would be less code for us to manage.
I suggest we do this after pre.2 release. Unless there's an immediate one-liner you know that allows us to use ADB search and change the assets icon as well. 
What do you think?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
jfreire-unity: Good catch, it was a debug leftover 👍 I will remove it

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
jfreire-unity: From my tests, it was needed. A lot of the Asset folder (and maybe even Packages) assets seem to show duplicated InputActionReferences if this check is not here.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: Hm, interesting. Good catch! Was this when doing a bunch of Undo/Redo actions? I'll try to replicate it as well.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: Yes. Since this is controlled by us in the InputManager.asset.
If there's an action, only one reference is created. We never create more than 1 reference for an action, at the moment.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
jfreire-unity: If this is not here and an Input Action is removed from the asset, it continues to reference a removed input action.
But this code has the implications you just mentioned. Ideally, we would like to remove this so that this property is updated once the action is removed from the asset. 

This is only needed for references of the project-wide actions asset. Normal assets work just fine.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -212,12 +210,51 @@ public override void OnImportAsset(AssetImportContext ctx)
...
+        {
+            //Get all InputActionReferences are stored at the same asset path as InputActionAsset
+            return AssetDatabase.LoadAllAssetsAtPath(AssetDatabase.GetAssetPath(asset)).Where(
+                o => o is InputActionReference && o.name != "InputManager").Cast<InputActionReference>();
+        }
+
+        // Get all InputActionReferences from assets in the project. By default it only gets the assets in the "Assets" folder.
+        internal static IEnumerable<InputActionReference> LoadInputActionReferencesFromAssetDatabase(string[] folderPath = null)
+        {
+            string[] searchInFolderPath = null;
```
jamesmcgill: As these variables are all arrays of paths, their naming should be plural.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -212,12 +210,51 @@ public override void OnImportAsset(AssetImportContext ctx)
...
+
+            // To find all the InputActionReferences, the GUID of the asset containing at least one action reference is
+            // used to find the asset path. This is because InputActionReferences are stored in the asset database as sub-assets of InputActionAsset.
+            // Then the whole asset is loaded and all the InputActionReferences are extracted from it.
+            // Also, the action references are duplicated to have backwards compatibility with the 1.0.0-preview.7. That
+            // is why we look for references withouth the `HideFlags.HideInHierarchy` flag.
+            var inputActionReferencesList = new List<InputActionReference>();
+            foreach (var guid in inputActionReferenceGUIDs)
+            {
+                var assetName = AssetDatabase.GUIDToAssetPath(guid);
```
jamesmcgill: `assetPath`

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
jfreire-unity: Ok, I just checked this in the latest 2022.3.11f, and this code seems to not be needed anymore. I can remove this code as it will have the same behavior. This is likely a leftover from doing most of the work with 2022.3.0f.....

jfreire-unity: Removed

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
jamesmcgill: Shouldn't it be `asset` instead of `ProjectWideActionsAsset.GetOrCreate()`?
Or did you really need to create another asset instance for some reason here?

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
jfreire-unity: We don't create a new one. It's just to get the project wide actions asset from `InputManager.asset` file.

jfreire-unity:
> * This PR (as the referenced PoC PR) needs unit tests for the added functionality to at least have some test coverage as part of CI.

Added at least one test now

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Not sure, my main test case have been creating, deleting, undo, redo, renaming, select again in picker, repeat.... Haven't found this particular one again though.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferenceSearchProviders.cs
ekcoh: Regarding the "incorrect InputAction icon" I think this is the expected behavior for the Search since this is what you get from Advanced Search via main menu. We should follow-up separately what to expect for package assets and sub-asset icons registered via importers. Might be a limitation of the Search functionality which is why I propose it could make sense to not add some custom add on to fix the icons. We should follow-up on this. The reason it fails (after disassembly of the Search code) is that AssetDatabase.CahedIcon doesn't return the icon for the asset if I recall correctly.

I tried to provide a custom IconFetcher on top of ADB before and it didn't work so I think we should reach out to the Search team. Your proposal makes sense.

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -210,6 +212,49 @@ public void ProjectWideActions_CanReplaceExistingActions()
         Assert.That(InputSystem.actions.actionMaps[1].actions.Count, Is.EqualTo(1));
         Assert.That(InputSystem.actions.actionMaps[1].actions[0].name, Is.EqualTo("replacedAction4"));
     }
+
+    [Test]
+    [Category(TestCategory)]
```
ekcoh: Great to get some testing on this in, we can likely extend this with Undo/Redo functionality later, e.g. https://docs.unity3d.com/ScriptReference/Undo.PerformUndo.html and https://docs.unity3d.com/ScriptReference/Undo.PerformRedo.html

ekcoh (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferenceSearchProviders.cs
jfreire-unity: Tracked possible improvements for this here https://jira.unity3d.com/browse/ISX-1681

jfreire-unity:
Known issues to follow up:
- Undo issues to be fixed: https://jira.unity3d.com/browse/ISX-1682
- Check if it's possible to use Asset Database search and customize the items icons: https://jira.unity3d.com/browse/ISX-1681

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
jfreire-unity: Ok, I was testing again, and without this code a removed action was still being referenced in the property. And we want to show None/Missing. I reintroduced this but now only checking for Project Wide Actions. I captured a task to remove this: https://jira.unity3d.com/browse/ISX-1683

jfreire-unity:
Adding @Pauliusd01 for some testing. Heads up that Undo/Redo doesn't work properly and was already captured as a [bug](https://jira.unity3d.com/browse/ISX-1682)

Pauliusd01:
https://github.com/Unity-Technologies/InputSystem/assets/54306142/6ed00a5f-b0ca-4091-9f8b-ed23705ae7e3

Are we able to keep these tabs always visible no matter the size of the window? Seems very odd in its current state

Also, a question about the tabs - why is NONE option only available in ALL and not in every tab? Is it because the tabs are not really tabs but just pre-made searches and it doesn't find "none" in the filter? :D

jfreire-unity:
> Unity_2023-10-31_15-40-21.mp4 
> Are we able to keep these tabs always visible no matter the size of the window? Seems very odd in its current state

I'll have a look if it's easy, otherwise, I'll poke the Search team about this. I wouldn't consider it a blocker but definitely something to improve.

Pauliusd01:
Is there a reason why if an action is deleted in Project Wide Actions the reference defaults to NONE instead of Missing (like for custom actions assets)? Could give the user a different impression - unfinished work vs broken work

Pauliusd01:
https://github.com/Unity-Technologies/InputSystem/assets/54306142/a3dbccee-b758-4584-97c1-5505993faefd

Are these area tags the same as the existing tabs? If so, why are they returning nothing?

Pauliusd01 (approved):
LGTM. Had non blocking minor things listed in comments (I can report those separately if needed, just let me know) Other than those and the undo redo issues did not notice anything broken, references seem to handle deleting,duplicating, basic renaming just fine for both ProjectWideActions and custom ones

jfreire-unity:
> LGTM. Had non blocking minor things listed in comments (I can report those separately if needed, just let me know) Other than those and the undo redo issues did not notice anything broken, reference seem to handle deleting,duplicating, basic renaming just fine for both ProjectWideActions and custom ones

Thanks a lot @Pauliusd01 🥇 

I'll have a look at the issues you reported and try to address them today. The ones that are not easily fixed, I'll create a ticket.

jfreire-unity:
> Unity_2023-10-31_16-02-44.mp4 
> Are these area tags the same as the existing tabs? If so, why are they returning nothing?

I removed this from the search in my last commit.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferenceSearchProviders.cs
```diff
@@ -0,0 +1,98 @@
...
+        // Note that this is overloaded to allow utilizing FetchLabel inside fetchItems to keep label formatting
+        // consistent between CreateItem and additional fetchLabel calls.
+        private static string FetchLabel(Object obj)
+        {
+            return obj.name;
+        }
+
+        private static string FetchLabel(SearchItem item, SearchContext context)
+        {
+            return FetchLabel((item.data as Object) !);
```
chris-massie, line 94: This line doesn't compile.

jfreire-unity: Could you clarify in what circumstances (editor version, OS, etc) ? I haven't seen any compile errors on my side and others, or CI.

chris-massie: Sorry this was my mistake, I was trying to enable global actions on an earlier Unity Editor version that isn't supported.

PR #1782, opened 2023-10-26
XRI package test

stefanunity:
Add editor test to add XRI package and check for errors after it has been added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,81 @@
...
+            else if (XRRemoveRequest.Status >= StatusCode.Failure)
+                Debug.Log(XRRemoveRequest.Error.message);
+
+            EditorApplication.update -= RemoveProgress;
+        }
+    }
+
+    void HandleLog(string logString, string stackTrace, LogType type) {
+        
+        if (type != LogType.Log) {
```
ekcoh: Assert.That(type, Is.EqualTo(LogType.Log))

ekcoh: (mintor)

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,81 @@
...
+        EditorApplication.update += RemoveProgress;
+        while (!XRRemoveRequest.IsCompleted)
+        {
+            yield return null;
+        }
+    }    	
+
+    [UnityTest]    
+    [Category("Integration")]
+    public IEnumerator AddingXRIPackageThrowsNoErrors()
```
ekcoh: AdddingLatestXRIPackageThrowsNoErrors()

ekcoh: (minor)

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,81 @@
...
+        while (!XRAddRequest.IsCompleted)
+        {
+            yield return null;
+        }       
+
+        AssetDatabase.Refresh();
+
+        yield return new WaitForDomainReload();
+
+        Assert.That(true);
```
ekcoh: This line can be removed

ekcoh: Evaluates to true by default

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,81 @@
...
+#pragma warning disable CA1707
+
+public class XRIPackageTests
+{
+	static AddRequest XRAddRequest;
+    static RemoveRequest XRRemoveRequest;
+	
+    [UnityTearDown]
+    public IEnumerator TearDown()
+    {
```
ekcoh, line 24: Maybe make a comment that this is associated to the test fetching the XRI package and might need to be reworked as a TearDown function if we add additional tests that doesn't operate on the package itself

ekcoh (approved):
Approving now, just correct the minors, at least the Asserts

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
stefanunity: done

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
stefanunity: done

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
stefanunity: done

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,82 @@
...
+#pragma warning disable CA1707
+
+public class XRIPackageTests
+{
+    static AddRequest XRAddRequest;
+    static RemoveRequest XRRemoveRequest;
+
+    /// <summary>
+    /// TearDown removes the added XRI package again.
+    /// If you are adding a 2nd test that needs the XRI package adjust this.
```
MindWrapper, line 19: As far as a remember unity. [UnityTearDown]/[UnityTearDown] are invoked before/after each test. So when  AdddingLatestXRIPackageThrowsNoError ends `com.unity.xr.interaction.toolkit` will be uninstalled. So comment should be adjusted.

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,82 @@
...
+        EditorApplication.update += AddProgress;
+
+        while (!XRAddRequest.IsCompleted)
+        {
+            yield return null;
+        }
+
+        AssetDatabase.Refresh();
+
+        yield return new WaitForDomainReload();
```
MindWrapper, line 49: Tests should assert something. So perhaps you should check that the log doesn't contain any errors

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    static void AddProgress()
+    {
+        if (XRAddRequest.IsCompleted)
```
MindWrapper, line 54: minor stylish stuff:  if (XRAddRequest.IsCompleted) can be inverted and early return can be done. This will reduce level of nesting.

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            if (XRAddRequest.Status == StatusCode.Success)
+                Debug.Log("Installed: " + XRAddRequest.Result.packageId);
+            else if (XRAddRequest.Status >= StatusCode.Failure)
```
MindWrapper, line 58: may be else else (wihtout if)? what are other options in Status code?

Assets/Tests/InputSystem.Editor/XRIPackageTest.cs
MindWrapper: Oh, I see assertion is hidden inside HaldleLog, in such case I would suggest writing all log messages into List and assert on them at the end of the test.

PR #1783, opened 2023-10-30
FIX: InputSystemProvider now uses DestroyImmediate() in the Editor instead of Destroy() to prevent Edit mode errors

Pauliusd01:
### Description

_Please fill this section with a description what the pull request is trying to address._

### Changes made

### Notes

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
@@ -640,9 +640,7 @@ namespace UnityEngine.InputSystem.Plugins.InputForUI
             m_InputActionAsset.Disable();
 
             // The Next/Previous action is not part of the input actions asset
-            UnregisterNextPreviousAction();
-
-            UnityEngine.Object.Destroy(m_InputActionAsset); // TODO check if this is ok
+            UnregisterNextPreviousAction();            UnityEngine.Object.DestroyImmediate(m_InputActionAsset); // TODO check if this is ok
         }
 
         public struct Configuration


Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -640,9 +640,7 @@ void UnregisterActions(Configuration cfg)
             m_InputActionAsset.Disable();
 
             // The Next/Previous action is not part of the input actions asset
-            UnregisterNextPreviousAction();
-
-            UnityEngine.Object.Destroy(m_InputActionAsset); // TODO check if this is ok
+            UnregisterNextPreviousAction();            UnityEngine.Object.DestroyImmediate(m_InputActionAsset); // TODO check if this is ok
```
ekcoh, line 643: Seems like formatting in this file is off, did you run formatter, /format.ps1 or /format.sh?

I noticed there is also another DestroyImmediate in this file. Why do you need DestroyImmediate here? Notice that its generally recommended against (See API docs) and should only be used in edit mode, so if we need it use

#if UNITY_EDITOR
... destroy immediate ....
#else
... destroy
#endif

However, I wonder why we can't use Destroy in both places in this file?

Pauliusd01: Huh that's weird, It didn't look like that in Visual Studio to me when I was editing it 😅 I will fix this
>I noticed there is also another DestroyImmediate in this file. Why do you need DestroyImmediate here? Notice that its generally recommended against (See API docs) and should only be used in edit mode, so if we need it use

I'm following instructions from Adrien for this bug: https://jira.unity3d.com/browse/ISXB-677 Sorry, I forgot to add the work in progress tag. I'm currently setting up the mentioned SRP tests to see if my changes actually fix the error in question and whether it causes any issues

ekcoh: Would you be able to consult Adrien regarding if #ifdef is needed or not? Basically based on remarks here: https://docs.unity3d.com/ScriptReference/Object.DestroyImmediate.html

ITR13:
This is an incorrect fix, the reason the error occurs is because Unity deallocates objects _after_ setting the current mode to "Edit Mode". This causes this part of the code to run too late and incorrectly flag it as an editor deletion. Though switching it to DestroyImmediate would hide the issue, it will also be incorrect in any scenario where [PanelSettings](https://docs.unity3d.com/Manual/UIE-Runtime-Panel-Settings.html) is unloaded while playing.

This is somewhat of an issue in Unity itself, and doesn't have any good fix. You could try to check Application.isPlaying, but that might fail if the code isn't running on the main thread, similar to what happens with destructors. It's also not well-documented if the ScriptableObject is deallocated when exiting Play mode or not, if it is you have to not Destroy it, if it isn't you have to call DestroyImmediate instead.

An alternative solution would be having Unity make sure deallocation happens before switching to "Edit Mode", which would solve a lot of similar issues.

Pauliusd01:
> This is an incorrect fix, the reason the error occurs is because Unity deallocates objects _after_ setting the current mode to "Edit Mode". This causes this part of the code to run too late and incorrectly flag it as an editor deletion. Though switching it to DestroyImmediate would hide the issue, it will also be incorrect in any scenario where [PanelSettings](https://docs.unity3d.com/Manual/UIE-Runtime-Panel-Settings.html) is unloaded while playing.
> 
> This is somewhat of an issue in Unity itself, and doesn't have any good fix. You could try to check Application.isPlaying, but that might fail if the code isn't running on the main thread, similar to what happens with destructors. It's also not well-documented if the ScriptableObject is deallocated when exiting Play mode or not, if it is you have to not Destroy it, if it isn't you have to call DestroyImmediate instead.
> 
> An alternative solution would be having Unity make sure deallocation happens before switching to "Edit Mode", which would solve a lot of similar issues.

Thank you Mikael, I was hoping to save developer time thinking this was just a quick fix but it turns out it isn't. I will close this PR and leave this for someone with actual experience 😓

PR #1785, opened 2023-10-31
DOCS: Added documentation for 1.8.0 pre-release 2

duckets:
### Description

Documentation for 1.8.0 pre-release 2.

### Changes made

Changes to documentation files only.


Packages/com.unity.inputsystem/Documentation~/ActionAssets.md
```diff
@@ -27,106 +27,8 @@ By default, Unity doesn't save edits you make in the Action Asset window when yo
...
-3. The right pane contains the properties of the currently selected Action or Binding.
 
-Use the following keyboard shortcuts to quickly trigger common operations:
-
-|Shortcut (Mac)|Shortcut (Windows)|Description|
-|--------------|------------------|-----------|
-|⌘X, ⌘C, ⌘V|Ctrl-X, Ctrl-C, Ctrl-V|Cut, Copy, and Paste. Can be used on Actions, Action Maps, and Bindings.|
-|⌘D|Ctrl-D|Duplicate. Can be used on Actions, Action Maps, and Bindings.|
-|⌘⌫|Del|Delete. Can be used on Actions, Action Maps, and Bindings.|
-|⌥S|Alt-S|Save.|
```
ekcoh, line 43: Out of curiosity are these the same shortcuts we've had in the older editor? What I am reacting to is that we have e.g. Alt+S instead of Ctrl+S which I would expect on e.g. Windows. If just following previous customs I think this is fine (and not really related to doc but rather functionality), but maybe we are blocked on e.g. Ctrl+S due to it being mapped to global Unity commands?

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -3,98 +3,73 @@ uid: input-system-actions
...
-
-Without Actions, the meaning of your input and the device controls end up hard-coded together in your scripts, which although quick to implement, is not very flexible and therefore not always desirable. For example, here the moveVector variable is hard-coded to read values from the right stick of a gamepad:
+**Actions** are an important concept in the Input System. They allow you to separate the purpose of an input from the device controls which perform that input. Actions allow you to associate the purpose and device controls together in a flexible way.
 
-```CSharp
-    // without Actions, you end up hard-coding device controls
-    // like this: (not always desirable)
-    moveVector = gamepad.rightStick.ReadValue<Vector2>();
-```
+For example, the purpose of an input in a game might be to make the player's character move around. The device control associated with that action might be the motion of the right gamepad stick.
```
ekcoh: Shouldn't we exemplify this with **left** gamepad stick instead since that is more common?

ekcoh: I noticed we used right stick in previous example but maybe left is better? (minor)

stefanunity (approved):

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -3,98 +3,73 @@ uid: input-system-actions
...
 
-When you use Actions, you do not need to refer to specific devices or their controls in your code. Instead the Action's binding defines which controls on which device is used to perform the action, and your code becomes simpler. In this example, `moveAction` would contain a reference to an Action, defined either in code, or in an Action Asset:
+The association between an Action and the device controls which perform that input is a **binding**, and you can set up bindings in the [Input Actions editor](ActionsEditor). When you use Actions in your code, you do not need to refer to specific devices because the binding defines which device's controls are used to perform the action.
 
 ```CSharp
-    // using Actions removes the need to refer directly to device controls
-    moveAmount = moveAction.ReadValue<Vector2>();
+    // this line of code reads the input from an action called "move"
+    // without referring to a specific device's controls
+    var moveVector = InputActions.player.move.value;
```
jamesmcgill: I'm afraid `.value` was part of the typesafe api, which is no longer available.
The user will need to continue using `.ReadValue<Vector2>()`

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -3,98 +3,73 @@ uid: input-system-actions
...
+When scripting with Actions in the Input System, there are number of key classes you can use, which are described here:
 
 |Class|Description|
 |-----|-----------|
-|`InputActionAsset`|An Asset that contains one or more Action Maps and, optionally, a sequence of Control Schemes. For more information on how to create, edit, and work with these Assets, see [Action Assets](ActionAssets.md).|
-|`InputActionMap`|A named collection of Actions.|
-|`InputAction`|A named Action that triggers callbacks in response to input.|
-
-Actions use `InputBinding` to refer to the inputs they collect. For more information about Bindings and how to use them, see [Action Bindings](ActionBindings.md).
+|[`InputActions`](../api/UnityEngine.InputSystem.InputActions.html)|A static class whose API dynamically updates to reflect the Action Maps, Actions, and properties that are currently configured in the [Input Actions editor](ActionsEditor). In many cases this is the only class you need to read input for typical games or apps.|
```
jamesmcgill: `InputActions` was also part of the typesafe API.

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -3,98 +3,73 @@ uid: input-system-actions
...
 
-Each Action Map has a name ([`InputActionMap.name`](../api/UnityEngine.InputSystem.InputActionMap.html#UnityEngine_InputSystem_InputActionMap_name)), which must be unique within the Action Asset that the Action Map belongs to, if any (see [`InputActionMap.asset`](../api/UnityEngine.InputSystem.InputActionMap.html#UnityEngine_InputSystem_InputActionMap_asset)). Each Action Map also has a unique ID ([`InputActionMap.id`](../api/UnityEngine.InputSystem.InputActionMap.html#UnityEngine_InputSystem_InputActionMap_id)), which you can use to reference the Action Map. The ID remains the same even if you rename the Action Map.
+Each Action Map has a name ([`InputActionMap.name`](../api/UnityEngine.InputSystem.InputActionMap.html#UnityEngine_InputSystem_InputActionMap_name)), which must also be unique with respect to the othe Action Maps present, if any. Each Action Map also has a unique ID ([`InputActionMap.id`](../api/UnityEngine.InputSystem.InputActionMap.html#UnityEngine_InputSystem_InputActionMap_id)), which you can use to reference the Action Map. The ID remains the same even if you rename the Action Map.
```
jamesmcgill: small typo `othe`

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -3,98 +3,73 @@ uid: input-system-actions
...
 
 ## Creating Actions
 
 You can create Actions in any of the following ways:
 
-- Use the dedicated editor for [Input Action Assets](ActionAssets.md).
-- [Embed them](Workflow-Embedded.md) in MonoBehaviour components, then set up bindings in the Inspector.
-- Manually load them from JSON.
-- Create them entirely in code, including setting up the bindings.
+- You can use the [Input Actions editor](ActionsEditor) in the Project Settings window (this is the primary recommended workflow). Actions and ActionMaps created here are automatically added to the InputActions class API.
```
jamesmcgill: `InputActions` class removed

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -212,9 +185,17 @@ void Start()
 }
 ```
 
-## Using Actions
+Actions defined in the [Input Actions editor](ActionsEditor) are automatically added to the InputActions class as API, and are ready to use immediately.
```
jamesmcgill: `InputActions` class removed

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -212,9 +185,17 @@ void Start()
...
 
-For an Action to do something, you must first enable it. You can do this either by individually enabling Actions, or by enabling them in bulk through Action Maps. The second method is more efficient in all scenarios.
+For example, for an Action called "Move" defined in an Action Map called "Player", you can read its value from your code like this:
+
+```
+    var moveVector = InputActions.player.move.value;
```
jamesmcgill: `.value` -> `.ReadValue<Vector2>()`

Packages/com.unity.inputsystem/Documentation~/ActionsEditor.md
```diff
@@ -0,0 +1,134 @@
...
+
+
+
+**Action Properties** define the type of input that relates to the action, such as whether it is a button press, or a movement axis.
+
+[**Control Schemes**](ActionsEditor.html#editing-control-schemes) refer to particular modes of input, such as "Joypad" or "Keyboard and Mouse". You can specify which bindings belong to particular control schemes. You might have one control scheme which is "Joypad", and another control scheme which is "Keyboard and Mouse". This allows you to determine which control scheme the user is currently using, so your game can respond to the user accordingly. This feature is often used to adapt the in-game UI to show the correct keys or buttons in on-screen prompts.
+
+
+### Configure Action Maps
+
```
jamesmcgill: Is the 3 lines of whitespace intentional here?

Packages/com.unity.inputsystem/Documentation~/ActionsEditor.md
```diff
@@ -0,0 +1,134 @@
...
+Input Action Assets can have multiple [Control Schemes](ActionBindings.md#control-schemes), which let you enable or disable different sets of Bindings for your Actions for different types of Devices.
+
+![Control Scheme Properties](Images/ControlSchemeProperties.png)
+
+To see the Control Schemes in the Input Action Asset editor window, open the Control Scheme drop-down list in the top left of the window. This menu lets you add or remove Control Schemes to your Asset. If the Asset contains any Control Schemes, you can select a Control Scheme, and then the window only shows bindings that belong to that Scheme. If you select a binding, you can now pick the Control Schemes for which this binding should be active in the __Properties__ view to the left of the window. When you add a new Control Scheme, or select an existing Control Scheme, and then select __Edit Control Scheme…__, you can edit the name of the Control Scheme and which devices the Scheme should be active for.
+
+## Using Input Action Assets
+
+### The Automaticaly-generated InputActions API
```
jamesmcgill: Auto generated `InputActions` API removed

Packages/com.unity.inputsystem/Documentation~/Workflow-Actions.md
```diff
@@ -0,0 +1,88 @@
...
+*The Input Actions editor in the Project Settings window*
+
+The [Input Actions editor](ActionsEditor.html) provides a way to define and manage Actions which relate to what you want users to be able to do in your game or app. It also allows you to create bindings between your Actions and one or more types of control on various devices, such as a stick on a gamepad, a key on a keyboard, or a button on an XR controller.
+
+Actions and Bindings allows you to separate the conceptual actions that you want in your game or app from the the specific decvices and controls that a player has to use to perform the actions, which can make your code simpler and more flexible. This is in contrast to [directly reading device states](Workflow-Direct.html) in your code, which can suit simple projects, but can become unwieldy and inflexible for more complex projects.
+
+### The default Actions
+
+The Input System comes pre-configured with some default Actions such as "Move", "Jump", and more, which suit many common app and game scenarios. They are configured to read input most types of input controller such asKeyboard, Mouse, Gamepad, Touchscreen and XR.
```
jamesmcgill: small typo `asKeyboard`

Packages/com.unity.inputsystem/Documentation~/Workflow-Actions.md
```diff
@@ -0,0 +1,88 @@
...
+Open the Input Actions editor by going to **Edit** > **Project Settings** > **Input System Package** > **Input Actions**
+
+![image alt text](./Images/ProjectSettingsInputActionsSimpleShot.png)
+*The Input Actions editor in the Project Settings window*
+
+The [Input Actions editor](ActionsEditor.html) provides a way to define and manage Actions which relate to what you want users to be able to do in your game or app. It also allows you to create bindings between your Actions and one or more types of control on various devices, such as a stick on a gamepad, a key on a keyboard, or a button on an XR controller.
+
+Actions and Bindings allows you to separate the conceptual actions that you want in your game or app from the the specific decvices and controls that a player has to use to perform the actions, which can make your code simpler and more flexible. This is in contrast to [directly reading device states](Workflow-Direct.html) in your code, which can suit simple projects, but can become unwieldy and inflexible for more complex projects.
+
+### The default Actions
```
jamesmcgill, line 23: Not sure if it needs mentioned or not, but there is already a `DefaultInputActions.cs` and `DefaultInputActions.inputactions` shipped with InputSystem.

It's a little confusing (even to me) that we decided to ship yet another default actions with project wide actions.
So maybe it could be confusing to users? And maybe this should be "default project wide actions" to make it clear?

Packages/com.unity.inputsystem/Documentation~/Workflows.md
```diff
@@ -4,18 +4,21 @@ uid: input-system-workflows
...
+There are multiple ways to use the Input System, however the primary and recommended workflow is to use the **Input Actions** panel in the **Project Settings window** to configure your project-wide Actions and Bindings, get references to those Actions in your code's Start method, then read the values for those actions in your Update method.
 
-To understand the different workflows so that you can choose between them, it’s important to first understand the [concepts and terms](Concepts.html) used to describe them.
+There are other workflows which can suit more unusual situations, for example you can use the **PlayerInput component** together with Actions and Bindings which adds a further layer of abstraction, allowing you to connect actions to your event handlers without requiring any intermediate code, and easily handle multiplayer scenarios.
+
+You can choose to configure Actions and Bindings in the Editor UI, or you can set up everything through scripting. Or you can take a more direct aproach by omitting the Actions and Bindings features altogether and instead use script to directly read the state of devices.
```
jamesmcgill: small typo `aproach`

jamesmcgill (approved):
LGTM.
Some left overs related to typesafe api need removed.

Failing CI jobs need to be fixed too:
- [Yamato] Check formatting  -> run the code formatter tool
- API_DocumentationManualDoesNotHaveMissingOrUnusedImages
- API_DocumentationManualDoesNotHaveMissingInternalLinks


Also, not sure if it's mentioned elsewhere, but Project-Wide Actions is only available in Unity 2022.3 and above (due to UITK usage).

ritamerkl (approved):

Packages/com.unity.inputsystem/Documentation~/ActionAssets.md
```diff
@@ -27,106 +27,8 @@ By default, Unity doesn't save edits you make in the Action Asset window when yo
...
-
-![Add 2D Vector Composite](./Images/Add2DVectorComposite.png)
-
-This creates multiple Binding entries for the Action: one for the Composite as a whole, and then, one level below that, one for each Composite part. The Composite itself doesn't have a Binding path property, but its individual parts do, and you can edit these parts like any other Binding. Once you bind all the Composite's parts, the Composite can work together as if you bound a single control to the Action.
-
-To change the type of a Composite retroactively, select the Composite, then select the new type from the **Composite Type** drop-down in the **Properties** pane.
-
-![Composite Type](./Images/CompositeType.png)
-
-To can change the part of the Composite to which a particular Binding is assigned, use the **Composite Part** drop-down in the Binding's properties.
```
ekcoh, line 115: Typo "To can change"

Packages/com.unity.inputsystem/Documentation~/ActionAssets.md
```diff
@@ -27,106 +27,8 @@ By default, Unity doesn't save edits you make in the Action Asset window when yo
...
-
-![Duplicated Part Bindings](./Images/DuplicatedPartBindings.png)
-
-### Editing Control Schemes
-
-Input Action Assets can have multiple [Control Schemes](ActionBindings.md#control-schemes), which let you enable or disable different sets of Bindings for your Actions for different types of Devices.
-
-![Control Scheme Properties](Images/ControlSchemeProperties.png)
-
-To see the Control Schemes in the Input Action Asset editor window, open the Control Scheme drop-down list in the top left of the window. This menu lets you add or remove Control Schemes to your Asset. If the Asset contains any Control Schemes, you can select a Control Scheme, and then the window only shows bindings that belong to that Scheme. If you select a binding, you can now pick the Control Schemes for which this binding should be active in the __Properties__ view to the left of the window. When you add a new Control Scheme, or select an existing Control Scheme, and then select __Edit Control Scheme…__, you can edit the name of the Control Scheme and which devices the Scheme should be active for.
```
ekcoh, line 129: Minor but I noticed we say "belong to" but I believe a binding may have multiple Control Schemes so maybe we should use the term "associated with" since 1:N and not 1:1

Packages/com.unity.inputsystem/Documentation~/Actions.md
ekcoh: Agree, this should be updated to reflect a world without type-safe API support

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -3,98 +3,73 @@ uid: input-system-actions
...
 ## Creating Actions
 
 You can create Actions in any of the following ways:
 
-- Use the dedicated editor for [Input Action Assets](ActionAssets.md).
-- [Embed them](Workflow-Embedded.md) in MonoBehaviour components, then set up bindings in the Inspector.
-- Manually load them from JSON.
-- Create them entirely in code, including setting up the bindings.
+- You can use the [Input Actions editor](ActionsEditor) in the Project Settings window (this is the primary recommended workflow). Actions and ActionMaps created here are automatically added to the InputActions class API.
+- You can create [Input Actions assets](ActionAssets) which define a set of action data similar to that defined in Project Settings, but is instead stored in a self-contained asset (this is an outdated workflow but is still supported).
```
ekcoh: Not sure we want to call it out as an outdated workflow at this point since there are use-cases for self-contained asset inputactions that cannot be supported with Project-wide settings at the moment, e.g. exporting self-contained asset packages in a reliable way. I rather see we leave the outdated part out for now, other opinions?

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -212,9 +185,17 @@ void Start()
...
-For an Action to do something, you must first enable it. You can do this either by individually enabling Actions, or by enabling them in bulk through Action Maps. The second method is more efficient in all scenarios.
+For example, for an Action called "Move" defined in an Action Map called "Player", you can read its value from your code like this:
+
+```
+    var moveVector = InputActions.player.move.value;
+```
+
+## Enabling actions
+
+Actions defined in the [Input Actions editor](ActionsEditor) are enabled by default and ready to use. For actions defined elsewhere, such as in code or in Action Assets, they begin in a disabled state, and you must enable them before they will respond to input. You can enable them individually, or as a group by enabling the Action Map which contains them.
```
ekcoh: Not related to the docs per-se but realize that this might be inconvenient for anyone that wants full control of what actions are enabled or not. Maybe we should consider making this a setting on actions created so that initially enabled could default to true for project-wide (or future major revision), but could be explicitly set to initially disabled.

Packages/com.unity.inputsystem/Documentation~/ActionsEditor.md
```diff
@@ -0,0 +1,134 @@
...
+    // read the default "move" action, which is a 2D vector axis
+    Vector2 move = InputActions.player.move.value;
+
+    // read the default "jump" action, which is a true/false button
+    bool jump = InputActions.player.jump.isPressed;
+```
+
+**Note**: The Action Map and Actions in this example configured as defaults when you install the Input System package.
+
+### Using Actions with the `PlayerInput` component
```
ekcoh: I recently had yet another case of confusion with PlayerInput. Maybe add a line here that conveys the message clearly that one instance of PlayerInput in a scene corresponds to a single Player. Each additional instance corresponds to another player in a local multiplayer game. And clarifying in the last sentence that it assigns unique input devices to each player based on the associated bindings and active control scheme.

Packages/com.unity.inputsystem/Documentation~/ActionsEditor.md
```diff
@@ -0,0 +1,134 @@
...
+
+The [Player Input](PlayerInput.md) component provides a convenient way to handle input for one or multiple players. It requires you to set up all your Actions in the Input Actions editor, which you can then select in the Player Input component. The Player Input component can then automatically handle activating Action Maps and selecting Control Schemes for you.
+
+![PlayerInput](Images/PlayerInput.png)
+
+### Modifying Input Action Assets at runtime
```
ekcoh: This should also be true for Project-wide Actions right? Not sure users will realize its treated as an asset

Packages/com.unity.inputsystem/Documentation~/Concepts.md
```diff
@@ -1,37 +1,18 @@
...
-|[**Action Map**](ActionAssets.md#editing-action-maps)|A named collection of Actions. You can simultaneously enable or disable all Actions in an action map, so it is useful to group Actions in Action Maps by the context in which they are relevant (for example: "gameplay").|
+|[**Control**](Controls.html)|The separate individual parts of an input device which each send input values into Unity. For example, a gamepad’s **controls** comprise multiple buttons, sticks and triggers, and a mouse’s controls include the two X and Y sensors on the underside, and the various buttons and scroll wheels on the top side.|
+|[**Action**](Actions.html)| Actions are a high-level concept that describe individual things that a user might want to do in your game or app, such as "Jump" within a game, or "Select" in an on-screen UI. They are things a user can do in your game or app as a result of input, regardless of what device or control they use to perform it. Actions generally have conceptual names that you choose to suit your project, and should usually be verbs. For example "Run", "Jump" "Crouch", "Use", "Start", "Quit".|
```
ekcoh, line 15: I am not sure I like the "things" term being used here. Would it make sense to relate or think of them as user-initiated actions or operations (similar to programming analogy of methods)? I struggle to find the right words, but tying to the analogy that an object in OOP or a component typically have methods allowing to modify the state of the object and actions typically map to these more or less 1:1

Packages/com.unity.inputsystem/Documentation~/Concepts.md
```diff
@@ -1,37 +1,18 @@
...
+|**Action Map**| Action Maps allow you to organise Actions into groups which represent specific situations where a set of actions make sense together. You can simultaneously enable or disable all Actions in an action map, so it is useful to group Actions in Action Maps by the context in which they are relevant. For example, you might have one action map for controlling a player, and another for interacting with your game's UI.|
+|[**Binding**](ActionBindings.html)| A connection defined between an **Action** and specific device controls. For example, your "Move" action might have bindings to the arrow keys and WSAD keys on the keyboard, and the left stick on a joypad, and the primary 2D axis on a VR controller. Multiple bindings like this means your game can accept cross-platform input. |
+|**Action Reference**| A reference in your script to an Input Action. Once you have a reference to an action, you can either read the current value or state of the action (also known as "polling"), or set up a callback to call your own method when actions are performed.|
```
ekcoh, line 18: Relating to "Action Reference", I notice we call this "polling" which is incorrect in this case. In one sense it could be perceived as such since you read the current value each frame so maybe o.k. but its not going to the device and fetching the latest value under the hood, it is only returning the value that was "polled" during InputSystem update.

Packages/com.unity.inputsystem/Documentation~/Installation.md
```diff
@@ -19,13 +19,13 @@ To install the new Input System, open Unity's package manager (menu: __Window >
 
 ## Enabling the new input backends
 
-By default, Unity's classic Input Manager (`UnityEngine.Input`) is active and support for the new Input System is inactive. This allows existing Unity Projects to keep working as they are.
+By default, Unity's classic Input Manager (`UnityEngine.Input`) is active, and support for the new Input System is inactive. This allows existing Unity Projects to keep working as they are.
 
-When you install the Input System package, Unity will ask whether you want to enable the new backends. If you click **Yes**, Unity will enable the new backends and disable the old backends, and the Editor will restart.
+When you install the Input System package, Unity will ask whether you want to enable the new backends. Click **Yes** to enable the new backends and disable the old backends. The Editor restarts during this process.
```
ekcoh, line 25: Is this still really correct? Isn't the default behavior that BOTH backends are active and the restart is not needed? We hope it is and if that is the case we might have to refer to this behavior for Unity version where this is still true and mention from which version default changed to BOTH.,

Packages/com.unity.inputsystem/Documentation~/PreReleaseNotes.md
```diff
@@ -2,91 +2,33 @@
...
-
-
-**Note**: The new features in this pre-release are **only documented on this page**. When reading the rest of the documentation in this package, please remember that it *will not mention these features*. So, for example, when another page in the documentation discusses **action assets**, or creating a reference to your action asset, you can in many cases use project-wide actions instead, as described on this page.
-
-## Project-wide actions and default actions
+The Input System now allows you to configure actions in the Project Settings window, in a new Input Actions Settings panel. The actions configured here apply project-wide. This means you no longer need to create an Actions asset and set up a reference to your asset to read input actions. Instead, you can configure actions in the Project Settings window, and read them directly from your scripts. You can still use Action assets if you like, but for many typical scenarios, they are no longer necessary.
```
ekcoh: minor: "the new" instead of "a new"?

ekcoh: "The actions configured here apply project-wide", would it make more sense to say "The actions configured here are available anywhere in the project"

Packages/com.unity.inputsystem/Documentation~/PreReleaseNotes.md
```diff
@@ -2,91 +2,33 @@
 
 ## Overview
 
-This pre-release contains updates to the Input System which simplifies and improves some of the main workflows compared with earlier versions of the Input System package.
+This pre-release contains updates to the Input System which simplify and improve some of the main workflows compared with earlier versions of the Input System package. This page describes the main differences introduced, and assumes you are familiar with the workflow in the previous versions of the Input System package.
 
-Because this is a pre-release, the rest of the documentation included with this version of the package has not yet been updated to reflect these changes. The documentation on this page explains the improvements and differences between these new features, and the previous version of the input system package. The improvements are as follows:
+## New project-wide actions
```
ekcoh, line 7: It seems we do not communicate the min version of Unity for this functionality to be available, maybe state that at least Unity 2022.3 is required?

Packages/com.unity.inputsystem/Documentation~/Workflow-Actions.md
ekcoh: James could we have a discussion around this separate to this review to discuss options?

ekcoh: @jamesmcgill

ekcoh (commented):
Fantastic work Ben, huge diff on the docs and many improvements and clarifications that were previously missing. I have done a review round on this now providing some feedback, questions and suggestions.

ekcoh (approved):

Packages/com.unity.inputsystem/Documentation~/PreReleaseNotes.md
duckets: Does the whole package (1.8.0) not work pre 2022.3, or just the project-wide actions feature?

jamesmcgill: Just the project-wide actions feature (and the new Project Settings UI for it).
Also we don't enable the new UI Toolkit for existing action asset editing and continue using the pre-existing IMGUI based editor.

PR #1789, opened 2023-11-02
Temporary fix to avoid exceptions being thrown by `InputControlPathEditor` editor inside UITK `IMGUIContainer` when deleting binding

Secticide:
### Description

When deleting the binding at the end of the list 2 exceptions would be thrown (Windows only). This is very likely to be something to do with how IMGUI `OnGUI` events are scheduled throughout the frame.

In my testing, we're looking to close down the area of the window (switch it to either another binding or the associated action if there are no bindings) but some `OnGUI` events are sneaking in before the UITK panel has a chance to cleanup.

This PR adds a try-catch block around the offending line to stop the exceptions from firing, it isn't the best fix but since there is a plan to reimplement the widget in UITK - this fix will tide us over until that task is complete.

### Changes made

- Added try-catch block around serialised property access to avoid exceptions when deleting a binding

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
@@ -109,7 +109,19 @@ namespace UnityEngine.InputSystem.Editor
             editButtonRect.width = 20;
             editButtonRect.height = 15;
 
-            var path = serializedProperty.stringValue;
+            var path = String.Empty;
+            try
+            {
+                path = serializedProperty.stringValue;
+            }
+            catch
+            {
+                // This try-catch block is a temporary fix for ISX-1436
+                // The plan is to convert InputControlPathEditor entirely to UITK and therefore this fix will
+                // no longer be required.
+                return;
+            }
+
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
```diff
@@ -109,7 +109,19 @@ public void OnGUI(Rect rect, GUIContent label = null, SerializedProperty propert
...
 
-            var path = serializedProperty.stringValue;
+            var path = String.Empty;
+            try
+            {
+                path = serializedProperty.stringValue;
+            }
+            catch
+            {
+                // This try-catch block is a temporary fix for: https://jira.unity3d.com/browse/ISX-1436
```
ekcoh: LGTM if this is a known issue in UITK/IMGUI integration. I take back what I said when we discussed this as in: Instead of providing an internal URL here linking to a page that is Unity internal, remove this URL (you may keep ISX-1436 reference) and instead link to this code from that issue in JIRA. Wouldn't that make more sense? I have now made the necessary linking in JIRA and made it a sub-task of that work to revert this workaround.

ekcoh (approved):

PR #1795, opened 2023-11-08
Add API additions to assist with polling workflow

chris-massie:
### Description

This adds new methods and properties to make the polling workflow easier. We plan on using these new methods in XRI.

- `InputAction.activeValueType` will allow us to call `ReadValue<TValue>()` without needing to wrap it in a `try`/`catch InvalidOperationException` for code that uses an `InputAction` that allows for either a Vector2 or float binding.
- `InputAction.GetMagnitude` will allow us to get the float value of a Vector2 action without having to recompute the expensive (sqrt) magnitude again. If you ignore processors, this lets us also use another method for getting a [0,1] float value with that Vector2 or float action.
- `InputAction.WasCompletedThisFrame` will allow us to know the frame that the action was released, but be more correct than using `WasReleasedThisFrame` for the non-default interactions like Hold or Sector in XRI.

### Changes made

To replicate the functionality of `WasReleasedThisFrame` where it does not return true when the currently pressed action is disabled, I needed to modify the signature of some `internal`/`private` methods to pass down the next phase after the target `toPhase`. This lets us skip setting the `uint` of the frame count when the action is changing from `performed` to `canceled` due to going to the `disabled` phase.

### Notes

We could potentially also add `WasCanceledThisFrame` and `WasStartedThisFrame`, but I didn't do so in this branch.

Pass-Through and Value type actions with the default interaction will never return `true` for `WasCompletedThisFrame`.
 - Pass-Through always stays in `Performed` even when returning back to the default value (i.e. it does not transition to `Canceled`). There is even a code `REVIEW` comment that questions that behavior. That seems like odd, unexpected behavior and is something that should be considered for change in a major version change.
 - Value actions once actuated go from `Waiting` to `Started`, then `Performed` in the same frame, and back to `Started` without firing an event. It stays in `Started` and only invokes `Performed` when the value changes while still actuated. Due to that, it's not clear how to track that the value is no longer performed, so Value types also always return `false` for `WasCompletedThisFrame`. It would be difficult to let the `ChangePhaseOfActionInternal` method know that it was going to `Canceled` from `Started` because of the default interaction Value type to mark it as unperformed, while not doing that for Button type actions or actions with button-like interactions like Press where it should only do so when in the `Performed` phase. So for that reason, default interaction Value types never return `true` for `WasCompletedThisFrame`.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1027,6 +1063,38 @@ public unsafe object ReadValueAsObject()
...
+        /// Magnitudes do not make sense for all types of controls. Controls that have no meaningful magnitude
+        /// will return -1 when calling this method. Any negative magnitude value should be considered an invalid value.
+        /// <br />
+        /// The magnitude returned by an action is usually determined by the
+        /// <see cref="InputControl"/> that triggered the action, i.e. by the
+        /// control referenced from <see cref="activeControl"/>.
+        /// <br />
+        /// However, if the binding that triggered is a composite, then the composite
+        /// will determine the magnitude and not the individual control that triggered (that
+        /// one just feeds values into the composite).
```
vrdave-unity:
```suggestion
        /// will determine the magnitude and not the individual control that triggered.
        /// Instead, the value of the control that triggered the action will be fed into the composite magnitude calculation.
```

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -3588,9 +3611,10 @@ public struct TriggerState
             [FieldOffset(26)] private ushort m_InteractionIndex;
             [FieldOffset(28)] private float m_Magnitude;
             [FieldOffset(32)] private uint m_LastPerformedInUpdate;
-            [FieldOffset(36)] private uint m_LastCanceledInUpdate;
-            [FieldOffset(40)] private uint m_PressedInUpdate;
-            [FieldOffset(44)] private uint m_ReleasedInUpdate;
+            [FieldOffset(36)] private uint m_LastUnperformedInUpdate;
+            [FieldOffset(40)] private uint m_LastCanceledInUpdate;
+            [FieldOffset(44)] private uint m_PressedInUpdate;
+            [FieldOffset(48)] private uint m_ReleasedInUpdate;
```
vrdave-unity: Isn't changing the field order of a struct with an explicit layout a breaking change?

```suggestion
            [FieldOffset(36)] private uint m_LastCanceledInUpdate;
            [FieldOffset(40)] private uint m_PressedInUpdate;
            [FieldOffset(44)] private uint m_ReleasedInUpdate;
            [FieldOffset(48)] private uint m_LastUnperformedInUpdate;
```

vrdave-unity (changes requested):
Couple of minor nitpicks

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
chris-massie: As far as I could tell, no native code was using this struct and it's internal to the package, so it wasn't a breaking change. But I can go with this suggestion to minimize the diff.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1616,6 +1616,289 @@ public void Actions_CanQueryIfPerformedInCurrentFrame()
         Assert.That(holdAction.WasPerformedThisFrame(), Is.False);
     }
 
+    [Test]
+    [Category("Actions")]
+    public void Actions_CanQueryIfUnperformedInCurrentFrame()
```
jamesmcgill: `WasUnperformedThisFrame()` is never tested that it can be `true` in this test

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1616,6 +1616,289 @@ public void Actions_CanQueryIfPerformedInCurrentFrame()
...
+            // since it was not held long enough to be performed.
+            Release(gamepad.buttonSouth);
+
+            Assert.That(defaultTrace, actionType != InputActionType.PassThrough
+                ? Canceled(defaultAction)
+                : Performed(defaultAction));
+            Assert.That(defaultAction.WasPressedThisFrame(), Is.False);
+            Assert.That(defaultAction.WasReleasedThisFrame(), Is.True);
+            Assert.That(defaultAction.WasPerformedThisFrame(), Is.EqualTo(actionType == InputActionType.PassThrough));
+            Assert.That(defaultAction.WasUnperformedThisFrame(), Is.EqualTo(actionType == InputActionType.Button));
```
jamesmcgill: On line 1726: `WasPerformedThisFrame()` was true in all cases.
Now it's only `Unperformed` for Button types.

Maybe this will make sense to me further into the review, but at first glance that doesn't look symmetrical.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1616,6 +1616,289 @@ public void Actions_CanQueryIfPerformedInCurrentFrame()
...
+            // Release button, the actions should cancel. The hold action should now be considered unperformed
+            // since it was held long enough to be performed.
+            Release(gamepad.buttonSouth);
+
+            Assert.That(defaultTrace, actionType != InputActionType.PassThrough
+                ? Canceled(defaultAction)
+                : Performed(defaultAction));
+            Assert.That(defaultAction.WasPressedThisFrame(), Is.False);
+            Assert.That(defaultAction.WasReleasedThisFrame(), Is.True);
+            Assert.That(defaultAction.WasPerformedThisFrame(), Is.EqualTo(actionType == InputActionType.PassThrough));
```
jamesmcgill, line 1942: What happens for PassThrough on the next frame after this?.
Does it become `Unperformed` at that point? Should it?

If there is anything interesting there maybe extend this test to show the state change.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/AxisComposite.cs
```diff
@@ -16,7 +16,7 @@ namespace UnityEngine.InputSystem.Composites
     /// <example>
     /// <code>
     /// var action = new InputAction();
-    /// action.AddCompositeBinding("Axis(minValue=0,maxValue=2")
+    /// action.AddCompositeBinding("Axis(minValue=0,maxValue=2)")
```
ekcoh, line 19: Thanks for correcting these

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
```diff
@@ -3,7 +3,7 @@
 
 There are two main techniques you can use to respond to Actions in your project. These are to either use **polling** or an **event-driven** approach.
 
-- The **Polling** approach refers to the technique of repeatedly checking the current state of the Actions you are interested in. Typically you would do this in the Update() method of a Monobehaviour script.
+- The **Polling** approach refers to the technique of repeatedly checking the current state of the Actions you are interested in. Typically you would do this in the `Update()` method of a `MonoBehaviour` script.
```
ekcoh, line 6: Not completely related to this PR, but maybe we should provide some additional guidance here that polling may be used when the order of actions performed is irrelevant, while event-driven callback approach is appropriate when order of actions during a frame is relevant or what do you think?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1834,6 +2119,658 @@ public void Actions_CanReadValueFromAction_AsButton(InputActionType actionType,
...
+                Assert.That(action.WasPerformedThisFrame(), Is.True);
+                Assert.That(action.WasUnperformedThisFrame(), Is.True);
+            }
+            else if (isHold)
+            {
+                Assert.That(trace, Is.Empty);
+                Assert.That(action.phase, Is.EqualTo(InputActionPhase.Started));
+                Assert.That(action.WasPerformedThisFrame(), Is.False);
+                Assert.That(action.WasUnperformedThisFrame(), Is.False);
+            }
```
jamesmcgill, line 2820: I was waiting to see the result of the Hold interaction being performed and unperformed but I think it's missing?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -608,6 +607,42 @@ public unsafe InputControl activeControl
...
+        /// control referenced from <see cref="activeControl"/>.
+        ///
+        /// However, if the binding that triggered is a composite, then the composite
+        /// will determine values and not the individual control that triggered (that
+        /// one just feeds values into the composite).
+        /// </remarks>
+        /// <seealso cref="InputControl.valueType"/>
+        /// <seealso cref="InputBindingComposite.valueType"/>
+        /// <seealso cref="activeControl"/>
+        public unsafe Type activeValueType
```
ekcoh, line 635: I would assume this type potentially changes multiple times during a frame, i.e. if using callbacks there might be multiple controls active at certain points in time. For a polling scenario I assume only the "most" or "last" actuated control would be influencing the type here or how is this intended to be interpreted? If this is the intention, maybe it should be mentioned in the doc?

ekcoh: Great you covered composite behavior in the doc

ekcoh: Not sure I completely understand how this is intended to be used though, would it be possible to incorporate it into the example or at least elaborate a bit on a typical use-case so I understand better?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1027,6 +1063,38 @@ public unsafe object ReadValueAsObject()
             return null;
         }
 
+        /// <summary>
+        /// Read the current amount of actuation of the control that is driving this action.
+        /// </summary>
+        /// <returns>Returns the current level of control actuation (usually [0..1]) or -1 if
```
ekcoh, line 1075: Based on the code below it seems to return 0f and not -1f if not supported?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1275,6 +1344,73 @@ public unsafe bool WasPerformedThisFrame()
             return false;
         }
 
+        /// <summary>
+        /// Check whether <see cref="phase"/> changed away from <see cref="InputActionPhase.Performed"/> at any point
```
ekcoh: Minor: Instead of "...whether phase changed away from Performed at any point in the current frame after being in performed state.", maybe "...whether phase transitioned from Performed to any other phase value at least once in the current frame." to emphasize that the evaluation is tied to the step function edge of high->low and that the number of occurrences doesn't matter, i.e. this may be true and phase may still be Performed?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1275,6 +1344,73 @@ public unsafe bool WasPerformedThisFrame()
...
+        /// in the current frame after being in performed state.
+        /// </summary>
+        /// <returns>True if the action unperformed this frame.</returns>
+        /// <remarks>
+        /// This method is different from <see cref="WasReleasedThisFrame"/> in that it depends directly on the
```
ekcoh, line 1362: Not really relevant for this PR but I believe the problem is the shortcut taken when introducing WasReleasedThisFrame since it could have been the same thing if "Release" was made a proper interaction which it maybe should have been...

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1275,6 +1344,73 @@ public unsafe bool WasPerformedThisFrame()
             return false;
         }
 
+        /// <summary>
```
ekcoh, line 1353: Thanks for writing an elaborate documentation that walks the reader through the definition with examples. This is very helpful. I have to admit I am not fully onboard with "unperformed" as a term here as in e.g. `wasUnperformedThisFrame()`. As a non native English speaker I had to look online for definitions and it seems like the definition is "not having been performed", which isn't in line with the behavior described here since its fired after something has been performed but is fired as soon as an interaction stops being performed. Hence I would suggest we consider alternative names.... maybe "Completed" would be an alternative, e.g. wasCompletedThisFrame() since the interaction has been through the phases and completed its cycle?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1275,6 +1344,73 @@ public unsafe bool WasPerformedThisFrame()
...
+        /// <see cref="Interactions.HoldInteraction"/> assigned to it. In the frame where the space bar
+        /// is pressed, <see cref="WasPressedThisFrame"/> will be true (because the button/key is now pressed)
+        /// but <see cref="WasPerformedThisFrame"/> will still be false (because the hold has not been performed yet).
+        /// If at that time the space bar is released, <see cref="WasReleasedThisFrame"/> will be true (because the
+        /// button/key is now released) but <c>WasUnperformedThisFrame</c> will still be false (because the hold
+        /// had not been performed yet). If instead the space bar is held down for long enough for the hold interaction
+        /// and then released, <c>WasUnperformedThisFrame</c> will be true (because the action is no longer performed)
+        /// and only in the frame where the hold performed.
+        ///
+        /// For another example where the action could be considered pressed but also unperformed, let's say the action
```
ekcoh: Its great to have another example here but this sits in Input System where there currently is no Sector interaction so it might be difficult for a user to relate to it. Is there a Sector interaction in XRI? Maybe link to the docs of that interaction then?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1275,6 +1344,73 @@ public unsafe bool WasPerformedThisFrame()
...
+        /// is bound to the thumbstick and that the binding has a Sector interaction assigned to it such that it only
+        /// performs in the forward sector area past a button press threshold. In the frame where the thumbstick is
+        /// pushed forward, both <see cref="WasPressedThisFrame"/> will be true (because the thumbstick actuation is
+        /// now considered pressed) and <see cref="WasPerformedThisFrame"/> will be true (because the thumbstick is in
+        /// the forward sector). If the thumbstick is then moved to the left in a sweeping motion, <see cref="IsPressed"/>
+        /// will still be true. However, <c>WasUnperformedThisFrame</c> will also be true (because the thumbstick is
```
ekcoh: Again this signals the interaction in its performed state completed

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1599,7 +1602,7 @@ private static bool ShouldIgnoreInputOnCompositeBinding(BindingState* binding, I
         /// If an action has multiple controls bound to it, control state changes on the action may conflict with each other.
         /// If that happens, we resolve the conflict by always sticking to the most actuated control.
         ///
-        /// Pass-through actions (<see cref="InputAction.passThrough"/>) will always bypass conflict resolution and respond
+        /// Pass-through actions (<see cref="InputActionType.PassThrough"/>) will always bypass conflict resolution and respond
```
ekcoh, line 1605: Thanks for correcting previous typo

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3308,6 +3308,23 @@ public static int ListEnabledActions(List<InputAction> actions)
...
+        /// <param name="id">ID of the action to look for.</param>
+        /// <returns>Returns the action with the given ID or <see langword="null"/> if no action could be found with the given ID.</returns>
+        /// <remarks>
+        /// For the action to be found, the action must be part of an <see cref="InputActionAsset"/>
+        /// that has had any of its actions enabled at some point. For embedded actions that are not part of an asset, the
+        /// action must have been enabled at some point.
+        /// </remarks>
+        /// <seealso cref="InputActionAsset.FindAction(Guid)"/>
+        /// <seealso cref="InputActionMap.FindAction(Guid)"/>
+        public static InputAction FindAction(Guid id)
```
ekcoh: Would this be applicable in both edit and playmode? If not I think we should call it out in docs.

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
```diff
@@ -39,27 +39,40 @@ public class Example : MonoBehaviour
...
 
 public class Example : MonoBehaviour
 {
-    InputAction jumpAction;
+    InputAction interactAction;
 
     private void Start()
     {
-        jumpAction = InputSystem.actions.FindAction("Jump");
+        interactAction = InputSystem.actions.FindAction("Interact");
```
ekcoh, line 63: Would this really be a valid example? It states in docs of FindAction that the action must have been previously enabled to be found (returned by that function) and I cannot see its enabled here?

ekcoh (commented):
Thanks for this PR @chris-massie and sorry for the delayed review and associated feedback. In general I think these additions make a lot of sense and adds value. However I have some comments and questions as can be seen in the review.

Adding review as a comment for now, approving after concluded questions/discussion aspects.

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: Could we clarify this further with the test name since its seems central to this test case to exploit problems with disable/enable without ticking time forwards?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1834,6 +2119,658 @@ public void Actions_CanReadValueFromAction_AsButton(InputActionType actionType,
...
+
+            // Press some more such that it crosses the press threshold.
+            Set(gamepad.leftTrigger, 0.75f);
+
+            // W = Waiting, S = Started, P = Performed, C = Canceled
+            // P/U = Performed/Unperformed
+            // (* means listeners not invoked)
+            // | Type   | Interaction | Phase Change     | P/U | isButtonLike | isHold |
+            // |--------|-------------|------------------|-----|--------------|--------|
+            // | Value  | Default     | S -> P, S*       | T/  |              |        |
```
ekcoh: What is the meaning of "T/" could this be explained in incline comment? Does it mean Performed=true, Unperformed=false? In that case maybe T/F is easier to understand?

ekcoh (commented):
Also thanks for investing in providing code coverage via automated tests. I believe its a necessity for this rather complex evaluation functions.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
jamesmcgill: I would prefer not to have this going looking into the runtime state.
As you put in the remarks there is a lot of caveats as to whether and when this function will work.
This exposes directly the innards of InputSystem which means the public API will now need to maintain this behaviour (e.g. Action must have been enabled etc). Changing ActionState might break that contract.

The other 'FindAction' methods are not using runtime state but rather the data registered with InputSystem (assets, action maps or whatever) that later feed into the runtime state. If this is really needed, then we would need to add some type of array `InputAction`s of registered single actions and aggregate that with action maps and actions, if we don't have it already.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1275,6 +1344,73 @@ public unsafe bool WasPerformedThisFrame()
...
+        /// Unlike <see cref="ReadValue{TValue}"/>, which will reset when the action goes back to waiting
+        /// state, this property will stay true for the duration of the current frame (that is, until the next
+        /// <see cref="InputSystem.Update"/> runs) as long as the action was unperformed at least once.
+        ///
+        /// <example>
+        /// <code>
+        /// var teleport = playerInput.actions["Teleport"];
+        /// if (teleport.WasPerformedThisFrame())
+        ///     InitiateTeleport();
+        /// else if (teleport.WasUnperformedThisFrame())
```
jamesmcgill: It's not actually clear to me what happens in the Hold case.
Does Unperformed happen always on the next frame after performed?
Or does it remain in the performed state for several frames until the user releases the control?
I don't think it was covered in the tests.

jamesmcgill (changes requested):
Thanks for this @chris-massie and also sorry for the long delay.
A lot of good changes especially the test coverage and docs.

I would ask to remove the new`FindAction` if you can live without it as it doesn't work quite like the others and exposes some internal behaviour.

As there is a lot of docs I would add @duckets for a quick review.
Maybe he also has some thoughts on `Unperformed` as a name. I feel it's somehow off, but I don't really have any better ideas.

jamesmcgill (approved):

jamesmcgill:
Approved after speaking to Chris and agreeing to remove FindAction for now and perhaps revisit in a future PR

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
chris-massie: The `magnitude` property can be -1 even though `haveMagnitude` is true. The `Actions_CanQueryMagnitudeFromAction_WithQuaternionControl_ReturnsInvalidMagnitude` test handles this scenario where the attitude Quaternion control is actuated but doesn't have a meaningful magnitude value, so it returns -1.

To me, it seemed like returning 0 in this method was preferable to returning -1 if no control is actuated, so that's why I used that value in this method for the default.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
chris-massie: Based on feedback from James, I removed this method.

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
chris-massie: This is calling the `FindAction(string)` method, which already existed.

The other example in this file for showing IsPressed does this same call and also doesn't enable the action, so the behavior still expects it to be enabled for it to work.

I've also removed the new `FindAction(Guid)` method I added from this PR.

Assets/Tests/InputSystem/CoreTests_Actions.cs
chris-massie: Yes it's True/False. I kept it blank because I thought it added visual noise and made it harder to see which ones are true at a glance, but I've switched it to F everywhere it was blank before. I also added an explanation to the legend above.

Assets/Tests/InputSystem/CoreTests_Actions.cs
chris-massie: Unperformed isn't that useful for Pass-Through actions since those actions don't go through any kind of button-like phase progression. It's just always Performed, even when the control goes back to default.

Assets/Tests/InputSystem/CoreTests_Actions.cs
chris-massie: Pass-Through actions are weird in that they always stay Performed, even when the control goes back to default. So it never leaves the Performed state to trigger Unperformed.

Assets/Tests/InputSystem/CoreTests_Actions.cs
chris-massie: That case is covered in `Actions_CanDistinguishCanceledAndUnperformedInCurrentFrame`. In this `Actions_CanReadPerformedFromAction_AsButton` test, I basically kept all the same setup and button change calls as the `Actions_CanReadValueFromAction_AsButton` test, and that didn't test that case either.

I added a new shorter test `Actions_CanQueryIfHoldInteractionUnperformedInCurrentFrame` that explicitly checks the hold interaction case.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
chris-massie: I believe one of the tests covered this, but I added a new shorter test `Actions_CanQueryIfHoldInteractionUnperformedInCurrentFrame` that explicitly checks the hold interaction case.

It remains in performed for several frames until they release, but `WasPerformedThisFrame` is only true for that one frame. `WasUnperformedThisFrame` is only true for one frame and then stays false.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
chris-massie: Yes, there is a Sector interaction in XRI, and I now added a link to it.

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
chris-massie: I thought it was still frame based? Does Input System actually know if A was pressed before B if they were both pressed in the same frame?

I don't think I know enough about the details to update the documentation about that in this PR.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
chris-massie: I added some notes to the xml-doc. Basically we have some input actions where we permit multiple bindings of different control types. In XRI, we allow either a Vector2 or float trigger for "select", but we can't call `ReadValue<float>` or `ReadValue<Vector2>` without it potentially throwing an exception. We have to wrap it in a try/catch. It would be nicer if we could query the expected type directly to know which version to call.

This value is already exposed in the `CallbackContext` with `CallbackContext.valueType`, but this property lets us get it in a polling way instead of having to be in the `performed` callback method.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
chris-massie: I renamed it to `WasCompletedThisFrame` as suggested.

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
ekcoh: When observing events yes, but not when using polling APIs. I will take a separate. note on addressing this.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: Totally agree or potentially in the long run avoid having non-typed actions in general. Thanks for explaining.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: ok

ekcoh (approved):

ekcoh:
@vrdave-unity OK to merge this one now that all checks pass or is your requested change from review still valid? If OK, would you mind approving this PR.

vrdave-unity (approved):

PR #1797, opened 2023-11-16
FIX: Make `DisableAndResetAllDevices` handle events while player runs in the background

jfreire-unity:
### Description

https://jira.unity3d.com/browse/ISXB-615

### Changes made

This PR addresses an issue when the player runs in the background (`InputSystem.runInBackground == true` )and `InputSystem.settings.backgroundBehavior = DisableAndResetAllDevices` is set. These settings were previously introduced in #1324.

With this change, `DisableAndResetAllDevices` has a behavior similar to `DisableAndResetNonBackgroundDevices`, where events that are enqueued while the app is in the background will be [discarded based on their type](https://github.com/Unity-Technologies/InputSystem/blob/bcf381095d5108aee26fbaaf2021efd1ccb909ad/Packages/com.unity.inputsystem/InputSystem/InputManager.cs#L3126). [`canEarlyOut`](https://github.com/Unity-Technologies/InputSystem/blob/bcf381095d5108aee26fbaaf2021efd1ccb909ad/Packages/com.unity.inputsystem/InputSystem/InputManager.cs#L2932) no longer depends on `DisableAndResetAllDevices`, which means we won't early out in this situation.

This solves the issue of having to process input events only when the focus was regained which could lead to the wrong triggering of callbacks. Also, the event buffer could grow to big such that we could see errors of type ` Exceeded budget for maximum input event throughput per InputSystem.Update()` is being raised, in case the player stayed out of focus for the time it took to fill the buffer to its max allowed size.

A test was added to test this behavior in the player.

### Notes

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -2933,9 +2933,10 @@ namespace UnityEngine.InputSystem
                 // Early out if there's no events to process.
                 eventBuffer.eventCount == 0
                 || canFlushBuffer
+
+#if UNITY_EDITOR
                 // If we're in the background and not supposed to process events in this update (but somehow
                 // still ended up here), we're done.
-#if UNITY_EDITOR
                 || ((!gameHasFocus || gameShouldGetInputRegardlessOfFocus) &&
                     ((m_Settings.backgroundBehavior == InputSettings.BackgroundBehavior.ResetAndDisableAllDevices && updateType != InputUpdateType.Editor)
                         || (m_Settings.editorInputBehaviorInPlayMode == InputSettings.EditorInputBehaviorInPlayMode.AllDevicesRespectGameViewFocus && updateType != InputUpdateType.Editor)
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -2686,9 +2686,7 @@ namespace UnityEngine.InputSystem
                 m_HasFocus = focus;
                 return;
             }
-            #endif
 
-            #if UNITY_EDITOR
             var gameViewFocus = m_Settings.editorInputBehaviorInPlayMode;
             #endif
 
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2934,22 +2932,20 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
                 // Early out if there's no events to process.
                 eventBuffer.eventCount == 0
-                || canFlushBuffer ||
+                || canFlushBuffer
                 // If we're in the background and not supposed to process events in this update (but somehow
                 // still ended up here), we're done.
-                ((!gameHasFocus || gameShouldGetInputRegardlessOfFocus) &&
-                    ((m_Settings.backgroundBehavior == InputSettings.BackgroundBehavior.ResetAndDisableAllDevices && updateType != InputUpdateType.Editor)
 #if UNITY_EDITOR
+                || ((!gameHasFocus || gameShouldGetInputRegardlessOfFocus) &&
```
jamesmcgill, line 2940: The 2 lines of comments above should probably move together with this.

jamesmcgill (approved):
Seems fine, the logic in this part always takes a while to get my head around.

We agreed about letting it flow through in this case so that `ResetAndDisableNonBackgroundDevices` and `ResetAndDisableAllDevices` follow the same path. I think that is still fine but I see that means removing the early out for the player entirely when out of focus and doing the whole update.

I wonder if we could just flush the buffer for the case you had?
I see for Editor it will do that for `ResetAndDisableAllDevices` but not in the Player.

Would something like this work:
```
            var canFlushBuffer =
                    // If out of focus and runInBackground is off (or all devices are disabled).
                    !gameHasFocus
                    && (!m_Runtime.runInBackground || m_Settings.backgroundBehavior == InputSettings.BackgroundBehavior.ResetAndDisableAllDevices)
#if UNITY_EDITOR
                    // .. and ExactlyAsInPlayer is on, discard input.
                    && m_Settings.editorInputBehaviorInPlayMode == InputSettings.EditorInputBehaviorInPlayMode.AllDeviceInputAlwaysGoesToGameView
#endif
                    ;
```

jfreire-unity:
> Seems fine, the logic in this part always takes a while to get my head around.
> 
> We agreed about letting it flow through in this case so that `ResetAndDisableNonBackgroundDevices` and `ResetAndDisableAllDevices` follow the same path. I think that is still fine but I see that means removing the early out for the player entirely when out of focus and doing the whole update.
> 
> I wonder if we could just flush the buffer for the case you had? I see for Editor it will do that for `ResetAndDisableAllDevices` but not in the Player.
> 
> Would something like this work:
> 
> ```
>             var canFlushBuffer =
>                     // If out of focus and runInBackground is off (or all devices are disabled).
>                     !gameHasFocus
>                     && (!m_Runtime.runInBackground || m_Settings.backgroundBehavior == InputSettings.BackgroundBehavior.ResetAndDisableAllDevices)
> #if UNITY_EDITOR
>                     // .. and ExactlyAsInPlayer is on, discard input.
>                     && m_Settings.editorInputBehaviorInPlayMode == InputSettings.EditorInputBehaviorInPlayMode.AllDeviceInputAlwaysGoesToGameView
> #endif
>                     ;
> ```

But if we flush the buffer, we lose events in case devices are removed/added in the background, right? 
The goal of not early out in the player is so that we can only process device configuration events and "flush" all the other events.

jfreire-unity:
> I see for Editor it will do that for ResetAndDisableAllDevices but not in the Player.

Yeah, I only did this to keep our unit tests green for this in the Editor, for now, to avoid breaking some specific use-case 🤷 
I didn't deep dive into the specifics of input event processing in the Editor & Player Mode, to be honest. But probably should.

ekcoh (approved):
LGTM, as a side note its a pity the Input module isn't aware of this setting. It would make sense to discard data events on lowest level when/if consumer is not emptying the queue. That would have avoided enqueuing the data in the first place.

PR #1799, opened 2023-11-21
FIX: Listen for standard events rather than hardcoding keybindings (ISX-1745).

graham-huws:
### Description

Fix for [ISX-1745](https://jira.unity3d.com/browse/ISX-1745). Previously, keyboard shortcuts for rename/delete/duplicate were hardcoded to receive specific keyboard presses, which meant users would not be able to make use of the standard customisable shortcuts across the editor, nor use the matching menu bar action (eg. Edit > Delete). This updates the code to receive [Command Events](https://docs.unity3d.com/2023.2/Documentation/Manual/UIE-Command-Events.html) instead.

### Changes made

Fixed: Menu bar/customisable keyboard shortcut support used when interacting with Actions and Action Maps.

### Notes

One downside is that F2 isn't (by default) set to Rename in Unity - which it probably _should_ be on Windows and Linux - so that no longer works to rename without editing the shortcuts manager. But Edit > Rename from the menu bar now works, or (on Mac) using the Return/Enter key, as well as other existing methods (double click an item; right click > Rename).

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -12,6 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - Fixed missing confirmation popup when deleting a control scheme.
+- Fixed support for menu bar/customisable keyboard shortcuts used when interacting with Actions and Action Maps.
 
 ## [1.8.0-pre.2] - 2023-11-09
 
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
@@ -1,4 +1,5 @@
 #if UNITY_EDITOR && UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
+using CmdEvents = UnityEngine.InputSystem.Editor.InputActionsEditorConstants.CommandEvents;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine.InputSystem.Utilities;
@@ -11,11 +12,6 @@ namespace UnityEngine.InputSystem.Editor
     /// </summary>
     internal class ActionMapsView : ViewBase<ActionMapsView.ViewState>
     {
-        private const string kCmd_Rename = "Rename";
-        private const string kCmd_Delete = "Delete";
-        private const string kCmd_SoftDelete = "SoftDelete";
-        private const string kCmd_Duplicate = "Duplicate";
-
         public ActionMapsView(VisualElement root, StateContainer stateContainer)
             : base(stateContainer)
         {
@@ -122,14 +118,14 @@ namespace UnityEngine.InputSystem.Editor
         {
             switch (evt.commandName)
             {
-                case kCmd_Rename:
+                case CmdEvents.Rename:
                     ((InputActionMapsTreeViewItem)m_ListView.GetRootElementForIndex(m_ListView.selectedIndex))?.FocusOnRenameTextField();
                     break;
-                case kCmd_Delete:
-                case kCmd_SoftDelete:
+                case CmdEvents.Delete:
+                case CmdEvents.SoftDelete:
                     ((InputActionMapsTreeViewItem)m_ListView.GetRootElementForIndex(m_ListView.selectedIndex))?.DeleteItem();
                     break;
-                case kCmd_Duplicate:
+                case CmdEvents.Duplicate:
                     ((InputActionMapsTreeViewItem)m_ListView.GetRootElementForIndex(m_ListView.selectedIndex))?.DuplicateItem();
                     break;
                 default:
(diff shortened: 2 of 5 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - Fixed missing confirmation popup when deleting a control scheme.
+- Fixed support for menu bar/customisable keyboard shortcuts used when interacting with Actions and Action Maps.
```
ekcoh, line 15: Great that changelog item is in place.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -12,6 +11,11 @@ namespace UnityEngine.InputSystem.Editor
     /// </summary>
     internal class ActionMapsView : ViewBase<ActionMapsView.ViewState>
     {
+        private const string kCmd_Rename = "Rename";
```
ekcoh: I see commands are referenced via a string name, isn't there any central place where these are defined that's accessible from the package instead of redefining these? I suspect it would break if these changed or got superseded by something else?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -25,6 +25,11 @@ internal class ActionsTreeView : ViewBase<ActionsTreeView.ViewState>
         //save TreeView element id's of individual input actions and bindings to ensure saving of expanded state
         private Dictionary<Guid, int> m_GuidToTreeViewId;
 
+        private const string kCmd_Rename = "Rename";
```
ekcoh: If these strings are really needed I rather see we use a common internal struct to keep the constants instead of duplicating them. I notice this is a duplicate of changes in ActionsMapsView?

ekcoh (changes requested):
I think its a great improvement to utilize the Unity Shortcut support for any shortcuts we have in the editor since it would otherwise diverge and behave differently than the editor for any user that customizes shortcuts. LGTM apart from duplicating string constants commented in review, ideally we could rely on a centralized place where these are defined or fallback on having a single package local definition of them.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -55,7 +59,8 @@ public ActionMapsView(VisualElement root, StateContainer stateContainer)
                 item.FocusOnRenameTextField();
             };
 
-            m_ListView.RegisterCallback<KeyDownEvent>(OnKeyDownEvent);
+            m_ListView.RegisterCallback<ExecuteCommandEvent>(OnExecuteCommand);
+            m_ListView.RegisterCallback<ValidateCommandEvent>(OnValidateCommand);
```
jfreire-unity, line 59: Nice change :)

jfreire-unity (commented):
Have the same comments as @ekcoh , otherwise looks good 👍

ritamerkl (approved):
Nice! Thank you for bringing this up. I wasn't aware of the existence of the ExecuteCommand flow.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
graham-huws-unity: So, unfortunately there _is_ a central place... currently in IMGUI code 😄 Thread about this issue here:
https://unity.slack.com/archives/C3414V4UV/p1698856135642649

I messaged Antoine about it and he hasn't had time to move it out. Would making use of the IMGUI code be preferable here?

jamesmcgill (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
jfreire-unity: I think it would be fine to use the IMGUI code, no?
Once they move to another place we could change them.
We currently still have some IMGUI dependencies anyway. And this specific one is not related to IMGUI functionality, just naming references.

graham-huws-unity: Had a look at this again today, and the existing definitions are also marked as internal, so I believe we'd have to jump through [quite a few hoops](https://confluence.unity3d.com/display/PAK/From+internal+to+public+API) just to get access to them - don't think that's worth it for four strings 😞

Agreed @ekcoh that it would be better to avoid duplication of the strings - what would be the best file for them to be moved to?

ekcoh: Sorry for delayed feedback here, must have overlooked you had an outstanding question to me here. I would suggest InputActionsEditorConstants.cs ?

graham-huws-unity: Fixed in 1bda88dac95a5a89f3ff692b85fb73db3f7fc59b.

ekcoh (approved):

Pauliusd01 (approved):
LGTM checked deleting on win and mac. Checked mentioned renaming issue before and after and whether it works when re-setting it back in the shortcut manager

PR #1804, opened 2023-11-27
FIX: fix Destroy may not be called from edit mode error

jamesmcgill:
### Description

Fixes the `Destroy may not be called from edit mode` error message in InputForUI

Described here:
https://forum.unity.com/threads/case-1426900-error-destroy-may-not-be-called-from-edit-mode-is-shown-when-stopping-play.1279895/#post-9479251

### Changes made

Change to use `DestroyImmediate()`

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
@@ -642,7 +642,11 @@ namespace UnityEngine.InputSystem.Plugins.InputForUI
             // The Next/Previous action is not part of the input actions asset
             UnregisterNextPreviousAction();
 
+#if UNITY_EDITOR
             UnityEngine.Object.DestroyImmediate(m_InputActionAsset);
+#else
+            UnityEngine.Object.Destroy(m_InputActionAsset);
+#endif
         }
 
         public struct Configuration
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -642,7 +642,7 @@ void UnregisterActions(Configuration cfg)
             // The Next/Previous action is not part of the input actions asset
             UnregisterNextPreviousAction();
 
-            UnityEngine.Object.Destroy(m_InputActionAsset); // TODO check if this is ok
+            UnityEngine.Object.DestroyImmediate(m_InputActionAsset);
```
ekcoh, line 646: Same qeustion/comment as when @Pauliusd01 looked into this, shouldn't this be conditioned with #UNITY_EDITOR and use Destroy in player build?
https://docs.unity3d.com/ScriptReference/Object.DestroyImmediate.html

ekcoh (approved):
Change looks fine, but should make sure its tested without side-effects in both editor and standalone editor, see comment.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
Pauliusd01: Related bug for context: https://jira.unity3d.com/browse/ISXB-677

jamesmcgill: I've made the change @ekcoh suggested.
Although, as the information given to @Pauliusd01 states, there really isn't a good solution here except for making changes to how the editor works (which of course we cannot do).
Testing doesn't reveal any issues with 'DestroyImmediate' (or with Hakans suggestion).

PR #1806, opened 2023-12-10
FIX: Select appropriate action/action map when right clicking (ISX-1743).

graham-huws:
### Description

Fix for [ISX-1743](https://jira.unity3d.com/browse/ISX-1743) - Right clicks would not change the selected item, giving a misleading context menu.

### Changes made

Fixed: Selection not changing when right clicking an Action Map or Action

### Notes

Unfortunately, right clicking a different Action and selecting Rename does nothing until right clicking a second time (on Mac) - There seems to be some voodoo about this which probably needs a #devs-uitoolkit wizard. The renameTextfield.SelectAll() call doesn't seem to do anything, and the Focus() call afterwards fails too. I wondered if this may be related to [this (closed as won't fix) bug](https://jira.unity3d.com/browse/UUM-1571) from a while ago, though changing the event to MouseDown didn't help in my case.

As is, I think it makes sense to push this PR ahead as-is, and a new lower prio bug should be logged for this remaining issue.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ekcoh (commented):
In general LGTM, however it sounds like the null behavior hasn't really been understood completely. When @ritamerkl is available again it would be great to get your perspective on this. Will likely provide additional feedback once I have been able to test the branch.

jamesmcgill (approved):
Looks like good changes to me (simpler/less code).
I agree we should get this merged rather than holding out for all issues to be resolved.
However I'll let @ritamerkl do the merge after she gets a chance to review this.

Pauliusd01:
Checking

Pauliusd01:
Besides the mentioned 2nd click being needed issue, this also introduces a new issue with multiple Input Action Windows present - if you select something in one window and then right-click in the other, the context menu will always do things on the first window's selection, no matter how many times you left click or right click in the second. (Adding a video of me left/right clicking between windows in your PR and then in current Develop) I'm not sure if this is as an easy issue to ignore as the one you already found

https://github.com/Unity-Technologies/InputSystem/assets/54306142/84983acd-bda3-46fa-9b4c-5a8d03157871

Pauliusd01 (changes requested):
My recommendation is leaning more to fix the multi window issue before merging but if someone else says It's fine I can change the status

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsTreeViewItem.cs
```diff
@@ -102,16 +97,6 @@ async void DelayCall()
...
 
-        public void DeleteItem()
-        {
-            OnDeleteItem?.Invoke(0);
-        }
-
-        public void DuplicateItem()
-        {
-            OnDuplicateItem?.Invoke(0);
-        }
```
ritamerkl: this is ok in my point of view as long as it is still working the intended way - even though I don't understand the context to the problem of the PR

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -293,6 +288,20 @@ private void OnValidateCommand(ValidateCommandEvent evt)
...
+            if (evt.button == (int)MouseButton.RightMouse && evt.clickCount == 1)
+            {
+                // Look upwards to the immediate child of the scroll view, so we know what Index to use
+                var element = (evt.target as VisualElement);
+                while (element.name != "unity-tree-view__item")
+                    element = element.parent;
+
+                m_ActionsTreeView.SetSelection(element.parent.IndexOf(element));
+            }
+        }
```
ritamerkl, line 450: I am a bit worried about setting selections like this. Is it even the expected behaviour to select things on right click? If then we should change it for actionmaps, actions and bindings to keep it consistent. Also I have the feeling there is a simpler way to do that then this, in particular looking at the while loop. (maybe there is even a way to allow right click selection in ActionsTreeView?) 
Also did we contact the UIToolkit team members here? Maybe they have a recommandation for this case.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed Documentation~/filter.yml GlobalNamespace rule removing all API documentation.
 - Fixed `Destroy may not be called from edit mode` error [ISXB-695](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-695)
 - Fixed possible exceptions thrown when deleting and adding Action Maps.
+- Fixed selection not changing when right clicking an Action Map or Action.
```
ekcoh: I would suggest adding some context here, i.e.

"Fixed selection not changing when right-clicking an Action Map or Action in the Project Settings Input Action Editor."

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ekcoh: I am also suprised there is not a out-of-the-box solution for this in UITK that would provide consistent behavior across the editor. I would rather keep this behavior instead of introducing inconsistency with the rest of the editor. Could we reach out to UITK team as suggested by Rita and possible find similar examples in the editor?

ekcoh: Some examples I found by quickly testing in the editor:
- Tags and layers in Project Settings list boxes (might be separate controls though), change focus/selection on right click
- Noticed that list in Package Manager settings do not select on right click but neither has a context menu either.
- Tree view in Project Settings (left panel) select on right click but do not have a context menu
- Project Hierarchy (probably most similar example), select on right click and has a context menu, so maybe UITK team can provide guidance on how that was achieved? Custom or some common mechanism? Also Game Object Hierarchy behaves this way

graham-huws-unity: Thanks for comments both!

> Is it even the expected behaviour to select things on right click?

I would think so - as well as the project/gameobject areas mentioned by Håkan, it's standard behaviour across operating systems, so I'd expect it to work here too.

> Also I have the feeling there is a simpler way to do that then this, in particular looking at the while loop.

Same! 😆 Maybe a question for #devs-uitoolkit if nobody has a suggestion? I'm still very new to it all.

Regarding asking UITK team about rightclick - the method here is one suggested by someone on their team just a few months ago, for the same issue: https://unity.slack.com/archives/C3414V4UV/p1693320293138709?thread_ts=1693306035.515169&cid=C3414V4UV and https://unity.slack.com/archives/C3414V4UV/p1694008889654529

Project and Hierarchy views are both IMGUI unfortunately, so can't make use of their methods 😞

ritamerkl: > I would think so - as well as the project/gameobject areas mentioned by Håkan, it's standard behaviour across operating systems, so I'd expect it to work here too.

Okay, wasn't aware of this, but that simplifies things!

I think #devs-uitoolkit is a good place to ask. Also both threads seem to be helpful, especially the trickle down recommendation for catching events before the context menu rightclick triggers.

Pauliusd01 (approved):
LGTM, the issue I've mentioned is fixed.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -345,6 +354,20 @@ private void OnValidateCommand(ValidateCommandEvent evt)
...
 
+        private void OnPointerDown(PointerDownEvent evt)
+        {
+            // Allow right clicks to select an item before we bring up the matching context menu.
+            if (evt.button == (int)MouseButton.RightMouse && evt.clickCount == 1)
+            {
+                // Look upwards to the immediate child of the scroll view, so we know what Index to use
+                var element = evt.target as VisualElement;
+                while (element.name != "unity-tree-view__item")
+                    element = element.parent;
```
ritamerkl, line 443: how about use panel.Pick(evt.mousePosition)? that could save a while loop here

ritamerkl (approved):
Looks good to me, I saw the [thread](https://unity.slack.com/archives/C3414V4UV/p1693320293138709?thread_ts=1693306035.515169&cid=C3414V4UV) about the selecting on right click in devs-uitoolkit. I just left one last comment on the while loop. I didn't test it, ok for me if @Pauliusd01 approved it. The [issues](https://github.com/Unity-Technologies/InputSystem/pull/1806#discussion_r1444683114) Håkan mentioned are to fix before though!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
graham-huws-unity: I tried that - unfortunately it just returns the same element as evt.target 😞
I think the issue is that in the scrollview, the hierarchy _should_ have everything descending from InputActionsTreeViewItem, which would mean I could get ancestor of type...

graham-huws:
@Pauliusd01 Could you have a check with the "add binding" button to make sure it's behaving - I added what should be a small improvement in 80d5916cd52fd5c3acccb6bfb7f0944c504aeebc.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ritamerkl: right, that would be a better option. If this is not working we should at least introduce a safety break in this loop, if no element is found. Something like element.parent !=null

Pauliusd01 (changes requested):
https://github.com/Unity-Technologies/InputSystem/assets/54306142/44d56755-7cf7-4435-a069-1422f5e2986f

Add binding on rightclick is broken. Steps are: open 2 input actions windows -> right click one of them and click add binding -> if the binding is added correctly then repeat the previous step in the other window -> now the binding should be added incorrectly

Also, sometimes I get this error thrown but it does not seem to be consistent, might be related to the number of actions in the asset:
```
Cannot add binding without an action and action map selected
UnityEditor.GenericMenu:CatchMenu (object,string[],int) (at D:/Gitrepo/unity/Editor/Mono/GUI/GenericMenu.cs:127)
```

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
graham-huws-unity: I _think_ in theory that shouldn't be possible, but sensible idea - added in b9f698fd36f6995569cb232ba3e6d26e0d1820bf.

graham-huws:
> Add binding on rightclick is broken. 

This seems to be broken on develop for me, going between windows - it gives "Cannot add binding without an action and action map selected" every time.

Pauliusd01 (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ekcoh: It think we should file a feature request for this. This might be a workaround now but it will for sure be needed  down the road for Project Hierarchy and then it would be nice to at least have a ticket to track/monitor.

ekcoh (approved):
If the workaround works then we should get it in. However I would advice we file a feature request on the UI team for this functionality since I would like to avoid having custom logic to handle such a common use-case. if there was a thread for this @graham-huws DM me and we can sort it out there.

For testing I would advice testing on multiple OS, e.g. Win + macOS since these behavior might be inconsistent (based on related bugs)

PR #1807, opened 2023-12-15
CHANGE: Update Input for UI provider to use project-wide input actions

jfreire-unity:
### Description

[ISX-1528
](https://jira.unity3d.com/browse/ISX-1528)
### Changes made

- Change InputSystemProvider for Input For UI module to work with the Project Wide Actions instead of DefaultInputActions. This makes UI TK use project-wide action bindings that can be modified by users.

- Added some warnings when the asset is saved with changes to the "UI" action maps. By doing this only in save, it makes sure that other parts of the code are decoupled from Project Wide Actions. 

❓
- ~Should there be a way to reset specific action maps? Besides having just "Reset", also have "Reset UI map" and "Reset Player map" as well?~
- ~Some might prefer to have an Editor UI visual warning (e.g. like a warning icon) when the UI action map is modified. But that would take more time than expected, as it involves UI TK UI changes. Please let me know if there should be any extra effort to change the UI~.

### Notes

ℹ️  The project to be used for QA is specified in the [JIRA ticket](https://jira.unity3d.com/browse/ISX-1528).

I'll add doc changes in another PR after talking with Ben Pitt.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -11,7 +11,7 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased]
 
 ### Changed
-- UI toolkit now uses the "UI" action map of project-wide actions as their default input actions. Removing bindings or renaming actions will break UI input for UI toolkit.
+- From 2023.2 forward: UI toolkit now uses the "UI" action map of project-wide actions as their default input actions. Previously, the actions were hardcoded and were based on `DefaultInputActions` asset which didn't allow user changes. Also, removing bindings or renaming the 'UI' action map of project wide actions will break UI input for UI toolkit.
 
 ### Fixed
 - Fixed missing confirmation popup when deleting a control scheme.
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -10,6 +10,9 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased]
 
+### Changed
+- UI toolkit now uses the "UI" action map of project-wide actions as their default input actions. Removing bindings or renaming actions will break UI input for UI toolkit.
+
 ### Fixed
 - Fixed missing confirmation popup when deleting a control scheme.
 - Fixed support for menu bar/customisable keyboard shortcuts used when interacting with Actions and Action Maps.
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
@@ -140,8 +140,8 @@ namespace UnityEngine.InputSystem.Editor
                     // "UI" actions have been modified.
                     if (uiMap.FindAction(action.name) == null)
                     {
-                        Debug.LogWarning($"The UI action \"{action.name}\" name has been modified.\r\n " +
-                            $"This will break the UI input at runtime. Please make sure the action name with \"{action.name}\" exists.");
+                        Debug.LogWarning($"The UI action '{action.name}' name has been modified.\r\n" +
+                            $"This will break the UI input at runtime. Please make sure the action name with '{action.name}' exists.");
                     }
                 }
             }
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
@@ -114,6 +114,7 @@ namespace UnityEngine.InputSystem.Editor
             }
         }
 
+        #if UNITY_2023_2_OR_NEWER
         /// <summary>
         /// Checks if the default UI action map has been modified or removed, to let the user know if their changes will
         /// break the UI input at runtime, when using the UI Toolkit.
@@ -141,11 +142,12 @@ namespace UnityEngine.InputSystem.Editor
                     {
                         Debug.LogWarning($"The UI action \"{action.name}\" name has been modified.\r\n " +
                             $"This will break the UI input at runtime. Please make sure the action name with \"{action.name}\" exists.");
-                        return;
                     }
                 }
             }
         }
+
+        #endif
 
         /// <summary>
         /// Updates the input action references in the asset by updating names, removing dangling references
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
@@ -95,7 +95,7 @@ namespace UnityEngine.InputSystem.Editor
 
         private static InputActionMap GetDefaultUIActionMap()
         {
-            var json = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath));
+            var json = File.ReadAllText(FileUtil.GetPhysicalPath(s_DefaultAssetPath));
             var actionMaps = InputActionMap.FromJson(json);
             return actionMaps[actionMaps.IndexOf(x => x.name == "UI")];
         }
@@ -136,7 +136,7 @@ namespace UnityEngine.InputSystem.Editor
                     if (uiMap.FindAction(action.name) == null)
                     {
                         Debug.LogWarning($"The UI action \"{action.name}\" name has been modified.\r\n " +
-                            $"This will break the UI input at runtime. Please make sure the action name with {action.name} exists.");
+                            $"This will break the UI input at runtime. Please make sure the action name with \"{action.name}\" exists.");
                         return;
                     }
                 }
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
@@ -16,7 +16,6 @@ namespace UnityEngine.InputSystem.Editor
 
         static string s_DefaultAssetPath = kDefaultAssetPath;
         static string s_AssetPath = kAssetPath;
-        static InputActionMap s_DefaultUIMap;
 
 #if UNITY_INCLUDE_TESTS
         internal static void SetAssetPaths(string defaultAssetPath, string assetPath)
@@ -60,8 +59,6 @@ namespace UnityEngine.InputSystem.Editor
             asset.LoadFromJson(json);
             asset.name = kAssetName;
 
-            GetDefaultUIActionMap();
-
             AssetDatabase.AddObjectToAsset(asset, s_AssetPath);
 
             // Make sure all the elements in the asset have GUIDs and that they are indeed unique.
@@ -95,12 +92,12 @@ namespace UnityEngine.InputSystem.Editor
 
             return asset;
         }
-        
-        private static void GetDefaultUIActionMap()
+
+        private static InputActionMap GetDefaultUIActionMap()
         {
             var json = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath));
             var actionMaps = InputActionMap.FromJson(json);
-            s_DefaultUIMap = actionMaps.FirstOrDefault(m => m.name == "UI");
+            return actionMaps[actionMaps.IndexOf(x => x.name == "UI")];
         }
 
         private static void CreateInputActionReferences(InputActionAsset asset)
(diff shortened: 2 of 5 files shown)


ritamerkl (approved):
LGTM

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -107,6 +117,36 @@ private static void CreateInputActionReferences(InputActionAsset asset)
...
+            if (asset != null)
+            {
+                if (s_DefaultUIMap == null)
+                    GetDefaultUIActionMap();
+                
+                var uiMap = asset.actionMaps.FirstOrDefault(m => m.name == "UI");
+                // "UI" action map has been removed or renamed.
+                if (uiMap == null)
+                {
+                    Debug.LogWarning("The default UI action map does not exist.\r\n " +
```
ekcoh: Good that there is some feedback to the user about this problem, but the requirement isn't really to have a "default UI action map" right, the requirement is to have "an action map named 'UI' " right?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -107,6 +117,36 @@ private static void CreateInputActionReferences(InputActionAsset asset)
...
+                        "This will break the UI input at runtime. Please revert the changes to restore the default UI action map.");
+                    return;
+                }
+
+                // "UI" action map has been modified.
+                foreach (var action in s_DefaultUIMap)
+                {
+                    if(uiMap.FindAction(action.name) == null)
+                    {
+                        Debug.LogWarning($"The UI action \"{action.name}\" has been modified.\r\n " +
```
ekcoh: Similar to previous comment on warning, instead of "modified" shouldn't this warning instead state that it "is missing" or "do not exist." The key isn't to restore or revert but rather to provide an action with the expected name/type?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -92,6 +95,13 @@ internal static InputActionAsset CreateNewActionAsset()
 
             return asset;
         }
+        
+        private static void GetDefaultUIActionMap()
+        {
+            var json = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath));
```
ekcoh: Given that the UI action map is cached into s_DefaultUIMap, wouldn't it make sense to move the check and early return into this function instead, e.g. have

```
if (s_DefaultUIMap != null) return s_DefaultUIMap;
var asset = GetOrCreate();
var json = ....
...
return s_DefaultUIMap;
```
inside this function? Another question would be, do we really need to cache it? It would occupy memory not really needed that often, so would suggest we could skip it? I think it would be fine loading it on demand when doing CheckForDefaultUIActionMapChanges()?

ekcoh: If we skip caching we can also get rid of the GetDefaultUIActionMap() function and move that code into CheckForDefaultUIActionMapChanges()

ekcoh (changes requested):
Looks good in general. I think we can skip the caching or was it taking a long time that drived the decision to cache in s_DefaultUIMap?

I also believe it would be good to add a test to cover CheckFirDefaultUIActionMapChanges.

ekcoh (commented):
Another one, CHANGELOG.md entry seems to be missing.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -107,6 +117,36 @@ private static void CreateInputActionReferences(InputActionAsset asset)
...
 
+        internal static void CheckForDefaultUIActionMapChanges()
+        {
+            var asset = GetOrCreate();
+            if (asset != null)
+            {
+                if (s_DefaultUIMap == null)
+                    GetDefaultUIActionMap();
+                
+                var uiMap = asset.actionMaps.FirstOrDefault(m => m.name == "UI");
```
jamesmcgill: LINQ usage.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -642,16 +641,12 @@ void UnregisterActions(Configuration cfg)
             // The Next/Previous action is not part of the input actions asset
             UnregisterNextPreviousAction();
 
-#if UNITY_EDITOR
-            UnityEngine.Object.DestroyImmediate(m_InputActionAsset);
-#else
-            UnityEngine.Object.Destroy(m_InputActionAsset);
-#endif
+            UnityEngine.Object.Destroy(m_InputActionAsset); // TODO check if this is ok
```
jamesmcgill: This is reverting a change in `develop` and re-adding the TODO.
I assume it's just down to a git merge issue.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -59,6 +60,8 @@ internal static InputActionAsset CreateNewActionAsset()
             asset.LoadFromJson(json);
             asset.name = kAssetName;
 
+            GetDefaultUIActionMap();
```
jamesmcgill: I would suggest passing in the `json` to this function rather than do another file read inside this function.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jamesmcgill: This line of of code should match the existing file read in `CreateNewActionAsset()` which is
`var json = File.ReadAllText(FileUtil.GetPhysicalPath(s_DefaultAssetPath));`

Additionally, as also mentioned in another comment, I think this line should be hoisted outside of this function and pass the json, so that there doesn't need to be 2 calls to this in `CreateNewActionAsset()`

`CheckForDefaultUIActionMapChanges()` will need this call added for that.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -92,6 +95,13 @@ internal static InputActionAsset CreateNewActionAsset()
 
             return asset;
         }
+        
+        private static void GetDefaultUIActionMap()
+        {
+            var json = File.ReadAllText(Path.Combine(Environment.CurrentDirectory, s_DefaultAssetPath));
+            var actionMaps = InputActionMap.FromJson(json);
+            s_DefaultUIMap = actionMaps.FirstOrDefault(m => m.name == "UI");
```
jamesmcgill: LINQ

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jamesmcgill: Also, `CreateNewActionAsset()` only gets called once when opening a project for the first time, so the cached value would mostly not exists for existing projects.

jamesmcgill:
I think for the first pass, it is enough that InputForUI just uses `InputSystem.actions`.

UI improvements like blocking editing the UI action names, warnings, or resetUI map, would take more time.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: I'll leave them for now since it only impacts the Editor and this class already has LINQ references.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: Now I'm not caching the template/default UI action map as suggested by @ekcoh. Thanks, makes more sense now.
This means that for the JSON, the alternative is either to cache it, to avoid more than 1 reading in the Editor, or then we need to read the file close to when we need it. Is there any other solution you're seeing @jamesmcgill ?

jamesmcgill: Will still need to be something like `var json = File.ReadAllText(FileUtil.GetPhysicalPath(path));` The `CurrentDirectory` way didn't work in some projects.

jamesmcgill: > Now I'm not caching the template/default UI action map as suggested by @ekcoh. Thanks, makes more sense now. This means that for the JSON, the alternative is either to cache it, to avoid more than 1 reading in the Editor, or then we need to read the file close to when we need it. Is there any other solution you're seeing @jamesmcgill ?

After your changes there is no longer a double file read, so not an issue anymore.

jfreire-unity:
@Pauliusd01 regarding testing, you can try to play around with removing the UI action map, UI actions, or changing the bindings like we discussed in the daily sync.
image


You should still see the events in the right column  the Visualizer test project scene. And should get appropriate warning and error messages. There are already some warning messages once the asset is saved and you rename the UI action map and its actions.

jfreire-unity:
@Pauliusd01 I put the wrong image before, I updated it now

Pauliusd01 (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: @ekcoh I believe I made the changes you requested. Have a look :)

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased]
 
+### Changed
+- UI toolkit now uses the "UI" action map of project-wide actions as their default input actions. Removing bindings or renaming actions will break UI input for UI toolkit.
```
ekcoh: I think this is expected given the solution we have previously discussed but its unfortunate since the action is basically required (driven by the implementation). In the long run I believe we should consider design changes that would allow the integration to own the action while the editor would allow reconfiguring it.

ekcoh: For the CHANGED message I believe it could be more clear to users if you added the delta to the description, e.g. "UI toolkit now uses the "UI" action map of project-wide actions as their default input actions instead of..." basically highlighting the switch from hard-coded to customisable.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -107,6 +114,41 @@ private static void CreateInputActionReferences(InputActionAsset asset)
...
+                }
+                var uiMap = asset.m_ActionMaps[uiMapIndex];
+                foreach (var action in defaultUIActionMap.actions)
+                {
+                    // "UI" actions have been modified.
+                    if (uiMap.FindAction(action.name) == null)
+                    {
+                        Debug.LogWarning($"The UI action \"{action.name}\" name has been modified.\r\n " +
+                            $"This will break the UI input at runtime. Please make sure the action name with \"{action.name}\" exists.");
+                    }
```
ekcoh, line 145: Shouldn't `{action.name}` here be within single quotes for clarity and consistency with above message?

ekcoh (approved):
In general looks good to me, happy to see some test coverage added. Two minor comments that I believe can improve the PR slightly.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
jfreire-unity: Good catch 👍  will change

PR #1812, opened 2024-01-04
FIX: Correctly select new ActionTree item after delete (ISX-1680)

timkeo:
### Description

Fix for [ISX-1680](https://jira.unity3d.com/browse/ISX-1680) - Deleting a Binding will correctly select a new item in the tree

### Changes made

Reworks the the logic to select a new Action/Binding after a delete operation by moving the functionality from Command.cs into ActionTreeView.cs.

Searches the ActionTreeView data to find the tree node that's about to be deleted and identify the siblings/parent node in the tree to select next and dispatches a separate Select command.

### Notes

Unlike Actions, the Binding IDs do not correspond to the indexes within the ActionTreeView and there's no way to determine a given Binding's siblings solely from the Action state. We must search through the visual tree to identify the new item to select (after delete) which is best executed within ActionTreeView.

For consistency (and safety should Action IDs change), this approach is applied both to Actions and Bindings.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -58,6 +58,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed case ISX-1436 (UI TK Input Action Asset Editor - Error deleting Bindings with DeleteKey on Windows).
 - Fixed issue with UI Toolkit based Input Action Editor not restoring it's selected items after Domain Reload.
 - Fixed the [`GetHapticCapabilitiesCommand`](xref:UnityEngine.InputSystem.XR.Haptics.GetHapticCapabilitiesCommand) always failing to execute due to a mismatch in the size in bytes of the payload and the size expected by XR devices. Changed [`HapticCapabilities`](xref:UnityEngine.InputSystem.XR.Haptics.HapticCapabilities) to include all properties returned by the XR input subsystem. This makes Input System match the functionality provided by the [XR](https://docs.unity3d.com/Manual/com.unity.modules.xr.html) module's [`InputDevice.TryGetHapticCapabilities`](https://docs.unity3d.com/ScriptReference/XR.InputDevice.TryGetHapticCapabilities.html) and [`HapticCapabilities`](https://docs.unity3d.com/ScriptReference/XR.HapticCapabilities.html).
+- Fixed case ISX-1680 (UITK Input Action Asset Editor - Deleting a binding selects a seemingly random binding)
 
 ## [1.8.0-pre.1] - 2023-09-04
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -58,7 +58,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed case ISX-1436 (UI TK Input Action Asset Editor - Error deleting Bindings with DeleteKey on Windows).
 - Fixed issue with UI Toolkit based Input Action Editor not restoring it's selected items after Domain Reload.
 - Fixed the [`GetHapticCapabilitiesCommand`](xref:UnityEngine.InputSystem.XR.Haptics.GetHapticCapabilitiesCommand) always failing to execute due to a mismatch in the size in bytes of the payload and the size expected by XR devices. Changed [`HapticCapabilities`](xref:UnityEngine.InputSystem.XR.Haptics.HapticCapabilities) to include all properties returned by the XR input subsystem. This makes Input System match the functionality provided by the [XR](https://docs.unity3d.com/Manual/com.unity.modules.xr.html) module's [`InputDevice.TryGetHapticCapabilities`](https://docs.unity3d.com/ScriptReference/XR.InputDevice.TryGetHapticCapabilities.html) and [`HapticCapabilities`](https://docs.unity3d.com/ScriptReference/XR.HapticCapabilities.html).
-- Fixed case ISX-1680 (UITK Input Action Asset Editor - Deleting a binding selects a seemingly random binding)
+- Fixed issue where deleting a binding in the Input Action Editor would usually result in an unexpected item being selected next.
 
 ## [1.8.0-pre.1] - 2023-09-04
 
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
@@ -157,19 +157,10 @@ namespace UnityEngine.InputSystem.Editor
                 var actionID = InputActionSerializationHelpers.GetId(action);
                 InputActionSerializationHelpers.DeleteActionAndBindings(actionMap, actionID);
                 state.serializedObject.ApplyModifiedProperties();
-                if (state.selectedActionIndex >= actionIndex)
-                    return SelectPrevAction(state, actionMap);
-                return state.SelectAction(state.selectedActionIndex);
-            };
-        }
 
-        private static InputActionsEditorState SelectPrevAction(InputActionsEditorState state, SerializedProperty actionMap)
-        {
-            var count = Selectors.GetActionCount(actionMap);
-            int index = -1;
-            if (count != null && count.Value > 0)
-                index = Math.Max(state.selectedActionIndex - 1, 0);
-            return state.SelectAction(index);
+                // ActionsTreeView will dispatch a separate command to select the previous Action
+                return state.With(selectedActionIndex: -1);
+            };
         }
 
         public static Command DeleteBinding(int actionMapIndex, int bindingIndex)
@@ -180,19 +171,10 @@ namespace UnityEngine.InputSystem.Editor
                 var binding = Selectors.GetCompositeOrBindingInMap(actionMap, bindingIndex).wrappedProperty;
                 InputActionSerializationHelpers.DeleteBinding(binding, actionMap);
                 state.serializedObject.ApplyModifiedProperties();
-                if (state.selectedBindingIndex >= bindingIndex)
-                    return SelectPrevBinding(state, actionMap);
-                return state.SelectBinding(state.selectedBindingIndex);
-            };
-        }
 
-        private static InputActionsEditorState SelectPrevBinding(InputActionsEditorState state, SerializedProperty actionMap)
-        {
-            var count = Selectors.GetBindingCount(actionMap);
-            var index = -1;
-            if (count != null && count.Value > 0)
-                index = Math.Max(state.selectedBindingIndex - 1, 0);
-            return state.SelectBinding(index);
+                // ActionsTreeView will dispatch a separate command to select the previous Binding
+                return state.With(selectedBindingIndex: -1);
+            };
         }
 
         public static Command ExpandCompositeBinding(SerializedInputBinding binding)
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
@@ -159,7 +159,7 @@ namespace UnityEngine.InputSystem.Editor
                 state.serializedObject.ApplyModifiedProperties();
 
                 // ActionsTreeView will dispatch a separate command to select the previous Action
-                return state.With(selectedActionIndex: -1);
+                return state;
             };
         }
 
@@ -173,7 +173,7 @@ namespace UnityEngine.InputSystem.Editor
                 state.serializedObject.ApplyModifiedProperties();
 
                 // ActionsTreeView will dispatch a separate command to select the previous Binding
-                return state.With(selectedBindingIndex: -1);
+                return state;
             };
         }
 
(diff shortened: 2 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -233,9 +233,24 @@ internal void AddComposite(string actionName, string compositeType)
...
                 Dispatch(Commands.DeleteBinding(data.actionMapIndex, data.bindingIndex));
+
+                if (bindingIndexToSelect >= 0)
+                    Dispatch(Commands.SelectBinding(bindingIndexToSelect));
+                else
+                    Dispatch(Commands.SelectAction(parentActionName));
+            }
+
+            // Deleting an item sometimes causes the UI Panel to lose focus; make sure we keep it
+            m_ActionsTreeView.Focus();
```
ritamerkl, line 253: Interesting, I guess this happens if a binding is deleted and no other item is selected directrly after -> then the TreeView looses focus because nothing is selected. I think it's okay to just manual retrieve the Focus here. 

If I am right with my guess another way to solve this would be combining the commands (deleting and selecting) to avoid a refresh of the UI in between where there is no selection. (in my view not urgently necessary)

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
```diff
@@ -180,19 +171,10 @@ public static Command DeleteBinding(int actionMapIndex, int bindingIndex)
...
 
-        private static InputActionsEditorState SelectPrevBinding(InputActionsEditorState state, SerializedProperty actionMap)
-        {
-            var count = Selectors.GetBindingCount(actionMap);
-            var index = -1;
-            if (count != null && count.Value > 0)
-                index = Math.Max(state.selectedBindingIndex - 1, 0);
-            return state.SelectBinding(index);
+                // ActionsTreeView will dispatch a separate command to select the previous Binding
+                return state.With(selectedBindingIndex: -1);
```
ritamerkl: How about just returning the state without changing the binding index? It will be overwritten by the ActionsTreeView anyway.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
```diff
@@ -157,19 +157,10 @@ public static Command DeleteAction(int actionMapIndex, string actionName)
...
 
-        private static InputActionsEditorState SelectPrevAction(InputActionsEditorState state, SerializedProperty actionMap)
-        {
-            var count = Selectors.GetActionCount(actionMap);
-            int index = -1;
-            if (count != null && count.Value > 0)
-                index = Math.Max(state.selectedActionIndex - 1, 0);
-            return state.SelectAction(index);
+                // ActionsTreeView will dispatch a separate command to select the previous Action
+                return state.With(selectedActionIndex: -1);
```
ritamerkl: How about just returning the state without changing the binding index? It will be overwritten by the ActionsTreeView anyway.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -293,6 +308,87 @@ private void OnValidateCommand(ValidateCommandEvent evt)
...
+                int index = treeView.FindIndex(item => item.data.name == actionName);
+                if (index > 0)
+                    index--;
+                else
+                    index++;
+
+                return treeView[index].data.name;
+            }
+
+            return string.Empty;
```
ritamerkl, line 330: what happens if the action was not found/the treeview is empty and the string is empty? I think in this case nothing should get selected?

ritamerkl (commented):
Good change. Just some thoughts to consider

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed case ISX-1436 (UI TK Input Action Asset Editor - Error deleting Bindings with DeleteKey on Windows).
 - Fixed issue with UI Toolkit based Input Action Editor not restoring it's selected items after Domain Reload.
 - Fixed the [`GetHapticCapabilitiesCommand`](xref:UnityEngine.InputSystem.XR.Haptics.GetHapticCapabilitiesCommand) always failing to execute due to a mismatch in the size in bytes of the payload and the size expected by XR devices. Changed [`HapticCapabilities`](xref:UnityEngine.InputSystem.XR.Haptics.HapticCapabilities) to include all properties returned by the XR input subsystem. This makes Input System match the functionality provided by the [XR](https://docs.unity3d.com/Manual/com.unity.modules.xr.html) module's [`InputDevice.TryGetHapticCapabilities`](https://docs.unity3d.com/ScriptReference/XR.InputDevice.TryGetHapticCapabilities.html) and [`HapticCapabilities`](https://docs.unity3d.com/ScriptReference/XR.HapticCapabilities.html).
+- Fixed case ISX-1680 (UITK Input Action Asset Editor - Deleting a binding selects a seemingly random binding)
```
ekcoh: We typically try to avoid referencing internal ticket IDs since they make little sense to users since not publicly available while the release notes and the repo is. Note that ISXB tickets are referenced via public issue tracker though. I would rather suggest explaining the fixed scenario rather than reusing the ticket title, e.g.
"Fixed an issue where deleting a binding in the Input Action Editor while some binding items where not expanded would result in an unexpected selection change that may expand unrelated bindings in the tree view."

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -293,6 +308,87 @@ private void OnValidateCommand(ValidateCommandEvent evt)
...
+                    continue;
+
+                if (FindBindingOrComponentTreeViewParent(action, data.bindingIndex, out var parentNode, out int childIndex))
+                {
+                    parentActionName = action.data.name;
+                    if (parentNode.children.Count() > 1)
+                    {
+                        int prevIndex = Math.Max(childIndex - 1, 0);
+                        var node = parentNode.children.ElementAt(prevIndex);
+                        retVal = node.data.bindingIndex;
```
ekcoh, line 359: Shouldn't we return `retVal `directly here? Or is the loop guaranteed to break indirectly by `if (!action.hasChildren)` above when this condition is true?

ekcoh (approved):
Looks good, added some minor comments regarding CHANGELOG and abort condition for loop. Happy to see the "algorithm/psuedo comment" in GetPreviousBindingIndexFromViewTree since it makes it easier to read the code by first understanding selected approach.

Pauliusd01:
Shouldn't composite bindings also have this behaviour? (Currently regular bindings select the binding above once deleted, composite bindings select nothing) This is also present in Develop, so It's not a regression or anything, just pointing out the differences

https://github.com/Unity-Technologies/InputSystem/assets/54306142/7a806b2d-c067-4f16-affc-0907cb717e0d

Pauliusd01 (approved):
LGTM. Checked deletion with the shortcut/context menu on different actions and bindings. Did not notice anything unexpected besides the mentioned inconsistency between regular bindings and composite bindings.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
timkeo: I wanted to avoid combining Delete/Select since I think they should be individual actions, but that was before discovering the UI refreshed between them. Maybe instead there's a way to "suspend" UI updates when executing multiple commands?

I do agree it's a minor issue and something can revisit later.

Packages/com.unity.inputsystem/CHANGELOG.md
timkeo: Updated CHANGELOG entry with fbacf6f958456ecc949d89a22634f77574cfc607.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
timkeo: If `actionName` isn't found then `FindIndex()` returns -1 which gets incremented to 0 in the `else` branch causing the first Action in the list to be selected. Typically I don't like taking this kind of dependency on an API and prefer to write more "defensive" code, but in this case `List<T>.FindIndex()` is a well established .NET API that won't change.

Still, I added a comment to clarify this reasoning with fbacf6f958456ecc949d89a22634f77574cfc607.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
timkeo: Yes I'm missing a `break` here; thanks for the catch!
Fixed with fbacf6f958456ecc949d89a22634f77574cfc607.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
timkeo: Fixed with fbacf6f958456ecc949d89a22634f77574cfc607.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
timkeo: I was thinking the same thing but unsure about possible side-effects, and so thanks for suggesting this.
Fixed with fbacf6f958456ecc949d89a22634f77574cfc607.

timkeo:
I opened [a bug ticket](https://jira.unity3d.com/browse/ISX-1814) for the issue with Composite Bindings @Pauliusd01 mentioned.
I discovered it in my own testing but forgot to open the bug.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ritamerkl: I think you can call the select command inside the delete command, what would mean we still have a delete and select command.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ritamerkl: And if the action list is empty? (Edge case, but still possible)

timkeo: The outer condition `if (treeView.Count > 1)` handles this case, i.e. if the Action list is empty this will return 0 and we jump to the default return.
But even if we didn't check, `FindIndex()` would still return -1 if the list was empty (didn't match actionName).

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
timkeo: I played around with this a bit more to try and "combine" the Select/Delete operations and avoid needing to call `Focus()` on the TreeView. My idea was to create a "CommandList" implementation, that executed multiple commands (in sequence) and only fire the `StateChanged` event once at the end. [This commit](https://github.com/Unity-Technologies/InputSystem/commit/f7e317d1f10cccbff13dbe9d9616e909127e1521) holds my quick, dirty prototype.

Unfortunately, it didn't work; deleting multiple Action/Bindings in a row still causes the TreeView to (somehow) lose focus. I also found this issue exists in `develop` without my changes, and so It appears to be a deeper issue (possibly a bug in UITK).

So I'm inclined to keep the `Focus()` call to work-around the problem.

ritamerkl: Interesting! Thank you for checking, I guess we are good to keep the Focus() work-around for now.

ritamerkl (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
timkeo: I opened [ticket ISX-1819](https://jira.unity3d.com/browse/ISX-1819) to follow up later.

PR #1816, opened 2024-01-16
Basic DS Edge support [ISX-1367]

stefanunity:
### Description

Dual Sense Edge controller is now matched the same way the normal Dual Sense controller works on Win, OSX, WSA, Editor.

### Changes made

DS Edge is mapped as DS controller. DS buttons and sticks work, but none of the extra functions of from the normal DS nor the extra bits from Edge.

Tested on Win+OSX editor + standalone.

### Notes

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


lewish-unity (approved):

lyndon-unity (approved):
Looks good - thanks Stefan

PR #1819, opened 2024-01-18
FIX: Don't show action properties view when there's no actions (ISX-1497).

graham-huws:
### Description

As per title, fixes [ISX-1497](https://jira.unity3d.com/browse/ISX-1497) to match pre-UITK behaviour.

### Changes made

Hide the action properties view based on number of actions.

image
image

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Pauliusd01 (approved):

lyndon-unity (approved):
Simple change and generally happy to approve but I think I'd prefer to have the Action Properties header still present even though the contents are hidden.

@Billreyn is likely to have an opinion on that aspect

Billreyn:
Interesting. AFAIK the only case where this behaviour seemed to be used in the IMGUI version is when there exists no actions?

I'd probably agree with @lyndon-unity about the Properties title, since the pane is still there, just with no information.

graham-huws:
Thanks for comments - Now updated!

Pauliusd01 (approved):
It does look nicer with the header visible, good call

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -205,7 +205,7 @@ public override void RedrawUI(ViewState viewState)
             addActionButton.SetEnabled(viewState.actionMapCount > 0);
 
             // Don't want to show action properties if there's no actions.
-            m_Root.Q<VisualElement>("properties-container").visible = m_ActionsTreeView.GetTreeCount() > 0;
+            m_Root.Q<VisualElement>("properties-container").Children().OfType<ScrollView>().First().visible = m_ActionsTreeView.GetTreeCount() > 0;
```
lyndon-unity: Hmm, this is 'ok' if it has the desired result, but I think I would prefer to see the sub area having a name in the UXML file so that you could do something like 

m_Root.Q<VisualElement>("properties-container-without-header").visible = m_ActionsTreeView.GetTreeCount() > 0;

lyndon-unity: I.e. adding here:
https://github.com/Unity-Technologies/InputSystem/blob/dac00ca42d47249b5e711360a326ada6cf6bbed2/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Resources/InputActionsEditor.uxml#L45

ekcoh: Agree, use an id/name to access and check for existence in code.

ekcoh (approved):
LGTM if the name access suggested by Lyndon is fixed, hence approving this now but that should be addressed. Addressing by type is brittle if design changes later.

ekcoh (approved):
LGTM

ekcoh:
Just think the CHANGELOG conflict needs to be resolved and then this should be ready to be merged.

PR #1820, opened 2024-01-23
Add main menu for Core Sample Scenes and QA Tests, Minor RebindingActionUI script changes - ISX-1776

Pauliusd01:
### Description

Minor changes to our main repo scenes for better testing quality of life as well as user requested changes to the Rebinding UI Sample. Example of the scene loader: 
https://github.com/Unity-Technologies/InputSystem/assets/54306142/029a3580-9b05-431a-ab8a-b018a041f891
Works, or at least should work both with Keyboard&Mouse and Controller cursor. (Esc and Start button bring up the menu to return to the loader)

### Changes made

- Adds user suggested changes to the RebindingUI sample (https://forum.unity.com/threads/rebinding-ui-sample-bugs-fixes.1307430/)
- Adds a main menu scene with a scene loader for easier access to the core platforms so you no longer need to rebuild the player to change/reset scenes
- Moved DocExamples scenes by Ben to QA folder as they are incomplete and not publicly accessible samples
- Some scenes had their event system objects updated to use the IS version. Seemed fine as the ones I changed did not mention to be Input Manager related.

### Notes

I've also noticed that there might be a bug with ProjectWideAction references when loading/unloading scenes: 
https://github.com/Unity-Technologies/InputSystem/assets/54306142/c39eda98-5b11-4a4b-99cc-dd002ae00085
Would like to report this separately for the developers to investigate after these changes merge.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/AddBuildScenes.cs
```diff
@@ -0,0 +1,31 @@
...
+public class AddBuildScenes : MonoBehaviour
+{
+
+    // Start is called before the first frame update
+    void Start()
+    {
+        List<SceneAsset> m_SceneAssets = new List<SceneAsset>();
+        // Find valid Scene paths and make a list of EditorBuildSettingsScene
+        List<EditorBuildSettingsScene> editorBuildSettingsScenes = new List<EditorBuildSettingsScene>();
+        foreach (var sceneAsset in m_SceneAssets)
```
lyndon-unity: Looks like this doesn't do anything - m_SceneAssets will be an empty list as its only just created.

Assets/QA/Tests/Core Platform Menu/AddScenesToBuild.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        return false;
+    }
+
+    private static void OnPlayModeStateChanged(PlayModeStateChange state)
+    {
+        if (state == PlayModeStateChange.ExitingPlayMode)
+        {
+            // Save the build settings before entering play mode
```
lyndon-unity: Isn't this the wrong way round. Sounds like its exiting play mode and we wouldn't want to be saving during that phase.

Assets/Samples/RebindingUI/RebindActionUI.cs
```diff
@@ -268,14 +268,16 @@ void CleanUp()
                     operation =>
                     {
                         m_RebindStopEvent?.Invoke(this, operation);
-                        m_RebindOverlay?.SetActive(false);
```
lyndon-unity, line 271: Why this change - its the same thing isn't it.

lyndon-unity (changes requested):
Added some comments as I'm not sure some of the changes are valid

Assets/Samples/RebindingUI/RebindActionUI.cs
Pauliusd01: Changes suggested by a user: https://forum.unity.com/threads/rebinding-ui-sample-bugs-fixes.1307430/

"The Elvis operator doesn't work with UnityEngine.Object types in some circumstances. It seems to be a long standing issue, see [this thread](https://forum.unity.com/threads/elvis-operator-null-coalescing-operator.433767/) for details."

Assets/AddBuildScenes.cs
Pauliusd01: Ah, this one wasn't supposed to be here at all. First attempt at it

Assets/QA/Tests/Core Platform Menu/AddScenesToBuild.cs
Pauliusd01: Leaving play mode in Editor would always clear my added build scenes so the next time I enter play mode I couldn't use the scene loader. Let me see if the issue is still there and what I can do

Assets/QA/Input_Test/IMETest.unity
lyndon-unity, line 1: There are a number of changed .unity files which seem just related to upgrading to a newer unity version. I'm not sure this need to be in the PR and may prevent running tests on older Unity versions. Can these be removed. 

Likewise some unneccessary .meta file changes

lyndon-unity (changes requested):
General concept of adding ui menu and moving some manual tests over to the QA area seems reasonable - comment about upgraded .unity files that seems risky if changes not needed.

Assets/QA/Input_Test/IMETest.unity
Pauliusd01: They've actually been modified intentionally by this:
>Some scenes had their event system objects updated to use the IS version. Seemed fine as the ones I changed did not mention to be Input Manager related.

If I don't make these changes those scenes will throw an error on load.

>Likewise some unneccessary .meta file changes

Can you point out which ones? I don't see any loose .meta files. Unless you mean the changes to the scenes again and I do agree that they do come with more than just that EventSystem component changes. I think I could ditch those although doesn't seem obvious which parts to keep.

ekcoh: I would advise to make any of those upgrades to .unity files and meta files using Unity 2019 since that is the min version required by Input System 1.x and hence should verify that there are no issues with 2019 due to any files being in a newer format (avoid downgrade scenarios). I believe comment by @lyndon-unity regarding .meta file changes is any of the diffs on the PR relating to already existing .meta files being modified but I couldn't spot any such changes in the diff so I let @lyndon-unity provide additional detail regarding that.

ekcoh (commented):
Did you confirm with @duckets that the move of DocExamples is purely structural and do not have any impact on doc generation or publishing?

Assets/QA/Input_Test/IMETest.unity
Pauliusd01: That is a good suggestion, I will redo those files using 2019.4

Pauliusd01:
> Did you confirm with @duckets that the move of DocExamples is purely structural and do not have any impact on doc generation or publishing?

Yes, he even suggested deleting them but I thought It's better to keep and modify as testing scenes for internal use

Pauliusd01:
@lyndon-unity @ekcoh I've reverted my event system changes to the old scenes and redone them again in 2019.4. I hope that feels safer than the previous attempt

Assets/Editor/AddScenesToBuild.cs
```diff
@@ -0,0 +1,90 @@
...
+        foreach (string path in scenePaths)
+        {
+            if (!IsPathInExcludedFolder(path))
+            {
+                filteredScenePaths.Add(path);
+            }
+        }
+
+        // Ensure "Core Platforms Menu" is at the beginning of the list
+        filteredScenePaths.Insert(0, "Assets/QA/Tests/Core Platform Menu/Core Platforms Menu.unity");
```
lyndon-unity: I was wondering if this gets added twice as scenePaths would pickup the scene too ?

lyndon-unity (approved):
This looks much better being based on 2019 Unity for the .unity / .meta files

Assets/Editor/AddScenesToBuild.cs
ekcoh: agreed, would recommend to either skip/filter it out in the foreach above and add it like this or detect in within the loop and use Insert(0 instead of the Add for that particular file.

Assets/Editor/AddScenesToBuild.cs
```diff
@@ -0,0 +1,90 @@
...
+using UnityEditor.SceneManagement;
+using UnityEngine;
+
+[InitializeOnLoad]
+public class AddScenesToBuild : EditorWindow
+{
+    [MenuItem("QA Tools/Open Core Scene Menu")]
+    static void OpenScene()
+    {
+        EditorSceneManager.OpenScene("Assets/QA/Tests/Core Platform Menu/Core Platforms Menu.unity");
```
ekcoh: This exact path is also present on line 37. I Would suggest you put it into a constant e.g.

`private const string corePlatformsMenu = "Assets/QA/Tests/Core Platform Menu/Core Platforms Menu.unity"`

and then reference it from all places used.

ekcoh (approved):
Approved, but would suggest switching to a constant for a string used more than once as well as address the comment by Lyndon about added twice.

ekcoh:
Needs to be updated with latest changes from develop and CI rerun before merge.

stefanunity (approved):

PR #1825, opened 2024-01-30
FIX: Avoid errors logged when opening settings menu (ISX-1721)

graham-huws:
### Description

It's possible for RedrawUI to be called when m_Root is empty, throwing errors in the console.
This can happen when the view is discarded immediately after being loaded, i.e. jumping to the input settings
page in Preferences when the last window shown there was the main view, which the editor (briefly) loads first. ([ISX-1721](https://jira.unity3d.com/browse/ISX-1721))

### Changes made

Compare against the number of m_ChildViews as a cheap way to know we should skip redrawing.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
@@ -13,11 +13,9 @@ namespace UnityEngine.InputSystem.Editor
     internal class ActionMapsView : ViewBase<ActionMapsView.ViewState>
     {
         public ActionMapsView(VisualElement root, StateContainer stateContainer)
-            : base(stateContainer)
+            : base(root, stateContainer)
         {
-            m_Root = root;
-
-            m_ListView = m_Root?.Q<ListView>("action-maps-list-view");
+            m_ListView = rootElement?.Q<ListView>("action-maps-list-view");
             m_ListView.selectionType = UIElements.SelectionType.Single;
 
             m_ListViewSelectionChangeFilter = new CollectionViewSelectionChangeFilter(m_ListView);
@@ -65,7 +63,7 @@ namespace UnityEngine.InputSystem.Editor
             ContextMenu.GetContextMenuForActionMapListView(this, m_ListView.parent);
         }
 
-        private Button addActionMapButton => m_Root?.Q<Button>("add-new-action-map-button");
+        private Button addActionMapButton => rootElement?.Q<Button>("add-new-action-map-button");
 
         public override void RedrawUI(ViewState viewState)
         {
@@ -183,7 +181,6 @@ namespace UnityEngine.InputSystem.Editor
 
         private readonly CollectionViewSelectionChangeFilter m_ListViewSelectionChangeFilter;
         private bool m_EnterRenamingMode;
-        private readonly VisualElement m_Root;
         private ListView m_ListView;
 
         internal class ViewState
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
@@ -15,7 +15,7 @@ namespace UnityEngine.InputSystem.Editor
         public ActionMapsView(VisualElement root, StateContainer stateContainer)
             : base(root, stateContainer)
         {
-            m_ListView = rootElement?.Q<ListView>("action-maps-list-view");
+            m_ListView = root.Q<ListView>("action-maps-list-view");
             m_ListView.selectionType = UIElements.SelectionType.Single;
 
             m_ListViewSelectionChangeFilter = new CollectionViewSelectionChangeFilter(m_ListView);
@@ -59,11 +59,10 @@ namespace UnityEngine.InputSystem.Editor
             CreateSelector(s => new ViewStateCollection<string>(Selectors.GetActionMapNames(s)),
                 (actionMapNames, state) => new ViewState(Selectors.GetSelectedActionMap(state), actionMapNames));
 
-            addActionMapButton.clicked += AddActionMap;
+            m_AddActionMapButton = root.Q<Button>("add-new-action-map-button");
+            m_AddActionMapButton.clicked += AddActionMap;
             ContextMenu.GetContextMenuForActionMapListView(this, m_ListView.parent);
         }
-
-        private Button addActionMapButton => rootElement?.Q<Button>("add-new-action-map-button");
 
         public override void RedrawUI(ViewState viewState)
         {
@@ -79,7 +78,7 @@ namespace UnityEngine.InputSystem.Editor
 
         public override void DestroyView()
         {
-            addActionMapButton.clicked -= AddActionMap;
+            m_AddActionMapButton.clicked -= AddActionMap;
         }
 
         private void RenameNewActionMaps()
@@ -181,7 +180,8 @@ namespace UnityEngine.InputSystem.Editor
 
         private readonly CollectionViewSelectionChangeFilter m_ListViewSelectionChangeFilter;
         private bool m_EnterRenamingMode;
-        private ListView m_ListView;
+        private readonly ListView m_ListView;
+        private readonly Button m_AddActionMapButton;
 
         internal class ViewState
         {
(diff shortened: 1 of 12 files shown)


graham-huws:
@ekcoh - I don't think this fix is "ideal", but it's a small change and it's a _very_ specific set of circumstances required to make the errors occur, and this catches that. If you've got any thoughts on improving though, I'm interested!

If there's a way to check for children by name in m_Root that doesn't produce errors, that's probably better 🤔

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -193,6 +193,13 @@ public override void DestroyView()
 
         public override void RedrawUI(ViewState viewState)
         {
+            // It's possible for RedrawUI to be called when m_Root is empty, throwing errors in the console.
+            // This can happen when the view is discarded immediately after being loaded, i.e. jumping to the input settings
+            // page in Preferences when the last window shown there was the main view, which the editor (briefly) loads first.
+            // Compare against the number of m_ChildViews as a cheap way to know we should skip redrawing. (ISX-1721)
+            if (m_Root.childCount != ChildViewCount())
```
ekcoh: Wouldn't it be better/cleaner to at least move this fix and the comments into the calcite of RedrawUI in ViewBase and handle it in a single place? At quick glance it at least looks like it could make sense to keep m_Root in ViewBase instead and then this can become completely private to ViewBase instead. There seems to be only one call-site inside ViewBase to RedrawUI (also includes other conditions).

ekcoh: I've seen another bug that indicate something similar which might indirectly be related to us not cleaning up properly after ourselves with dispatched UI actions, event handlers etc. But this could at least mitigate this for now and then it would be nice to keep it in the base class.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
```diff
@@ -98,6 +98,13 @@ private void OnAutoSaveToggle(ChangeEvent<bool> evt)
 
         public override void RedrawUI(ViewState viewState)
         {
+            // It's possible for RedrawUI to be called when m_Root is empty, throwing errors in the console.
```
ekcoh: Redundant comment and code, recommend moving to ViewBase as previous example.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ViewBase.cs
```diff
@@ -65,6 +65,11 @@ public void UpdateView(InputActionsEditorState state)
             view.DestroyView();
         }
 
+        public int ChildViewCount()
+        {
+            return m_ChildViews.Count;
```
ekcoh: Probably direct access of this member is the only thing needed inside ViewBase and you may skip the function if changed as suggested above?

ekcoh (changes requested):
Thanks for providing this fix. I think it can be improved by moving all of it into `ViewBase#UpdateView` method.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -16,9 +16,8 @@ namespace UnityEngine.InputSystem.Editor
     /// </summary>
     internal class ActionsTreeView : ViewBase<ActionsTreeView.ViewState>
     {
-        private readonly VisualElement m_Root;
         private readonly TreeView m_ActionsTreeView;
-        private Button addActionButton => m_Root?.Q<Button>("add-new-action-button");
+        private readonly Button m_AddActionButton;
```
graham-huws-unity, line 22: @ritamerkl, is moving addActionButton around here okay? I wasn't sure why it was set up differently to eg. m_ActionsTreeView. I made the change because otherwise it seemed to be accessing the root too early when following the repro steps for the bug which this PR fixes. ("_Expected a visual element called 'add-new-action-button' of type 'UnityEngine.UIElements.Button' to exist but none was found._")

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -65,7 +63,7 @@ public ActionMapsView(VisualElement root, StateContainer stateContainer)
             ContextMenu.GetContextMenuForActionMapListView(this, m_ListView.parent);
         }
 
-        private Button addActionMapButton => m_Root?.Q<Button>("add-new-action-map-button");
+        private Button addActionMapButton => rootElement?.Q<Button>("add-new-action-map-button");
```
ekcoh: Is there any reason for querying this lazily? If not I would advise initializing the reference as readonly in the constructor using the passed `root` argument instead of querying it on each call.

ekcoh: Right now its adressed differently in ActionMapsView vs ActionPropertiesView.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ekcoh: To me the change makes sense but would be good to get info from @ritamerkl whether there was any reason to do this lazily via getter indirection. If not I propose we look it up once and case it as a readonly var as suggested by @graham-huws.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -13,11 +13,9 @@ namespace UnityEngine.InputSystem.Editor
     internal class ActionMapsView : ViewBase<ActionMapsView.ViewState>
     {
         public ActionMapsView(VisualElement root, StateContainer stateContainer)
-            : base(stateContainer)
+            : base(root, stateContainer)
         {
-            m_Root = root;
-
-            m_ListView = m_Root?.Q<ListView>("action-maps-list-view");
+            m_ListView = rootElement?.Q<ListView>("action-maps-list-view");
```
ekcoh: Why use rootElement when you already have root passed as arg?

ekcoh: This is a pattern in all the constructors. It probably doesn't matter at the moment but if accessing base within constructor (while constructing the object) can be avoided its usually good to avoid. (This exists in all affected classes but haven't commented on all of them).

Pauliusd01 (approved):
LGTM, tried opening/editing different settings assets both active and inactive ones and wasn't able to repro the error anymore.

ekcoh (changes requested):

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ritamerkl: Ok for me!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
graham-huws-unity: > This is a pattern in all the constructors. It probably doesn't matter at the moment but if accessing base within constructor (while constructing the object) can be avoided its usually good to avoid.

Interesting, is that a style thing? I assumed performance-wise it would be the same; my thinking was that it's better to have everything accessing rootElement since then it's consistent across the files (eg. it's more clear that the root in the constructor and in RedrawUI is the same). Happy to change, either way!

ekcoh: Its probably just in the category nit-picking and could be ignored, but since there is already a local (potentially) in a register, why call into a getter that needs to fetch the reference from base?

ekcoh (approved):
LGTM

Pauliusd01 (approved):
Don't see the difference from when I last checked, lgtm

ritamerkl (approved):

PR #1829, opened 2024-02-02
Add editor mode tests to editor testin

stefanunity:
### Description

Finallly adding editor tests to be run on our CI.

### Changes made
-upgraded com.unity.test-framework to 1.4.3 which is the latest version supported on all versions
-added editor tests to be run on all editor test runs (named "win" & "mac" in our CI)
--results can be seen in yamato and the artifacts, listed next to normal playmode tests
-split DuplicateControlSchemeCommand_CreatesCopyOfControlSchemeWithUniqueName into two separate tests to cover two main cases:
--if the ControlScheme name ends on a string a "1" is added to the name of the new duplicate
--if the ControlScheme name ends on an int the is incremented by 1 in the name of the new duplicate
-disabled two test for deleting ControlSchemes which don't work in batchmode yet
-disabled InGameHints test since my fix is not stable yet

-future TODO:
--fix disabled editor tests
--add one more duplication test for the current untested case of:
---when duplicating a name with ending with an int the duplicated name will skip as many numbers as needed if there are already existing control schemes ending on the "next" number

### Checklist

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/Tests/Samples/InGameHintsTests.cs
```diff
@@ -15,6 +15,7 @@ public class InGameHintsTests : CoreTestsFixture
 {
     [UnityTest]
     [Category("Samples")]
+    [Ignore("Disabled: Unstable and running into Debug.Assert in ~@InGameHintsActions() on rare occasions.")]
```
ritamerkl: I am not familiar with this syntax. Is this to ignore test failure?

stefanunity: the whole test is going to get ignored aka not run. is there a better way? never did this before myself.

ritamerkl: Interesting, me neither. So unfortunately I don't know.

ekcoh: @stefanunity this is the way :)

Packages/manifest.json
```diff
@@ -7,7 +7,7 @@
     "com.unity.ide.visualstudio": "2.0.15",
     "com.unity.probuilder": "4.2.3",
     "com.unity.settings-manager": "1.0.3",
-    "com.unity.test-framework": "1.1.33",
+    "com.unity.test-framework": "1.4.3",
```
ekcoh, line 10: Is this manifest only referenced in testing? If not, does it work well with Unity 2019 since Input System goes back to 2019?

Assets/Tests/Samples/InGameHintsTests.cs
ekcoh: @stefanunity I would also recommend filing a ticket on any unstable tests disabled in this PR when its merged.

ekcoh (commented):
Nice to see this coming together @stefanunity ! Looks like a merge conflict needs to be resolved for CI to run on this branch though.

Packages/manifest.json
stefanunity: The test framework is only used when running the tests. Some of the tests are broken when running in 2019 using either version of that framework. 

1.4.3 is listed as supported from 2019.2 upwards.

Assets/Tests/Samples/InGameHintsTests.cs
stefanunity: https://jira.unity3d.com/browse/ISX-1896 for followup

ritamerkl (approved):

Packages/manifest.json
ekcoh: ok, thanks

ekcoh (approved):

PR #1832, opened 2024-02-05
FIX: Properly reset ActionMap state in OnAfterDeserialize (ISXB-737)

timkeo:
### Description

Simple fix for [ISXB-737](https://jira.unity3d.com/browse/ISXB-737) and [ISX-1861](https://jira.unity3d.com/browse/ISX-1861).

### Changes made

The `m_EnabledActionsCount = 0` is used to determine if the ActionMap is "enabled" but wasn't being reset within `OnAfterDeserialize()`. This caused the ActionMap to fall into a bad state: it was effectively disabled but calling `Enable()` would early out because `m_EnabledActionsCount` indicated it was still enabled. Setting this field 0 along with nulling `m_State` fixes the bug.

### Notes

This issue only seems to occur when importing the InputSystem package; this code path isn't hit when opening a project that already has InputSystem imported. Currently this bug doesn't cause any (visible) problems, since the domain reload when entering PlayMode wipes out the bad state, but Project Wide Actions and Fast Enter PlayMode are now exposing the problem.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -39,6 +39,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed "Listen" functionality for selecting an input sometimes expecting the wrong input type.
 - Fixed console errors that can be produced when opening input package settings from the Inspector.
 - Fixed InputManager.asset file growing in size on each Reset call.
+- Fixed Opening InputDebugger throws 'Action map must have state at this point' error
 
 ## [1.8.0-pre.2] - 2023-11-09
 
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
@@ -1984,6 +1984,7 @@ namespace UnityEngine.InputSystem
         {
             m_State = null;
             m_MapIndexInState = InputActionState.kInvalidIndex;
+            m_EnabledActionsCount = 0;
 
             // Restore references of actions linking back to us.
             if (m_Actions != null)
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -11923,4 +11923,36 @@ partial class CoreTests
     {
         Assert.Fail();
     }
+
+    // Validate OnAfterDeserialize() completely disables ActionMap
+    // https://jira.unity3d.com/browse/ISXB-737
+    [Test]
+    [Category("Actions")]
+    public void Actions_ActionMapDisabledDuringOnAfterSerialization()
+    {
+        var map = new InputActionMap("MyMap");
+        var action = map.AddAction("MyAction");
+        action.AddCompositeBinding("2DVector")
+            .With("Up", "<Keyboard>/w")
+            .With("Down", "<Keyboard>/s")
+            .With("Left", "<Keyboard>/a")
+            .With("Right", "<Keyboard>/d");
+
+        map.Enable();
+
+        Assert.That(map.enabled, Is.True);
+        Assert.That(map.FindAction("MyAction", true).enabled, Is.True);
+        Assert.Throws<System.InvalidOperationException>(() => map.AddAction("Something"));
+
+        // Calling this directly is contrived (and not supported usage) but it should
+        // effectively cover this regression scenario.
+        map.OnAfterDeserialize();
+
+        Assert.That(map.enabled, Is.False);
+
+        map.Enable();
+
+        Assert.That(map.enabled, Is.True);
+        Assert.That(map.FindAction("MyAction", true).enabled, Is.True);
+    }
 }
--- a/Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
@@ -253,6 +253,48 @@ internal partial class CoreTests
         Assert.That(InputSystem.actions.actionMaps[1].actions.Count, Is.EqualTo(1));
         Assert.That(InputSystem.actions.actionMaps[1].actions[0].name, Is.EqualTo("replacedAction4"));
     }
-}
 
+#if UNITY_EDITOR
+    [Test]
+    [Category(TestCategory)]
+    public void ProjectWideActions_ThrowsWhenAddingOrRemovingWhileEnabled()
+    {
+        var asset = ProjectWideActionsAsset.GetOrCreate();
+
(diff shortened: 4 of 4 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed "Listen" functionality for selecting an input sometimes expecting the wrong input type.
 - Fixed console errors that can be produced when opening input package settings from the Inspector.
 - Fixed InputManager.asset file growing in size on each Reset call.
+- Fixed Opening InputDebugger throws 'Action map must have state at this point' error
```
ekcoh, line 42: Since this is a public issue provide ISXB-737 reference with hyperlink to public issue tracker. Also missing full-stop at end of line.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
```diff
@@ -1984,6 +1984,7 @@ public void OnAfterDeserialize()
         {
             m_State = null;
             m_MapIndexInState = InputActionState.kInvalidIndex;
+            m_EnabledActionsCount = 0;
```
ekcoh, line 1987: Nice small resolution to this one :) I would recommend considering adding a unit test that captures this scenario if this wasn't previously covered in the unit tests for InputActionMap. Do you think that would be feasible?

ekcoh (changes requested):
Fix LGMT, basically looking for link in CHANGELOG link and potentially a test case covering this scenario.

Pauliusd01:
ISX-1861 still reproduces with no changes to the repro steps

Packages/com.unity.inputsystem/CHANGELOG.md
timkeo: It was reported internally (not public) and doesn't have any issuetracker page.
Should we make it public?

timkeo:
@Pauliusd01
I'm curious to what your verification steps were.

This is how I verified the fix:
1. Start with a "clean" copy of the repro project downloaded from the ticket
2. Open "manifest.json" in a text editor
3. Change the "com.unity.inputsystem" package to point to the local repro, e.g.
```"com.unity.inputsystem": "file:C:/UnitySrc/InputSystem/Packages/com.unity.inputsystem",```
4. Open the project in a 2023.3 build of the Editor
5. Open InputDebugger

You should be able to verify the fix by switching branches before opening the project, i.e. it repros with develop but doesn't repro with my branch. The _important_ thing: it only seems to (reliably) repro when "upgrading" a project between Unity versions. Therefore, you must completely delete the project and start from a fresh copy between repro attempts.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Ah, sorry then we are all good. Assumed it was external since it was an ISXB, but then leave it as-is.

ekcoh (approved):
I marked the previous comment on CHANGELOG as resolved since this was not a public issue so leave CHANGELOG as-is. Consider adding a unit test covering this scenario though or at least a target test on the InputActionMap reset to get coverage. Since fix is legit I will approve this PR as it currently sits but any test coverage improvement would be valuable to avoid future regressions.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
timkeo: I added 2 test cases:
- An Actions test covering the specific scenario
- A ProjectWideActions test validating Maps/Actions cannot be added/removed while ActionMap is enabled

The only other time ActionMap's `m_State` is reset is when processing Setup changes, i.e. `OnSetupChanged()`, which is only allowed when it's already disabled. I didn't find any other test validating this specific behavior and so added the second case.

Pauliusd01:
> @Pauliusd01 I'm curious to what your verification steps were.
> 
> This is how I verified the fix:
> 
>     1. Start with a "clean" copy of the repro project downloaded from the ticket
> 
>     2. Open "manifest.json" in a text editor
> 
>     3. Change the "com.unity.inputsystem" package to point to the local repro, e.g.
>        `"com.unity.inputsystem": "file:C:/UnitySrc/InputSystem/Packages/com.unity.inputsystem",`
> 
>     4. Open the project in a 2023.3 build of the Editor
> 
>     5. Open InputDebugger
> 
> 
> You should be able to verify the fix by switching branches before opening the project, i.e. it repros with develop but doesn't repro with my branch. The _important_ thing: it only seems to (reliably) repro when "upgrading" a project between Unity versions. Therefore, you must completely delete the project and start from a fresh copy between repro attempts.

The same way listed on the case itself. I pull your changes to the repo -> disable domain reload -> open and run the projectwideactions test scene ->keyboard does not move the cube. Contrary to what it says in the description I can repeat this as many times as I want and not just once. By the way it sounds like you might be referring to the wrong issue - ISXB-737, I am talking about ISX-1861 which is listed as fixed in your PR description

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs
ekcoh: Sounds great. Thank you.

timkeo:
> 
> The same way listed on the case itself. I pull your changes to the repo -> disable domain reload -> open and run the projectwideactions test scene ->keyboard does not move the cube. Contrary to what it says in the description I can repeat this as many times as I want and not just once. By the way it sounds like you might be referring to the wrong issue - ISXB-737, I am talking about ISX-1861 which is listed as fixed in your PR description

Ah sorry my bad! You're right I was thinking of the other ticket and not [ISX-1861](https://jira.unity3d.com/browse/ISX-1861).

Alright then, for FEPM ProjectWideActions are also broken by this bug [ISX-1840](https://jira.unity3d.com/browse/ISX-1840). I know the ticket is marked "Done" but the fix hasn't landed yet (since FEPM work being set aside), and so you'll need to add the workaround described in the bug to the repro project. Basically manually call Enable() on the ActionMap from a script. With the FEPM work-around in place, you should be able to repro the InputDebugger bug and then verify my change fixes it.

I had meant to call this out in the PR but completely forgot about it.
Sorry for the confusion.

Pauliusd01 (approved):
LGTM, the issue I mentioned is fixed correctly now after I applied additional changes to the projectwideactions script from ISX-1840.

PR #1822, opened 2024-01-29
ISX-1851 Remove title from editor window

ekcoh:
### Description

Removes the header title from UITK editor window when used outside Project Settings.
Improves styling of header title in Project Settings.

### Changes made

Removes the header title from editor window as called out in user feedback here: https://forum.unity.com/threads/project-wide-actions-1-8-0-pre-2-release-feedback-wanted-new-update-released.1488843/ by Walter_hulsebos.

Modifies header title styling to look more similar to other Settings. Not perfect but closer. Also alignes menu button with container window menu button.

Screenshot 2024-01-29 at 12 29 47

Screenshot 2024-01-29 at 12 16 04

### Notes

-

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #1833, opened 2024-02-06
Isx 1851 remove title from editor window rename - extended

ekcoh:
### Description

Note that this branch is a sub-branch of https://github.com/Unity-Technologies/InputSystem/pull/1822 and hence it contains all changes from that PR but also contains additional fixes called out under *Changes made* below.

This PR is a tentative fix to the regressive behaviour reported in https://github.com/Unity-Technologies/InputSystem/pull/1822 and hence if verified to eliminate the problems mentioned in that PR and fulfilling the functionality intended on https://github.com/Unity-Technologies/InputSystem/pull/1822 this PR might be merged instead and the previous PR closed.
If not successful this either needs more work or needs to be discarded while the original PR remains intact with the UI changes.

### Changes made

Everything mentioned in https://github.com/Unity-Technologies/InputSystem/pull/1822

But also:
- Modified RenameNewActionMaps to no longer be triggered from RedrawUI. Instead Rename is triggered as a shortcut command or via AddActionMap (indirectly through state not in sync with view/state). However this now happens as a continuation of AddActionMap if successful. This is likely the fix corrected the regression.
- Modified RenameNewActionMap to scroll to target.
- Extended StateContainer Dispatch to allow taking an optional action to be executed on the UI scheduler queue directly after the command if and only if the command is successful. The intention is increased exception safety.
- Extended ViewBase with a isRedrawInProgress that may be used to discard any View to Model changes that would not lead to any change early. This basically just eliminates need to unhook/modify/hook patterns in RedrawUI since RedrawUI should only reflect the model/state to the view.
- Made some minor improvements to CollectionViewSelectionChangeFilter so it uses less memory when disabled. Also fixed a compiler error when disabled (dev mode only). 
- Made a bunch of methods protected that were public for no reason (only internal code)
- Made a bunch of fields readonly when intention is to not allow modification of them.

### Notes

Additional issues found during development. Previously Action Map panel never scrolled to newly added items when list is larger than panel. This has been corrected logically but for some reason the list is always one item off. Unclear why.

Found an issue where adding a number of Actions and directly after exiting renaming mode with enter the up key is pressed causes an exception in UITK code. Is this a new bug?

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Resources/InputActionsEditor.uxml
```diff
@@ -1,7 +1,7 @@
 <ui:UXML xmlns:ui="UnityEngine.UIElements" xmlns:uie="UnityEditor.UIElements" xsi="http://www.w3.org/2001/XMLSchema-instance" engine="UnityEngine.UIElements" editor="UnityEditor.UIElements" noNamespaceSchemaLocation="../../../../../UIElementsSchema/UIElements.xsd" editor-extension-mode="True">
     <Style src="project://database/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Resources/InputActionsEditorStyles.uss?fileID=7433441132597879392&amp;guid=7dac9c49a90bca4499371d0adc9b617b&amp;type=3#InputActionsEditorStyles" />
     <ui:VisualElement name="header" class="header">
-        <ui:Label text="Input Actions" display-tooltip-when-elided="true" name="title-label" class="header-label" style="width: auto; margin-left: 0; margin-right: 0; margin-top: 0; margin-bottom: 0;" />
+        <ui:Label text="Input Actions" display-tooltip-when-elided="true" name="title-label" class="header-label" />
```
timkeo, line 4: Question: Was removing this part of the "header" fix or simply removing unnecessary attributes?
I'm a noob with UITK and curious to this change.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -57,43 +62,49 @@ public ActionMapsView(VisualElement root, StateContainer stateContainer)
...
 
         private void DeleteActionMap(int index)
         {
-            Dispatch(Commands.DeleteActionMap(index));
+            Dispatch(Commands.DeleteActionMap(index), () =>
+            {
+                // WORKAROUND: For some reason after delete (haven' identified a pattern), m_ListView would
+                // lose focus preventing further navigation or commands associated with the view.
+                // Hence we explicitly reclaim focus here post the delete command as a workaround.
+                m_ListView.Focus();
```
timkeo, line 106: FYI: This is [similar to what I did for deleting ActionTree item](https://github.com/Unity-Technologies/InputSystem/pull/1812/files#r1444433459).

timkeo (approved):
LGTM.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/StateContainer.cs
```diff
@@ -42,7 +43,13 @@ public void Dispatch(Command command)
                 // catch exceptions here or the UIToolkit scheduled event will keep firing forever.
                 try
                 {
+                    // Note that continueWith is only executed if provided and only if StateChanged does not
+                    // throw and command didn't throw.
+                    // This provides increased exception safety since if continueWith was just scheduled
+                    // to be executed after command, any exception in command might fail continueWith since
+                    // state may be undefined by partial execution of command.
                     StateChanged?.Invoke(m_State);
+                    continueWith?.Invoke();
```
ritamerkl, line 52: good idea!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -18,10 +18,15 @@ public ActionMapsView(VisualElement root, StateContainer stateContainer)
...
+            // Setup a selection change filter that prevents the user from deselecting items.
+            // This is desirable since deselecting an Action Map (ESC) would make the Action panel empty with
+            // the current UI layout. Note that we also avoid dispatching select commands if callback
+            // is indirectly triggered by a redraw (instead of constantly hooking and unhooking listener).
             m_ListViewSelectionChangeFilter = new CollectionViewSelectionChangeFilter(m_ListView);
             m_ListViewSelectionChangeFilter.selectedIndicesChanged += (selectedIndices) =>
             {
-                Dispatch(Commands.SelectActionMap((string)m_ListView.selectedItem));
+                if (!isRedrawInProgress)
+                    Dispatch(Commands.SelectActionMap((string)m_ListView.selectedItem));
```
ritamerkl, line 29: I am a bit concerned. Is it possible we miss selections like that? For example is we click around very fast?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -57,43 +62,49 @@ public ActionMapsView(VisualElement root, StateContainer stateContainer)
...
+        protected override void RedrawUI(ViewState viewState)
         {
-            m_ListView.itemsSource = viewState.actionMapNames?.ToList() ?? new List<string>();
-            if (viewState.selectedActionMap.HasValue)
+            m_ListView.itemsSource = viewState.actionMapNames;
+            m_ListView.Rebuild();
+
+            // Update view to reflect model selection
+            var desiredSelectedIndex = viewState.selectedActionMapIndex;
+            if (desiredSelectedIndex >= 0 && desiredSelectedIndex != m_ListView.selectedIndex)
```
ritamerkl, line 80: Especially with this code (see comment above)- could it be we reset a selection that was made in the list but did not reach the state yet?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -57,43 +62,49 @@ public ActionMapsView(VisualElement root, StateContainer stateContainer)
...
         }
 
-        private void RenameNewActionMaps()
+        private void RenameNewActionMap()
         {
-            if (!m_EnterRenamingMode)
-                return;
-            m_ListView.ScrollToItem(m_ListView.selectedIndex);
             var element = m_ListView.GetRootElementForIndex(m_ListView.selectedIndex);
+            m_ListView.ScrollToItem(m_ListView.selectedIndex);
```
ritamerkl, line 95: Why are we scrolling to the item twice? Isn't this triggered from RedrawUI already after adding the new Action Map?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ritamerkl: Is this known/ is there an explanation from UIToolkit folks?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -123,8 +134,7 @@ private void ChangeActionMapName(int index, string newName)
 
         private void AddActionMap()
         {
-            Dispatch(Commands.AddActionMap());
-            m_EnterRenamingMode = true;
+            Dispatch(Commands.AddActionMap(), continueWith: RenameNewActionMap);
```
ritamerkl, line 137: Good change, much better like this!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -120,6 +120,9 @@ public ActionsTreeView(VisualElement root, StateContainer stateContainer)
             m_ActionsTreeViewSelectionChangeFilter = new CollectionViewSelectionChangeFilter(m_ActionsTreeView);
             m_ActionsTreeViewSelectionChangeFilter.selectedIndicesChanged += (_) =>
             {
+                if (isRedrawInProgress)
+                    return;
```
ritamerkl, line 124: same concerns- could it be we miss selections?

ritamerkl (commented):
I am happy to see lots of good changes.
I left some comments, while my major concern is that we might miss selections if we interrupt the Selection Callback when a redraw is in process. Maybe I am overseeing something.

Pauliusd01 (changes requested):
I am still reproducing https://github.com/Unity-Technologies/InputSystem/pull/1822#issuecomment-1914690387 is that expected?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Resources/InputActionsEditor.uxml
ekcoh: It was just a minor cleanup, moving styling to the .uss file (corresponding to e.g. web css) instead of declaring it inline. It was part of the initial purpose of the PR this is based on.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ekcoh: I haven't asked them but probably should. Did you follow up on this @timkeo ? 
However, I think its not uncommon behavior on delete operations in UI frameworks to move focus to the parent control/container. E.g. select a file in Explorer, press delete. Tentative focus in moved to the directory window (container) and hence pressing down key selects the first child element in the list. However, here we try to do something different due to the UI design of this window which would otherwise make Actions panel and Bindings panel empty after delete due to missing selection. I believe this is more similar to how Finder in macOS behaves.

ekcoh:
> I am happy to see lots of good changes.
> I left some comments, while my major concern is that we might miss selections if we interrupt the Selection Callback when a redraw is in process. Maybe I am overseeing something.

Not sure I follow here @ritamerkl could you elaborate a bit since it sounds like you have spotted something about this PR I might have overlooked?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ekcoh: The only purpose with this is when Selection Change is coming from the State which triggers a Redraw, which in turn updates m_ListView selection. In that case Dispatch here would only schedule a command that would lead to the very same update, hence yet an unnecessary update. isRedrawInProgress would only be true if this code is triggered from within Redraw (since single threaded). Hence this wouldn't block selection changes to m_ListView triggered by the user. Doing something really fast shouldn't matter (but could get throttled by UI framework) since single threaded and operating from a queue. Or might I be wrong about these expectations? At least this is what I saw when doing these changes on my Mac.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ekcoh: Ah, I believe I understand what you are after now. Redraw seems to be scheduled via UI scheduler so I would expect user initiated and callback events being processed through the same scheduler queue but I might be wrong. I did not notice any such behavior while testing this but I know too little about UITK internals to be sure.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ekcoh: Thanks for pointing this out, agree it seems to be done twice. However I still think its possible to select an item in the list, then scroll so the selected item isn't visible, then hit Rename shortcut. If the item being renamed is not visible then it might get confusing?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
ekcoh: Same answer as previous, could easily be removed which only leads to additional states and redraws being generated.

ekcoh:
> I am still reproducing [#1822 (comment)](https://github.com/Unity-Technologies/InputSystem/pull/1822#issuecomment-1914690387) is that expected?

I hoped this would eliminate that which is the whole point of this "sub-branch". Seems like I need to revisit on Windows again then.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ritamerkl: A redraw is triggered automatically as soon as the serializedObject changed as well, I think the scheduler could collide with this behavior, that's my doubt. If so this would mean we discard the selection of the list/treeView at a point. But I don't know certainly as well how this would be working, maybe worth to investigate with devs-uitoolkit.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ritamerkl: right, I didn't consider this case, anyway I think it won't harm at least.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
ritamerkl: As far as I know the redraw is triggered at two places: after we Dispatch() a command and after we change the serializedObject of the state. Interesting is that this happens twice as soon as we do both? So if we change a SerializedObject and do a (eg.) selection after (in the same command) there is a possibility that we abort the second redraw that is actually necessary. Maybe I am missing something, it seems to be working, hopefully it is not just good timing.

ritamerkl:
> Not sure I follow here @ritamerkl could you elaborate a bit since it sounds like you have spotted something about this PR I might have overlooked?

I tried to bring a bit more clarity in my comments at the according threads

ritamerkl:
> > I am still reproducing [#1822 (comment)](https://github.com/Unity-Technologies/InputSystem/pull/1822#issuecomment-1914690387) is that expected?
> 
> I hoped this would eliminate that which is the whole point of this "sub-branch". Seems like I need to revisit on Windows again then.

I was not able to reproduce this error in 2022.3.19 and the same editor version shown in the video. 
The first error seems to start the exception row, it would be interesting to see the whole log of this one. If you can reproduce this bug again @Pauliusd01, could you look if you can get the whole stack trace?

ekcoh:
I believe this PR is obsolete now due to other changes. Will close this PR but leave branch existing for a while longer if there is need to pick any cherries from it.

PR #1834, opened 2024-02-07
CHANGE: ISX-1865 Switch from InputManager.asset to asset-based storage for Project-wide actions

ekcoh:
### Description

This branch switches from using InputManager.asset for storing project-wide actions to instead store them in an appointed .inputactions asset. 

### Changes made

- Introduces new UX in _Project Settings_ > _Input Actions_ that prompts the user to either select an existing asset or to create a new one in order to get setup with Project-wide actions. If a new asset is created or selected this is at the same time assigned to the static property `InputSystem.actions`. If the default target filename already exist the user is prompted to overwrite or cancel the operation.
- Introduces new UX via Inspector when a .inputactions asset is selected to display whether its the current Project-wide actions or not and to also make it the active asset if its not currently the active asset. Open button for regular assets opens assets in a Input Actions Editor window. When the project-wide asset is selected there is also an option to Edit In Project Settings.
- Double clicking a .inputactions asset opens it in a Input Actions Editor window regardless if its the active Project-wide asset or not.
- When InputSystem.actions is assigned either via editor or via code, the appointed project-wide asset is changed and also persisted in Editor Build Settings as an asset reference (Config-object). This object reference is fetched during build to embed a preloaded asset for player builds that can be loaded via `Resources`. This is similar to how `InputSettings` are handled.
- After a round of Unity assert import has finished, Input System now attempts to extract any actions residing in `InputManager.asset` into a project-wide Input Actions asset instead. This is currently guarded with a static variable to only happen once per domain reload (throttling).
- All write operations to InputManager.asset have been removed.
- Adds a new API `InputSystem.onActionsChanged` which allows monitoring any changes to the property.
- All previous code for writing Project-wide actions into ProjectSettings/InputManager.asset has been removed.
- Project-wide actions are enabled on entering play-mode () and disabled when exiting play-mode (OnEnterEditMode).
- Added "Remove All Action Maps" option to Project-wide Input Action Editor context menu to allow a user to quickly remove default template if desired.
- Removed tests from CoreTests_ProjectWideActions which no longer makes sense to tests.
- Removed tests from CoreTests_ProjectWideActions which overlap with tests in CoreTests_Actions.cs.
- Modified Input Actions Editor Reset operation to only operate on the current in-memory version of the asset making it Undoable.
- Added logic to handle the case of Project-wide .inputactions being deleted while the editor is open.
- Reworked object pickers for InputActionAsset (affecting e.g. PlayerInput editor) and InputActionReference to also visualise Project-wide assets in a separate tab compared to regular assets.

Minor/technical:
- Removed redundant code for saving InputActionAssets and relocated to InputActionAssetManager. Eliminated redundant dialog strings by utilising a common utility function regardless of editor framework.
- Refactored InputActionImporter.
Modified CopyPasteHelpers, SerializationHelpers to allow using buffer modification for any string buffer to enable utilising this functionality when replacing action maps within an existing asset.
- InputForUI monitors changes to project-wide actions configuration and handles case of an initial null-asset

Open/known issues that might need fixing before merge:
- Changelog needs update based on this list.
- Temporarily disabled test `CoreTests_ProjectWideActions#ProjectWideActions_ThrowsWhenAddingOrRemovingWhileEnabled` (added recently by @timkeo) since its unclear  how it should be treated in this new concepts. Needs to be determined.
- Not directly related to changes on this branch but it seems InputForUI assumes Project-wide actions is available and the separate assembly lacks any symbol definition to conditionally compile. 
- Consider adding a mechanic to handle the scenario of project-wide asset being open simultaneously in regular editor and Project-Settings. In this case, and to simplify implementation, we could hide the editor in Project Settings and replace it with a disclaimer, e.g. "The asset is currently being edited in a separate Input Action Editor" with a shortcut to move to that window. Basically Project Settings could back off if another editor with the asset is open and otherwise show the UI again. Probably a UX discussion topic  @Billreyn . 
- Need to investigate/test properly that Project-Settings editor and windowed-editor both react to external asset updates when already open. Basically, if source file is modified while editor is open there are two scenarios. 1) Asset unmodified, and editor should reload latest content and mark itself as not-dirty. Other scenario is that in-memory changes exist and in that case user need to be prompted whether to overwrite file on disc with editor version or discard changes and replace editor content with asset from disk.

### Notes

Suggestion for QA:
- Make sure to cover usage of Project-wide actions in built player.
- Make sure to cover usage of Project-wide actions when entering/exiting play-mode.
- Make sure to cover editor workflows relating to Project-wide asset.
- Good to check corner cases like moving, deleting, renaming project-wide-asset and doing such operations when editors are open.
- Test in game UI toolkit control with valid and with null Project-wide actions

For any issues found we need to decide on a case-by-case basis whether it should be handled as a separate issue (and get this merged) or whether its an issue that may risk this concept not holding up and then we should fix it on this branch before merge.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


lyndon-unity:
If I follow these steps I see a problem:

1. Create new project and add input system package
2. Open project settings and click the create asset button - this works to create the asset, but the create button remains as the asset contents are displayed
3. Close project settings
4. Reopen project settings - asset is found and shown, no button for create
5. Delete the asset from the database
6. Open project settings again and as no asset, click create button - works to create it and show it
7. Close project settings
8. Reopen - **asset is not found/shown in UI** - so create button still present on its own  (but asset is present in the database)

Problem seems to be related to the InputSystem.asset assignment. The check to see if the asset is already assigned passes as the assigned value has a instanceId that matches.


Note - restarting editor and the asset is found again when opening the project settings
9 - restart editor
10 - open project settings - asset is found !


In step 5 deleting the asset from the database, the InputSystem.actions still has a reference to that removed file.
It seems that when assigning the newly created file in 6 it checks if InputSystem.actions is the same as the new asset before assigning and the instanceID matches so it doesn't assign it over.

if we selected 'none' or a different asset and then reselect the newly created asset then the UI updates.

lyndon-unity:
Also the UX is different to the flow for creating the input settings - these don't pop up a location dialog

Input settings don't ask for location when button clicked - they only ask for location when using the version from the menu on the cog icon (which isn't present in the project wide input actions editor ATM)

^ I've addressed this by making the button write the asset to a fixed location.
Adding the selector to be able to pick an asset from other locations has also helped.
Currently you still can't create an asset in a random location (unless you use the project browser)

Adding a drop down like the IMGUI settings version is more work and not critical in first pass.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/AssetContext.cs
```diff
@@ -0,0 +1,106 @@
...
+        public string sourcePath => m_SourcePath;
+
+        public void AddObjectToAsset(string identifier, Object subAsset, Texture2D icon)
+        {
+            AssetDatabase.AddObjectToAsset(subAsset, m_AssetPath);
+        }
+
+        public void SetMainObject(Object obj)
+        {
+            AssetDatabase.AddObjectToAsset(obj, assetPath); // Note that adding is necessary here compared to importer
```
lyndon-unity: Can this object get added more than once though this call ?
Does this need to remove any old object ?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/AssetContext.cs
```diff
@@ -0,0 +1,106 @@
...
+            m_Context.SetMainObject(obj);
+        }
+
+        public void LogError(string message)
+        {
+            m_Context.LogImportError(message);
+        }
+    }
+
+    struct AssetDatabaseAssetContext : IAssetContext
```
lyndon-unity: I don't see this being used anywhere - why the extra complexity here?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
```diff
@@ -425,9 +425,11 @@ public static Command ResetGlobalInputAsset(Action<InputActionAsset> postResetAc
         {
             return (in InputActionsEditorState state) =>
             {
+                // TODO Here we would instead modify the asset based on an instance loaded from defaults. Note that we cannot use importer here since we modify an in-memory asset as serialized object.... needs some thought to not break Undo.
```
lyndon-unity: Its not immediately obvious in the UX that the reset doesn't work yet so need to make sure this is revisited.

That said the 'reset' button is currently missing from the UX.

Pauliusd01:
Was asked to take a look early, here's my initial notes:

- Does the pre-filled asset name have to be a generic "InputSystem.Actions.inputactions" ? Why not something more clear? [Optional]
- ProjectWideActions test scene in main repo does not work with this PR's changes (with newly created asset that has the default name)
- PlayerInput component is broken if you don't already have the ProjectWideActions asset created. You should be able to use the dropdown to use a custom actions asset if you don't want to create a projectwideactions but currently dropdown does nothing. Also, if you do use the built in button to create a projectwideactions asset then the component is still broken - showing you the custom actions view instead.
- ProjectWideActions reset button does nothing.
- The name of the currently used asset should be visible somewhere in the Project Settings -> ProjectWideActions view as you can load different ones now. [Optional]
- Text explaining what ProjectWideActions are and why the user might want it next to the "Create new a new Project Wide Actions Asset" button would be good to have [Optional]

lyndon-unity:
> * The name of the currently used asset should be visible somewhere in the Project Settings -> ProjectWideActions view as you can load different ones now. [Optional]

This is now added

lyndon-unity:
> * ProjectWideActions reset button does nothing.

This is now fixed

lyndon-unity:
I'm seeing some issues in 2022.3.12f1 when selecting an input action asset

NullReferenceException: Object reference not set to an instance of an object
UnityEngine.InputSystem.Editor.InputActionImporterEditor.OnInspectorGUI () (at ./Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporterEditor.cs:47)
UnityEditor.UIElements.InspectorElement+<>c__DisplayClass72_0.<CreateInspectorElementUsingIMGUI>b__0 () (at <53ddbed73faf4fe3b980a493ab4e6639>:0)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr, Boolean&)

image

Appears to be an import error

image


My latest commit fixes this

lyndon-unity:
> * ProjectWideActions test scene in main repo does not work with this PR's changes (with newly created asset that has the default name)

I think you mean ProjectWideActionsSampleScene
I've just fixed this to show a message if the InputSystem.actions is null and also to enabled them if they are not enabled (which they are not in the PR at present)

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
lyndon-unity: ^ I've fixed the reset button

ekcoh:
> Also the UX is different to the flow for creating the input settings - these don't pop up a location dialog
> 
> Input settings don't ask for location when button clicked - they only ask for location when using the version from the menu on the cog icon (which isn't present in the project wide input actions editor ATM)
> 
> ^ I've addressed this by making the button write the asset to a fixed location. Adding the selector to be able to pick an asset from other locations has also helped. Currently you still can't create an asset in a random location (unless you use the project browser)
> 
> Adding a drop down like the IMGUI settings version is more work and not critical in first pass.

Makes sense, the initial behavior was as explained only a quick fix for having some kind of UI. However, it is likely wise to align these two UIs to not have project-wide selection and InputSettings having different selection UIs and behavior. It could make sense to update InputSettings UI to behave the same.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/AssetContext.cs
ekcoh: I will likely undo this part as mentioned in a previous comment about this work, it was helpful during the transition phase to remove code bloat.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/AssetContext.cs
ekcoh: In theory yes, similar to other comment either this goes away since no longer needed when doing final cleanup of this work or a check should be in place.

ekcoh:
> If I follow these steps I see a problem:
> 
> 1. Create new project and add input system package
> 2. Open project settings and click the create asset button - this works to create the asset, but the create button remains as the asset contents are displayed
> 3. Close project settings
> 4. Reopen project settings - asset is found and shown, no button for create
> 5. Delete the asset from the database
> 6. Open project settings again and as no asset, click create button - works to create it and show it
> 7. Close project settings
> 8. Reopen - **asset is not found/shown in UI** - so create button still present on its own  (but asset is present in the database)
> 
> Problem seems to be related to the InputSystem.asset assignment. The check to see if the asset is already assigned passes as the assigned value has a instanceId that matches.
> 
> Note - restarting editor and the asset is found again when opening the project settings 9 - restart editor 10 - open project settings - asset is found !
> 
> In step 5 deleting the asset from the database, the InputSystem.actions still has a reference to that removed file. It seems that when assigning the newly created file in 6 it checks if InputSystem.actions is the same as the new asset before assigning and the instanceID matches so it doesn't assign it over.
> 
> if we selected 'none' or a different asset and then reselect the newly created asset then the UI updates.

I believe this relates to the currently missing connection between the importer and the UITK editors, currently looking into this. It seems it was never properly implemented as it is for IMGUI editor.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/AssetContext.cs
ekcoh: This has now been removed. Will mark this resolved.

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1834) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1834) before we can accept your contribution.<br/><hr/>**Lyndon Homewood** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/InputSystem?pullRequest=1834) it.</sub>

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -1,10 +1,13 @@
 #if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
 
 using System.IO;
+using System.Text.RegularExpressions;
```
lyndon-unity, line 5: Is this really needed - looks like its added without using it

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -9,44 +9,69 @@
...
-        }
-    }
 
-    internal static class ProjectWideActionsAsset
-    {
-        private const string kDefaultAssetPath = "Assets/InputSystem_Actions.inputactions";
-        private const string kDefaultTemplateAssetPath = "Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsTemplate.json";
+            internal static void DeleteActionAssetAndActionReferences()
+            {
+                var objects = AssetDatabase.LoadAllAssetsAtPath(kAssetPath);
```
ekcoh: This looks redundant, either pass `objects` array from `MoveInputManagerAssetActionsToProjectWideInputActionAsset()` into this function or combine the functions together since I believe its the only place where this is used?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -9,44 +9,69 @@
...
+            // TODO Consider preserving GUIDs to potentially enable references to stay intact.
+            // TODO 3. Let InputActionImporter do its job on importing and configuring the asset.
+            // TODO 4. Assign to InputSystem.actions
 
-            // Handle deleting all InputActionAssets as older 1.8.0 pre release could create more than one project wide input asset in the file
-            foreach (var obj in objects)
+            class ProjectSettingsPostprocessor : AssetPostprocessor
             {
-                if (obj is InputActionReference)
+                static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths, bool didDomainReload)
```
ekcoh, line 30: This will be retriggered on every asset import. I believe we could keep a project specific setting here or at least a static to avoid excessive unnecessary processing?

ekcoh: @AlexTyrer

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: @AlexTyrer

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: I've added a static as a quick fix here (aee68129f9765b77ca559bcc0c9c39aa3f7b1a72)

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: Good call - I've removed the DeleteActionAssetAndActionReferences() function and just put the implementation inside MoveInputManagerAssetActionsToProjectWideInputActionAsset().

Commit: aee68129f9765b77ca559bcc0c9c39aa3f7b1a72

ekcoh:
> If I follow these steps I see a problem:
> 
> 1. Create new project and add input system package
> 2. Open project settings and click the create asset button - this works to create the asset, but the create button remains as the asset contents are displayed
> 3. Close project settings
> 4. Reopen project settings - asset is found and shown, no button for create
> 5. Delete the asset from the database
> 6. Open project settings again and as no asset, click create button - works to create it and show it
> 7. Close project settings
> 8. Reopen - **asset is not found/shown in UI** - so create button still present on its own  (but asset is present in the database)
> 
> Problem seems to be related to the InputSystem.asset assignment. The check to see if the asset is already assigned passes as the assigned value has a instanceId that matches.
> 
> Note - restarting editor and the asset is found again when opening the project settings 9 - restart editor 10 - open project settings - asset is found !
> 
> In step 5 deleting the asset from the database, the InputSystem.actions still has a reference to that removed file. It seems that when assigning the newly created file in 6 it checks if InputSystem.actions is the same as the new asset before assigning and the instanceID matches so it doesn't assign it over.
> 
> if we selected 'none' or a different asset and then reselect the newly created asset then the UI updates.

I believe this is fixed now.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/AssetContext.cs
ekcoh: No longer relevant since I removed the abstracted/decopled importer interfaces.

stefanunity:
PWA icon asset

If the only thing I can choose there is an Input Actions asset the icon should be an Input Actions one and not a script?

ekcoh:
> PWA icon asset
> 
> If the only thing I can choose there is an Input Actions asset the icon should be an Input Actions one and not a script?

Checked this with debugger and ObjectField is correctly assigned to InputActionAsset type so not sure whether this is default UI Toolkit ObjectField behavior for null value or if something else is missing. Will consult.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -8,230 +8,183 @@
...
         /// </summary>
-        internal static void UpdateInputActionReferences()
+        internal static bool CheckForDefaultUIActionMapChanges(InputActionAsset asset, IReportInputActionAssetValidationErrors reporter = null)
         {
-            var asset = GetOrCreate();
-            var existingReferences = InputActionImporter.LoadInputActionReferencesFromAsset(asset).ToList();
+            reporter ??= new DefaultInputActionAssetValidationReporter();
+
+            var defaultUIActionMap = GetDefaultUIActionMap();
+            var uiMapIndex = asset.actionMaps.IndexOf(x => x.name == "UI");
```
ritamerkl, line 156: looks dangerous comparing to a name that can potentially be changed internally or renamed on user side

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -8,230 +8,183 @@
...
 
-            // Check if referenced input action exists in the asset and remove the reference if it doesn't.
-            foreach (var actionReference in existingReferences)
+            // "UI" action map has been removed or renamed.
```
ritamerkl, line 158: would be better to find the UI action map from ID or similar to avoid getting an error after rename?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
```diff
@@ -543,13 +544,30 @@ public static Command ChangeCompositeName(int actionMapIndex, int bindingIndex,
...
+        {
+            return (in InputActionsEditorState state) =>
+            {
+                InputActionSerializationHelpers.DeleteAllActionMaps(state.serializedObject);
+                var temp = InputActionAsset.FromJson(inputActionAssetJsonContent);
+                using (var tmp = new SerializedObject(temp))
+                {
+                    InputActionSerializationHelpers.AddActionMaps(state.serializedObject, tmp);
+                }
+                state.serializedObject.ApplyModifiedProperties();
```
ritamerkl, line 575: ApplyModifiedProperties is called twice. In DeleteAllActionMaps and here. This could lead to double updates of the UI.

ritamerkl (commented):
Hard to grasp the full logic at once. Just leaving minor comments, I think this is worth testing wisely (also considering the changes in CopyPaste).

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Yes I agree this is a brittle design, this was however added already in https://github.com/Unity-Technologies/InputSystem/pull/1807/files. I have just moved the code around. The integration between UITK and ISX is not ideal here. Ideal would have been that the UITK part owned the actions and ISX only was allowed to rebind them and nothing else.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: I do not understand how that could be done since Input Actions are based on strings to begin with? How would you determine when you have a valid reference or how would you now the initial ID? How would you detect which one is valid? Also comes from  https://github.com/Unity-Technologies/InputSystem/pull/1807/files I just moved it around a bit.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
ekcoh: hmmm checked the code and I cannot spot the place you are not calling out? Would you mind post a link to the other invocation?

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Basically the integration as currently done is based on a string/name convention which is very brittle indeed.

ritamerkl: okay, if this is old code we are good for now I guess, maybe worth creating a ticket for this anyway

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
ritamerkl: nvm, I was looking at the wrong method

ekcoh: always happy for a false alarm :) will mark resolved.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Will mark this one resolved since outside the scope of this PR. This PR does not attempt to improve UITK / ISX integration.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: We discussed this quite a lot during Q4 just before it got implemented and given how InputForUI is implemented I see little improvement we can make due to this string based restriction. Will however file a ticket on visual validation. Filed ISX-1884 internally.

ekcoh: Will mark this resolved.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -106,6 +106,23 @@ public InputSettings settings
...
+        public InputActionAsset actions
+        {
+            get
+            {
+                return m_Actions;
+            }
+
+            set
+            {
+                /*if (m_Actions == value)
```
AlexTyrer: Some detritus left over from debugging?

AlexTyrer:
First pass read-through looks good (long though, there's a lot to look at!)

Will try some more testing on Monday.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -8,230 +8,183 @@
 
 namespace UnityEngine.InputSystem.Editor
 {
+    // Placeholder for converting InputManager.asset actions into regular asset to support conversion from 1.8.0-pre1 and 1.8.0-pre2 to asset based Project-wide actions
```
lyndon-unity: I assume not a placeholder anymore?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
```diff
@@ -2,16 +2,24 @@
 using System.Collections.Generic;
 using UnityEditor;
 using UnityEngine.UIElements;
+using UnityEditor.UIElements;
 
 namespace UnityEngine.InputSystem.Editor
 {
+    // TODO This editor should react to InputSystem.actions being reassigned from outside
```
lyndon-unity: Is this TODO still valid/

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3632,15 +3675,18 @@ internal static void TrackDirtyInputActionAsset(InputActionAsset asset)
...
 
 #else
-        private static void InitializeInPlayer(IInputRuntime runtime = null, InputSettings settings = null)
+        private static void InitializeInPlayer(IInputRuntime runtime = null, InputSettings settings = null, InputActionAsset actions = null)
         {
             if (settings == null)
                 settings = Resources.FindObjectsOfTypeAll<InputSettings>().FirstOrDefault() ?? ScriptableObject.CreateInstance<InputSettings>();
 
+            if (actions == null)
+                actions = Resources.FindObjectsOfTypeAll<InputActionAsset>().FirstOrDefault();
```
lyndon-unity: Do we really want to pick one if a user has set this to null.
I don't think we do for the actions.

lyndon-unity: This was causing 4 tests to fail so I've removed it.

ekcoh:
@duckets There seems to be a test failure on this branch (missing tag / broken link) that needs to be addressed: https://unity-ci.cds.internal.unity3d.com/job/34173140?utm_source=github

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Good catch, will clean that up

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -116,7 +116,7 @@ private void Destroy(Object obj)
 #if UNITY_EDITOR
         Object.DestroyImmediate(obj);
 #else
-        Object.Destroy(actions);
+        Object.DestroyImmediate(actions);
```
ekcoh, line 119: According to documentation of Object.DestroyImmediate and Object.Destroy that should not be done? https://docs.unity3d.com/ScriptReference/Object.DestroyImmediate.html

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3680,9 +3680,6 @@ private static void InitializeInPlayer(IInputRuntime runtime = null, InputSettin
             if (settings == null)
                 settings = Resources.FindObjectsOfTypeAll<InputSettings>().FirstOrDefault() ?? ScriptableObject.CreateInstance<InputSettings>();
 
-            if (actions == null)
```
ekcoh: Removing this would completely bypass player loading of preloaded asset? Cannot see how this change is valid and would like to revert it. This area anyway need to be addressed additionally since there are existing problems with this an InputForUI integration. There are also related problems to static object stack push/pop within InputSystem/InputManager.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -565,39 +565,43 @@ void RegisterActions(Configuration cfg)
...
-            m_MoveAction = InputActionReference.Create(m_InputActionAsset.FindAction(m_Cfg.MoveAction));
-            m_SubmitAction = InputActionReference.Create(m_InputActionAsset.FindAction(m_Cfg.SubmitAction));
-            m_CancelAction = InputActionReference.Create(m_InputActionAsset.FindAction(m_Cfg.CancelAction));
-            m_LeftClickAction = InputActionReference.Create(m_InputActionAsset.FindAction(m_Cfg.LeftClickAction));
-            m_MiddleClickAction = InputActionReference.Create(m_InputActionAsset.FindAction(m_Cfg.MiddleClickAction));
-            m_RightClickAction = InputActionReference.Create(m_InputActionAsset.FindAction(m_Cfg.RightClickAction));
-            m_ScrollWheelAction = InputActionReference.Create(m_InputActionAsset.FindAction(m_Cfg.ScrollWheelAction));
-
-            if (m_PointAction.action != null)
+            if (m_InputActionAsset != null)
```
ekcoh, line 580: I believe this just addresses the null checking. There is also a need to monitor any changes to project wide actions in run-time and reconfigure based on that or the concept needs modified to be more strict to not allow changes during run-time but then its an editor only feature and would basically imply that a setter on InputSystem.actions shouldn't exist.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3524,6 +3555,12 @@ internal static void InitializeInEditor(IInputRuntime runtime = null)
             }
             s_SystemObject.newInputBackendsCheckedAsEnabled = true;
 
+#if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
+            // Make sure project wide input actions are enabled
+            if (actions != null)
```
ekcoh, line 3560: This is the second time actions are enabled in this code path since also done via Reset above. Needs to be checked which of these is correct since Reset is also invoked when static state stack is pushed/popped.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: I have removed it in pending commit

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Removed

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
ekcoh: Yes, but I will remove the TODOs and cover in Open Issues part of PR instead.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -80,6 +80,8 @@ public void Initialize()
 
             m_Cfg = Configuration.GetDefaultConfiguration();
             RegisterActions(m_Cfg);
+
```
ekcoh, line 83: InputSystem.actions and InputSystem.onActionsChange are only available when UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS is defined, but there are no symbolic checks here. In addition this is compiled into its own assembly so cannot detect symbols from Input System defines in asmdef and this seems to only rely on UNITY version which makes this broken if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS is undefined?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
ekcoh: Removed

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
lyndon-unity: Good spot.

I've reveted this and fixed 2 of the tests instead, with the knowledge that this can pickup an Asset from the test itself.

I am concerned with this way of loading the InputActionAsset though. If there are multiple in the project then it seems this could load the wrong one.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
lyndon-unity: The one in Reset is not used in Editor during the initialisation flow as its too early and at that point the value is null. So this line is needed then.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
lyndon-unity: True, but that issue is not introduced by this PR.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
lyndon-unity: Reverted - see later comment

lyndon-unity (approved):
Approving to land. This addresses a lot of the main issues of the InputManager.asset storage.
Any further improvements can land in follow up PR's

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: Agree that this is too brittle, seems to be the same case for InputSwttings from previously. Likely needs an internal property serialized that allows tagging it unless there is a better way or use a dedicated path. This needs a ticket and a test case.

PR #1835, opened 2024-02-08
Add UI Toolkit matching control bindings UI

lyndon-unity:
### Description

Implemented UI TK version of the matching control bindings recently added to the IMGUI editor
https://github.com/Unity-Technologies/InputSystem/pull/1672

Example of it working in the project wide input actions. This is the same UX used in the individual assets.
image

### Changes made

The code to populate the UX is the same as the IMGUI code - but split out to generate a tree before then generating the UI specific tree.

Also fixed up some styling on the control scheme info which is in the binding UI to be more consistent with the IMGUI version. Bolded and fixed margins.

### Notes

Note this version doesn't store the full tree state - just the expanded/not level.
This means I didn't need to add the static dictionary which was present in the IMGUI version.
I didn't feel this was critical and reduced static state.

I also had to make the top level expansion based on e checkbox rather than a 'fold out' / top of tree as the UI TK treeView doesn't have API to react to the expand/collapse state.
^ This comment is old - I've added a Foldout to replace the Toggle now

### Checklist

Before review:

- Changelog entry added.
- No tests added/changed
- No new doc

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/BindingPropertiesView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/BindingPropertiesView.cs
@@ -76,37 +76,43 @@ namespace UnityEngine.InputSystem.Editor
 
         private void DrawMatchingControlPaths(ViewState viewState)
         {
-            var checkbox = new Toggle("Show Matching Paths")
-            {
-                value = viewState.showPaths
-            };
-            rootElement.Add(checkbox);
-
             List<MatchingControlPath> matchingControlPaths = CollectMatchingControlPaths(viewState.selectedBindingPath.stringValue, viewState);
             m_MatchingControlPaths = BuildMatchingControlPaths(matchingControlPaths);
 
-            var treeView = new TreeView();
-            rootElement.Add(treeView);
-            treeView.visible = viewState.showPaths;
-            treeView.style.flexGrow = viewState.showPaths ? 1 : 0;
-
-            checkbox.RegisterValueChangedCallback(changeEvent =>
+            if (m_MatchingControlPaths.Count > 0)
             {
-                viewState.showPaths = changeEvent.newValue;
-                treeView.visible = viewState.showPaths;
-                treeView.style.flexGrow = viewState.showPaths ? 1 : 0;
+                var treeView = new TreeView();
+                rootElement.Add(treeView);
+                treeView.fixedItemHeight = 20;
+                treeView.SetRootItems(m_MatchingControlPaths);
 
-                Dispatch(Commands.ShowMatchingPaths(viewState.showPaths));
-            });
+                // Set TreeView.makeItem to initialize each node in the tree.
+                treeView.makeItem = () =>
+                {
+                    var label = new Label();
+                    label.RegisterCallback<ClickEvent>((evt) => OnItemClicked(evt, viewState));
+                    label.AddToClassList("matching-paths");
+                    return label;
+                };
 
-            treeView.SetRootItems(m_MatchingControlPaths);
+                // Set TreeView.bindItem to bind an initialized node to a data item.
+                treeView.bindItem = (VisualElement element, int index) =>
+                {
+                    var label = (element as Label);
+                    label.text = treeView.GetItemDataForIndex<MatchingControlPath>(index).path;
+                };
 
-            // Set TreeView.makeItem to initialize each node in the tree.
-            treeView.makeItem = () => new Label();
+                //if (viewState.showPaths)
+                   treeView.ExpandRootItems();
+            }
+        }
 
-            // Set TreeView.bindItem to bind an initialized node to a data item.
-            treeView.bindItem = (VisualElement element, int index) =>
-                (element as Label).text = treeView.GetItemDataForIndex<MatchingControlPath>(index).path;
+        private void OnItemClicked(ClickEvent evt, ViewState viewState)
+        {
+            var element = evt.target as VisualElement;
+               
+            // TODO: Move this to the expend/collapse state of the first node - rather than the label click state
+            // Dispatch(Commands.ShowMatchingPaths(viewState.showPaths));
         }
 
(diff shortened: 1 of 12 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/BindingPropertiesView.cs
```diff
@@ -64,10 +69,261 @@ public override void RedrawUI(ViewState viewState)
...
                 rootElement.Add(controlPathContainer);
 
+                DrawMatchingControlPaths(viewState);
                 DrawControlSchemeToggles(viewState, binding.Value);
             }
         }
 
+        private void ShowDerivedBindings(ViewState viewState)
+        {
+        }
```
ritamerkl: Is this method unused?

ritamerkl (commented):
Good to see this working! 
I just have one major point- I think it could be worth outsourcing the biggest part of the DrawMatchingControlPath and connected methods to an external script (or other related existing classes). Like that we could keep the PropertiesView clean and "logic free". 
I also added one minor comment on a potential unused method to remove.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/BindingPropertiesView.cs
lyndon-unity: Yes I think I forgot to remove - will do that now.

lyndon-unity: Removed now.

lyndon-unity:
> Good to see this working! I just have one major point- I think it could be worth outsourcing the biggest part of the DrawMatchingControlPath and connected methods to an external script (or other related existing classes). Like that we could keep the PropertiesView clean and "logic free". I also added one minor comment on a potential unused method to remove.

Yes I agree I can take a look at spitting those out

ritamerkl (approved):
LGTM now

Pauliusd01:
Is the style used wrong? Why is the derived binding text selectable? It wasn't like that in the IMGUI version and I don't see why that should be changed

Pauliusd01:
https://github.com/Unity-Technologies/InputSystem/assets/54306142/d77c6a20-4fc2-43e3-b976-4d44f3b3cc78

Since this is a toggle now I think its state should be saved, seems annoying to have to re-enable it between window restarts

Pauliusd01 (changes requested):
Just minor notes from me. Functionality seems the same and works without issues

lyndon-unity:
> Is the style used wrong? Why is the derived binding text selectable? It wasn't like that in the IMGUI version and I don't see why that should be changed

I've fixed that - it was just the default for the UI TK tree view.
I've set it to none to match IMGUI version

lyndon-unity:
> I think its state should be saved

I'd like to do that but I don't have a quick fix for that so might land the update so far and look at that later.

Pauliusd01:
> > I think its state should be saved
> 
> I'd like to do that but I don't have a quick fix for that so might land the update so far and look at that later.

Does it have to be a toggle then? If a foldout would solve this issue I'd prefer that, and later on we can switch it to a toggle once state saving is there

lyndon-unity:
> > > I think its state should be saved
> > 
> > 
> > I'd like to do that but I don't have a quick fix for that so might land the update so far and look at that later.
> 
> Does it have to be a toggle then? If a foldout would solve this issue I'd prefer that, and later on we can switch it to a toggle once state saving is there

State is now maintained over open/close to match IMGUI.
I've just switched to a foldout now too.

ekcoh (approved):
LGTM, is all logic in MatchingControlPaths new? If it is then its fine as-is, otherwise it might be worthwhile avoiding duplicating logic with existing feature implementation for IMGUI? It seems like MatchingControlPaths is not tied to UI framework and would be possible to cover with unit tests?

lyndon-unity:
> LGTM, is all logic in MatchingControlPaths new? If it is then its fine as-is, otherwise it might be worthwhile avoiding duplicating logic with existing feature implementation for IMGUI? It seems like MatchingControlPaths is not tied to UI framework and would be possible to cover with unit tests?

The MatchingControlPaths isn't all new code - its adapted from the IMGUI version. It drops some of the code for retaining the expand/collapse state of individual parts of the hierarchy as that had a static dictionary that seemed lower priority for the memory overheads. To use in IMGUI directly we would have to add back in the static dictionary and store extra state in the returned data. I decided to duplicate as the IMGUI version would be deprecated in future.
^ but I have now removed duplicated code and refactored a little so it can be shared with IMGUI and UI TK.

Tests should be added but I felt given time constraints that part could be added later.

Pauliusd01 (approved):
LGTM, checked that the matching bindings are correct for different bindings in both ProjectWideActions and custom action assets.

ekcoh:
> > LGTM, is all logic in MatchingControlPaths new? If it is then its fine as-is, otherwise it might be worthwhile avoiding duplicating logic with existing feature implementation for IMGUI? It seems like MatchingControlPaths is not tied to UI framework and would be possible to cover with unit tests?
> 
> The MatchingControlPaths isn't all new code - its adapted from the IMGUI version. It drops some of the code for retaining the expand/collapse state of individual parts of the hierarchy as that had a static dictionary that seemed lower priority for the memory overheads. To use in IMGUI directly we would have to add back in the static dictionary and store extra state in the returned data. I decided to duplicate as the IMGUI version would be deprecated in future. ^ but I have now removed duplicated code and refactored a little so it can be shared with IMGUI and UI TK.
> 
> Tests should be added but I felt given time constraints that part could be added later.

I see, great to avoid duplicating too much. OK to postpone the test part temporarily but in that case please file a ticket on the remaining part from your perspective.

PR #1836, opened 2024-02-09
FIX: Input Action file is marked as dirty when opening after renaming it (ISXB-749)

timkeo:
### Description

Fixes issue [ISXB-749](https://jira.unity3d.com/browse/ISXB-749) and also another bug [ISX-1874](https://jira.unity3d.com/browse/ISX-1874) I discovered along the way

### Changes made

Adds AssetPostprocessor handler that checks if the serialized "Name" field in the JSON (for the InputAction being imported) matches the in-memory name. If it doesn't match, then the proper name is written to the file. I discovered this bug also occurs if an InputAction asset is copied (not just renamed) and so a post-process handler was the best way to catch all the scenarios.

I added a new test to validate the asset file contents do indeed match the in memory name during these operations.

I also added a simple fix for another issue I encountered during my investigation: while opening the InputActionEditor window, checks if the target asset is valid. If not it immediately closes the window.

### Notes

Closing the InputActionEditor window from within the CreateGUI() handler will cause a minor visual artifact: it'll briefly display for a frame or two causing a "flash". Since this is such a minor corner-case, and I couldn't find a safe alternative, I figured this would be fine.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Editor.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Editor.cs
@@ -743,6 +743,73 @@ partial class CoreTests
         Assert.That(map.bindings[0].groups, Is.EqualTo(""));
     }
 
+    struct AssetFileTestConstants
+    {
+        public const string kOriginalAssetName = "zzMyInputActions";
+        public const string kOriginalDirectory = "zzStartingDirectory";
+
+        public const string kNewAssetName = "NEWactions";
+        public const string kNewDirectory = "NewDirectory";
+
+        public const string kOriginalAssetPath = "Assets/" + kOriginalDirectory + "/" + kOriginalAssetName + "." + InputActionAsset.Extension;
+        public const string kOriginalAssetContents = "{\"name\": \"" + kOriginalAssetName + "\",\"maps\": [],\"controlSchemes\": []}";
+    }
+
+    [Test]
+    [TestCase("Assets/" + AssetFileTestConstants.kOriginalDirectory + "/" + AssetFileTestConstants.kNewAssetName + "." + InputActionAsset.Extension, AssetFileTestConstants.kNewAssetName, false)]      // Move - Same directory but new filename - expect changed name
+    [TestCase("Assets/" + AssetFileTestConstants.kNewDirectory + "/" + AssetFileTestConstants.kNewAssetName + "." + InputActionAsset.Extension, AssetFileTestConstants.kNewAssetName, false)]           // Move - New directory and new filename - expect changed name
+    [TestCase("Assets/" + AssetFileTestConstants.kNewDirectory + "/" + AssetFileTestConstants.kOriginalAssetName + "." + InputActionAsset.Extension, AssetFileTestConstants.kOriginalAssetName, false)] // Move - New directory but same filename - expect original name
+    [TestCase("Assets/" + AssetFileTestConstants.kOriginalDirectory + "/" + AssetFileTestConstants.kNewAssetName + "." + InputActionAsset.Extension, AssetFileTestConstants.kNewAssetName, true)]       // Copy - Same directory but new filename - expect changed name
+    [TestCase("Assets/" + AssetFileTestConstants.kNewDirectory + "/" + AssetFileTestConstants.kNewAssetName + "." + InputActionAsset.Extension, AssetFileTestConstants.kNewAssetName, true)]            // Copy - New directory and new filename - expect changed name
+    [TestCase("Assets/" + AssetFileTestConstants.kNewDirectory + "/" + AssetFileTestConstants.kOriginalAssetName + "." + InputActionAsset.Extension, AssetFileTestConstants.kOriginalAssetName, true)]  // Copy - New directory but same filename - expect original name
+    [Category("Editor")]
+    public void Editor_InputActions_AssetFileUpdatedAfterMoveOrCopy(string newAssetPath, string expectedAssetName, bool executeCopy)
+    {
+        try
+        {
+            AssetDatabase.CreateFolder("Assets", AssetFileTestConstants.kOriginalDirectory);
+            AssetDatabase.CreateFolder("Assets", AssetFileTestConstants.kNewDirectory);
+
+            File.WriteAllText(AssetFileTestConstants.kOriginalAssetPath, AssetFileTestConstants.kOriginalAssetContents);
+            AssetDatabase.ImportAsset(AssetFileTestConstants.kOriginalAssetPath);
+
+            var asset = AssetDatabase.LoadAssetAtPath<InputActionAsset>(AssetFileTestConstants.kOriginalAssetPath);
+            Assert.NotNull(asset, "Could not load asset: " + AssetFileTestConstants.kOriginalAssetPath);
+
+            try
+            {
+                if (executeCopy)
+                {
+                    AssetDatabase.CopyAsset(AssetFileTestConstants.kOriginalAssetPath, newAssetPath);
+                    AssetDatabase.Refresh();
+                }
+                else AssetDatabase.MoveAsset(AssetFileTestConstants.kOriginalAssetPath, newAssetPath);
+
+                var newAsset = ScriptableObject.CreateInstance<InputActionAsset>();
+                var fileContents = File.ReadAllText(newAssetPath);
+                newAsset.LoadFromJson(fileContents);
+
+                Assert.That(newAsset.name, Is.EqualTo(expectedAssetName));
+            }
+            finally
+            {
+                AssetDatabase.DeleteAsset(AssetFileTestConstants.kOriginalAssetPath);
+                AssetDatabase.DeleteAsset(newAssetPath);
+            }
+        }
+        finally
+        {
+            const string kOriginalPath = "Assets/" + AssetFileTestConstants.kOriginalDirectory;
+            const string kNewPath = "Assets/" + AssetFileTestConstants.kNewDirectory;
+
+            FileUtil.DeleteFileOrDirectory(kOriginalPath);
+            FileUtil.DeleteFileOrDirectory(kOriginalPath + ".meta");
+            FileUtil.DeleteFileOrDirectory(kNewPath);
+            FileUtil.DeleteFileOrDirectory(kNewPath + ".meta");
+            AssetDatabase.Refresh();
+        }
+    }
+
     [Test]
     [Category("Editor")]
     public void Editor_InputActionAssetManager_CanMoveAssetOnDisk()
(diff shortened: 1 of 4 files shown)


Pauliusd01 (changes requested):
Creating new input action assets throws this warning:
```
Import Error Code:(4)
Message: Build asset version error: assets/new controls.inputactions in SourceAssetDB has modification time of '2024-02-09T08:37:49.2661504Z' while content on disk has modification time of '2024-02-09T08:37:49.3020947Z'

UnityEngine.GUIUtility:ProcessEvent (int,intptr,bool&)
```

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -743,6 +743,73 @@ public void Editor_InputAsset_CanReplaceBindingGroupThroughSerialization()
         Assert.That(map.bindings[0].groups, Is.EqualTo(""));
     }
 
+    struct AFTestConstants
```
ekcoh: Is this "AF" capitalised prefix representing "Asset File" or what is it intended to represent? I would recommend renaming it and avoid "unknown" acronyms if this is the case and call it "AssetFileTestConstants"

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -743,6 +743,73 @@ public void Editor_InputAsset_CanReplaceBindingGroupThroughSerialization()
...
+        public const string kOriginalDirectory = "zzStartingDirectory";
+
+        public const string kNewAssetName = "NEWactions";
+        public const string kNewDirectory = "NewDirectory";
+
+        public const string kOriginalAssetPath = "Assets/" + kOriginalDirectory + "/" + kOriginalAssetName + "." + InputActionAsset.Extension;
+        public const string kOriginalAssetContents = "{\"name\": \"" + kOriginalAssetName + "\",\"maps\": [],\"controlSchemes\": []}";
+    }
+
+    [Test]
```
ekcoh, line 758: Great you added these test cases and permutations of tests to cover affected functionality.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -256,6 +256,39 @@ public static void CreateInputAsset()
             ProjectWindowUtil.CreateAssetWithContent("New Controls." + InputActionAsset.Extension,
                 InputActionAsset.kDefaultAssetLayoutJson, InputActionAssetIconLoader.LoadAssetIcon());
         }
+
+        // When an action asset is renamed, copied, or moved in the Editor, the "Name" field in the JSON will
+        // hold the old name and won't match what's in memory (asset looks "dirty"). To work around this, we
+        // must flush the updated JSON to the file whenever this operation occurs.
+        // https://jira.unity3d.com/browse/ISXB-749
+        private class PostProcessAssetModifications : AssetPostprocessor
```
ekcoh: Interesting this was the missing piece for this particular bug. I stumbled upon this myself while working on https://jira.unity3d.com/browse/ISX-1865 but from another angle. The missing functionality I identified is what is being done here for the "legacy" IMGUI based editor: https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs#L856-L914
Also of relevance is: https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
I haven't fully landed in how this should work given that the functionality present there seems mainly focused on updating the editor if the file on disc changes while the editor content is unmodified. 
So in addition to the scenarios you have covered here reflecting asset modifications via files it is relevant to consider what to expect for scenarios where the file modifications happen and there is also in-memory modifications of the asset in Unity editors. I am afraid that is not possible to solve in a good way apart from selecting either update-from-disc or overwrite-to-disc. Due to this I only comment for now and try to think about this a bit more and get back to you tomorrow. Feel free to share your perspective as well based on existing implementation above. It clearly looks like the UITK editor was constructed overlooking these aspects from the older editor implementation.

ekcoh (commented):
In general the fix looks really good but I am afraid there is more to this, see current comment. Will circle back to this PR tomorrow.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
timkeo: Unfortunately I've been unable to find a real solution to the Warnings logged to the console (which Paulinus pointed out) when creating a new asset and in a couple other scenarios. This leads me to believe this current approach is wrong, and we should not be modifying the asset file during the import process. So, even if we decide the fix itself is correct, it needs to be executed somewhere else and not during import.

I do see this more as a fundamental problem with InputActionAsset handling and not just an oversight with UITK's implementation. I did try a scenario in which the asset file was moved while the ActionEditor window was open (with unsaved changes), resulting in a FileNotFound exception with the in-memory changes failing to save.

However, I don't think this is the most important problem to solve right now, and as you said there isn't a good solution. It might be better to implement a work-around fix for the UITK ActionsEditor and revisit the larger issue later.

timkeo:
I pushed a rebased branch (to resolve conflicts) and updated my implementation to properly update the asset file outside of Asset Importing operation (fixes the warning logged to console).

Basically, I perform the same logic to determine if a file as been renamed/copied, but store the assetPath in a list (don't modify the file). Later within OnPostprocessAllAssets(), where it's safe to make asset file changes, I write updated JSON contents for each "tagged" asset in the list.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -256,6 +256,74 @@ public static void CreateInputAsset()
...
+                    foreach (var assetPath in assetFilesNeedingRefresh)
+                    {
+                        var newAsset = ScriptableObject.CreateInstance<InputActionAsset>();
+
+                        try
+                        {
+                            var fileContents = File.ReadAllText(assetPath);
+
+                            newAsset.LoadFromJson(fileContents);
+                            newAsset.name = Path.GetFileNameWithoutExtension(assetPath);
```
ekcoh, line 310: II do not understand why we need to do this at all. This is always unconditionally assigned to InputActionAsset when being imported? The name present in the JSON is, as far as I can tell completely irrelevant or am I missing something?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -155,10 +155,20 @@ private void CreateGUI()
                 if (m_State.serializedObject == null)
                 {
                     var asset = GetAssetFromDatabase();
-                    m_AssetPath = AssetDatabase.GetAssetPath(asset);
-                    m_AssetJson = File.ReadAllText(m_AssetPath);
-                    var serializedAsset = new SerializedObject(asset);
-                    m_State = new InputActionsEditorState(m_State, serializedAsset);
+                    if (asset != null)
```
ekcoh, line 158: Makes sense, I believe something similar is done in InputActionEditorWindow (notice the lack of an 's' which is the previous IMGUI based editor)

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
ekcoh: In the importer the JSON name is completely discarded. I guess it only is relevant when loading JSON files manually and parsing them into InputActionAssets if some kind of identification is needed? Not sure where this caused a problem?

ekcoh: I see now, it basically generates a diff just because the name is different since the name is adjusted when converted back into JSON..... I guess this fix makes sense then.

ekcoh: Wouldn't be cheaper and make more sense to use this kind of processor? https://docs.unity3d.com/ScriptReference/AssetModificationProcessor.OnWillMoveAsset.html

ekcoh: As an example see its usage in legacy editor. https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs#L896

ekcoh (approved):
After some thinking I realize the fix makes sense. Its just unfortunate we need to process assets again just for this reason. I guess one alternative could have been to ignore the name in dirty comparison. I believe this is one of the reasons its ignored in InputAssetManager.FromJson/ToJson functions... I will approve this fix but leaving some comments with changes/suggestions for you to judge if relevant or not.

timkeo:
>Wouldn't be cheaper and make more sense to use this kind of processor?

That was actually my original implementation, which worked well enough for Move/Rename but did not handle the Copy scenario.

With that interface you have to use [OnWillCreateAsset](https://docs.unity3d.com/ScriptReference/AssetModificationProcessor.OnWillCreateAsset.html) to handle copying an asset, but since this occurs before the file is even created we can't compare the serialized name with the filename. Also, there doesn't appear to be a straightforward way to determine the source asset in the copy scenario using that handler.

I considered using both interfaces and handle the Move/Rename and Copy scenarios separately, but that would make the implementation more complex, and I decided it wasn't worth it. Unlike textures or meshes, we shouldn't have a lot of .inputactions files to process and the scenarios that trigger re-importing should be infrequent.

I agree this is a heavy-handed process just to fix this rename/move/copy issue, but I'm thinking (later) we can build some additional logic into this to determine if the asset was edited outside the editor. Then, if the ActionsEditor window is open, we can refresh the UI with the new contents (if there aren't any local changes), or prompt the user if is there are unsaved changes.

I tried implementing some of this logic yesterday but couldn't get it fully working, and so set it aside for now.

timkeo:
@Pauliusd01

We need to land this today so @ekcoh can move ahead with [ProjectWideActions re-implementation PR](https://github.com/Unity-Technologies/InputSystem/pull/1834).

The Warnings logged to the console are fixed, and the other tests are green. If you do discover anything else wrong, please file a bug and assign it to me.

PR #1843, opened 2024-02-15
FIX: ArgumentNullExceptions thrown when deleting items quickly in the…

timkeo:
### Description

The issue occurs because the Delete command gets "repeated" when holding down the keyboard key, which is dispatched to the view before the UI has updated from the previous one. This causes the execution of the Delete to reference an Action that no longer exists within the backing data leading to a null SerializedPropery reference when we try to retrieve it.

To guard against this, block UI commands from executing until the UI has updated form the previous one. This won't impact human usage (occurs far too quickly) but will block program driven commands like key repeat and other programmatic sources.

### Changes made

Added a flag to act as a type of Semaphore and only allow a single UI command to execute per UI Update. Since it's single-threaded we don't need an _actual_ Semaphore.

### Notes

This could affect UI tests, if they send multiple UI commands per frame, causing them to fail.
Might be some other side-effects with corner cases.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
@@ -148,31 +148,38 @@ namespace UnityEngine.InputSystem.Editor
             var selectedItem = m_ListView.GetRootElementForIndex(m_ListView.selectedIndex);
             if (selectedItem == null)
                 return;
-            switch (evt.commandName)
+
+            if (allowUICommandExecution)
             {
-                case CmdEvents.Rename:
-                    ((InputActionMapsTreeViewItem)selectedItem).FocusOnRenameTextField();
-                    break;
-                case CmdEvents.Delete:
-                case CmdEvents.SoftDelete:
-                    ((InputActionMapsTreeViewItem)selectedItem).DeleteItem();
-                    break;
-                case CmdEvents.Duplicate:
-                    ((InputActionMapsTreeViewItem)selectedItem).DuplicateItem();
-                    break;
-                case CmdEvents.Copy:
-                    CopyItems();
-                    break;
-                case CmdEvents.Cut:
-                    CutItems();
-                    break;
-                case CmdEvents.Paste:
-                    var isActionCopied = CopyPasteHelper.GetCopiedClipboardType() == typeof(InputAction);
-                    if (CopyPasteHelper.HasPastableClipboardData(typeof(InputActionMap)))
-                        PasteItems(isActionCopied);
-                    break;
-                default:
-                    return; // Skip StopPropagation if we didn't execute anything
+                switch (evt.commandName)
+                {
+                    case CmdEvents.Rename:
+                        ((InputActionMapsTreeViewItem)selectedItem).FocusOnRenameTextField();
+                        break;
+                    case CmdEvents.Delete:
+                    case CmdEvents.SoftDelete:
+                        ((InputActionMapsTreeViewItem)selectedItem).DeleteItem();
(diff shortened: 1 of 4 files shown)


ritamerkl (approved):
LGTM. My first thought was to call the boolean isProcessingUICommand or something similar - I think it is a bit more descriptive of what it really is, but that's just a minor food for thought.

timkeo:
> to call the boolean isProcessingUICommand or something similar

I actually struggled a bit to decide on what to call this and considered something similar.

Ultimately I landed on `allowUICommandExecution` because it makes the check "positive", That is with `isProcessingUICommand` the check becomes "negative": `if (!isProcessingUICommand) {...}`. Basically it comes down to an old coding standard to use "positive" condition statements whenever possible because they're easier to reason through.

timkeo:
@Pauliusd01

My concern with this change is it could break some corner-case scenario in which UITK dispatches 2 UI commands simultaneously. I can't think of a scenario (outside of Unit Tests) in which this would occur, but could you please verify other commands like Cut/Copy/Paste, Rename, etc. all still work, especially when executed from a keyboard shortcut or a menu selection.

Thanks!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionMapsView.cs
```diff
@@ -148,31 +148,38 @@ private void OnExecuteCommand(ExecuteCommandEvent evt)
...
-                    CutItems();
-                    break;
-                case CmdEvents.Paste:
-                    var isActionCopied = CopyPasteHelper.GetCopiedClipboardType() == typeof(InputAction);
-                    if (CopyPasteHelper.HasPastableClipboardData(typeof(InputActionMap)))
-                        PasteItems(isActionCopied);
-                    break;
-                default:
-                    return; // Skip StopPropagation if we didn't execute anything
+                switch (evt.commandName)
```
ekcoh, line 157: Would recommend moving this switch statement into its own function for readability when there such a long switch statement

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -373,35 +373,41 @@ private void OnExecuteCommand(ExecuteCommandEvent evt)
...
-                    break;
-                case CmdEvents.Paste:
-                    var hasPastableData = CopyPasteHelper.HasPastableClipboardData(data.isAction ? typeof(InputAction) : typeof(InputBinding));
-                    if (hasPastableData)
-                        PasteItems();
-                    break;
-                default:
-                    return; // Skip StopPropagation if we didn't execute anything
+                var data = (ActionOrBindingData)m_ActionsTreeView.selectedItem;
+                switch (evt.commandName)
```
ekcoh, line 388: Same thing here

ekcoh (approved):
Looks like a valid mitigation to the reported issue. Some minor comments.

ekcoh:
I resolved the merge conflict on this branch to rerun CI @timkeo

Pauliusd01 (approved):
LGTM. Deleting quick via shortcut or context menu seemed fine for both Project Wide Actions and custom ones, also does not seem to affect undo/redo negatively.  As for Tim's comment, did not see any issues with renaming/cutting/copying/duplicating or drag and dropping

PR #1844, opened 2024-02-15
FIX: Allow Copy/Paste between windows (ISX-1823).

graham-huws:
### Description

Fix for [ISX-1823](https://jira.unity3d.com/browse/ISX-1823).

### Changes made

This was previously not working because the Paste command would only apply to the relevant Action Maps/Actions List/Tree, so if neither of those were focused, nothing would happen. We now also listen for Paste at the top level of the element hierarchy and send paste commands from there.

### Notes

Issues:
• Undoing adding an Action Map/Action leaves the list unselected. Pasting at that point fails because the state thinks there's still a valid item selected. Mentioned [here](https://unity.slack.com/archives/C04RDRXRJ1L/p1707842657730249?thread_ts=1707220445.607469&cid=C04RDRXRJ1L), and should be added as a separate bug.
• Currently won't work if the window is tabbed to - user has to click somewhere in the window. Think I'll need to ask UITK team about this.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed InputManager.asset file growing in size on each Reset call.
 - Fixed Opening InputDebugger throws 'Action map must have state at this point' error
 - Fixed Cut/Paste behaviour to match Editor - Cut items will now be cleared from clipboard after pasting.
+- Fixed Pasting items between windows not working until an item in the Action Map/Action lists was selected.
```
ekcoh: Would be good to be more specific and clarify that "windows" here refer to "Input Action Editor windows". Also good to be more specific about "not working", instead suggest mentioning what would happen.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
```diff
@@ -288,13 +292,14 @@ private static void PasteAction(SerializedProperty arrayProperty, string jsonToI
         private static int PasteBindingOrComposite(SerializedProperty arrayProperty, string json, int index, string actionName, bool createCompositeParts = true)
         {
             var pastePartOfComposite = IsPartOfComposite(json);
-            if (index > 0)
+            if (index > 0 && arrayProperty.arraySize > 0 && index - 1 < arrayProperty.arraySize)
```
ekcoh: I am confused about an index that must be greater than zero and at least two elements from the end of the anway? Would recommend adding an inline comment if this is correct but some kind of special case.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
```diff
@@ -351,7 +356,7 @@ private static bool IsPartOfComposite(string json)
         private static SerializedProperty AddElement(SerializedProperty arrayProperty, string name, int index = -1)
         {
             var uniqueName = InputActionSerializationHelpers.FindUniqueName(arrayProperty, name);
-            if (index < 0)
+            if (index < 0 || index > arrayProperty.arraySize)
```
ekcoh, line 391: Is index == arrayProperty.arraySize here?

ekcoh:
As a side-note, when https://github.com/Unity-Technologies/InputSystem/pull/1834 lands there will exist an interesting scenario for this functionality when the same asset is open simultaneously in two windows, both in Project Settings and in a free-floating Input Action Editor.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
```diff
@@ -204,6 +204,10 @@ private static void PasteActionsFromClipboard(InputActionsEditorState state, boo
             var index = state.selectedActionIndex;
             if (addLast)
                 index = actionArray.arraySize - 1;
+
+            if (index < 0)
+                index = 0;
```
ritamerkl, line 219: consider use Math.Clamp

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
```diff
@@ -159,6 +169,49 @@ public class ViewState
             public IEnumerable<InputControlScheme> controlSchemes;
             public int selectedControlSchemeIndex;
         }
+
+        void OnExecuteCommand(ExecuteCommandEvent evt)
```
ritamerkl, line 252: Feels like double work - do we handle events twice here? (additional to handling them in the InputActionsTreeView / InputActionMapsView)

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
graham-huws-unity: This somewhat doubled up from merging develop - should be clearer now :)

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
graham-huws-unity: Not sure what you mean by this sorry?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
graham-huws-unity: If you mean, could both be called with one action - I don't believe this is possible since we StopPropagation.

If you mean there's code overlap - I think I've now reduced the code in this section to the point where reducing it further would just mean adding an extra string comparison and function call to jump to the maps/actions views' OnExecuteCommand calls, but let me know what you think 🙂

On a somewhat related note to that - do we need the HasPastableClipboardData calls in OnExecuteCommand in ActionsTreeView/ActionMapsView? Won't that be checked anyway in one of the functions that PasteItems calls? It feels like that's logically where it should happen, if it doesn't already.

ritamerkl: Is it? Even with using the events on TrickleDown? If the second event is used OnBubbleUp this means we could catch the event beforehand. 
I thought about a delegation concept when I said "do we handle events twice", a event handler somewhere that delegates to the list/treeView. It was more a rough idea, if the trickledown/bubbleup does not work out. 
About your second comment: I think there is a safety check, and if we could indeed skip the one in the list/treeviews.

graham-huws-unity: I think I'm not sure what the problem you want to solve is - do you believe there's a way the event would be acted upon twice? If so, how? (I'm definitely new to the events system!) If it's just that you'd prefer the actual Paste command is only in one place, fair enough, but an intermediary handler to decide which View to send the command to would surely be more complex than what's here in the editor View?
Apologies if I'm entirely misunderstanding!

graham-huws-unity: Ah - were you suggesting removing the Paste event handlers for the Action Maps/Actions, and having a single handler for the entire window? And then events from anywhere lower in the hierarchy can bubble up to the root?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
ekcoh: Sorry for the unclear question, I was wondering if index == arrayProperty.arraySize is actually a valid index in this function? Its typically not which catches my attention

graham-huws-unity: Ah, gotcha - yes it's fine, it's used for InsertArrayElementAtIndex where it's valid to add an element at the end of the array

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
ritamerkl: yes, that was what I was thinking of as one solution. The other way would be to catch the events at actionMap/actionTreeView on trickleDown, and stop the event propagation and then only if nothing was registered there we catch the rest of the events on bubble up on the root element to handle them from there. 
For your reference: https://docs.unity3d.com/2021.2/Documentation/Manual/UIE-Events-Handling.html
This describes the event phases pretty well.

stefanunity (approved):
Finished for now. Two bugs not yet fixed are:

1. BUG: trying to paste an action into empty map throws error: 
    --ArgumentException: '0' cannot be greater than -1.
2. BUG: trying to paste a  modifier into empty map throws errror	
    --NullReferenceException: Object reference not set to an instance of an object

Once they are fixed it looks good.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
graham-huws-unity: Thanks Rita - Have pushed a couple changes, looks much better now I think!

ritamerkl: I think like that the logic to paste actions into action maps (and add them as last element) is broken because it is impossible to determine from outside of the actionMapView in wich list the paste was performed. Added @Pauliusd01

graham-huws:
> Finished for now. Two bugs not yet fixed are:
> 
>     1. BUG: trying to paste an action into empty map throws error:
>        --ArgumentException: '0' cannot be greater than -1.
> 
>     2. BUG: trying to paste a  modifier into empty map throws errror
>        --NullReferenceException: Object reference not set to an instance of an object
> 
> 
> Once they are fixed it looks good.

Thanks @stefanunity - I believe these should both be fixed now, can you confirm?

Pauliusd01:
https://github.com/Unity-Technologies/InputSystem/assets/54306142/6953ce5a-e7d8-4695-b417-014c8351781a

Cutting and pasting into an empty asset somehow "resurrects" all of the previously deleted actions. (I do deselect ProjectWideActions to confirm the auto save but that does not seem to help)

stefanunity:
> Thanks @stefanunity - I believe these should both be fixed now, can you confirm?

Both are fixed now.

ekcoh:
The bug called out @Pauliusd01 doesn't really make any sense to me but should be fixed. Where is this coming from if not the copy buffer?

Pauliusd01 (changes requested):
Updating status - waiting for fixes

Pauliusd01:
Closing as the proposed fix introduces a regression risk for a very low pain bug. Copy and pasting between windows is available, only issue is that it requires you to focus an action/map first

PR #1846, opened 2024-02-15
FIX: Improve window layout to avoid elements being hidden (ISX-1848).

graham-huws:
### Description

Fix for [ISX-1848](https://jira.unity3d.com/browse/ISX-1848) - It was noted that the initial layout when opening the Input windows hid elements.

### Changes made

Tweaked initial values for pane column widths, set a minimum width for the Properties pane, and extended the binding properties "path" section to use some of the empty space, meaning much more is readable even at minimum width.

Before:
image

After:
image

Before:
https://github.com/Unity-Technologies/InputSystem/assets/102745393/dcae6c55-7608-42d3-a539-db7a4b74b49b

After:
https://github.com/Unity-Technologies/InputSystem/assets/102745393/39f01ae0-0071-4780-a7b9-b467d6173147

### Notes

For anyone else who wants to work with defaults - the values for window position, size and layout for the Input windows are stored in [Editor Prefs](https://docs.unity3d.com/ScriptReference/EditorPrefs.html). I ended up backing up my .plist on Mac, and just deleting the freshly made one every time I wanted to test new defaults.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed InputManager.asset file growing in size on each Reset call.
 - Fixed Opening InputDebugger throws 'Action map must have state at this point' error
 - Fixed Cut/Paste behaviour to match Editor - Cut items will now be cleared from clipboard after pasting.
+- Improved initial window layout to avoid elements being hidden.
```
ekcoh: I would suggest being more specific in the CHANGELOG about which window we are referring to, e.g. "Project Settings - Input Actions".

ekcoh (approved):
LGTM change wise. A small suggestion to improve CHANGELOG message. I should mention I haven't tested the branch, only reviewed the changes and observed the screenshots. In general what I have found the most annoying with the sizes is the sizing policy of the three panels. Basically when extending the window it would be ideal to not let Action Maps steal from the new available space. This might relate to nesting/hierarchy of split panes I assume.

graham-huws:
> Basically when extending the window it would be ideal to not let Action Maps steal from the new available space. This might relate to nesting/hierarchy of split panes I assume.

Hmm, for me it's Actions that takes all the space when extending. Would the preferred solution be that all three sections expand equally?

ekcoh:
> > Basically when extending the window it would be ideal to not let Action Maps steal from the new available space. This might relate to nesting/hierarchy of split panes I assume.
> 
> Hmm, for me it's Actions that takes all the space when extending. Would the preferred solution be that all three sections expand equally?

Not sure what is ideal to be honest, could you loop the question via internal team-thread with @Billreyn for discussion/input? I also believe we can get this in as-is and reiterate on any feedback since that may delay further insights. Given current behavior this seems to improve the situation.

graham-huws:
Updated "Changes made" section to add videos showing before/after for the elements no longer getting hidden.

stefanunity:
Looks good, some minor things found but might be out of scope of this specific PR?

1. Compacting window to minimum allows movement of map<>actions border
2. Can expand maps sections to the right moving the rest off the window space
3. Drag map<>actions border to the right so bindings are off screen
   -- Click & slightly drag actions<>bindings border to the left
   -- It jumps to the left on top of actions
   sctioneditor_drag_binding_border_from_off_screen

4. Collapsing project settings window removes most of action editor, but probably by design since it does the same for other setting sections too

stefanunity (approved):

graham-huws:
Thanks @stefanunity! I would say that unless they were introduced in this PR (I definitely reproduced the second one you mentioned without my changes), the points raised should be a separate Jira issue, as ISX-1848 was specifically about the initial state of the windows when users open them for the first time - @ekcoh would you agree?

ekcoh:
If @stefanunity agree I would say that if initial sizing is considered improved and sizing restrictions or overflow behavior is unaffected by this PR we could merge this and file the remaining work as new tickets. If its rather a regression of this PR (I fail to see how) its better fixed on this PR before merged.

stefanunity:
Yes, can be merged. Will make tickets later.

lyndon-unity (approved):
Good step forward, especially the min-width

I think the actions and properties pane would benefit from being slightly wider to see a little more of the text in the Path entry but its still an improvement for the defaults.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Resources/InputActionsEditor.uxml
```diff
@@ -27,7 +27,7 @@
                         <ui:ListView focusable="true" name="action-maps-list-view" />
                     </ui:VisualElement>
                 </ui:VisualElement>
-                <ui:TwoPaneSplitView name="actions-and-properties-split-view" fixed-pane-index="1" fixed-pane-initial-dimension="400" style="height: auto;">
+                <ui:TwoPaneSplitView name="actions-and-properties-split-view" fixed-pane-index="1" fixed-pane-initial-dimension="230" style="height: auto;">
```
lyndon-unity: I think this could have a min-wdith to prevent scrolling off screen by moving the maps/actions separator.

(e.g. 450 min)

graham-huws:
Thanks all - Agreed @lyndon-unity, I think that can be part of the new tickets Stefan will add, also relevant to the question [here](https://unity.slack.com/archives/C04RDRXRJ1L/p1708090595118869).

graham-huws:
Updated to add the min-width fix @lyndon-unity suggested. Also lowered the gaps between dropdowns, here's what they look like now with the Properties pane at min width:

image
image

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionPropertiesView.cs
```diff
@@ -41,6 +42,12 @@ public override void RedrawUI((SerializedInputAction ? , List<string>) viewState
             {
                 tooltip = inputAction.actionTypeTooltip
             };
+
+            // Tighten up the gap between the label and dropdown so the latter is more readable when the parent pane is at min width.
+            var actionLabel = actionType.Q<Label>();
+            actionLabel.style.minWidth = m_DropdownLabelWidth;
```
lyndon-unity, line 48: Can this be in the uxml rather than the code?

lyndon-unity (approved):
Happy to approve the end result i saw.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionPropertiesView.cs
graham-huws-unity: I wondered about that, but there doesn't seem to be uxml for the Properties view, it's entirely programmatic - Not sure if there's a nice way around that?

lyndon-unity: Fair enough, let's move forward with this then.

PR #1848, opened 2024-02-19
FIX: Conditional API signature selection for asset processor

ekcoh:
### Description

FIX: Conditional API signature selection for asset processor based on Unity version whether domain reload boolean is available or not. This fix makes it compatible back to Unity 2019. This corrects a problem introduced with https://github.com/Unity-Technologies/InputSystem/pull/1836. 

### Changes made

Adds compile-time dynamic selection of API signature.

### Notes

This should have been part of the original solution if 2019 was covered in QA/testing.

Decided to not include a CHANGELOG since this should have been part of the original user-facing fix.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #1849, opened 2024-02-20
[Input] Improve Input System action migration from InputManager.asset file

AlexTyrer:
### Description
Improved Input System action migration from InputManager.asset file

### Changes made

o Ensure InputSystem.actions is set after migrating InputActionAssets
o Ensure AssetDatabase is aware of newly created .inputaction asset
o Don't overwrite existing Input System action file - always migrate settings to a new unique file

### Notes

### Checklist

Before review:

- Changelog entry added
- No tests added - required manual testing with different Input System (pre-release) versions
- No documentation required

During merge:

(notes for later...)
CHANGE: Migrate any project-wide input actions found in the InputManager.asset file to a new InputSystem_Actions.inputactions asset file.


Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -39,14 +34,39 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
...
+                            List<string> names = new List<string>();
+                            for (int i = 0; i < files.Length; i++)
+                            {
+                                names.Add(System.IO.Path.GetFileNameWithoutExtension(files[i]));
+                            }
+                            string unique = ObjectNames.GetUniqueName(names.ToArray(), "InputSystem_Actions");
+                            path = "Assets/" + unique + ".inputactions";
+                        }
+
+                        var json = inputActionsAsset.ToJson();
```
lyndon-unity: This line fixed the JSON data and was critical :)

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -39,14 +34,39 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
...
+                        File.WriteAllText(path, json);
+
+                        AssetDatabase.Refresh();
+                        InputSystem.actions = (InputActionAsset)AssetDatabase.LoadAssetAtPath(path, typeof(InputActionAsset));
```
lyndon-unity, line 90: Excellent - this makes it show up now in project settings

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -63,6 +83,8 @@ internal static void MoveInputManagerAssetActionsToProjectWideInputActionAsset()
                             AssetDatabase.RemoveObjectFromAsset(obj);
                         }
                     }
+
+                    AssetDatabase.SaveAssets();
```
lyndon-unity: Great - this made the InputManager.asset file change at this point so we could see it cleaned up

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -39,14 +34,39 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
...
+                            //      Assets/InputSystem_Actions (2).inputactions ...
+                            //
+                            string[] files = Directory.GetFiles("Assets", "*.inputactions");
+                            List<string> names = new List<string>();
+                            for (int i = 0; i < files.Length; i++)
+                            {
+                                names.Add(System.IO.Path.GetFileNameWithoutExtension(files[i]));
+                            }
+                            string unique = ObjectNames.GetUniqueName(names.ToArray(), "InputSystem_Actions");
+                            path = "Assets/" + unique + ".inputactions";
```
lyndon-unity: Great - this prevents overwriting user files

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -15,15 +16,9 @@ internal static class ProjectWideActionsAsset
 
         internal static class ProjectSettingsProjectWideActionsAssetConverter
         {
-            internal const string kAssetPath = "ProjectSettings/InputManager.asset";
+            internal const string kAssetPathInputManager = "ProjectSettings/InputManager.asset";
```
lyndon-unity, line 20: Easier to read this name below - thanks

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -15,15 +16,9 @@ internal static class ProjectWideActionsAsset
...
             internal const string kAssetName = InputSystem.kProjectWideActionsAssetName;
```
lyndon-unity: Could name this kAssetNameProjectWide but not critical

lyndon-unity (approved):
Great update - much more robust

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Not sure why defining a constant set to another constant? However...
InputSystem.kProjectWideActionsAssetName should no longer exist, so rather move that constant here instead. Its only relevant for the previous solution. It seems to not be used anywhere else....

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: +1 better name for what it implies

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -39,14 +34,39 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
 
             internal static void MoveInputManagerAssetActionsToProjectWideInputActionAsset()
             {
-                var objects = AssetDatabase.LoadAllAssetsAtPath(kAssetPath);
+                var objects = AssetDatabase.LoadAllAssetsAtPath(kAssetPathInputManager);
```
ekcoh: I think this needs to use FileUtils.PhysicalPath to be future proof, DM me for further elaboration. See EditorHelpers.cs for how its used conditionally. I have extracted the #ifdef into its own function on a pending PR but its not available in this version.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -39,14 +34,39 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
...
-                    if (inputActionsAsset != null)
+                    if (inputActionsAsset != default)
                     {
-                        var json = JsonUtility.ToJson(inputActionsAsset, prettyPrint: true);
-                        File.WriteAllText(ProjectWideActionsAsset.kDefaultAssetPath, json);
+                        // Found some actions in the InputManager.asset file
+                        //
+                        string path = ProjectWideActionsAsset.kDefaultAssetPath;
+
+                        if (File.Exists(path))
```
ekcoh: Minor: Would suggest doing inverted ifs in this function in general if all of them are early-outs since there is a lot of nesting making it harder to read.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Great to see the "finding an available name" logic to avoid the clash. Would however recommend using InputActionImporter.cs constant for extension though rather than hardcoding a string copy here. That way this just works if it would change in the future.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
```diff
@@ -39,14 +34,39 @@ static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAsse
...
+                            for (int i = 0; i < files.Length; i++)
+                            {
+                                names.Add(System.IO.Path.GetFileNameWithoutExtension(files[i]));
+                            }
+                            string unique = ObjectNames.GetUniqueName(names.ToArray(), "InputSystem_Actions");
+                            path = "Assets/" + unique + ".inputactions";
+                        }
+
+                        var json = inputActionsAsset.ToJson();
+                        File.WriteAllText(path, json);
```
ekcoh: I would suggest removing this File.WriteAllText + AssetDatabase.Refresh() with a call to InputActionAssetManager.SaveAsset instead. In addition that method handles any underlying single-checkout versioning system which could otherwise generate a File-access exception here.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh, line 90: I wonder if we shouldn't provide a Debug.Info message here informing about the conversion that have been done and that the asset has been assigned to InputSystem.actions?

Also, what if InputSystem.actions is already assigned? In that case I do not think we should assign it, only port to a new asset and inform about the conversion that has been made.

ekcoh (changes requested):
Good improvement of the import/porting logic. I noticed however that the AssetProcessor in this file is using the callback with the `didDomainReload` argument. This is not present in versions less than Unity 2021.2. I did a similar fix recently to support compilation for 2019, see https://github.com/Unity-Technologies/InputSystem/pull/1848
I advice to do a similar change here, and I cannot understand how this could pass CI without this being updated on that PR, do you know? Or was the fix I did unnecessary somehow and was a false positive form IDE? (Confused)

ekcoh (commented):
This branch is missing a CHANGELOG.md entry so that needs to be added.

I also wonder whether it would be feasible to add a unit test exercising the porting logic? It would be helpful to avoid any future problems. I believe that could be done by writing an asset containing a InputManager.asset copy to any test path and verify the conversion result and that it is cleaned up. Such a test would of course be editor only so conditionally compiled for UNITY_EDITOR.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: That's probably a result of starting in trunk and then moving to your branch - will define the constant here and remove the old one in InputSystem.cs

AlexTyrer:
> Good improvement of the import/porting logic. I noticed however that the AssetProcessor in this file is using the callback with the `didDomainReload` argument. This is not present in versions less than Unity 2021.2. I did a similar fix recently to support compilation for 2019, see #1848 I advice to do a similar change here, and I cannot understand how this could pass CI without this being updated on that PR, do you know? Or was the fix I did unnecessary somehow and was a false positive form IDE? (Confused)

I am no sure either.

I see other instances of functions with this parameter in the repo (without a  #if UNITY_2021_2_OR_NEWER conditional)
eg
**RenderPipelineGlobalSettingsPostprocessor.cs**

https://github.cds.internal.unity3d.com/unity/unity/blob/15f36247668085dfe54fbf0301812c4e89fff643/Editor/Mono/Graphics/RenderPipelineGlobalSettingsPostprocessor.cs#L75

**TerrainTransientToolbarOverlay.cs**
https://github.cds.internal.unity3d.com/unity/unity/blob/15f36247668085dfe54fbf0301812c4e89fff643/Modules/TerrainEditor/Overlays/TerrainTransientToolbarOverlay.cs#L21

Update: of course these are in the Unity repo - the package needs to be compatible with multiple releases - have added the pre-processor check.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: I can add a constant, but there's already some duplication here:

        private const string kDefaultAssetPath = "Assets/InputSystem_Actions.inputactions";

Do you mean to define something like:

        private const string kInputActionsBaseName = InputSystem_Action";

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: Will make this change..

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: Happy to add a message, but I can't find any example of "Debug.Info" - can you provide an example?

As for the assignment, we are in this block of code because we found project wide settings in the input manager asset file.

When testing this with @lyndon-unity we worked on the assumption that the user probably did want the assignment to be updated.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: Just change to this?

```
#if UNITY_2021_2_OR_NEWER
                var objects = AssetDatabase.LoadAllAssetsAtPath(FileUtil.GetPhysicalPath(kAssetPathInputManager));
#else
                var objects = AssetDatabase.LoadAllAssetsAtPath(kAssetPathInputManager);
#endif
```

AlexTyrer: Will make this change for the other file path too - will a local conversion function to avoid so many pre-processor blocks.

```
                        string path = ProjectWideActionsAsset.kDefaultAssetPath;

                        if (File.Exists(path))

```

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Sounds good.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Sounds good

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
ekcoh: Thinking about this, maybe just let it be as is

ekcoh (approved):

ekcoh:
Might have run into instabilities with CI, recommend opening All Tests in Yamato and reschedule failed jobs.

ekcoh:
Approved under condition that CHANGELOG.md is added and CI is green.

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: I've added a Debug.Log() for both of these: migration (and assignment if required)

Packages/com.unity.inputsystem/InputSystem/Editor/ProjectWideActions/ProjectWideActionsAsset.cs
AlexTyrer: I've tidied these names up.

Pauliusd01:
Checking

Pauliusd01:
Two things I noticed when upgrading an older 1.8.0-pre.2 project:

- It correctly creates a mirrored projectwideactions asset for you but that asset reference is not updated in any of the playerinput components, that component still points to the old, now non-existent projectwideactions. Should those automatically update? (See video, the name of the asset referenced is that of the old one I have to manually select the one your PR created for me)

https://github.com/Unity-Technologies/InputSystem/assets/54306142/aea93302-7e03-4ccd-a167-7155d261064d

- The newly created asset is dirty on creation, I was prompted to save "changes" right after migrating and opening it

I don't think either is a blocker but just mentioning any odd things I come across

Pauliusd01 (approved):
LGTM, did not notice any major issues with upgrading projects that use ProjectWideActions or existing functionality. Only left 2 minor notes above for consideration

PR #1850, opened 2024-02-20
FIX: ISX-1890 + Improved asset management in Input Action Asset editors

ekcoh:
### Description

This branch introduces improved asset management of .inputactions files open in editors and fixes a bug where pressing Save Asset in Input Action Editor window would throw an exception if the file was moved while the editor was being open. For details on changes introduced see Changes made below.

### Changes made

The following changes have been made to this branch:
- Fixed an issue where deleting an `.inputactions` asset with the asset open for editing in a Input Action Editor window (UITK) would cause the editor to crash with `FileNotFoundException`.
- Fixed an issue where UITK and IMGUI editors have different behaviours when moving an asset with editor open at the same time. (Move/drag from within Unity + external file system rename). 
- Fixed an issue where UITK and IMGUI editors have different behaviours when deleting an asset with editor open at the same time. (Delete from within Unity + external file system delete)
- Fixed an issue where UITK and IMGUI editors have different behaviours when renaming an asset with editor open at the same time. (Rename from within Unity + external file system rename)
- Refactored and added same AssetModificationProcessor and AssetPostProcessor monitors to both IMGUI and UITK editors to avoid overlapping functionality. This is implemented as `InputActionAssetPostprocessor` and `InputActionAssetModificationProcessor` replacing existing processor in IMGUI and introducing it for UITK editor.
- Refactored and moved all dialog logic into Dialogs.cs and converted all outcomes to enum values for increased readability. Also added a layer of abstraction to allow mocking and "head-less" test execution.
- Modified UITK dispatch mechanism to show full exception traces in log instead of truncating tracing information to messages only to simplify tracking errors and bugs.
- Modified `InputActionsEditorState` to rely on GUID-to-index and index-to-GUID remapping when reimporting an asset and it is different from previous incarnation of the asset. Since renaming, restructuring, deleting and additions may happen to any map, action, binding, interaction, processor or control scheme during reimport it makes sense to rely on GUIDs for identification rather than names since not stable identifiers.
- Fixed an issue where the UITK Input Action Editor would show incorrect modification state if open during domain reload.

Test related changes:
- Added `AssetDatabaseUtils.cs` which simplifies writing test cases involving assets since it removes the burden from the developer to find unique file and directory names as well as automatically cleans-up all associated files and folders as part of InputTestFixture teardown.
- Corrected incorrect one-time test fixture setup and teardown in `CoreTest_ProjectWideActions.cs` which resulted in backup folder being left in /Assets after running tests.
- Fixed test case `Editor_InputActionAssetManager_CanMoveAssetOnDisk `which was testing InputActionAssetManager monitoring of external file system changes but failed to move .meta files along with assets generating warnings in test runner.
- Expanded test coverage of InputActionAssetManager via tests called out in checklist below. This was done to support refactoring of the class with less worry about breaking existing behavior without improving it. Also moved those tests into its own .cs file targeting this specific class.
- Updated `InputTestFixture` to apply default mocking and restore for all editor test cases to avoid running into problems with user dialogs affecting execution flow.

### Notes

Project-wide settings editor still operates without a copy/instance of the imported/source asset which might have implications or cause inconsistencies in behavior, but difficult to handle when auto-save is enforced.

I had the ambition to refactor the UITK editor to also use `InputActionAssetManager` as part of this PR but gave up on this after partial progress on improving it and adding unit tests for it which at least gets us one step closer to eliminate even more code. This could be resumed in the future as a separate task.

Recommended to cover during quality assurance:
- Copying, renaming, moving and deleting .inputactions assets from within editor with and without editors open.
- Copying, renaming, moving and deleting .inputactions assets from outside editor (OS) with and without editors open.
- Functional test of UITK Input Action Editor with respect to above and general functionality.
- Functional test of IMGUI Input Action Editor with respect to above and general functionality.
- Functional test of UITK Project Settings Input Action Editor with respect to above and general functionality.

### Checklist

Before review:

- [x] Changelog entry added.
    - Functional tests added `Actions_NameIsSetToNameFromJson_IfCreatedFromJsonWithName`, `Actions_NameIsSetToEmptyString_IfCreatedFromJsonWithoutName`, `Actions_NameInJsonIsSetToObjectName_IfCreatedFromObjectWithGivenName()`, `Actions_NameInJsonIsSetToEmptyString_IfCreatedFromObjectWithEmptyName`, `Actions_NameInJsonIsSetToEmptyString_IfCreatedFromObjectWithNullName`, `Editor_InputActionAssetManager_ThrowsIfCreatedFromNullReference`,
`Editor_InputActionAssetManager_ThrowsIfCreatedFromNonPersistedAsset`,
`Editor_InputActionAssetManager_SerializedObjectIsAWorkingCopyOfImportedAsset`,
`Editor_InputActionAssetManager_isDirtyReflectsWhetherWorkingCopyIsDifferentFromImportedAsset`,
`Editor_InputActionAssetManager_CanMoveAssetOnDisk` (modified/simplified),
`Editor_InputActionAssetManager_CanDeleteAssetOnDisk`.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Pauliusd01:
Entering Play Mode in main repo project throws this error (scene does not matter): 
`'UnityEngine.InputSystem.Editor.EnableUITKEditor' is missing the class attribute 'ExtensionOfNativeClass'!`

Pauliusd01:
Trying to add new control schemes throws errors

https://github.com/Unity-Technologies/InputSystem/assets/54306142/f55ae476-26c7-48b0-b094-640144fe5b28

ekcoh:
> Trying to add new control schemes throws errors
> 
>  Unity_2024-02-28_14-10-23.mp4

ekcoh:
> ExtensionOfNativeClass

I am not getting this error. How did you manage to get this? (Running 2023.3.b04 with cleaned Library folder and checkout)

Pauliusd01:
> > ExtensionOfNativeClass
> 
> I am not getting this error. How did you manage to get this? (Running 2023.3.b04 with cleaned Library folder and checkout)

I don't see it on latest commit either, will let you know if I ever get it again

ekcoh:
> Trying to add new control schemes throws errors
> 
>  Unity_2024-02-28_14-10-23.mp4

Found why this is happening. Basically range validation should happen at construction time since I now allow swapping out the serializedobject, fix in the works.

Pauliusd01:
Nullrefs thrown when renaming an already opened input actions asset. (On Develop you don't get these types of errors but you do get a FileNotFoundException instead)
https://github.com/Unity-Technologies/InputSystem/assets/54306142/d4e46c7a-f840-479f-932e-e5e8ca2bb665

ekcoh:
> > Trying to add new control schemes throws errors
> > Unity_2024-02-28_14-10-23.mp4
> 
> Found why this is happening. Basically range validation should happen at construction time since I now allow swapping out the serializedobject, fix in the works.

@Pauliusd01 this should now be fixed with this commit https://github.com/Unity-Technologies/InputSystem/pull/1850/commits/3215aaa813fa45d9d66d438aa61f1108554b826b along with improved selection index transformation index->GUID and then GUID->index between asset instances when reimported from disc with editor open. This also triggers on save since disc write will trigger reimport so to the editor save and external file changes are similar.

ekcoh:
> Nullrefs thrown when renaming an already opened input actions asset. (On Develop you don't get these types of errors but you do get a FileNotFoundException instead)
> https://github.com/Unity-Technologies/InputSystem/assets/54306142/d4e46c7a-f840-479f-932e-e5e8ca2bb665

This was one of my key test cases just before putting this PR in review so must be some mistake I did in last clean-up of the branch. Investigating....

ekcoh:
> > Nullrefs thrown when renaming an already opened input actions asset. (On Develop you don't get these types of errors but you do get a FileNotFoundException instead)
> > https://github.com/Unity-Technologies/InputSystem/assets/54306142/d4e46c7a-f840-479f-932e-e5e8ca2bb665
> 
> This was one of my key test cases just before putting this PR in review so must be some mistake I did in last clean-up of the branch. Investigating....

This should be fixed in https://github.com/Unity-Technologies/InputSystem/pull/1850/commits/9e4dc181d0eec337e0aba540c87a1df70122887f. Also simplified editor code a bit and fixed an issue where editor would indicate incorrect modification/dirty flag (*) if editor is open during domain reload.

ekcoh:
For visibility I am currently looking into the test failures here which do not reproduce for me locally inside Unity but seem to reproduce here on CI

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -4980,6 +4980,80 @@ public void Actions_CanConvertAssetToAndFromJson()
         Assert.That(asset.controlSchemes[1].deviceRequirements[2].isOR, Is.True);
     }
 
+    static string MinimalJson(string name = null)
+    {
+        if (name != null)
+            return "{\n    \"name\": \"" + name + "\",\n    \"maps\": [],\n    \"controlSchemes\": []\n}";
```
lyndon-unity, line 4986: I find this a bit hard to parse, perhaps using string literal might help

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated?redirectedfrom=MSDN#interpolated-raw-string-literals

E.g.

string s = $"""
    "name": {name},
    "maps": [],
	"controlSchemes": []
"""

For some reason this didn't work for me

string s = $"""
{{
    "name": {name},
    "maps": [],
	"controlSchemes": []
}}
"""

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -4980,6 +4980,80 @@ public void Actions_CanConvertAssetToAndFromJson()
...
+        ScriptableObject.Destroy(asset);
+        Assert.That(content, Is.EqualTo(expected));
+    }
+
+    [Test]
+    [Category("Actions")]
+    public void Actions_NameInJsonIsSetToEmptyString_IfCreatedFromObjectWithNullName()
+    {
+        // When serializing JSON from object without a given name, Unity forces the name to empty string.
+        // Basically Unity prevents us from doing serialization an omit optional members in a convenient way.
```
lyndon-unity, line 5045: Text here looks slightly off

Should the **an** be **and**:
Basically Unity prevents us from doing serialization **and** omit**s** optional members in a convenient way.

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputActionAssetManager_CanMoveAssetOnDisk()
+    {
+        const string filename = "InputAsset." + InputActionAsset.Extension;
+        var directoryBeforeMove = AssetDatabaseUtils.CreateDirectory();
+        var directoryAfterMove = AssetDatabaseUtils.RandomDirectoryPath();
```
lyndon-unity: Why switched to random path here.
Probably fine - but seems unnecessary refactoring for this pr

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        const string kDefaultContents = "{}";
+        var asset = AssetDatabaseUtils.CreateAsset<InputActionAsset>(directoryPath: directoryBeforeMove, filename: filename, content: kDefaultContents);
+
+        using (var inputActionAssetManager = new InputActionAssetManager(asset))
+        {
+            inputActionAssetManager.Initialize();
+
+            FileUtil.MoveFileOrDirectory(directoryBeforeMove, directoryAfterMove); // TODO Wouldn't move .meta files
+            FileUtil.MoveFileOrDirectory(directoryBeforeMove + ".meta", directoryAfterMove + ".meta");
```
lyndon-unity: Original test didn't move the meta file - not sure why that's changed either

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -834,36 +831,6 @@ public void Editor_InputActions_AssetFileUpdatedAfterMoveOrCopy(string newAssetP
...
-        AssetDatabase.CreateFolder("Assets", "DirectoryBeforeRename");
-        File.WriteAllText(kAssetPath, kDefaultContents);
-        AssetDatabase.ImportAsset(kAssetPath);
-
-        var asset = AssetDatabase.LoadAssetAtPath<InputActionAsset>(kAssetPath);
-        Assert.NotNull(asset, "Could not load asset: " + kAssetPath);
-
-        var inputActionAssetManager = new InputActionAssetManager(asset);
-        inputActionAssetManager.Initialize();
-        inputActionAssetManager.onDirtyChanged = (bool dirty) => {};
```
lyndon-unity, line 854: This is removed when refactoring. Not sure why this was present - perhaps it was previously here to avoid any dialog boxes ?

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
```diff
@@ -0,0 +1,134 @@
...
+            AssetDatabase.Refresh();
+
+            inputActionAssetManager.SaveChangesToAsset();
+
+            var fileContents = File.ReadAllText(AssetDatabase.GetAssetPath(asset));
```
lyndon-unity: Previously this was checking the target path specifically - we now no longer check that so potentially its not been moved.
May be safe but there seems to be more refactoring that I would expect in this test.

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -834,36 +831,6 @@ public void Editor_InputActions_AssetFileUpdatedAfterMoveOrCopy(string newAssetP
...
-
-        FileUtil.MoveFileOrDirectory("Assets/DirectoryBeforeRename", "Assets/DirectoryAfterRename");
-        AssetDatabase.Refresh();
-
-        Assert.DoesNotThrow(() => inputActionAssetManager.SaveChangesToAsset());
-
-        var fileContents = File.ReadAllText(kAssetPathAfterMove);
-        Assert.AreNotEqual(kDefaultContents, fileContents, "Expected file contents to have been modified after SaveChangesToAsset was called.");
-
-        AssetDatabase.DeleteAsset("Assets/DirectoryAfterRename");
```
lyndon-unity, line 864: This clean up is removed from the test - so doesn't it now leak the asset ?

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
```diff
@@ -0,0 +1,134 @@
...
+            Assert.AreNotEqual(kDefaultContents, fileContents, "Expected file contents to have been modified after SaveChangesToAsset was called.");
+        }
+    }
```
lyndon-unity: No clean up of asset so is this now leaking the test?

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
```diff
@@ -0,0 +1,134 @@
...
+            UnityEngine.Object.DestroyImmediate(asset);
+        }
+    }
+
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputActionAssetManager_SerializedObjectIsAWorkingCopyOfImportedAsset()
+    {
+        var asset = AssetDatabaseUtils.CreateAsset<InputActionAsset>();
+        using (var sut = new InputActionAssetManager(asset))
```
lyndon-unity: sut is a bit of an odd variable name.
I don't know what that actually is supposed to mean

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
```diff
@@ -0,0 +1,134 @@
...
+        // Start from an asset, since we haven't done modifications expect it to not be marked dirty
+        var asset = AssetDatabaseUtils.CreateAsset<InputActionAsset>();
+        using (var sut = new InputActionAssetManager(asset))
+        {
+            Assert.That(sut.dirty, Is.False);
+
+            // Add a map to editable object and apply changes, expect it to not be dirty since we haven't explicitly updated dirty state
+            var mapProperty = InputActionSerializationHelpers.AddActionMap(sut.serializedObject);
+            sut.ApplyChanges();
+            Assert.That(sut.dirty, Is.False); // TODO Note: dirty flag is not updated only by applying changes
```
lyndon-unity: Is this is expected then I don't know why the TODO is present

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -29,8 +29,8 @@ internal partial class CoreTests
     // Note that any existing default created asset is preserved during test run by moving it via ADB.
 
     const string TestCategory = "ProjectWideActions";
-    const string m_AssetBackupDirectory = "Assets/~TestBackupFiles";
-    const string s_DefaultProjectWideAssetBackupPath = "Assets/~TestBackupFilesDefaultProjectWideAssetBackup.json";
+    const string kAssetBackupDirectory = "Assets/~TestBackupFiles";
+    const string kDefaultProjectWideAssetBackupPath = kAssetBackupDirectory + "/DefaultProjectWideAssetBackup.json";
```
lyndon-unity, line 33: Looks like a missing / was added here. I think thats intentional

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -44,10 +44,10 @@ public void OneTimeSetUp()
         // This is for verifying the default output of templated actions from editor tools.
         if (File.Exists(ProjectWideActionsAsset.defaultAssetPath))
         {
-            if (!Directory.Exists(m_AssetBackupDirectory))
-                Directory.CreateDirectory(m_AssetBackupDirectory);
+            if (!Directory.Exists(kAssetBackupDirectory))
+                AssetDatabase.CreateFolder("Assets", "~TestBackupFiles");
```
lyndon-unity, line 48: Why the change from CreateDirectory to CreateFolder ?
If a user did it out side the test via folder browser the wouldn't be using AssetDatabase API.

Assets/Tests/InputSystem/CoreTests_ProjectWideActions.cs
```diff
@@ -57,14 +57,13 @@ public void OneTimeTearDown()
...
+        if (File.Exists(kDefaultProjectWideAssetBackupPath))
         {
             if (File.Exists(ProjectWideActionsAsset.defaultAssetPath))
                 AssetDatabase.DeleteAsset(ProjectWideActionsAsset.defaultAssetPath);
-            AssetDatabase.MoveAsset(oldPath: s_DefaultProjectWideAssetBackupPath,
+            AssetDatabase.MoveAsset(oldPath: kDefaultProjectWideAssetBackupPath,
                 newPath: ProjectWideActionsAsset.defaultAssetPath);
-            Directory.Delete("Assets/~TestBackupFiles");
-            File.Delete("Assets/~TestBackupFiles.meta");
+            AssetDatabase.DeleteAsset(kAssetBackupDirectory);
```
lyndon-unity, line 66: Probably fine but not sure why changed - seems more change than needed in this PR.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1,7 +1,6 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
```
lyndon-unity, line 4: Nice removal :)

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -297,14 +296,12 @@ public IEnumerable<InputBinding> bindings
         /// <seealso cref="FromJson"/>
         public string ToJson()
         {
-            var fileJson = new WriteFileJson
+            return JsonUtility.ToJson(new WriteFileJson
```
lyndon-unity, line 299: Is this just a style choice or are you doing this for performance ?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -948,6 +945,13 @@ internal struct WriteFileJson
             public InputControlScheme.SchemeJson[] controlSchemes;
         }
 
+        [Serializable]
```
lyndon-unity, line 948: Doesn't seem used unless I'm missing something

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -1,5 +1,6 @@
 #if UNITY_EDITOR
 using System;
+using System.Diagnostics.CodeAnalysis;
```
lyndon-unity, line 3: Is this needed?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -51,20 +63,29 @@ private InputActionAsset importedAsset
...
             {
+                // Note that this may be null after deserialization from domain reload
                 if (m_ImportedAssetObject == null)
                     LoadImportedObjectFromGuid();
 
                 return m_ImportedAssetObject;
             }
         }
 
+        public InputActionAsset editedAsset => m_AssetObjectForEditing; // TODO Remove if redundant
```
lyndon-unity, line 74: Why added if then also asking if should be removed ?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -51,20 +63,29 @@ private InputActionAsset importedAsset
...
         public InputActionAssetManager(InputActionAsset inputActionAsset)
         {
+            if (inputActionAsset == null)
+                throw new NullReferenceException(nameof(inputActionAsset));
+            m_AssetGUID = EditorHelpers.GetAssetGUID(inputActionAsset);
+            if (m_AssetGUID == null)
+                throw new Exception($"Failed to get asset {inputActionAsset.name} GUID");
+
             m_ImportedAssetObject = inputActionAsset;
-            bool isGUIDObtained = AssetDatabase.TryGetGUIDAndLocalFileIdentifier(importedAsset, out m_AssetGUID, out long _);
```
lyndon-unity, line 66: Seem to be deciding to change from warning to Exception. Not sure why though

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -91,45 +114,65 @@ public bool Initialize()
 
         public void Dispose()
         {
+            if (m_SerializedObject == null)
+                return;
             m_SerializedObject?.Dispose();
```
lyndon-unity, line 119: No longer need ?. if checking above

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -91,45 +114,65 @@ public bool Initialize()
...
+            var json = importedAsset.ToJson();
             if (m_ImportedAssetJson == json)
                 return false;
 
             CreateWorkingCopyAsset();
             return true;
         }
 
-        private void CreateWorkingCopyAsset()
+        public static InputActionAsset CreateWorkingCopy(InputActionAsset source)
```
lyndon-unity, line 133: Do these need to be public - I don't see them called outside the file

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/EditorHelpers.cs
```diff
@@ -60,31 +69,61 @@ public static bool CheckOut(string path)
...
-            if (asset == null)
-                throw new ArgumentNullException(nameof(asset));
-            var path = AssetDatabase.GetAssetPath(asset);
-            CheckOut(path);
+            // Attempt to checkout the file path for editing and inform the user if this fails.
+            if (!CheckOut(assetPath))
+                return false;
+
+            // (Over)write file text content.
+            File.WriteAllText(GetPhysicalPath(assetPath), text);
```
lyndon-unity, line 88: I don't really understand what the refactoring solves here
EditorHelpers.WriteAllText -> File.WriteAllText

Seems ok, but not clear why needed

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -191,44 +219,30 @@ internal static bool SaveAsset(string assetPath, string assetJson)
...
-            {
-                Debug.LogError($"Unable save asset to \"{assetPath}\" since the asset-path could not be checked-out as editable in the underlying version-control system.");
-                return false;
-            }
+            // Attempt to write asset to disc (including checkout the file) and inform the user if this fails.
+            if (EditorHelpers.WriteAsset(assetPath, assetJson))
+                return true;
 
-            return true;
+            Debug.LogError($"Unable save asset to \"{assetPath}\" since the asset-path could not be checked-out as editable in the underlying version-control system.");
```
lyndon-unity, line 226: Logic change to remove ! seems ok but not really sure why needed in this PR.
A few changes seem more than necessary for focused fixes.
I wouldn't block the PR for it but it makes it harder to review and seems a style choice.

ekcoh:
> For visibility I am currently looking into the test failures here which do not reproduce for me locally inside Unity but seem to reproduce here on CI

I noticed the asset manager tests where running in player/play mode which make little sense so moved them to editor assembly. Will see if this helps the problem since only failed in play mode (likely to sandbox nature of it mixed with ADB and file operations).

Update: Above didn't work but noticed only Windows failures and native Windows paths to added a fix to convert to Unix path separators when processing paths in CreateDirectories. New attempt to fix test errors in https://github.com/Unity-Technologies/InputSystem/pull/1850/commits/f9740ab8f9f76c85baf0a75c2cd0e15b56df66be

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
```diff
@@ -263,7 +246,7 @@ private void SetAsset(InputActionAsset asset)
                 return;
 
             m_ActionAssetManager = new InputActionAssetManager(asset) {onDirtyChanged = OnDirtyChanged};
-            m_ActionAssetManager.Initialize();
+            //m_ActionAssetManager.Initialize(); // TODO No longer needed when using constructor
```
lyndon-unity, line 249: Seems like this should be deleted ?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionEditorWindow.cs
```diff
@@ -854,64 +838,22 @@ public void Dispose()
         private Vector2 m_PropertiesScroll;
         private bool m_ForceQuit;
 
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1812:AvoidUninstantiatedInternalClasses", Justification = "Intantiated through reflection by Unity")]
-        private class ProcessAssetModifications : UnityEditor.AssetModificationProcessor
```
lyndon-unity, line 858: So this is moved to IInputActionAssetEditor

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -331,38 +319,72 @@ public static void CreateInputAsset()
                 InputActionAsset.kDefaultAssetLayoutJson, InputActionAssetIconLoader.LoadAssetIcon());
         }
 
-        // When an action asset is renamed, copied, or moved in the Editor, the "Name" field in the JSON will
-        // hold the old name and won't match what's in memory (asset looks "dirty"). To work around this, we
-        // must flush the updated JSON to the file whenever this operation occurs.
+        // File extension of the associated asset
+        private const string kFileExtension = "." + InputActionAsset.Extension;
```
lyndon-unity, line 323: Seems reasonable adding null check but the extra movement in the file made it harder to diff.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -58,7 +50,7 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
             string content;
             try
             {
-                content = EditorHelpers.ReadAllText(context.assetPath);
```
lyndon-unity, line 61: Some of this extra refactoring seemed unnecessary for this pr - tidy maybe but had made the pr longer to review

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/EditorHelpers.cs
```diff
@@ -1,5 +1,6 @@
 #if UNITY_EDITOR
 using System;
+using System.Diagnostics.CodeAnalysis;
```
lyndon-unity, line 3: Is this needed?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
```diff
@@ -483,7 +483,9 @@ public static Command SaveAsset(Action postSaveAction)
         {
             return (in InputActionsEditorState state) =>
             {
-                InputActionAssetManager.SaveAsset(state.serializedObject.targetObject as InputActionAsset);
+                // TODO This needs to callback to owning editor to save or have asset GUID
+                // TODO It makes more sense to call back to editor since editor owns target object?
+                //InputActionAssetManager.SaveAsset(state.serializedObject.targetObject as InputActionAsset);
```
lyndon-unity, line 488: Can we clean up the TODO here
Delete this now ? If SaveAction is handing it ?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/Commands.cs
```diff
@@ -498,7 +500,7 @@ public static Command ToggleAutoSave(bool newValue, Action postSaveAction)
                     // If it changed from disabled to enabled, perform an initial save.
                     if (newValue)
                     {
-                        InputActionAssetManager.SaveAsset(state.serializedObject.targetObject as InputActionAsset);
+                        //InputActionAssetManager.SaveAsset(state.serializedObject.targetObject as InputActionAsset);
```
lyndon-unity, line 503: Delete this now ? If SaveAction is handing it ?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
```diff
@@ -52,6 +55,16 @@ public InputActionsEditorView(VisualElement root, StateContainer stateContainer,
             autoSaveToggle.value = InputEditorUserSettings.autoSaveInputActionAssets;
             autoSaveToggle.RegisterValueChangedCallback(OnAutoSaveToggle);
 
+            // Hide save toolbar if there is no save action provided since we cannot support it
```
lyndon-unity, line 58: Fine to move this but didn't seem like a critical refactoring

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
```diff
@@ -14,12 +14,12 @@ internal struct InputActionsEditorState
...
         public int selectedBindingIndex { get {return m_selectedBindingIndex; } }
         public SelectionType selectionType { get {return m_selectionType; } }
-        public SerializedObject serializedObject { get; }
+        public SerializedObject serializedObject { get; } // Note that state doesn't own this disposable object
 
         // Control schemes
         public int selectedControlSchemeIndex { get { return m_selectedControlSchemeIndex; } }
         public int selectedDeviceRequirementIndex { get  {return m_selectedDeviceRequirementIndex; } }
-        public InputControlScheme selectedControlScheme => m_ControlScheme;
+        public InputControlScheme selectedControlScheme => m_ControlScheme; // TODO Bad this either po
```
lyndon-unity, line 22: TODO seems to have missing text

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
```diff
@@ -54,34 +56,94 @@ internal struct InputActionsEditorState
...
-                var controlSchemeSerializedProperty = serializedObject
-                    .FindProperty(nameof(InputActionAsset.m_ControlSchemes))
-                    .GetArrayElementAtIndex(m_selectedControlSchemeIndex);
-
-                m_ControlScheme = new InputControlScheme(controlSchemeSerializedProperty);
+                if (m_selectedControlSchemeIndex >= controlSchemesArrayProperty.arraySize)
+                    m_selectedControlSchemeIndex = 0;
+                m_ControlScheme = new InputControlScheme(
+                    controlSchemesArrayProperty.GetArrayElementAtIndex(other.m_selectedControlSchemeIndex));
+                // TODO Preserve device requirement index
```
lyndon-unity, line 133: Incomplete TODO - not sure if this needs doing before landing. Seems no worse than it was

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -123,67 +124,86 @@ private static InputActionsEditorWindow GetOrCreateWindow(int id, out bool isAlr
...
+            var text = asset.name + " (Input Actions Editor)";
+            if (isDirty)
+                text = "(*) " + text;
+            return new GUIContent(text);
+        }
+
         private void SetAsset(InputActionAsset asset, string actionToSelect = null, string actionMapToSelect = null)
         {
-            m_AssetPath = AssetDatabase.GetAssetPath(asset);
-            var serializedAsset = new SerializedObject(asset);
```
lyndon-unity, line 129: ok looks like this moved to TryUpdateFromAsset

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -260,34 +271,48 @@ private void OnLostFocus()
...
-                case InputActionsEditorWindowUtils.ConfirmSaveChangesDialogResult.Cancel:
+                case Dialog.Result.Cancel:
                     // Cancel editor quit. (open new editor window with the edited asset)
                     ReshowEditorWindowWithUnsavedChanges();
                     break;
-                case InputActionsEditorWindowUtils.ConfirmSaveChangesDialogResult.DontSave:
+                case Dialog.Result.Discard:
                     // Don't save, quit - reload the old asset from the json to prevent the asset from being dirtied
-                    AssetDatabase.ImportAsset(m_AssetPath);
+                    // TODO I doubt this is needed since we were operating on a copy?!
```
lyndon-unity: Delete this then ?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
```diff
@@ -13,18 +15,35 @@ internal class InputActionsEditorWindowUtils
...
-            Cancel = 1,
-            DontSave = 2
+            public InputActionMap.WriteMapJson[] maps;
+            public InputControlScheme.SchemeJson[] controlSchemes;
         }
 
-        public static ConfirmSaveChangesDialogResult ConfirmSaveChanges(string path)
+        // Similar to InputActionAsset.ToJson() but converts to JSON excluding the name property and any additional JSON
+        // content that may be part of the file not recognized as required data.
+        public static string ToJsonWithoutName(InputActionAsset asset)
```
lyndon-unity, line 28: Are these actually used and do we really need them ?

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
```diff
@@ -140,6 +140,13 @@ public virtual void Setup()
                 InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseOptimizedControls, true);
                 InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kUseReadValueCaching, true);
                 InputSystem.settings.SetInternalFeatureFlag(InputFeatureNames.kParanoidReadValueCachingChecks, true);
+
+                // Default mock dialogs to avoid unexpected cancellation of standard flows
```
lyndon-unity, line 144: Handy to keep testing flowing

Packages/com.unity.inputsystem/Tests/TestFixture/ScopedDisposable.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using UnityEditor;
+
+namespace UnityEngine.InputSystem
+{
+    // Utility allowing access to object T as well as a dispose delegate to clean-up any resources associated with it.
+    // Useful with the 'using' keyword to provide scoped RAII-like cleanup of objects in tests without having a
+    // dedicated fixture handling the clean-up.
+    public sealed class ScopedDisposable<T> : IDisposable
```
lyndon-unity, line 9: This doesn't seem to be used unless I'm missing something

lyndon-unity (commented):
Seems ok - few comments about some todo's that can be removed.
Overall seems to be better, but couple of failing tests need fixing.

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: I tried that but ended up with so much escaping it was even more unreadable. Will try your provided sample though. The benefit with this one is that its comparable to the formatting generated by the JSON parser. I guess another alternative is to make it readable, convert an object from it, then back to json and store that as a "constant" to be used later.

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: Thanks, fixed in https://github.com/Unity-Technologies/InputSystem/pull/1850/commits/e8fb5dd4c1b6870163e809ae8bdd15e76270be85

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
ekcoh: Switching to random paths was not strictly necessary but the reason why I introduced that utility is the asset file mess that currently exists in our tests leaving all sorts of files as left overs. Hence why I created AssetDatabaseUtil.cs. Random paths may be annoying but originated as that to avoid all the clashes test cases caused when not removing things properly in teardown. The added util takes care of this and guarantees that there are never clashes even if file cleanup would fail.

ekcoh: My vision with it is to reuse in other tests to gradually clean this up down the road. Used it here since these tests were contributing to leaving test files and directories generating warnings in console after test run and leaving a dirty asset folder after test run.

Assets/Tests/InputSystem/CoreTests_Editor_InputActionAssetManager.cs
ekcoh: This is true, and this resulted in a warning, something like "XXX.meta file not removed". I changed this to move them in pairs instead to mimic what you should do in file system and what AssetDatabase does under the hood. The above is now a single method AssetDatabaseUtil.ExternalMoveFileOrDirectory. A corresponding exist using ADB instead.

Assets/Tests/InputSystem/CoreTests_Editor.cs
ekcoh: Yes, it was an odd design I modified since it didn't make sense to me and provided a somewhat messy API. Basically when this type is constructed via a constructor it doesn't make sense to require a call to Initialize after. This is now done as part of the constructor which is cleaner in my opinion. Initialize() remains available for serialisation scenarios when it should be called, but then no constructor is called.
The other odd thing here was that SaveChangesToAsset() would suprisingly throw if onDirtyChanged callback was not set. This didn't sit well I thought so I made it optional to set a callback and if you have registered one you get a callback on Save (transition to false), while its also fine to save without a callback.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: Correct. Why repeat the same code in multiple places.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindowUtils.cs
ekcoh: Yes they are used to convert an asset to JSON but exclude the name field. It is used for comparing two assets content to update dirty flags. @timkeo made a change previously which aligns JSON and SerializedObject names when desynchronised, but there is no need for editor to go dirty if that is the only change, similar to us discarding other "augmented" fields in json that users might enter. This also allowed removing a previous workaround for comparing whether an asset was empty or not that was a bit clunky.

ekcoh: I noticed now however that the Read version is not used anymore (refactored that one away) so I removed it in a coming commit.

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
ekcoh: Or to set them and actually test the dialogs in the future which currently has zero test coverage from auto tests.

Packages/com.unity.inputsystem/Tests/TestFixture/ScopedDisposable.cs
ekcoh: Correct, it was initially used but I managed to replace all its use with AssetDatabaseUtil.cs. I would like to keep it for now if OK since there are many use-cases for it in existing ill-formed tests that leak resources if any Assertion fails.

ekcoh: It works really nice with C# 8 "using var", but doesn't seem to work well with Unity 2019 then... :/ it will be usable as soon as modern C# can be used in the tests since "using var scoped = Scoped.CreateObject<T>();" would auto clean-up regardless of where test execution was aborted and avoids writing tons of TearDown functions to handle it for each different scenario.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: Removed in https://github.com/Unity-Technologies/InputSystem/pull/1850/commits/a8c93599cbc79b3f13fa0a0cad68d20483954d32

ekcoh:
> Seems ok - few comments about some todo's that can be removed.
> Overall seems to be better, but couple of failing tests need fixing.

Seems like the update on test utility and tests made those tests pass now. Other CI issues (also seen on other branches remain)

Pauliusd01:
Is it okay to get this "failed to delete" prompt? Cancel should cancel deletion instead of letting it fail. Although this does not seem to break anything so I'm fine with keeping it as is, if It's not worth it

https://github.com/Unity-Technologies/InputSystem/assets/54306142/ac77f253-9142-4a96-abee-3d175afbdfea

Pauliusd01 (approved):
LGTM. Checked deleting and renaming input assets while they're open in different ways, checked basic input functionality in Editor/Player and checked Input Action Window functionality after Editor restarts (This PR also fixes ISX-1904 so I've closed that one)

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -33,12 +41,16 @@ public string path
...
+        /// <summary>
+        /// Returns the name of the associated imported asset.
+        /// </summary>
         public string name
         {
             get
             {
-                if (m_ImportedAssetObject != null)
-                    return m_ImportedAssetObject.name;
+                var asset = importedAsset;
```
ritamerkl, line 51: looks like we could skip creating var asset here.

Packages/com.unity.inputsystem/InputSystem/Editor/Dialog.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            // Unsaved changes should be discarded and NOT be saved to the associated asset.
+            Discard = 2,
+
+            // User decided to delete the associated resource.
+            Delete = 3,
+        }
+
+        // User UI dialog windows related to InputActionAssets
+        public static class InputActionAsset
```
ritamerkl, line 30: naming: InputActionAssetDialog?

Packages/com.unity.inputsystem/InputSystem/Editor/Dialog.cs
```diff
@@ -0,0 +1,119 @@
...
+            // Shows a dialog prompting the user to discard changes or cancel the operation.
+            // May return Result.Discard or Result.Cancel.
+            public static Result ShowDiscardUnsavedChanges(string path)
+            {
+                return discardUnsavedChanges(path);
+            }
+        }
+
+        // User UI dialog windows related to InputControlSchemes
+        public static class ControlScheme
```
ritamerkl, line 90: naming: ControlSchemeDialog?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
```diff
@@ -242,18 +304,32 @@ private HashSet<int> GetOrCreateExpandedStateInternal()
...
         private SerializedProperty GetSelectedActionMap()
         {
-            return serializedObject
-                .FindProperty(nameof(InputActionAsset.m_ActionMaps))
-                .GetArrayElementAtIndex(selectedActionMapIndex);
+            return GetActionMap(serializedObject, selectedActionMapIndex);
         }
 
-        private readonly Dictionary<(string, string), HashSet<int>> m_ExpandedCompositeBindings;
+        private static SerializedProperty GetSelectedAction(SerializedProperty map, int actionIndex)
```
ritamerkl: Selectors has this method already.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
```diff
@@ -221,13 +234,23 @@ public class ViewState
             public int selectedControlSchemeIndex;
             public int selectedDeviceIndex;
         }
+
+        public override void DestroyView()
+        {
+            base.DestroyView();
+
+            Debug.Log("Destroy View");
```
ritamerkl: Maybe remove method? looks like this is redundant and just for debugging purpose.

ritamerkl (approved):
LGTM. The comments and explanation blocks are helpful and a good inspiration for documenting code. Added some minor things to consider.

ekcoh:
> LGTM. Checked deleting and renaming input assets while they're open in different ways, checked basic input functionality in Editor/Player and checked Input Action Window functionality after Editor restarts (This PR also fixes ISX-1904 so I've closed that one)

Awesome, always happy to have some positive side-effects on the open bug count :)

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
ekcoh: I guess you are right, its kind of intentional though since importedAsset when not just cached is a costly operation. Hence I keep a local to avoid any side-effects or cost introduced by calling importedAsset multiple times. Since there are many access to asset variable in this getter though and importedAsset is non-trivial I rather just keep the local. Otherwise if someone is inventive down the line with importedAsset this getter could be causing problems.

Packages/com.unity.inputsystem/InputSystem/Editor/Dialog.cs
ekcoh: Am I understanding correctly that you would you prefer `Dialog.InputActionAssetDialog.ShowSaveChanges(path)` in favour of `Dialog.InputActionAsset.ShowSaveChanges(path)`?  
Since already prefixed and scope reduced by outer `Dialog` class my intention was to avoid excessive long names and also avoid scope leakage into outer scope which happens when you just use namespace and auto-import. The idea here was to access Dialog since that is what you want to use and then for what kind of type and then use-case. I can change but just want to know if I understand correctly.

Packages/com.unity.inputsystem/InputSystem/Editor/Dialog.cs
ekcoh: Would fall under same comment/question as above.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
ekcoh: Yes but only using a InputActionState and it would be bad practise to pass this before object is fully constructed. I guess we could refactor Selectors to have to GetSelectedAction functions and also refactor the ActionMap ones to accept SerializedObject as an overload...

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
ekcoh: This should be removed yes, will fix

ekcoh:
> LGTM. The comments and explanation blocks are helpful and a good inspiration for documenting code. Added some minor things to consider.

Happy to hear that if it helped readability of the code. Just don't take the size and scattered focus of this PR as inspiration since its bad practise to create such a big PR covering too much ground.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
ritamerkl: Your suggestion sounds reasonable to me.

Packages/com.unity.inputsystem/InputSystem/Editor/Dialog.cs
ritamerkl: I think it's good to avoid any confusion with the names in all scenarios, if we are forced to always use Dialog.xx I think this is sufficient.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
ekcoh: Fixed in https://github.com/Unity-Technologies/InputSystem/pull/1850/commits/7541e17d9342e40e515f3f7637b3160ca139aeaf

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
ekcoh: Fixed in https://github.com/Unity-Technologies/InputSystem/pull/1850/commits/7541e17d9342e40e515f3f7637b3160ca139aeaf
Also found another method I could remove in this class

Packages/com.unity.inputsystem/InputSystem/Editor/Dialog.cs
ekcoh: ok will leave it as is for now then and we can reconsider in the future

PR #1853, opened 2024-02-22
change downloader-cli options to use --fast for il2cpp builds

stefanunity:
### Description

Il2cpp builds sometimes take a long time or even time out waiting for the latest to be built for the platform components.
Switch to use '--fast' in order to speed the builders up the same way the rest of the platform builders work.

### Changes made

-enabled --fast for il2cpp builder
-spelled out --wait option

### Notes

Internal change only.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/.yamato/config.metadata
+++ b/.yamato/config.metadata
@@ -28,7 +28,7 @@ platforms_win:
     flavor: b1.large
     runtime: StandaloneWindows64
     scripting-backend: Il2Cpp
-    installscript: unity-downloader-cli -c editor -c StandaloneSupport-IL2CPP -w -u
+    installscript: unity-downloader-cli -c editor -c StandaloneSupport-IL2CPP --wait --fast -u 
 platforms_nix:
   - name: mac
     type: Unity::VM::osx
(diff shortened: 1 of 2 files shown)


.yamato/upm-ci.yml
```diff
@@ -85,7 +85,7 @@ build_ios_{{ editor.version }}:
   commands:
     - {{ utr_install_nix }}
     - {{ unity_downloader_install }}
-    - unity-downloader-cli -c Editor -c iOS -u {{ editor.version }} --fast -w
+    - unity-downloader-cli -c Editor -c iOS -u {{ editor.version }} --fast --wait
```
ekcoh, line 88: This is no difference right? -w is equal to --wait? (Spelled out from description)

ekcoh: What does --wait imply?

.yamato/config.metadata
```diff
@@ -28,7 +28,7 @@ platforms_win:
     flavor: b1.large
     runtime: StandaloneWindows64
     scripting-backend: Il2Cpp
-    installscript: unity-downloader-cli -c editor -c StandaloneSupport-IL2CPP -w -u
+    installscript: unity-downloader-cli -c editor -c StandaloneSupport-IL2CPP --wait --fast -u
```
ekcoh, line 31: Seems like this and change below fast was added, what does it imply?

ekcoh (approved):
Would love to understand what --wait and --fast here implies for future reference. Still build took 27 minutes all-in-all which I guess is close enough to our average 23 minutes to be a valid fix. At least not 5 hour run like recent PRs. We should keep an eye out after this is merged to see if the problem seems gone or not. Good catch!

PR #1856, opened 2024-02-24
FIX: Bindings pasted to Action appear at top of the list (ISX-1899)

timkeo:
### Description

Fixes a couple copy/paste issues in the UITK Editor in which pasted bindings appeared at the top of the list instead of the bottom.

### Changes made

Modified the logic within `PasteBindingsFromClipboard()` and `PasteBindingOrComposite()` to update the "index" of pasted items so it appears at the bottom of their respective sets when their "root" node is selected:

- Bindings pasted to an Action are inserted at the end of the Action's current Binding list
- CompositePart Bindings are inserted at the end of the Component Binding's current set

When a Binding or CompositePart is selected during paste, the pasted item will still appear immediately below the selected item.

### Notes

I was only able to do a little of bit of testing myself, and would've like to do more before opening the PR but ran out of time. I figured it's better to just get the PR open and deal with any issues next week.

Having automated tests to cover these scenarios would be really good, but that'll take some time/effort, which I didn't think is appropriate at this point.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
@@ -1,6 +1,7 @@
 #if UNITY_EDITOR && UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using UnityEditor;
 
@@ -221,9 +222,17 @@ namespace UnityEngine.InputSystem.Editor
         {
             var actionMap = Selectors.GetSelectedActionMap(state)?.wrappedProperty;
             var bindingsArray = actionMap?.FindPropertyRelative(nameof(InputActionMap.m_Bindings));
-            var actions = actionMap?.FindPropertyRelative(nameof(InputActionMap.m_Actions));
-            var index = state.selectionType == SelectionType.Action ? Selectors.GetBindingIndexBeforeAction(actions, state.selectedActionIndex, bindingsArray) : state.selectedBindingIndex;
-            PasteData(EditorHelpers.GetSystemCopyBufferContents(), new[] {index}, bindingsArray);
+
+            int newBindingIndex;
+            if (state.selectionType == SelectionType.Action)
+            {
+                var bindingsForAction = Selectors.GetBindingsForAction(state, actionMap, state.selectedActionIndex);
+                newBindingIndex = bindingsForAction.Count > 0 ? bindingsForAction.Select(b => b.GetIndexOfArrayElement()).Max() : 0;
+            }
+            else
+                newBindingIndex = state.selectedBindingIndex;
+
+            PasteData(EditorHelpers.GetSystemCopyBufferContents(), new[] { newBindingIndex }, bindingsArray);
         }
 
         private static void PasteData(string copyBufferString, int[] indicesToInsert, SerializedProperty arrayToInsertInto)
@@ -241,7 +250,7 @@ namespace UnityEngine.InputSystem.Editor
             foreach (var transmission in copyBufferString.Substring(k_CopyPasteMarker.Length + k_TypeMarker[copiedType].Length)
                      .Split(new[] {k_EndOfTransmission}, StringSplitOptions.RemoveEmptyEntries))
             {
-                indexOffset += 1;
+                indexOffset++;
                 foreach (var index in indicesToInsert)
                     PasteBlocks(transmission, index + indexOffset, arrayToInsertInto, copiedType);
             }
@@ -298,14 +307,30 @@ namespace UnityEngine.InputSystem.Editor
         private static int PasteBindingOrComposite(SerializedProperty arrayProperty, string json, int index, string actionName, bool createCompositeParts = true)
         {
             var pastePartOfComposite = IsPartOfComposite(json);
+            bool currentPartOfComposite = false;
+            bool currentIsComposite = false;
+
             if (index > 0)
             {
                 var currentProperty = arrayProperty.GetArrayElementAtIndex(index - 1);
-                var currentIsComposite = IsComposite(currentProperty) || IsPartOfComposite(currentProperty);
+                currentPartOfComposite = IsPartOfComposite(currentProperty);
+                currentIsComposite = IsComposite(currentProperty) || currentPartOfComposite;
                 if (pastePartOfComposite && !currentIsComposite) //prevent pasting part of composite into non-composite
                     return index;
             }
-            index = pastePartOfComposite || s_State.selectionType == SelectionType.Action ? index : Selectors.GetSelectedBindingIndexAfterCompositeBindings(s_State) + 1;
+
+            // Update the target index for special cases
+            if (pastePartOfComposite && !currentPartOfComposite)
+            {
+                // Pasting into a Composite and CompositePart isn't the target, i.e. Composite "root" selected, paste at the end of the composite
+                index = Selectors.GetSelectedBindingIndexAfterCompositeBindings(s_State) + 1;
+            }
+            else if (!pastePartOfComposite && s_State.selectionType != SelectionType.Action)
+            {
+                // Pasting with a Binding selected needs to skip all the CompositeParts (if any)
+                index = Selectors.GetSelectedBindingIndexAfterCompositeBindings(s_State) + 1;
+            }
+
             if (json.Contains(k_BindingData)) //copied data is composite with bindings - only true for directly copied composites, not for composites from copied actions
                 return PasteCompositeFromJson(arrayProperty, json, index, actionName);
             var property = PasteElement(arrayProperty, json, index, out var oldId, "", false);
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
@@ -1,7 +1,6 @@
 #if UNITY_EDITOR && UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Text;
 using UnityEditor;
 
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
```diff
@@ -298,14 +307,30 @@ private static void PasteAction(SerializedProperty arrayProperty, string jsonToI
...
+            // Update the target index for special cases
+            if (pastePartOfComposite && !currentPartOfComposite)
+            {
+                // Pasting into a Composite and CompositePart isn't the target, i.e. Composite "root" selected, paste at the end of the composite
+                index = Selectors.GetSelectedBindingIndexAfterCompositeBindings(s_State) + 1;
+            }
+            else if (!pastePartOfComposite && s_State.selectionType != SelectionType.Action)
+            {
+                // Pasting with a Binding selected needs to skip all the CompositeParts (if any)
+                index = Selectors.GetSelectedBindingIndexAfterCompositeBindings(s_State) + 1;
```
ritamerkl: the if/else have identical content. maybe declare boolean variables before and check them into one combined if statement?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
```diff
@@ -1,6 +1,7 @@
 #if UNITY_EDITOR && UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
ritamerkl: The new guidelines recommend to not use Linq anymore. I know it is on various places in our code already, but I think it makes sense to try not using it in the new/changed code.

ritamerkl (commented):
LGTM! I think there is a changelog entry missing.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
ekcoh: Agreed, lets try to avoid introducing more.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
```diff
@@ -221,9 +222,17 @@ private static void PasteBindingsFromClipboard(InputActionsEditorState state)
...
-            PasteData(EditorHelpers.GetSystemCopyBufferContents(), new[] {index}, bindingsArray);
+
+            int newBindingIndex;
+            if (state.selectionType == SelectionType.Action)
+            {
+                var bindingsForAction = Selectors.GetBindingsForAction(state, actionMap, state.selectedActionIndex);
+                newBindingIndex = bindingsForAction.Count > 0 ? bindingsForAction.Select(b => b.GetIndexOfArrayElement()).Max() : 0;
+            }
+            else
+                newBindingIndex = state.selectedBindingIndex;
```
ekcoh, line 229: Minor: Would suggest not mixing brackets and non brackets, i.e. use brackets on the else.

ekcoh (approved):
Looks good, some minor comments. Comments from @ritamerkl should be addressed but given they are this looks good from my perspective.

Pauliusd01 (changes requested):
This breaks pasting composite bindings into non composite bindings checks. Steps to repro: Create a composite binding action -> Copy or Cut and paste it to a non composite binding -> errors thrown (No repro in Develop)

https://github.com/Unity-Technologies/InputSystem/assets/54306142/11d74de8-94bb-4f45-af2f-77368b94c81f

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
timkeo: The "Select()" call from `bindingsForAction.Select`, which I copied over from `/UITKAssetEditor/Views/Selectors.cs` is what's using it.

I'll restore the function `Selectors.GetLastBindingIndexForSelectedAction()`, which contained this logic (Selectors.cs already uses LINQ) and call into that from here.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
timkeo: The logic was kinda confusing for me, and so I did it this way because I found it easier to reason through, but I do agree it's not optimal. I'll see if can rework it so it's not duplicating checks but still not too confusing.

ritamerkl: How about if you take the content of the if statements into two booleans with comments?

timkeo: Yeah that's a good suggestion, and I think I'll go that way.

timkeo: Reworked the conditions to remove duplicated line, and also fix regression pasting Composite: 630fe8d551ecfe45a0245c77583f76b8faf1f285

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
timkeo: Moved the code using LINQ back into Selectors.cs with: 630fe8d551ecfe45a0245c77583f76b8faf1f285

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/CopyPasteHelper.cs
timkeo: Fixed with 630fe8d551ecfe45a0245c77583f76b8faf1f285.

ritamerkl (approved):
As far as I see still no changelog or do we skip that for this fix? 
Looks good for me, would be fine to merge after Paulius approved as well.

Pauliusd01:
Pasting composite bindings into non composite actions is now correctly blocked but it also clears the clipboard when trying to paste incorrectly. For example: You cut a composite binding and try to paste it into a non composite action -> nothing happens -> you try again on a composite action -> still nothing happens (On Develop it pastes correctly in this step)

Pauliusd01 (changes requested):
forgot to update status

timkeo:
> As far as I see still no changelog or do we skip that for this fix? Looks good for me, would be fine to merge after Paulius approved as well.

Ah you're right, I completely missed that!
I guess I see this fix as a Multi-Copy-Paste feature detail and probably not something that needs to be called out.

timkeo:
> Pasting composite bindings into non composite actions is now correctly blocked but it also clears the clipboard when trying to paste incorrectly. For example: You cut a composite binding and try to paste it into a non composite action -> nothing happens -> you try again on a composite action -> still nothing happens (On Develop it pastes correctly in this step)

I was unable to repro this, but I did find and fix the other issue you came across (errors logged to the console in when copy/pasting CompositeParts). This was due to a flaw in `GetSelectedBindingIndexAfterCompositeBindings()`, which wasn't checking for `index` going beyond the array size. This issue only occurs when the Composite binding you're trying to paste into is last in the Actions.

The specific issue you pointed out didn't occur in develop because we weren't trying to paste at the end of the Composite list before. However, I did some additional testing and found another case where this error occurs (in Develop), and verified the change fixes those as well.

In general, I see a lot of places where we call `GetArrayElementAtIndex()` for a Bindings property without checking the `index` against `arraySize`. I think this is generally fine if/when we expect `bindingIndex` to be valid, but we should be testing cut/copy/paste/delete scenarios against Bindings (regular and Composite) at the bottom of the Actions list.

Pauliusd01:
Cutting and pasting bindings into an empty action throws nullrefs

https://github.com/Unity-Technologies/InputSystem/assets/54306142/8ce93b01-bd44-4b64-a0b7-ccf975455e06

Trying to cut a 2nd time after pasting bindings with composites somehow only cuts the action properties and throws nullrefs 

https://github.com/Unity-Technologies/InputSystem/assets/54306142/f2b9aead-6e3d-41a0-b37c-33964e6de7a3

Pauliusd01 (changes requested):
Updating status - waiting for fixes

timkeo:
> Cutting and pasting bindings into an empty action throws nullrefs
> Unity_2024-02-28_12-54-31.mp4
> 
> Trying to cut a 2nd time after pasting bindings with composites somehow only cuts the action properties and throws nullrefs
> Unity_2024-02-28_12-53-26.mp4

Both of these issues seems to stem from the same problem: we don't properly set the `selectedBindingIndex` value after cutting a Binding. This _should_ be solved with @ritamerkl [Cut/Paste/Delete fix PR](https://github.com/Unity-Technologies/InputSystem/pull/1858), since Cut doesn't delete the item immediately and avoids us getting into this bad state. 

I tested my changes with Rita's branch and wasn't able to repro either of these issues, However, I decided to fix the 1st issue (pasting to an empty Action) by adding another empty Bindings Array check. Although it should be unnecessary with Rita's work it's a simple, safe change. The 2nd issue is much more complex and can't be easily fixed without duplicating/conflicting with Rita's changes.

I propose opening a ticket to re-investigate these issue (and `selectedBindingIndex` in general) after both our PRs are merged and address any remaining problems then.

Pauliusd01:
> > Cutting and pasting bindings into an empty action throws nullrefs
> > Unity_2024-02-28_12-54-31.mp4
> > Trying to cut a 2nd time after pasting bindings with composites somehow only cuts the action properties and throws nullrefs
> > Unity_2024-02-28_12-53-26.mp4
> 
> Both of these issues seems to stem from the same problem: we don't properly set the `selectedBindingIndex` value after cutting a Binding. This _should_ be solved with @ritamerkl [Cut/Paste/Delete fix PR](https://github.com/Unity-Technologies/InputSystem/pull/1858), since Cut doesn't delete the item immediately and avoids us getting into this bad state.
> 
> I tested my changes with Rita's branch and wasn't able to repro either of these issues, However, I decided to fix the 1st issue (pasting to an empty Action) by adding another empty Bindings Array check. Although it should be unnecessary with Rita's work it's a simple, safe change. The 2nd issue is much more complex and can't be easily fixed without duplicating/conflicting with Rita's changes.
> 
> I propose opening a ticket to re-investigate these issue (and `selectedBindingIndex` in general) after both our PRs are merged and address any remaining problems then.

Thank you Tim, It's a bit confusing to review when there's multiple ongoing PRs on the same thing. I'll approve this, we will have to QA the full release anyway and hopefully any missed bugs should spring up then

Pauliusd01 (approved):

PR #1861, opened 2024-02-27
FIX: Processors/Interactions not added to selected item after reordering (ISX-1873)

timkeo:
### Description

When reordering Actions/Bindings, the selection doesn't change and neither do any of the other values used by PropertyView's Selectors to detect state changes. That is, PropertyView didn't detect the reordering and therefore didn't update it's UI.

However, with this operation, the SerializedProperty's `propertyPath` changes, but PropertyView (since it wasn't updated) was still referencing the old `propertyPath`. This caused the Processor/Interaction to be added to the wrong item.

### Changes made

To fix this issue, I added the `propertyPath` string to `SerializedInputAction` and `SerializedInputBinding` equality checkers. This allows reordering of Actions/Bindings to trigger state change for PropertyView (and other views).

### Notes

This isn't the change I proposed at Standup; turns out my original idea caused other problems. But I think this is a better and more eloquent fix.

Since this fix is related to Drag/Drop feature, I'll forego adding a CHANGELOG entry; I see this as an implementation detail for the larger feature.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/SerializedInputAction.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/SerializedInputAction.cs
@@ -59,7 +59,8 @@ namespace UnityEngine.InputSystem.Editor
                 && processors == other.processors
                 && initialStateCheck == other.initialStateCheck
                 && actionTypeTooltip == other.actionTypeTooltip
-                && expectedControlTypeTooltip == other.expectedControlTypeTooltip;
+                && expectedControlTypeTooltip == other.expectedControlTypeTooltip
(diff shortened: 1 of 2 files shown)


Pauliusd01 (approved):

ekcoh (changes requested):
Fix looks good. Think its fine to land this one without changelog since no public report.

ekcoh (approved):

PR #1863, opened 2024-02-29
[Input] Stop accumulation of duplicate InputActions (case ISX-1902)

AlexTyrer:
### Description

FIX: Stop the accumulation of duplicate Actions being created when loading/unloading the Input System test scenes.

### Changes made

o ProjectWideInputActions are now only enabled in playmode - the state is controlled by EditorApplication.playModeStateChanged listener PlayStateNotifier.ModeChanged()

o Only create DefaultInputActions once in AssignDefaultActions() (controlled with class static defaultActions)

o Improve Input Debug window Action item display (show which action map asset each action was created from)

### Notes

InputSystem.actions are now only enabled in playmode.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
```diff
@@ -927,6 +927,15 @@ private unsafe void AddActionItem(TreeViewItem parent, InputAction action, ref i
                 var name = action.actionMap != null ? $"{action.actionMap.name}/{action.name}" : action.name;
                 if (!action.enabled)
                     name += " (Disabled)";
+                if (action.actionMap != null && action.actionMap.m_Asset != null)
```
ekcoh, line 930: I think this is a helpful change. Since its also user facing I think it should be added under "Added" title in CHANGELOG.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3011,6 +3011,41 @@ internal static bool ShouldDrawWarningIconForBinding(string bindingPath)
 
 #if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
 
+#if UNITY_EDITOR
+        [InitializeOnLoad]
+        public static class PlayStateNotifier
```
ekcoh, line 3016: Its likely fine to keep this isolated but I also noticed we have both monitoring and evaluations of playmode state changes in https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs#L341-L362 and https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/InputManager.cs#L323

Wanted to share just for awareness. Might be better to keep this isolated.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3011,6 +3011,41 @@ internal static bool ShouldDrawWarningIconForBinding(string bindingPath)
...
+            static void ModeChanged(PlayModeStateChange playModeState)
+            {
+                // EnteredEditMode  Occurs during the next update of the Editor application if it is in edit mode and was previously in play mode.
+                // ExitingEditMode  Occurs when exiting edit mode, before the Editor is in play mode.
+                // EnteredPlayMode  Occurs during the next update of the Editor application if it is in play mode and was previously in edit mode.
+                // ExitingPlayMode  Occurs when exiting play mode, before the Editor is in edit mode.
+                switch (playModeState)
+                {
+                    case PlayModeStateChange.EnteredEditMode:
+                        actions?.Disable();
```
ekcoh, line 3038: Was there a reason for using this rather than ExitingPlayMode?

ekcoh: Just wondering if there might be any implications of doing it after the mode change "edge"?

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3011,6 +3011,41 @@ internal static bool ShouldDrawWarningIconForBinding(string bindingPath)
...
+#if UNITY_EDITOR
+        [InitializeOnLoad]
+        public static class PlayStateNotifier
+        {
+            static PlayStateNotifier()
+            {
+                EditorApplication.playModeStateChanged += ModeChanged;
+            }
+
+            static void ModeChanged(PlayModeStateChange playModeState)
```
ekcoh, line 3023: If not already checked it would be good to investigate and document along with the already added inline docs (which is great by the way) when the actions may be expected to be enabled. E.g.

Are they enabled when entering Awake() on a MonoBehavior?
Are they enabled when entering Start() on a MonoBehavior?
Are they enabled when entering OnEnable() on a MonoBehavior?
Are they disabled when entering OnDisable() on a MonoBehavior?
Are they disabled when entering OnDestroy() on a MonoBehavior? (When destroy by exiting play mode)

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1388,6 +1394,7 @@ public void AssignDefaultActions()
         /// <seealso cref="AssignDefaultActions"/>
         public void UnassignActions()
         {
+            defaultActions = default;
```
ekcoh, line 1398: If assigning null to default actions like this it seems like we will leak memory. DefaultInputActions seems to implement IDisposable so should be?
```

defaultActions?.Dispose();
defaultActions = default;
```

ekcoh (approved):
In general changes look good. I have some questions and some suggestions related to making sure its solid.

ekcoh:
It might be good for @lyndon-unity to review this PR since he was the may author of adding those enable() and disable() calls into InputSystem.cs in a previous PR.

ekcoh (changes requested):
Filing a request change for updating CHANGELOG.md, there should be one ADD entry for the Input Debugger addition, and a FIX entry for the accumulated actions from InputUIModule I believe.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
AlexTyrer: I tried both ways and didn't see any difference.

It wasn't immediately clear to me if one was preferable to the other.

From reading the docs perhaps I should make this change to: EnteredPlayMode and ExitingPlaymode.

I can see an argument for both ways - it's unclear which are the most "balanced" - the enums and the comments about the enums balance differently.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
AlexTyrer: I see that other option now, thanks - was just focussed on InputSystem in isolation.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
AlexTyrer: Right - that's a good call, will change.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
AlexTyrer: Will add those lines.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3011,6 +3011,41 @@ internal static bool ShouldDrawWarningIconForBinding(string bindingPath)
...
+                {
+                    case PlayModeStateChange.EnteredEditMode:
+                        actions?.Disable();
+                        break;
+
+                    case PlayModeStateChange.ExitingEditMode:
+                        break;
+
+                    case PlayModeStateChange.EnteredPlayMode:
+                        actions?.Enable();
```
lyndon-unity, line 3045: I think MonoBehaviour's Start() is called before PlayModeStateChange.EnteredPlayMode
So when I (previously - not with this PR) tested to see if actions were enabled in my Start function, it was not enabled.
This is why I added the enable in 'InitializeInEditor' too. So it should already be enabled by here.
(history is in this PR : https://github.com/Unity-Technologies/InputSystem/pull/1834/files ) 

And another comment:
PlayModeStateChange.ExitingPlayMode gets hit before OnDisable 
(But I see you are using EnteredEditMode which is later)

lyndon-unity: So I think there is a test in here to update:
Assets/Samples/ProjectWideActions/ProjectWideActionsExample.cs

lyndon-unity: ^ That test probably doesn't need updating as the Enable in the InitializeInEditor code is still present.

lyndon-unity (approved):
I think this looks good overall

ekcoh:
Thought about this a bit and generally this is just connecting an assets enable() and disable() logic to edit mode and play mode callbacks. I think we should at least consider (may be done after this PR lands) moving this out of InputSystem.cs and into InputActionAsset instead. This way, when the callbacks trigger we could easily just check if `InputSystem.actions == this` to determine if we should act. But more interestingly, it would open up for making this a setting/property on the asset instead and allow basically any InputActionAsset to have this behavior, while at the same time make it easy for users to opt-in or opt-out of that behavior both for project-wide or Any InputActionAsset.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
AlexTyrer: OK, I've tested the various combinations here:

Using PlayModeStateChange.**EnteredEditMode** / PlayModeStateChange.**EnteredPlayMode**

```
    InputSystem.actions are
        Awake       enabled
        OnEnable    enabled
        Start       enabled
        OnDisable   enabled
        OnDestroy   enabled
```

Using PlayModeStateChange.**ExitingPlayMode** / PlayModeStateChange.**ExitingEditMode**

```
    InputSystem.actions are
        Awake       DISABLED
        OnEnable    DISABLED
        Start       DISABLED
        OnDisable   enabled
        OnDestroy   enabled
```
(they show as enabled in the OnDisable / OnDestroy as they are explicitly enabled in the sample's Start method if they are not already)
   
So I believe using the **Entered*** calls is the right thing to do here.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
AlexTyrer: I have added an explanatory to the source (see comment I made to @lyndon-unity [below](https://github.com/Unity-Technologies/InputSystem/pull/1863#discussion_r1508913457))

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
AlexTyrer: The **Entered*** calls are the right ones I think - see my comments below.

AlexTyrer:
> Filing a request change for updating CHANGELOG.md, there should be one ADD entry for the Input Debugger addition, and a FIX entry for the accumulated actions from InputUIModule I believe.

I've adde those here in commit **[b54385a](https://github.com/Unity-Technologies/InputSystem/pull/1863/commits/b54385a1692b0ccb92560921b662b5be0bf835e2)**

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -57,6 +58,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed Input Actions Editor losing reference to current ControlScheme upon entering Play Mode [ISXB-770](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-770).
 - Fixed Save shortcut (ctrl/cmd + S by default) not saving changes in Input Actions Editor windows. [ISXB-659](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-659).
 - Fixed headers in InputActionEditor windows becoming squashed when there is a large number of Action Maps/Actions.
+- Fixed duplication of project wide input actions when loading/unloading scenes [ISX-1902]
```
ekcoh: Remove ISX-1902 from this since its an internal ticket. We only link public bug tickets ISXB and for them use issue tracker links.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDebuggerWindow.cs
ekcoh: thanks

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: I think we should keep it isolated, will resolve this

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: Awesome, thanks, that is helpful

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: This should ideally be covered in tests and I will try to cover that as part of modifying the tests which I currently do on a pending PR. Will resolve this.

ekcoh (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: @AlexTyrer This one needs to be fixed before merge, but approved to not block this PR from landing.

ekcoh:
@AlexTyrer this branch should also get some QA assessment I believe. Adding @Pauliusd01

Packages/com.unity.inputsystem/CHANGELOG.md
AlexTyrer: OK - I saw there are other some internal ISX- issues in the file (with no links) so assumed that was the thing to do.

Removed in 1da1021

AlexTyrer:
> @AlexTyrer this branch should also get some QA assessment I believe. Adding @Pauliusd01

Good idea, thanks.

@Pauliusd01 you should just be able to run the samples and go in and out of all the scenes - you will see the actions in the "Input Debug" window come and go as each scene loads / unloads.

Exiting playmode should go back to the original state.

Note now that the project wide input actions are now **never** enabled in edit mode (only in playmode / the player) - so they won't be shown in the Input Debug window in edit mode.

Pauliusd01:
This also fixes ISX-1602 which was closed for being low priority

Pauliusd01 (approved):

PR #1865, opened 2024-02-29
FIX: ISX-1889 Improved robustness for selection of project-wide input actions for player builds.

ekcoh:
### Description

A fix for ticket ISX-1889 which is about improving the determination of which of any preloaded InputActionAssets is the selected project-wide input action asset. This is trivial in edit mode since we are able to store a reference to the asset as a ConfigObject in EditorBuildPrefs, but in a built-player we cannot rely on ConfigObject. 

I initially attacked this problem with a JSON-based tag/property solution which turned out to be quite cumbersome, branch here for reference: https://github.com/Unity-Technologies/InputSystem/tree/isx-1889-project-wide-robustness-json-tag. When more or less done with it and testing it with a built player I realised something and did some additional testing and bypassing that solution.

This led me to the simple solution being included in this PR which do not include JSON at all but rather relies on Unity built-in serialisation which can be taken advantage of when dealing with a non-file preloaded asset.

### Changes made

Added an internal serialised boolean field to track whether an asset is project-wide or not.
This field is set during build by our build settings provider and since its internal a Unity user cannot set it.
Modified code in InputSystem applying to Player builds that iterates over all preloaded assets and picks the first one with the flag set. The flag is never used in edit mode. This should be sufficient to guard against any user-added preloaded assets and we do not need to guard against our own since we only include one. 

### Notes

Difficult to test without creating scripts to preload additional assets.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsBuildProvider.cs
```diff
@@ -25,8 +25,12 @@ public void OnPreprocessBuild(BuildReport report)
 
 #if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
             // Determine if we need to preload project-wide InputActionsAsset.
+            // Also make sure we set project wide flag on object that gets serialized into build.
+            // It will allow us to find the right object if multiple preloaded assets
+            // Input System asset + user assets.
             var actions = InputSystem.actions;
             var actionsMissing = NeedsToBeAdded(preloadedAssets, actions, ref newSize);
+            actions.m_IsProjectWide = true;
```
lyndon-unity: Should we be making sure to set this and remove this each time an asset is assigned to the 'actions' in general ?

I'm just not sure if this could end up persisting on the asset after its unassigned from actions.

ekcoh: I did exactly that when I put this together and then I realised that we only bundle a single preloaded asset anyway so it doesn't matter. All other assets are accessed from either disk/json and when deserialised/imported that way that field will always be false (not mapped from json). If this would be set on an InputActionAsset object as part of a build and then the object is swapped to another you are right that both of them might have the flag set in Unity default serialisation mechanism. But there wouldn't be any harm since the only access to this is internally and in the Resource.LoadAllResources during initialisation which will only retrieve preloaded assets. Hence I believe the only risk would be that a user first create an asset and use it as project-wide within a build. Then switches it, then add their own BuildSettingsProvider to inject it as a preloaded asset. I guess what we could do would be to add into InputActionAsset OnValidate function to set it to false if not equal to InputSystem.asset to make it even harder if you believe its necessary?

Also note that I decided to leave InputSettings alone (still uses the brittle approach which it has since for as long as it has existed), wonder if that should be filed as a separate future ticket to correct with same pattern.

ekcoh: And to answer your question, it could until its reimported I believe.

Pauliusd01 (approved):
Checked ProjectWideActions functionality in Editor and Player, did not notice any issues. If you have any suggestions on what specific areas need breaking let me know I'll check them.

ekcoh:
> Checked ProjectWideActions functionality in Editor and Player, did not notice any issues. If you have any suggestions on what specific areas need breaking let me know I'll check them.

Thanks but I believe the only good way to test this is via auto-tests with internal visibility into source assembly.

Packages/com.unity.inputsystem/InputSystem/Editor/Settings/InputSettingsBuildProvider.cs
ekcoh: I did the change you requested but wonder if we should really do it that way since only build related.

PR #1879, opened 2024-03-06
Remove Probuilder Package from the project to unblock trunk CI

Pauliusd01:
### Description

Opened and ran every scene in the project to check for object dependencies. None were found so the package isn't even used anywhere in the project and is safe to remove.

### Changes made

Removed the unneeded Probuilder package from the package manifest.json

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ekcoh (approved):
Code diff looks good, any unnecessary references we can get rid of is good. Would suggest looking through scenes and also check with 2019 that everything is ok there.

stefanunity (approved):

PR #1880, opened 2024-03-06
FIX: Visual glitch in Actions list when scrolling the view while renaming an item (ISXB-748)

timkeo:
### Description
In UITK, scrolling List/Tree items off the visible area causes them to be "unbound" from the active item set, which causes [issues with the Rename functionality](https://unity.slack.com/archives/C3414V4UV/p1683713461312229).

In this case, although Reset is called for `ActionTreeViewItem` being unbound, which resets rename tracking fields, the Rename TextEdit control remains active causing items at the top of the list to appear like they've been renamed. Their names haven't _actually_ been changed (just a visual glitch) but it's quite ugly.

This change fixes the problem by canceling the Rename operation (removes Rename TextEdit box) when a `WheelEvent` is received.

### Changes made

- Hooks up a `WheelEvent` handler to ActionTreeView to cancel Rename operation (if one is active)
- Added a static `s_EditingItem` to the Item class for tracking item currently being edited
- Applied same fix to ActionMapsTreeView (has the same bug)
- Some minor refactoring within ActionsTreeView

### Notes

This fix only covers scrolling from the Mouse Wheel and (probably) won't work if the list is scrolled by some other means. For some strange reason, there doesn't appear to be a `ScrollEvent` or a (public) OnScroll() override on the ScrollView.

Note that Rename TextEdit has keyboard focus (so arrow keys aren't a concern) and a left click cancels rename, but dragging via Touch could still be a problem. This fix should be good enough for now, but the Rename functionality should be revisited later.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -77,6 +77,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue in the InputActionAsset Editor where ControlType wasn't updated when ActionType changed.
 - Fixed an issue in the InputActionAsset Editor where Canceling ControlScheme changes didn't reset the values in the UI.
 - Fixed an issue where newly created action map names were not editable.
+- Fixed a visual glitch in the InputActionAsset Editor when scrolling the Actions list with a rename in progress. [ISXB-748](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-748)
 
 ## [1.8.0-pre.2] - 2023-11-09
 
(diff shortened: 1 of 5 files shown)


ekcoh (approved):
Changes LGTM. Would you be able to reach out to UITK team to confirm that there is no scroll event we could utilise for a long term fix?

Pauliusd01:
Not sure if this always existed and your changes just made it easier to notice but I can mess up the context menu if I open the context menu quickly enough after scrolling. I do not think this is important though, unlikely to happen naturally

https://github.com/Unity-Technologies/InputSystem/assets/54306142/cc7fdc49-eda5-4918-80d2-a81b34e22d25

Pauliusd01 (approved):

PR #1881, opened 2024-03-07
FIX: Give an appropriate error message if an asset went missing (ISX-1921)

graham-huws:
### Description

Fix confusing nullref messages if user deleted/moved an input asset which was previously open while the editor was closed ([ISX-1921](https://jira.unity3d.com/browse/ISX-1921))

### Changes made
More understandable error is now shown:
image

### Notes
This apparently started happening recently, so no need for changelog

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1881) <br/>All committers have signed the CLA.

Pauliusd01:
I got this error instead of the one in the screenshot:
```

ArgumentException: Object at index 0 is null
UnityEditor.SerializedObject.InternalCreate (UnityEngine.Object[] monoObjs, UnityEngine.Object context) (at <3e9bfdebc22c4d658ec054a816b3689c>:0)
UnityEditor.SerializedObject..ctor (UnityEngine.Object obj) (at D:/Gitrepo/unity/Editor/Mono/SerializedObject.bindings.cs:22)
UnityEngine.InputSystem.Editor.InputActionsEditorWindow.CreateGUI () (at ./Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs:177)
UnityEditor.RetainedMode:UpdateSchedulers() (at D:/Gitrepo/unity/Modules/UIElementsEditor/RetainedMode.cs:57)
```

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/InputActionAssetManager.cs
```diff
@@ -198,7 +198,7 @@ internal void SaveChangesToAsset()
             // If this is invoked after a domain reload, importAsset will resolve itself.
             // However, if the asset do not exist importedAsset will be null and we cannot complete the operation.
             if (importedAsset == null)
-                throw new Exception("Unable to save changes. Associated asset do not exist.");
+                throw new Exception("Unable to save changes. Associated asset does not exist.");
```
ekcoh, line 201: Thanks for correcting this, but just wanted to highlight that InputActionAssetManager is currently only used for IMGUI editor, i.e. InputActionEditorWindow

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -169,6 +169,10 @@ private void SetAsset(InputActionAsset asset, string actionToSelect = null, stri
                 {
                     var assetPath = AssetDatabase.GUIDToAssetPath(m_AssetGUID);
                     var asset = AssetDatabase.LoadAssetAtPath<InputActionAsset>(assetPath);
+
+                    if (asset == null)
```
ekcoh, line 174: Hmmm is throwing the right thing to do here? Check the other places where we have fallbacks, we typically Debug.LogError + call Close() then return. Otherwise it will leak memory and maybe also relates to Paulius error? Or maybe we are good here since coming back from domain reload and all we have are the serialised objects. But at least m_AssetForEditing would leak.

ekcoh (changes requested):
Requesting changes to avoid memory leak. Paulius has also reported an interesting but hard to decipher error.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: The Close() would in turn trigger Destroy which would cleanup memory, destroy UI etc.

timkeo:
> I got this error instead of the one in the screenshot:
> 
> ```
> 
> ArgumentException: Object at index 0 is null
> UnityEditor.SerializedObject.InternalCreate (UnityEngine.Object[] monoObjs, UnityEngine.Object context) (at <3e9bfdebc22c4d658ec054a816b3689c>:0)
> UnityEditor.SerializedObject..ctor (UnityEngine.Object obj) (at D:/Gitrepo/unity/Editor/Mono/SerializedObject.bindings.cs:22)
> UnityEngine.InputSystem.Editor.InputActionsEditorWindow.CreateGUI () (at ./Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs:177)
> UnityEditor.RetainedMode:UpdateSchedulers() (at D:/Gitrepo/unity/Modules/UIElementsEditor/RetainedMode.cs:57)
> ```

I investigate this and discovered it occurs when the asset's .meta file is deleted (while Editor is closed) but not the actual asset; `m_AssetObjectForEditing` is null in this case and calling `new SerializedObject(m_AssetObjectForEditing)` throws the exception. A simple null check fixes this issue and allows the Asset Editor window to stay open after Unity is launched.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
timkeo: > Hmmm is throwing the right thing to do here? Check the other places where we have fallbacks, we typically Debug.LogError + call Close() then return. Otherwise it will leak memory and maybe also relates to Paulius error? Or maybe we are good here since coming back from domain reload and all we have are the serialised objects. But at least m_AssetForEditing would leak.

This code is enclosed in a `try...catch` block that handles exceptions as you stated: logs an error and calls `Close()` on the window. I verified this does indeed happen when reproing this scenario and also noted this pattern is used elsewhere in the source file.

So, all that to say, I think this is good :)

graham-huws:
> > I got this error instead of the one in the screenshot:
> > ```
> > 
> > ArgumentException: Object at index 0 is null
> > UnityEditor.SerializedObject.InternalCreate (UnityEngine.Object[] monoObjs, UnityEngine.Object context) (at <3e9bfdebc22c4d658ec054a816b3689c>:0)
> > UnityEditor.SerializedObject..ctor (UnityEngine.Object obj) (at D:/Gitrepo/unity/Editor/Mono/SerializedObject.bindings.cs:22)
> > UnityEngine.InputSystem.Editor.InputActionsEditorWindow.CreateGUI () (at ./Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs:177)
> > UnityEditor.RetainedMode:UpdateSchedulers() (at D:/Gitrepo/unity/Modules/UIElementsEditor/RetainedMode.cs:57)
> > ```
> 
> I investigate this and discovered it occurs when the asset's .meta file is deleted (while Editor is closed) but not the actual asset; `m_AssetObjectForEditing` is null in this case and calling `new SerializedObject(m_AssetObjectForEditing)` throws the exception. A simple null check fixes this issue and allows the Asset Editor window to stay open after Unity is launched.

Fab, thanks Tim!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: Thanks @timkeo , realised I overlooked that in this review but realised when I checked out the branch. I also think this is all good. Will approve this now. Thanks for the collaboration on this PR.

ekcoh (approved):
LGTM now, great to see some collaborative PRs @graham-huws @timkeo

Pauliusd01 (changes requested):
If the asset ISN'T deleted after a restart then it is opened blank with no UI headers, actions, etc

As for the deleted scenario it is also opened as blank for 2-3 secs and then automatically closed

graham-huws:
I believe this is all now fixed - I copied what TryUpdateFromAsset does later in the file, does this look correct @ekcoh?

ekcoh (approved):

ekcoh:
> I believe this is all now fixed - I copied what TryUpdateFromAsset does later in the file, does this look correct @ekcoh?

I believe so, lets refactor those two into one post-release. Will file a ticket.

Pauliusd01 (approved):
LGTM. Checked having the input actions window open/docked/docked,unfocused and restarting the editor with that asset being unchanged/deleted. (For docked,unfocused windows that were deleted they still show up as a tab that you can click but as soon as you try to open it you get the error that the asset has been deleted, which I think is fine)

PR #1883, opened 2024-03-08
FIX: Default control scheme should be "All" (ISX-1929)

graham-huws:
### Description

Default control scheme should be "All" in InputActionsEditor windows - [ISX-1929](https://jira.unity3d.com/browse/ISX-1929).

### Changes made

The issue is that the initial InputActionsEditorState is created indirectly via default constructor, setting all values to 0/null, when looking at the rest of the class' constructors we can see that's not what we want for all values.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


graham-huws:
@ekcoh What are your thoughts here? I think the _correct_ course of action would be to make InputActionsEditorState a class instead so that we can override the default constructor, but I'm not sure what other effects that would have (serialisation knock-ons?), so I'm quite wary to make such a change, given the timing!

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
```diff
@@ -123,6 +123,14 @@ public InputActionsEditorState(InputActionsEditorState other, SerializedObject a
             // if the asset would be modified during domain reload.
             serializedObject = asset;
 
+            if (other.Equals(default(InputActionsEditorState)))
```
ekcoh, line 126: I am not sure about this fix, it might solve this particular problem but.... wouldn't the correct solution to actually introduce an explicit default constructor initialising the struct with valid appropriate initial values for all members so that default(InputActionsEditorState) includes the ones below but also appropriate defaults for all the other fields?

ekcoh (changes requested):
Suggested to introduce a proper default constructor and not an implicit one providing faulty defaults?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
graham-huws-unity: Right, but the struct is created by default constructor as part of the Window being created, and you can't override the default constructor for a struct in C# - hence why I said we'd need to swap it to being a class instead if we want that, and I don't know if that would affect anything else

ekcoh: You are right, I always forget that. Then I approve this.

ekcoh (approved):

Pauliusd01 (approved):
Looks good to me, I checked that the control scheme isn't reset in unwanted cases such as unfocusing the window, editor or modifying actions.

PR #1866, opened 2024-03-01
ISX-1901, ISX-1520: Project-wide actions API docs, capabilities update and improved test coverage

ekcoh:
### Description

This PR is mainly a tidy-up of previous major changes introduced with regards to project-wide actions.

### Changes made

The following key changes have been made:
- ISX-1901: Updated documentation of `InputSystem.actions` related properties added to public API surface. Also improved test coverage for requirements related to these properties.
- ISX-1520: Added a restriction on `InputSystem.actions` so its only allowed to be assigned during edit-time. This makes the feature less flexible but also allows for providing stronger guarantees on what to expect from the associated Project-wide `InputActionsAsset`.
- Fixed an issue where `InputSystem.actions` could be assigned a non-persisted `InputActionAsset` instance which would not be possible to store in EditorPlayerSettings and hence not get included in the build. Such objects are now detected when `InputSystem.actions` is assigned which now throws `ArgumentException`.
- Fixed an issue where InputActionAsset validation (InputForUI UI map) where not triggered for Project-wide input actions when the project-wide asset was edited in a regular windowed Input Action Asset editor window.
- Reverted removed code for `InputSettingsBuildProvider` removed by @jamesmcgill in a56878c4 commit which seems to undo a bug fix previously done by @jimon. I assume this was a mistake. Also removed all project-wide actions code from this class and moved to new single-purpose class `ProjectWideActionsBuildProvider`. Both classes now have all common code in `BuildProviderHelpers` inside Internal source folder.
- Removed code enabling `InputSystem.actions` explicitly if not auto-enabled from sample code. Even though valid to do that, I believe it defeats the key point of having them auto-enabled by the system since this check would always be true since actions are not enabled since after Start(). The contract here is now also explicitly mentioned in InputSystem.actions script API doc.
- Fixed an issue where `InputSystem.actions` instance could be destroyed if the `InputSystem` would reset.
- Fixed an issue where `InputSystem.actions` where incorrectly evaluated using null-propagation operator resulting in incorrect execution flow by not relying on `UnityEngine.Object` equality and inequality overloads. This generated warnings in Rider so likely overlooked if not warned about in VS IDE.
 - Fixed an issue where `InputSystem.actions` would not be initialized during `InputSystem` initialization routines causing incorrect serialized manager state. This was likely mainly an issue in test environment, but prevented auto-testing the same way in editor play-mode and player.
 - Inspector UI when selecting an .inputactions asset that is not the project-wide asset now disables the "Assign button" while in play mode.
[...]

PR #1885, opened 2024-03-08
FIX: (Tentative) Modified constraint for enabling project-wide actions

ekcoh:
### Description

Modified constraint for enabling project-wide actions to allow similar behaviour in player and play-mode.

### Changes made

Modified constraint to enable project-wide actions.

### Notes

Not sure whether we want to include this in 1.8.0 or not but if CI is happy I lean to saying yes. Putting it up for discussion. Fix originally suggested by @AlexTyrer in https://github.com/Unity-Technologies/InputSystem/pull/1866.

The idea is to provide consistent behavior between play-mode and player. Without this fix the following happens:

```
void Awake() { Debug.Log(InputSystem.actions.enabled); } // false in play-mode, true in player
void Start() { Debug.Log(InputSystem.actions.enabled); } // false in play-mode, true in player
void OnEnable() { Debug.Log(InputSystem.actions.enabled); } // false in play-mode, true in player
```

with this fix its consistently true in both.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Editor.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Editor.cs
@@ -2890,27 +2890,10 @@ partial class CoreTests
         Assert.That(action.ReadValue<float>(), Is.EqualTo(0.75f).Within(0.00001f));
     }
 
-    private static void DisableProjectWideActions()
-    {
-#if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
-        // If the system has project-wide input actions they will also trigger enable/disable via
-        // play mode change triggers above. Hence we adjust extra variable to compensate of
-        // state allocated by project-wide actions.
-        if (InputSystem.actions)
-        {
-            Assert.That(InputActionState.s_GlobalState.globalList.length, Is.EqualTo(1));
-            InputSystem.actions.Disable();
-            InputActionState.DestroyAllActionMapStates();
-        }
-#endif
-    }
-
     [Test]
     [Category("Editor")]
     public void Editor_LeavingPlayMode_DestroysAllActionStates()
     {
-        DisableProjectWideActions();
-
         // Initial state
         Assert.That(InputActionState.s_GlobalState.globalList.length, Is.EqualTo(0));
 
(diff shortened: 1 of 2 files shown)


AlexTyrer (approved):
I think this change is worth it - users will expect playmode to behave the same as in the player.

Having project wide input actions not become enabled until after the Start() method is unexpected.

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -2890,27 +2890,10 @@ public void Editor_WhenRunUpdatesInEditModeIsEnabled_InputActionsTriggerInEditMo
...
-            Assert.That(InputActionState.s_GlobalState.globalList.length, Is.EqualTo(1));
-            InputSystem.actions.Disable();
-            InputActionState.DestroyAllActionMapStates();
-        }
-#endif
-    }
-
     [Test]
     [Category("Editor")]
     public void Editor_LeavingPlayMode_DestroysAllActionStates()
```
ekcoh: It seems this change allows actions to be enabled earlier which doesn't convolute test case so removed changes I did in recent PR and was also able to remove initial disable added previously. I do not yet fully understand how this is possible but its definitely much better if other state is preserved. I suspect its pushed earlier to the static stack so action states are pushed prior to the test.

Pauliusd01 (approved):

Pauliusd01 (approved):

ekcoh:
> I think this change is worth it - users will expect playmode to behave the same as in the player.
> 
> Having project wide input actions not become enabled until after the Start() method is unexpected.

Yeah my worry is that a difference there might be devastating since user code could potentially not be consistent. I am less worried about slight changes to our test cases since they use the InputTestFixture.

Assets/Tests/InputSystem/CoreTests_Editor.cs
timkeo: >I suspect its pushed earlier to the static stack so action states are pushed prior to the test.

I also think the Push/Pop of state between tests is the likely culprit.

timkeo (approved):
LGTM

PR #1893, opened 2024-03-20
FIX: Composite binding isn't triggered after ResetDevice() called during Action handler (ISXB-746)

timkeo:
### Description

Whew! This was a tricky one.
My full analysis was added as a comment [to the bug ticket](https://jira.unity3d.com/browse/ISXB-746), but here it is again for convenience.

Ultimately, the problem stems from a focus change that occurs when the display window is changed (as you'd expect), which triggers a `InputManager.ResetDevice()` operation. This is supposed to reset all actions and controls, but in the case of Composite fails to do so.

The Reset operation generates a "fake" `InputEvent` that, although injected, is intended to appear like any normal event. However, when `InputActionState.ProcessControlStateChange()` is called for controls part of a Composite, this check is performed:

```
// If the composite has already been triggered from the very same event, ignore it.
// Example: KeyboardState change that includes both A and W key state changes and we're looking
//          at a WASD composite binding. There's a state change monitor on both the A and the W
//          key and thus the manager will notify us individually of both changes. However, we
//          want to perform the action only once.
if (ShouldIgnoreInputOnCompositeBinding(compositeBindingPtr, eventPtr))
    return;
```
and within `ShouldIgnoreInputOnCompositeBinding()` this logic is executed:
```
var eventId = eventPtr->eventId;
if (eventId != 0 && binding->triggerEventIdForComposite == eventId)
    return true;
binding->triggerEventIdForComposite = eventId;
return false;
```
That is: Has this event already been processed by a Control within this (Composite) Binding? If so ignore it.

The problem in this scenario: the same Event object is being used to reset all the Controls but this function only allows the "first" Control in a Composite to get Reset. The other Controls retain their values. In this scenario the SHIFT modifier is reset but the F1 key is skipped.

But you may ask: "Releasing the keys will reset the state, and so why does the bug still occur?"

This leads to the second part of the problem: Composite `pressTime`.

When processing Action states, this check is performed (in a few different locations):
```
if (bindingStatePtr->pressTime == default || bindingStatePtr->pressTime > trigger.time)
    bindingStatePtr->pressTime = trigger.time;
```
 
this basically updates the `pressTime` value for a given Binding only if the current time is 0 or smaller than the current time. Normally, the `pressTime` is 0 (reset) and updated when the Modifier and Key (of a Composite Binding) are pressed. However, since the F1 key wasn't properly reset, it's still holding the "old" `pressTime`. Not a huge issue on it's own, but this leads to the third part of the problem.

Within `OnModifierComposite.ModifierIsPressed()` this logic is executed:
```
// When the modifiers are gating a button, we require the modifiers to be pressed *first*.
if (modifierDown && m_BindingIsButton && !overrideModifiersNeedToBePressedFirst)
{
    var timestamp = context.GetPressTime(binding);
    var timestamp1 = context.GetPressTime(modifier);
        return timestamp1 <= timestamp;
} 
```
which checks that the Modifier was pressed before the (function) key. However, we didn't properly update the `pressTime` for F1 because it was holding a old value from when we previously executed the Action.

Therefore, even though both keys are pressed, the `pressTime` for the SHIFT modifier is _after_ the F1 key which fails this check preventing the Action from firing.

There's one final element to this bug: the `overrideModifiersNeedToBePressedFirst` flag must be set.

This flag enforces the "modifier must be pressed first" behavior within `OneModifierComposite.ModifierIsPressed()`, and if it's not set the `pressTime` comparison between the modifier and binding is skipped thereby avoiding the bug.

### Changes made

I made a surgical modification to `ProcessControlStateChange()` that removes the early return after `ShouldIgnoreInputOnCompositeBinding()` and instead flags if the Composite has been triggered already. This flag prevents (most) of the remaining processing but still allows `ProcessButtonState()` to be called; the essential function to update/reset state for each Control within the Composite.

Also adds a test to cover the corner-case scenario and also verify the fix.

### Notes

I used other Action tests to verify didn't regress functionality, and in fact had to adjust my fix because it originally was breaking things. Specifically, I looked at the tests covering "Conflicting" bindings, especially those that set both `isConflictingInput` and `compositeAlreadyTriggered` flags.

Suggested QA Testing:

This change specifically impacts "Composite Button" bindings; non-composite bindings and Value bindings shouldn't be affected. So, here's some additional testing I recommend for validating Actions with the following types of Composite bindings:

- Composite binding with "several" different button controls and Modifiers
- Composite binding with single button control
- Composite binding mixing Button and Value controls
- Composite binding with Button controls and Processors
- Multiple Composite bindings that use the same Button controls with different Modifiers and Processors

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Pauliusd01:
For this one I checked the composite binding behaviour on dual monitors from the user project and tried other ways to trigger the "reset" such as losing focus of the built app or using different bindings. @ekcoh Is there anything else in your opinion that I should check? It's a scary sounding change and I'd like to feel confident that It's safe

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -12029,4 +12029,99 @@ public void Actions_ActionMapDisabledDuringOnAfterSerialization()
...
         Assert.That(map.FindAction("MyAction", true).enabled, Is.True);
     }
+
+    // ResetDevice wasn't properly clearly Composite key state, i.e. BindingState.pressTime
+    // https://jira.unity3d.com/browse/ISXB-746
+    [Test]
+    [TestCase(false)]
+    [TestCase(true)]
+    [Category("Actions")]
+    public void Actions_CompositeBindingResetWhenResetDeviceCalledWhileExecutingAction(bool useTwoModifierComposite)
```
ekcoh, line 12039: Thank you for adding these tests. Important to get this covered.

ekcoh (approved):
Changes look good to me and excellent you improved test coverage.
I think we should consider landing this in next package release and not in the coming targeted patch to allow for additional direct and/or indirect test exposure. Lets align on this aspect.

Pauliusd01:
Is this good to go? I wasn't really given any new things to test so if you're satisfied with what was already checked I can approve

timkeo:
> Is this good to go? I wasn't really given any new things to test so if you're satisfied with what was already checked I can approve

Sorry, I forgot to post update on this fix.

Due to the inherent risk of this change, we decided it should _not_ land for the 1.8.2 release, since it contains critical bug fixes that cannot be jeopardized with a regression. Once that release goes out, this fix can be evaluated for 1.8.3 release.

jfreire-unity:
Thank you for such a detailed PR! It was really good to understand the problem and the fix. And great to see tests added to cover this edge-case. 👏 💪

Pauliusd01:
Just a status update, I am currently testing this. Should be done by the end of ~tomorrow~ thursday (forgot tomorrow is a public holiday)

Pauliusd01 (approved):
I used the boat attack demo project for testing, editing its controls to add new composites and processors to them, did not encounter any issues. Listing the things mentioned in the notes and my results on them:

- Composite binding with "several" different button controls and Modifiers (Boat attack project already had this and it worked just as before)
- Composite binding with single button control (Worked fine)
- Composite binding mixing Button and Value controls (Pasted value composite into an action that already had button composites, both worked fine)
- Composite binding with Button controls and Processors (Tried a few processors that made sense with the given Value/Button controls like invert, scale, clamp. Seemed to work the same as in develop)
- Multiple Composite bindings that use the same Button controls with different Modifiers and Processors (This one I was not able to check properly, mostly because I'm not sure if I'm testing it right, I duplicated some composite bindings and added different processors to the duplicated one. For example: invert, I expected my controls to do nothing such as if you move right you get right movement + inverted right which is left movement = should equal no movement? But it just moved the same as before. The duplicate bindings just seem to be ignored but It's the same in Develop, whether the test makes sense or not)

timkeo:
The tvOS failure is an infrastructure issue that's impacting CI and is unrelated to this PR; the builder had been green before previous update).

PR #1896, opened 2024-03-26
FIX: Avoid cases on Android where keys can get stuck set as pressed (ISXB-475).

graham-huws:
### Description

Original bug reproduction - open linked project (in [ISXB-475](https://jira.unity3d.com/browse/ISXB-475)) and build for Android, connecting a physical keyboard (either bluetooth or USB) to the Android device. Hold the W key for about a second then let go. After several attempts the app will say that the W key is pressed, even though it isn't. For me (Pixel 7 Pro) this usually reproduces by the time the counter gets to about ~3000, but definitely usually happens within a few minutes.

### Changes made

See comment added to the code. I don't know the area well enough to expect any negative effects, but the change certainly seems to fix the issue!

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3343,6 +3343,14 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
 #if UNITY_EDITOR
                                 m_Diagnostics?.OnEventTimestampOutdated(new InputEventPtr(currentEventReadPtr), device);
+#elif UNITY_ANDROID
+                                // Android keyboards can send events out of order: Holding down a key will send multiple
+                                // presses after a short time, like on most platforms. Unfortunately, on Android, the
+                                // last of these "presses" can be timestamped to be after the event of the key release.
+                                // If that happens, we'd skip the keyUp here, and the device state will have the key
+                                // "stuck" pressed. So, special case here to not skip keyboard events on Android. ISXB-475
+                                // N.B. Android seems to have similar issues with touch input (OnStateEvent, Touchscreen.cs)
+                                if (!(device is Keyboard))
```
ekcoh, line 3353: Before accepting this I would like to understand some aspects of it. How can the timestamp / out-of-order behaviour happen? Is the source/producer multi-threaded or how do we end up with the press coming after the release event? Did I understand you correctly that order is consistent, but timestamps are not?

ekcoh: I would also love to understand why this would be Android specific unless something is broken in platform implementation?

Pauliusd01:
I tried this and the testing phone I have at home does not pick up my keyboard (either via usb dock or bluetooth) Maybe @stefanunity could help QA this

stefanunity:
I can't repro the issue on my Pixel 8a neither on the old unifixed Input System nor on the PR version using my MS Natural Keyboard :(

Pauliusd01:
Added area QA - @IGuscin

IGuscin (commented):
I wasn't able to reproduce initial issue, so would prefer to ask a person from CQA to verify the fix.
I've tested with couple devices:
Galaxy s10+ Android 10
Galaxy s23+ Android 13
Galaxy zFlip2 Android 11

Pauliusd01:
Requested help from the filing CQA

Pauliusd01:
From CQA:
"I tested this on 2023.2.17f1 with 1.7.0f1 input system - and it repro'd. Then I added the package from this PR and tried to repro again until the timer hit 6000 which it didn't repro even once.
Checked with this device: VLNQA00267, Samsung Galaxy S10+ (SM-G975F), Android 12, CPU: Exynos 9 (9820), GPU: Mali-G76"

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
graham-huws-unity: > How can the timestamp / out-of-order behaviour happen? Is the source/producer multi-threaded or how do we end up with the press coming after the release event? Did I understand you correctly that order is consistent, but timestamps are not?

How exactly it happens, I don't know. But tracing the inputs back to their source as raw Android events, they always arrive in the correct order, just sometimes with incorrectly ordered timestamps for the keyup and final repeated key press. As for why it would be Android specific - I don't know, but it's not been reported on any other platform, and there's the existing known case of oddness from how Android deals with touch input event ordering, so there's some history there 😄

ekcoh: Based on the comment it sounds like key repeats are co-evaluated with press/release which sounds incorrect in general. Given the positive CQA feedback I will approve this, but I have a suspicion something is incorrectly evaluate in the package with respect to this.

ekcoh (approved):

PR #1900, opened 2024-04-04
FIX: Update DualSense HID support so OutputReports function properly (ISXB-787)

timkeo:
### Description

The OutputReport size for the DualSense Edge controller is larger than the basic DS controller: 64 bytes vs. 32 bytes. This causes IOCTL commands for setting bumble and light bar values to be rejected by the Windows back-end, i.e. the command buffer size didn't match the feature report.

To fix this problem we need to dispatch output commands with the same buffer size reported by the HID descriptor.

### Changes made

I modified all the DualShockHID output commands to take an `outputReportSize` parameter, which is used instead of the constant `kSize` when creating the output buffer. This value should, of course come from the controller's HID descriptor. Note that the const `kSize` is still used (and unchanged) for the field serialization, but the parameter is used for initializing the command buffer.

I added an (internal) property to `DualShockGamepad` returning said HID descriptor parsed from InputDevice's Description, if the DS controller is connected via HID.

### Notes

I verified the fix using the (wired) DualSense Edge and PS4 DualShock controllers.
I tested these controllers over Bluetooth, but that still doesn't work; the WriteFile API fails with INVALID_ARGUMENT error, which I guess is expected.

I did _not_ test with a standard DualSense controller (I don't currently have one), and I also didn't test with the PS5 package.

I noticed the SwitchPro HID controller has a similar implementation: output commands use a hard-coded size instead of the value from HID descriptor. It could potentially have a similar issue if Nintendo released a new rev. of the hardware, but it's not worth the risk right now and so I left it alone.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -15,6 +15,7 @@ however, it has to be formatted properly to pass verification tests.
 - Physical keyboards used on Android/ChromeOS could have keys "stuck" reporting as pressed after a long press and release [ISXB-475](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-475).
 - NullReferenceException thrown when right-clicking an empty Action Map list in Input Actions Editor windows [ISXB-833](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-833).
 - Fixed an issue where `System.ObjectDisposedException` would be thrown when deleting the last ActionMap item in the Input Actions Asset editor.
+- Fixed DualSense Edge's vibration and light bar not working on Windows
 
 ### Changed
 - For Unity 6.0 and above, when an `EventSystem` GameObject is created in the Editor it will have the
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepad.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepad.cs
@@ -86,7 +86,13 @@ namespace UnityEngine.InputSystem.DualShock
         /// <summary>
         /// The last used/added DualShock controller.
         /// </summary>
+        /// <value>Equivalent to <see cref="Gamepad.leftTrigger"/>.</value>
         public new static DualShockGamepad current { get; private set; }
+
+        /// <summary>
+        /// If the controller is connected over HID, returns <see cref="HID.HID.HIDDeviceDescriptor"/> data parsed from <see cref="InputDeviceDescription.capabilities"/>.
+        /// </summary>
+        internal HID.HID.HIDDeviceDescriptor hidDescriptor { get; private set; }
 
         /// <inheritdoc />
         public override void MakeCurrent()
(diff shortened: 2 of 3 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - NullReferenceException thrown when right-clicking an empty Action Map list in Input Actions Editor windows [ISXB-833](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-833).
+- Fixed DualSense Edge's vibration and light bar not working on Windows
```
timkeo, line 18: The ISXB ticket isn't public.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepad.cs
```diff
@@ -118,6 +124,9 @@ protected override void FinishSetup()
             R2 = rightTrigger;
             L3 = leftStickButton;
             R3 = rightStickButton;
+
+            if (m_Description.interfaceName == "HID")
+                hidDescriptor = HID.HID.HIDDeviceDescriptor.FromJson(m_Description.capabilities);
```
timkeo, line 129: I decided to put this into the "base" DualShock class instead of copy-pasting it amongst the various DualShock/DualSense HID classes. I think ideally we'd have a DualShockHID common class that all the variants inherit from.

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockGamepad.cs
```diff
@@ -118,6 +124,9 @@ protected override void FinishSetup()
             R2 = rightTrigger;
             L3 = leftStickButton;
             R3 = rightStickButton;
+
+            if (m_Description.capabilities != null && m_Description.interfaceName == "HID")
+                hidDescriptor = HID.HID.HIDDeviceDescriptor.FromJson(m_Description.capabilities);
```
timkeo, line 129: I decided to put this into the "base" DualShock class instead of copy-pasting it amongst the various DualShock/DualSense HID classes. I think ideally we'd have a DualShockHID common class that all the variants inherit from.

Pauliusd01:
I do not have the special controller, adding @stefanunity instead

stefanunity:
-Tested normal DS, DS Edge and DS4 all wired and they all rumble
-don't rumble via BT as mentioned by Tim
-there were errors with the PS5 package in the repro project about accessing  (internal) properties, but adding latest from their repro made the errors go away

stefanunity (approved):

ekcoh (approved):

PR #1901, opened 2024-04-04
FIX: ObjectDisposedException regression when deleting actions maps (ISXB-831)

timkeo:
### Description

Fixes regression introduced with this bug fix: https://github.com/Unity-Technologies/InputSystem/pull/1861

The original fix added `propertyPath` to the SerializedInputAction/Binding Equals method, but it retrieved the value directly from the cached SerializedProperty. In this scenario (deleting last ActionMap) the underlying Property is disposed and causes this exception. Instead, this change saves `propertyPath` to a separate string during initialization when the SerializedProperty is known to be valid.

### Changes made

Added a `propertyPath` property to SerializedInputAction/Binding which holds the value retrieved from `wrappedProperty`.
This string is used during Equals() method.

### Notes

I verified this change doesn't break https://jira.unity3d.com/browse/ISX-1873 but there's still a small concern about caching `propertyPath`. If the it changes in the underlying Property it won't be updated here. But I don't think this is a real problem given how this struct is used.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -13,6 +13,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 - Physical keyboards used on Android/ChromeOS could have keys "stuck" reporting as pressed after a long press and release [ISXB-475](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-475).
 - NullReferenceException thrown when right-clicking an empty Action Map list in Input Actions Editor windows [ISXB-833](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-833).
+- Fixed an issue where `System.ObjectDisposedException` would be thrown when deleting the last ActionMap item in the Input Actions Asset editor.
 
 ## [1.8.1] - 2024-03-14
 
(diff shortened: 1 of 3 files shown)


Pauliusd01 (approved):
LGTM, tried deleting, undoing, redoing action maps, actions and bindings. Did not get the error anymore or any other issues

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - NullReferenceException thrown when editing a binding path in InputActionAsset windows.
+- Fixed ObjectDisposedException when deleting last ActionMap.
```
ekcoh: Minor: Add code formatting to ObjectDisposedException, e.g. `ObjectDisposedException`, would also suggest to write something a bit more elaborate e.g. "Fixed an issue where `System.ObjectDisposedException` would be thrown when deleting the last action map item in the input actions editor." or equivalent.

ekcoh (approved):
LGTM

PR #1902, opened 2024-04-08
CHANGE: Override UGUI EventSystem with InputSystemUIInputModule

jfreire-unity:
### Description

Unity 6 will come with the "new" Input System enabled by default. The `EventSystem` of UGUI uses `StandaloneInputModule` regardless of the enabled input backend.
This PR overrides the `EventSystem` game object default InputModule using a new class,  `InputModuleComponentFactory` , that will be available on Unity 6.

### Changes made

- Event System Input Module is overridden with `InputSystemUIInputModule` when the Input System native backend is enabled. Only with "Legacy Input"  in "Project Settings > Player > Active Input Handling", no override is done.
- Adds an Editor test to validate the behavior we want for `EventSystem`. Once the Input System package is enabled, we want the `EventSystem` to be created with `InputSystemUIInputModule`.

### Notes

PR is in a good state for reviewing the changes it does. After that, it still needs: 
- Check if there is documentation that needs to be updated.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
@@ -2,14 +2,25 @@
 using System;
 using System.Linq;
 using UnityEditor;
+using UnityEditor.EventSystems;
 
 ////TODO: add button to automatically set up gamepad mouse cursor support
 
 namespace UnityEngine.InputSystem.UI.Editor
 {
     [CustomEditor(typeof(InputSystemUIInputModule))]
+    [InitializeOnLoad]
     internal class InputSystemUIInputModuleEditor : UnityEditor.Editor
     {
+        SerializedProperty activeInputBackend;
+        static InputSystemUIInputModuleEditor()
+        {
+#if UNITY_6000_0_OR_NEWER
+            //TODO: only add this if backend selected is ONLY InputSystem
+            InputModuleComponentFactory.SetInputModuleComponentOverride(go => go.AddComponent<InputSystemUIInputModule>());
+#endif
+        }
+
         private static InputActionReference GetActionReferenceFromAssets(InputActionReference[] actions, params string[] actionNames)
         {
             foreach (var actionName in actionNames)
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
@@ -15,10 +15,9 @@ namespace UnityEngine.InputSystem.UI.Editor
     {
         static InputSystemUIInputModuleEditor()
         {
-#if UNITY_6000_0_OR_NEWER
-            if (EditorPlayerSettingHelpers.newSystemBackendsEnabled && !EditorPlayerSettingHelpers.oldSystemBackendsEnabled)
-                InputModuleComponentFactory.SetInputModuleComponentOverride(
-                    go => ObjectFactory.AddComponent<InputSystemUIInputModule>(go));
+#if UNITY_6000_0_OR_NEWER && ENABLE_INPUT_SYSTEM
+            InputModuleComponentFactory.SetInputModuleComponentOverride(
+                go => ObjectFactory.AddComponent<InputSystemUIInputModule>(go));
 #endif
         }
 
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
@@ -3,6 +3,7 @@ using System;
 using System.Linq;
 using UnityEditor;
 using UnityEditor.EventSystems;
+using UnityEngine.InputSystem.Editor;
 
 ////TODO: add button to automatically set up gamepad mouse cursor support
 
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
```diff
@@ -2,14 +2,25 @@
...
 namespace UnityEngine.InputSystem.UI.Editor
 {
     [CustomEditor(typeof(InputSystemUIInputModule))]
+    [InitializeOnLoad]
     internal class InputSystemUIInputModuleEditor : UnityEditor.Editor
     {
+        static InputSystemUIInputModuleEditor()
+        {
+#if UNITY_6000_0_OR_NEWER
+            if (EditorPlayerSettingHelpers.newSystemBackendsEnabled && !EditorPlayerSettingHelpers.oldSystemBackendsEnabled)
```
benoitalain: OK, yes, makes for sense this way!

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
```diff
@@ -2,14 +2,25 @@
...
 
 namespace UnityEngine.InputSystem.UI.Editor
 {
     [CustomEditor(typeof(InputSystemUIInputModule))]
+    [InitializeOnLoad]
     internal class InputSystemUIInputModuleEditor : UnityEditor.Editor
     {
+        static InputSystemUIInputModuleEditor()
+        {
+#if UNITY_6000_0_OR_NEWER
```
benoitalain: Indeed! Any reason to worry about 6000.0.0b15 and below not working with this version of the InputSystem package?

benoitalain (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
```diff
@@ -2,14 +2,25 @@
...
+            if (EditorPlayerSettingHelpers.newSystemBackendsEnabled && !EditorPlayerSettingHelpers.oldSystemBackendsEnabled)
+                InputModuleComponentFactory.SetInputModuleComponentOverride(go => go.AddComponent<InputSystemUIInputModule>());
```
benoitalain: I did some more testing on uGUI. This will have to be replaced by
```
ObjectFactory.AddComponent<InputSystemUIInputModule>(go)
```
, otherwise we have a weird bug with Undo. See gif below:

Unity_9y36X5MqV9

benoitalain (changes requested):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
jfreire-unity: Good catch ⚡ Replaced it

benoitalain (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
jfreire-unity: Changed this to check for `ENABLE_INPUT_SYSTEM` instead.

jfreire-unity:
⚠️ Note that this PR requires changes on `trunk` to land first. That's why trunk tests are currently failing.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
jfreire-unity: Good point. So far I believe this kind of pattern is what we've been doing, which means we "don't care" about patch versions that don't have these changes. I believe we always assume the latest patch version of an Editor version because that's where the latest Input System package gets pulled cc @ekcoh .
I don't think for example: for a 6b1 version we would pull the latest 1.8.2 to be released for example. Or am I wrong? cc @stefanunity @Pauliusd01

ekcoh (approved):
Thanks for addressing this issue. Changes looks good to me.

benoitalain:
Tested with https://github.cds.internal.unity3d.com/unity/unity/pull/47422 changes to trunk and this works. Now we only need to wait for that PR to land.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModuleEditor.cs
```diff
@@ -2,14 +2,25 @@
...
 using UnityEditor;
+using UnityEditor.EventSystems;
+using UnityEngine.InputSystem.Editor;
 
 ////TODO: add button to automatically set up gamepad mouse cursor support
 
 namespace UnityEngine.InputSystem.UI.Editor
 {
     [CustomEditor(typeof(InputSystemUIInputModule))]
+    [InitializeOnLoad]
```
timkeo, line 13: Is this functionality affected by Fast Enter Play Mode, i.e. will this still work if Domain Reloads are disabled?

timkeo (approved):
LGTM

PR #1905, opened 2024-04-17
CHANGED: Expand Optimized Controls performance testing

jfreire-unity:
### Description

As part of investigations to fix [ISXB-533](https://issuetracker.unity3d.com/issues/input-system-enabling-both-optimization-options-of-input-system-reduces-performance-in-built-players) I decided to expand our performance testing around optimized controls. The goals of the PR are:
- Have a systematic way to showcase that there are issues with our feature flags for optimizing CPU performance.
- Have a repetitive way of measuring the impact of changes in this area.
- Have results necessary to make an informed decision around the future of these features.

The PR and the review does not have the goal to solve the issues reported in the bug ticket yet. That will come in a follow-up PR.

### Changes made

Using feature flags such as `USE_OPTIMIZED_CONTROLS` and `USE_READ_VALUE_CACHING` has the intention to improve performance. However, we have reports by users that that's not the case.

Previously, our performance testing around this area only showed isolated improvements without calling `InputSystem.Update()`. I added more tests that show the cases where performance is improved and degraded when using these flags. Specifically:
1. Calling `InputControl.ReadValue()` and `InputSystem.Update()` for Gamepad. Every 100 frames the read value is changed.
2. Calling `InputControl.ReadValue()` and `InputSystem.Update()` for Mouse. Every 100 frames the read value is changed.
3. Calling **only** `InputSystem.Update()`.
4. Calling `InputControl.ReadValue()`, with new values every frame, and `InputSystem.Update()` for a Gamepad.

Currently, in all these tests `USE_OPTIMIZED_CONTROLS` always performs worse. That's [due to an extra check being](https://github.com/Unity-Technologies/InputSystem/blob/3f4d780cf8e850be00f6852a86cf75405f6e7908/Packages/com.unity.inputsystem/InputSystem/InputManager.cs#L2974-L2976) done when a build is created with `DEVELOPMENT_BUILD`. On release builds we do not have this performance cost, but it does show up when the profiler is run. 

However, `USE_READ_VALUE_CACHING`, has performance improvements and also degradation. In essence, when control values change every frame, it will almost likely always result in a **performance cost.**

### Notes

Feel free to comment on the way testing is done. And if there's more tests that you feel like we could have, please do give feedback.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
[...]

PR #1906, opened 2024-04-18
FIX: `USE_OPTIMIZED_CONTROLS` feature flag reduces performance in built players

jfreire-unity:
### Description

This PR tries to make a partial fix for https://issuetracker.unity3d.com/issues/input-system-enabling-both-optimization-options-of-input-system-reduces-performance-in-built-players, related to `USE_OPTIMIZED_CONTROLS`. 
Regarding `USE_READ_VALUE_CACHING` there are no fixes to be done with this PR.

#### Optimized control read value changes

 `InputManager.OnUpdate()` was doing an extra recursive check to evaluate if the format of the control has changed to a format that cannot be optimized. This check had the goal to help users be aware of a problem and it would also fix it for them. However, even though this check was only be done in `DEVELOPMENT_BUILD`, the gains could not be noticed when using a profiler which requires `DEVELOPMENT_BUILD` to be set.
 
This leads me to conclude that a `DEVELOPMENT_BUILD` variant is not the right place for it even though we had good intention with it.
I made the check run on `EDITOR` code only (`PlayMode` will see a performance cost anyway) and updated the documentation so that users are more aware of which cases this won't work in case they start having problems. Also, I mentioned the performance cost on `PlayMode` due to these checks.

The alternative to fix this for good would be to have a **breaking change** so that `AxisControl.invert` and other parameters could call `InputControl.ApplyParameterChanges()` to tell the control if they could be optimized or not.
Since we don't want to introduce a breaking change, we could also introduce somekind of `INPUT_SYSTEM_DEBUG` that would run for `DEVELOPMENT_BUILD` but not when the profiler is used. But this would be more time consuming and the gains would be similar in my view.

#### Control value caching

There's really no changes to this code path. I added more performance tests to show examples where there are positive and negative performance impacts. Although these tests are very specific, it does prove this performance path doesn't provide performance improvements in **every use case**. The reason why this happens is that looking for "stale" controls ([doing a binary search whenever we are writing a state change](https://github.com/Unity-Technologies/InputSystem/blob/93e5db963c4499dbeaf58679bb5da918bcdf3347/Packages/com.unity.inputsystem/InputSystem/InputManager.cs#L3690-L3701)) has some overhead that might not serve any purpose in some cases.

I believe that to have better performance from this path, we would need to replace state monitors and use the logic we have to look for "stale" controls instead. 

The documentation was updated to attempt to improve communication to users with regards to expectations of this feature flag.

### Notes

These changes can be seen in the performance tests introduced in #1905 .
 Player builds now show `USE_OPTIMIZED_CONTROLS` being more performant than in `PlayMode` where the performance cost is still visible (and was still present in development player builds as well, before this PR).

PlayMode:
image

Player build (MacOSXStandalone):
image

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


stefanunity (approved):
Looks like a very targeted step fix.
Really appreciate the changes to the docs being more explicit what these flags do or don't do.

Packages/com.unity.inputsystem/Documentation~/Controls.md
```diff
@@ -243,7 +243,7 @@ Use [`InputControl<T>.value`](../api/UnityEngine.InputSystem.InputControl-1.html
...
+When the `'USE_READ_VALUE_CACHING'` internal feature flag is set, the Input System will switch to an optimized path for reading control values. This path efficiently marks controls as 'stale' when they have been actuated and subsequent calls to [`InputControl<T>.ReadValue`](../api/UnityEngine.InputSystem.InputControl-1.html#UnityEngine_InputSystem_InputControl_1_ReadValue) will only apply control processing when absolutely necessary. Control processing in this case can mean any hard-coded processing that might exist on the control, such as with [`AxisControl`](../api/UnityEngine.InputSystem.Controls.AxisControl.html) which has built-in inversion, normalisation, scaling etc, or any processors that have been applied to the controls' [processor stack](Processors.md#processors-on-controls). This can have a significant positive impact on performance, especially when using complex composite input actions with many composite parts, such as a movement input action that could be bound to W, A, S, and D on the keyboard, two gamepad sticks and a DPad.
```
ekcoh: > This can have a significant positive impact on performance, especially when using complex composite input actions with many composite parts, such as a movement input action that could be bound to W, A, S, and D on the keyboard, two gamepad sticks and a DPad.

Regarding this sentence, maybe mentioned why? That the performance gain is coming from caching a somewhat "expensive" computation instead of doing it on the fly? And that this isn't recommended fin values are only consumed once?

Do you have any insight into the actual cost here? Is it coming from memory allocation or is it memcpy or where is the cost allocated?

Packages/com.unity.inputsystem/Documentation~/Controls.md
```diff
@@ -256,10 +256,21 @@ If there are any non-obvious inconsistencies, 'PARANOID_READ_VALUE_CACHING_CHECK
...
+> __Please note__: This optimization has a performance impact on `PlayMode` as we do extra checks to ensure that the controls have the correct memory representation during development. Don't be alarmed if you see a performance drop in `PlayMode` when using this optimization as it's expected at this stage.
```
ekcoh, line 270: Maybe suggest to always check actual performance with a profiler in release mode build where such checks are not active?

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
```diff
@@ -952,8 +952,9 @@ internal void SetOptimizedControlDataTypeRecursively()
         // This is mainly to AxisControl fields being public and capable of changing at any time even if we were not anticipated such a usage pattern.
         // Also it's not clear if InputControl.stateBlock.format can potentially change at any time, likely not.
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        // Only do this check in development builds and editor in hope that it will be sufficient to catch any misuse during development.
-        [Conditional("DEVELOPMENT_BUILD"), Conditional("UNITY_EDITOR")]
+        // Only do this check in and editor in hope that it will be sufficient to catch any misuse during development.
+        // It is not done in debug builds because it has a performance cost and it will show up when profiled.
```
ekcoh, line 956: I am slightly confused here, why would it be bad to have this in a debug build? Debug mode code shouldn't be profiled unless some kind of rough ballpark number is of interest? I think checking in editor-only might be fine, but I am confused about caring about debug build and profiling numbers. I mean many non-UNity tools shows you many red warnings when trying to profile or benchmark something in a debug build.

Pauliusd01:
I tried running the original repro project and I don't notice a difference in the mentioned "PreUpdate.NewInputUpdate" times higher than 0.01ms there - in other words I couldn't even repro the original bug. Tried to run the test from your notes both in this PR and copied to Develop - again no difference. So either I'm doing something wrong or it could be platform specific since I used Win 10 for both projects. I think I'll refrain from reviewing pointlessly

Packages/com.unity.inputsystem/Documentation~/Controls.md
jfreire-unity: I didn't change this area. It was a formatting change AFAIK.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
jfreire-unity: You can only use the Unity Profiler with development builds. See https://docs.unity3d.com/Manual/profiler-profiling-applications.html.

jfreire-unity:
> I tried running the original repro project and I don't notice a difference in the mentioned "PreUpdate.NewInputUpdate" times higher than 0.01ms there - in other words I couldn't even repro the original bug. Tried to run the test from your notes both in this PR and copied to Develop - again no difference. So either I'm doing something wrong or it could be platform specific since I used Win 10 for both projects. I think I'll refrain from reviewing pointlessly

Thanks. I'll try to understand your issues in Slack 👍

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
jfreire-unity: It seems the only way to do release builds profiling on Unity is by using native profilers. I didn't not look into those and the reported issue was found using the Unity profiler as well. The issue likely didn't exist in release builds but users could not measure them with the Unity profiler.

The degradation of performance with development builds (which are the only ones users can profile) due to using an optimization flag is also not ideal . That's why I think the best is to remove that code variant at this moment. This way at least when users measure the performance of the code using optimizations, they will see some performance improvement in player builds (or at least no performance degradation in the worst case).

Packages/com.unity.inputsystem/Documentation~/Controls.md
jfreire-unity: Resolving this comment as there is no way to profile release builds in Unity at the moment.

ekcoh: OK, if only considering Unity Profiler then that seems true. I was thinking about profiling outside the scope of what is built into Unity as in development activity using any tool.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControl.cs
ekcoh: OK, I would recommend taking a look with e.g. Superluminal to make sure we understand behavior in release build.

ritamerkl (approved):

jfreire-unity:
@ekcoh @stefanunity I added some new commits with regards to caching read values documentation and performance tests. Feel free to comment and approve/request changes based on my last changes. Thanks!

Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -766,6 +766,66 @@ void CallUpdate()
...
+
+        var value = 0f;
+        var value2d = Vector2.zero;
+        Measure.Method(() =>
+        {
+            for (var i = 0; i < 1000; ++i)
+            {
+                // Make sure state changes are different from previous state so that we mark the controls as
+                // stale.
+                InputSystem.QueueStateEvent(gamepad, new GamepadState { leftTrigger = i / 1000f, rightTrigger = i / 1000f});
```
ekcoh: Wouldn't it make sense to not include the queue state event part in the benchmark and instead put this on a background thread? (applies to all the benchmarks). If you do, you can spawn that thread and then let it busy-wait on an atomic until measurement thread signals it and then let them execute. Otherwise this might be dominated by the part that is not subject for testing?

Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -766,6 +766,66 @@ void CallUpdate()
...
+            .Run();
+
+
+        Measure.Method(() =>
+        {
+            for (var i = 0; i < 1000; ++i)
+            {
+                // Make sure state changes are different from previous state so that we mark the controls as
+                // stale.
+                InputSystem.QueueStateEvent(gamepad,
```
ekcoh, line 839: Same comment as previous test.

ekcoh (commented):
In general looks good, but I think we should move the queuing part to another thread to not colour the benchmark with work that might be a majority of the work involved. Otherwise see if there is framework support to pause/resume benchmark and handle that injection outside the profiled statement.

lyndon-unity (approved):
Thanks for the extra documentation additions to add some more clarity.

Moving the checks to editor only (rather than development mode on player) seem reasonable, especially for the opt in optimization.

Assets/Tests/InputSystem/CorePerformanceTests.cs
jfreire-unity: In theory that makes sense and it's maybe something we should do. But I think it should be out of scope of this PR as it will be more complex. 
Also, I'm not sure the impact of `InputSystem.QueueStateEvent` matters that much here. It will be the same regardless of the optimization type. Also, the goal here is to do a relative benchmark rather than a absolute one. Even if there's something "coloring" the test with some extra work, it colors all the tests in the same way.

Pauliusd01 (approved):
Tried testing this again, seems like the performance test results window was the problem for me showing very similar results but if I copy the results directly from the generated XML file I see the improvement vs Develop. Looks good to me

jfreire-unity:
I updated the tests and the Controls documentation with regards to the control value caching optimization path.
I was being wrongly biased by the project-wide actions in my performance measurements.

PR #1908, opened 2024-04-19
NEW: Add tests for default UI input actions (ISX-1915).

graham-huws:
### Description

Extend tests to check all default UI actions work with and without a set project-wide asset - [ISX-1915](https://jira.unity3d.com/browse/ISX-1915).

### Changes made

As above

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/InputForUITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/InputForUITests.cs
@@ -33,6 +33,7 @@ public class InputForUITests : InputTestFixture
     private const string kTestCategory = "InputForUI";
 
     readonly List<Event> m_InputForUIEvents = new List<Event>();
+    private int m_CurrentInputEventToCheck;
     InputSystemProvider m_InputSystemProvider;
 
     private InputActionAsset storedActions;
@@ -41,6 +42,7 @@ public class InputForUITests : InputTestFixture
     public override void Setup()
     {
         base.Setup();
+        m_CurrentInputEventToCheck = 0;
 
         storedActions = InputSystem.actions;
 
@@ -66,6 +68,11 @@ public class InputForUITests : InputTestFixture
 #endif
 
         base.TearDown();
+    }
+
+    internal Event GetNextRecordedUIEvent()
+    {
+        return m_InputForUIEvents[m_CurrentInputEventToCheck++];
     }
 
     private bool InputForUIOnEvent(in Event ev)
@@ -103,20 +110,12 @@ public class InputForUITests : InputTestFixture
         Update();
 
         Assert.AreEqual(1, m_InputForUIEvents.Count);
-        Assert.That(m_InputForUIEvents[0] is Event
+        Assert.That(GetNextRecordedUIEvent() is
         {
             type: Event.Type.PointerEvent,
             asPointerEvent: { type: PointerEvent.Type.ButtonPressed,
                               eventSource: EventSource.Touch }
         });
-    }
-
-    void PressUpdateReleaseUpdate(ButtonControl button)
-    {
-        Press(button);
-        Update();
-        Release(button);
-        Update();
     }
 
     #region UI_Input_Actions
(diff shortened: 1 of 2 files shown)


Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -118,7 +101,7 @@ public void TouchIsPressedAndMouseEventsAreIgnored()
         Move(mouse.position, new Vector2(100f, 0.5f));
         Update();
 
-        Assert.IsTrue(m_InputForUIEvents.Count == 1);
+        Assert.AreEqual(1, m_InputForUIEvents.Count);
```
ekcoh, line 112: I was not part of establishing the convention so do not now why, but Assert.That(m_InputForUIEvents.Count, Is.EqualTo(1)); seems preferred instead. Likely due to not mixing up expected and actual argument order...

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -127,78 +110,389 @@ public void TouchIsPressedAndMouseEventsAreIgnored()
...
+    // When there are no project-wide actions asset, the InputSystemProvider should still work as it currently gets
+    // the actions from DefaultActionsAsset().asset.
+
     [Test]
     [Category(kTestCategory)]
-    // Presses a gamepad left stick left and verifies that a navigation move event is dispatched
-    public void NavigationMoveWorks()
+    [TestCase(true)]
+    [TestCase(false)]
+    public void UI_Input_NavigationActionsWork(bool useProjectWideActionsAsset)
```
ekcoh: What does "work" mean? I would suggest renaming this test to provide more clarity, e.g. "NavigationActionFiresUINavigationEvent_RegardlessIfInputIsGamepadJoystickOrKeyboard"

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -127,78 +110,389 @@ public void TouchIsPressedAndMouseEventsAreIgnored()
...
+        }
+
+
+        var joystick = InputSystem.AddDevice<Joystick>();
+        PressUpdateReleaseUpdate(joystick.stick.left);
+        PressUpdateReleaseUpdate(joystick.stick.up);
+        PressUpdateReleaseUpdate(joystick.stick.right);
+        PressUpdateReleaseUpdate(joystick.stick.down);
+
+        Assert.That(m_InputForUIEvents[12] is
```
ekcoh: All of these specific offsets into event buffer, how likely do you think they are to change and break this test? I notice we do not check in-between events, would it make more sense to assert in-between each actuation above?

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -127,78 +110,389 @@ public void TouchIsPressedAndMouseEventsAreIgnored()
...
+            asPointerEvent: { type: PointerEvent.Type.PointerMoved, eventSource: EventSource.Touch }
+        });
     }
 
     [Test]
     [Category(kTestCategory)]
-    public void SendWheelEvent()
+    [TestCase(true)]
+    [TestCase(false)]
+    public void UI_Input_ClickActionWorks(bool useProjectWideActionsAsset)
```
ekcoh: @graham-huws did you notice any difference in behavior on these clicks? I see left/middle/right is configured differently in the PWA asset

ekcoh (approved):
In general looks good to me, some minor comments. Thanks for improving the test coverage on the UI integration.

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
graham-huws-unity: Oh, interesting - I assumed everyone was using Rider which shows the arguments:
image
Though I guess github lacks that 😄

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
graham-huws-unity: Good points, I'll look into this

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
graham-huws-unity: Not sure what you mean here sorry - They produce the different PointerEvent.Button (left/middle/right), is there something else I should be checking for?

ekcoh: Sorry for vagueness, I was basically referring to the fact that they have different configuration...

Click has initialStateCheck = true while Middle and Right Clicks have initialStateCheck = false, doesn't make sense to me....

PR #1909, opened 2024-04-19
Fix resource leaks in Input Actions Editor window.

AlexTyrer:
### Description

This PR addresses resource leaks in the Input Actions Editor window.

These resource leaks could lead to unexpected exceptions being thrown.

### Changes made

InputActionsEditorSettingsProvider is now handled by a single static instance to work around UIToolkit callback order issues identified.

Fix VisualElement handling on the StateContainer that manages the tracked serialized object representing the Input Actions.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

    - FIX: Fixed Input Actions Editor window resource leak that could result in unexpected exceptions [ISXB-865](https://jira.unity3d.com/browse/ISXB-865)

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -17,6 +17,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where `System.ObjectDisposedException` would be thrown when deleting the last ActionMap item in the Input Actions Asset editor.
 - Fixed DualSense Edge's vibration and light bar not working on Windows
 - Fixed Project-wide Actions asset failing to reload properly after deleting project's Library folder.
+- Fixed Input Actions Editor window resource leak that could result in unexpected exceptions [ISXB-865](https://jira.unity3d.com/browse/ISXB-865)
 
 ### Changed
 - For Unity 6.0 and above, when an `EventSystem` GameObject is created in the Editor it will have the
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -17,7 +17,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where `System.ObjectDisposedException` would be thrown when deleting the last ActionMap item in the Input Actions Asset editor.
 - Fixed DualSense Edge's vibration and light bar not working on Windows
 - Fixed Project-wide Actions asset failing to reload properly after deleting project's Library folder.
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where `System.ObjectDisposedException` would be thrown when deleting the last ActionMap item in the Input Actions Asset editor.
 - Fixed DualSense Edge's vibration and light bar not working on Windows
 - Fixed Project-wide Actions asset failing to reload properly after deleting project's Library folder.
+- Fixed Input Actions Editor window resource leak that could result in unexpected exceptions [ISXB-865](https://jira.unity3d.com/browse/ISXB-865)
```
ekcoh: This should not use internal URL. If internal only provide issue reference number or skip.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/StateContainer.cs
```diff
@@ -51,9 +43,17 @@ public void Dispatch(Command command)
             });
         }
 
-        public void Initialize()
+        public void Initialize(VisualElement rootVisualElement)
         {
+            m_RootVisualElement = rootVisualElement;
```
ekcoh, line 51: Maybe add a comment here explaining why we do this little dance?
E.g. "We need to use a root element for TrackSerializedObjectValue that is destroyed with the view. Using a root element from the settings window would not enable the tracking callback to be destroyed or garbage collected."

ekcoh (changes requested):
Two small comments to address, otherwise LGTM. Will approve once updated if QA is positive.

Packages/com.unity.inputsystem/CHANGELOG.md
AlexTyrer: Of course - I've removed the internal link in 5e872116617b56d20a930efac85404e519874a53

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/StateContainer.cs
AlexTyrer: Good idea - have added the comment in 5e872116617b56d20a930efac85404e519874a53

ekcoh (approved):

ekcoh:
I think this can land after QA assessment is available with approval.

Pauliusd01 (approved):
Since I wasn't able to repro the original issues in Unity 6 I only checked basic ProjectWideActions and custom actions window functionality. Did not notice any new issues from these changes. LGTM

PR #1921, opened 2024-05-07
DRAFT - NEW: Add setting for logging processed input events

jfreire-unity:
### Description

This PR adds a setting to improve debugging of received device events on the Input System.

It helps track down devices that might be sending a lot of events per frame. 
It can also help when looking for issues with the Input System.

The idea came while trying to fix https://unity3d.atlassian.net/servicedesk/customer/portal/2/IN-61472.
The customer needed more information to track down a problem with `InputSettings.maxEventBytesPerUpdate` being exceeded in certain situations. 
Our engine support engineers had to go back and forth to help the customer add logging of events to their Input System package. 

This PR pretends to improve that and fulfill the customer request to have more logging information about which device is causing the problem.

### Changes made

Added a new setting to `InputSettings`.

When enabled, the Console output will be "polluted" with data like this (Player build logging example).
image

### Notes

📣 
This is a very naive solution to avoid making customers have to add this kind of logs manually to `InputManager`. It is not "super elegant", but it works.
I did not find a better location for this: I thought about `InputMetrics` or `InputDiagnostics`, but they all depend on the `Input Debugger` window which might not be ideal as it lacks saving history AFAIK.

With this PR I'm mostly looking for comments on this solution. The goal is to improve observability when we have problems in the Input System and allow for developers to have more information about input devices event processing.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -3380,7 +3380,10 @@ namespace UnityEngine.InputSystem
                                 haveChangedStateOtherThanNoise = UpdateState(device, eventPtr, updateType);
                             }
 
+
                             totalEventBytesProcessed += eventPtr.sizeInBytes;
+                            if (Debug.isDebugBuild)
+                                LogProcessedEventForDevice(device, totalEventBytesProcessed, eventPtr);
 
                             // Update timestamp on device.
                             // NOTE: We do this here and not in UpdateState() so that InputState.Change() will *NOT* change timestamps.
@@ -3491,6 +3494,17 @@ namespace UnityEngine.InputSystem
             ////       same goes for events that someone may queue from a change monitor callback
             InvokeAfterUpdateCallback(updateType);
             m_CurrentUpdate = default;
+        }
+
+        private void LogProcessedEventForDevice(InputDevice device, uint totalEventBytesProcessed, InputEventPtr eventPtr)
+        {
+            if (m_Settings.debugLogProcessedEventForDevice && (m_CurrentUpdate != InputUpdateType.Editor))
+            {
+                Debug.Log("Frame Count: " + Time.frameCount +
+                    ", Device: " + device.name +
+                    ", Bytes Processed: " + totalEventBytesProcessed +
+                    ", Event Size: " + eventPtr.sizeInBytes);
+            }
         }
 
         // Only do this check in editor in hope that it will be sufficient to catch any misuse during development.
--- a/Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
@@ -702,6 +702,27 @@ namespace UnityEngine.InputSystem
             }
         }
 
+
+        /// <summary>
+        /// Enables or disables logging of amount of processed input events for each device. This is done per
+        /// InputSystem.Update() call.
+        /// </summary>
+        ///
+        /// This should only be used for debugging purposes. It is useful to track down devices that are sending
+        /// a lot of input events per frame. It will pollute the console log with a lot of messages.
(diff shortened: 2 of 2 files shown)


stefanunity:
So could the customer have solved their problem with access to this information? They could at least then just not use the offending device?

ritamerkl (commented):
After only a quick glance for me it is unclear: What is the EventSize and what is Bytes Processed? Is the event size included into the processed bytes? I am just thinking about the info the user should take from that and if there is possibly a better naming.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3493,6 +3496,17 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             m_CurrentUpdate = default;
         }
 
+        private void LogProcessedEventForDevice(InputDevice device, uint totalEventBytesProcessed, InputEventPtr eventPtr)
+        {
+            if (m_Settings.debugLogProcessedEventForDevice && (m_CurrentUpdate != InputUpdateType.Editor))
```
ekcoh, line 3501: I would move this if statement into the outer if statement to avoid doing a potentially non-inlined function call for just doing an early return.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
```diff
@@ -702,6 +702,27 @@ public bool shortcutKeysConsumeInput
...
 
+
+        /// <summary>
+        /// Enables or disables logging of amount of processed input events for each device. This is done per
+        /// InputSystem.Update() call.
+        /// </summary>
+        ///
+        /// This should only be used for debugging purposes. It is useful to track down devices that are sending
+        /// a lot of input events per frame. It will pollute the console log with a lot of messages.
+        public bool debugLogProcessedEventForDevice
```
ekcoh, line 713: I do not think such a setting belong in InputSettings since InputSettings are shared among developers on a project. E.g. if you turn on debugging I wouldn't expect to get debugging turned on if I checkout the same repo? I would instead suggest that this is saved under Preferences, e.g. https://docs.unity3d.com/ScriptReference/EditorPrefs.html or what do you think?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: I wonder if this couldn't be replaced with a callback instead and let a Debug.Log implementation of a handler exist that could be added by a corresponding setting. This would allow other subscribers as well, e.g. Debugger window, statistical computations etc... I wonder if this isn't already covered somehow, might need to checking into how recording system or debug window does similar things.

Packages/com.unity.inputsystem/InputSystem/InputSettings.cs
jfreire-unity: Good point, it wouldn't make sense to use InputSettings then.

jfreire-unity:
> So could the customer have solved their problem with access to this information? They could at least then just not use the offending device?

Likely. It would help them understand their issue better since they are using manual updates.

jfreire-unity:
> After only a quick glance for me it is unclear: What is the EventSize and what is Bytes Processed? Is the event size included into the processed bytes? I am just thinking about the info the user should take from that and if there is possibly a better naming.

EventSize is the amount of data of a **single event** to be processed.
Bytes processed is the sum of all the EventSize's processed during `InputManager.onUpdate()`.

So maybe these names would have to be changed.

jfreire-unity:
I'm closing this PR and moving into something that provides more data to the error message instead.

PR #1923, opened 2024-05-10
CHANGE: prevent editing of composite parts

smnwttbr:
### Description

When deleting parts of a composite action, the deletion of the last part raises an exception.

### Changes made

Preventing changes to a composite part via context menu and commands.

### Notes

We decided to do this as it does not make sense from a user perspective to modify a composite action. Composite actions are created and configured with the appropriate parts using menu items, which create create the composite using pre-determined patterns. The user is already prevented from adding parts to the composite by the absence of the '+' menu, so it makes sense to remove the corresponding cut/paste/delete items.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -11,6 +11,9 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased] - yyyy-mm-dd
 - Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746)
 
+### Changed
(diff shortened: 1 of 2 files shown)


Pauliusd01:
Still looking at it but my initial thoughts are that deletion, duplication and the context menu should all be available. For composites you can have different bindings for the same action that you might no longer need, for example WASD and arrow keys under one action, just like in our own ProjectWideActions asset. With these changes you can no longer delete or create them. How were the extra bindings created? Duplication I assume. Why isn't there a direct way to just create one? I don't know but in my opinion there should be one. Which kinda questions the whole need for these changes

And the context menu should instead grey out options instead of being blocked from showing up entirely if we decide that this change is needed. A greyed out context menu at least signals that the action being blocked is intentional while a right-click that does nothing may appear as a bug

PR #1925, opened 2024-05-13
FIX: fixed logic when checking for next composite part in the TreeView

smnwttbr:
### Description

Deleting the last part of a composite action raised an exception in the editor. This PR fixes this issue.
[ISXB-804](https://jira.unity3d.com/browse/ISXB-804)

### Changes made

The treeview logic formerly looked for the next item to process without checking if that item exists. Changes made to address this.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -608,33 +608,41 @@ public static List<TreeViewItemData<ActionOrBindingData>> GetActionsAsTreeViewDa
                 {
                     var serializedInputBinding = actionBindings[i];
                     var inputBindingId = new Guid(serializedInputBinding.id);
-
+
```
ekcoh: It seems these changes prevent any exception from being thrown which is an improvement so it could land in that sense but I think we should instead prevent the last instance of each composite part being supported by the composite class from being deleted. If we do not prevent this we have a non-reversible behavior in the editor where deleting a composite part cannot be re-added. We allow duplicating the parts which is fine I guess for adding multiple bindings per part but I believe we should avoid deleting the last instance of each part.

Pauliusd01:
Not reproing the issue anymore so would be willing to approve but I see there's discussion on the implementation so will hold off. Poke me if It's fine to pass and I'll approve

Pauliusd01 (changes requested):
Just updating status that I'm waiting for a resolution on whether the approach here is correct. I can approve it is fine

simonwittber:
@ekcoh is this PR ok? IIRC we discussed accepting this PR, and scheduling other work (to add a + button to a composite) for a future time.

ekcoh:
@simonwittber Sorry for the heavily delayed feedback. Let's land it as a fix for the bug.

ekcoh (approved):

Pauliusd01 (approved):

PR #1926, opened 2024-05-13
[ISXB-505] Fixed disabling composite binding while pressed

bmalrat:
### Description

Fixed Composite binding isn't triggered after disabling action while there are in progress [ISXB-505](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-505).

Within the settings shortcutKeysConsumeInput at true,
The bug was due to the fact that when using modifier it require that the modifier is pressed before. But if it's disabled while the key is pressed the binding state will keep the old time and and when reactivated it will discard any modifier since it will arrive after the old key press time which was not reset.

### Changes made

When disabling an action, it now reset the pressTime of the binding state to ensure that if re-enabled it will not prevent modifier key to trigger.

### Notes

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -1445,6 +1445,40 @@ partial class CoreTests
 
     [Test]
     [Category("Actions")]
+    public void Actions_CanDisableAndEnable_FromCallbackWhileOtherCompositeBindingIsProgress()
+    {
+        // Enables "Modifier must be pressed first" behavior on all Composite Bindings
+        InputSystem.settings.shortcutKeysConsumeInput = true;
+
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        var map = new InputActionMap("map");
+
+        var withModiferReceivedCalls = 0;
+        var actionWithModifier = map.AddAction("OneModifier", type: InputActionType.Button);
+        actionWithModifier.AddCompositeBinding("OneModifier")
+            .With("Binding", "<Keyboard>/space")
+            .With("Modifier", "<Keyboard>/ctrl");
+        actionWithModifier.performed += _ => ++withModiferReceivedCalls;
+
+        var actionWithoutModifier = map.AddAction("One", type: InputActionType.Button, binding: "<Keyboard>/space");
+        actionWithoutModifier.performed += _ => actionWithModifier.Disable();
+
+        map.Enable();
+       
+        PressAndRelease(keyboard.spaceKey);
+        InputSystem.Update();
+        Assume.That(actionWithModifier.enabled, Is.False);
+
+        // Re-enable action which has been disabled by space
+        actionWithModifier.Enable();
+        Press(keyboard.leftCtrlKey, queueEventOnly: true);
+        Press(keyboard.spaceKey);
+        InputSystem.Update();
+        Assert.That(withModiferReceivedCalls, Is.EqualTo(1));
+    }
+
+    [Test]
+    [Category("Actions")]
     public void Actions_WhenEnabled_TriggerNotification()
     {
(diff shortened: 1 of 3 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1926) <br/>All committers have signed the CLA.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1443,6 +1443,40 @@ public void Actions_CanDisableAndEnableOtherAction_FromCallback()
         Assert.That(receivedCalls, Is.EqualTo(2));
     }
 
+    [Test]
+    [Category("Actions")]
+    public void Actions_CanDisableAndEnable_FromCallbackWhileOtherCompositeBindingIsProgress()
```
ekcoh, line 1448: Awesome you covered the particular scenario with a test case!

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -1443,6 +1443,40 @@ public void Actions_CanDisableAndEnableOtherAction_FromCallback()
...
+            .With("Binding", "<Keyboard>/space")
+            .With("Modifier", "<Keyboard>/ctrl");
+        actionWithModifier.performed += _ => ++ withModiferReceivedCalls;
+
+        var actionWithoutModifier = map.AddAction("One", type: InputActionType.Button, binding: "<Keyboard>/space");
+        actionWithoutModifier.performed += _ => actionWithModifier.Disable();
+
+        map.Enable();
+
+        PressAndRelease(keyboard.spaceKey);
```
ekcoh, line 1471: Minor: Could be helpful to describe with a one liner what the stimuli and assertion here is expected to do to make it easier to come back to this test later on.

ekcoh (approved):
Thanks for submitting this PR and finding the root cause of this issue, fix and test looks good to me. If QA outcome is positive to verify bug is resolved this may land from my perspective.

ekcoh:
The tvOS CI failure is unfortunately a currently known issue with the test environment and may be ignored for now. Note that the test is also not marked "Required".

Pauliusd01 (approved):
LGTM, checked modifier functionality and the bug repro with 1 and 2 modifiers, did not see any issues.

Assets/Tests/InputSystem/CoreTests_Actions.cs
bmalrat: I added more comments

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1164,6 +1164,9 @@ private void DisableControls(int mapIndex, int controlStartIndex, int numControl
                     SetInitialStateCheckPending(bindingStatePtr, false);
                 manager.RemoveStateChangeMonitor(controls[controlIndex], this, mapControlAndBindingIndex);
 
+                // Ensure that pressTime is reset if the composite binding is reenable. ISXB-505
+                bindingStatePtr->pressTime = default;
+
```
timkeo, line 1169: I believe this is a good fix.
the `pressTime` field was central to the other bug I fixed, and so resetting it here makes perfect sense. I just wonder if there are _other_ places this fields should be reset.

timkeo (approved):

PR #1927, opened 2024-05-13
FIX: Avoid potential for NullReferenceException in FireStateChangeNotifications (ISXB-724).

graham-huws:
### Description

A tentative fix, given no known repro, for [ISXB-724](https://jira.unity3d.com/browse/ISXB-724).

### Changes made

Added early returns in cases we'd otherwise crash.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -11,10 +11,10 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased] - yyyy-mm-dd
 
 ### Fixed
-- Avoid potential crashes from NullReferenceException in FireStateChangeNotifications.
-- Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746).
-- Fixed resource designation for "d_InputControl" icon to address CI failure
-- Fixed Composite binding isn't triggered after disabling action while there are in progress [ISXB-505](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-505)
+- Avoid potential crashes from `NullReferenceException` in `FireStateChangeNotifications`.
+- Fixed an issue where a composite binding would not be consecutively triggered after ResetDevice() has been called from the associated action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746).
+- Fixed resource designation for "d_InputControl" icon to address CI failure.
+- Fixed an issue where a composite binding would not be consecutively triggered after disabling actions while there are action modifiers in progress [ISXB-505](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-505).
 
 ## [1.8.2] - 2024-04-29
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -11,8 +11,8 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased] - yyyy-mm-dd
 
 ### Fixed
-- Avoid potential crashes from NullReferenceException in FireStateChangeNotifications [ISXB-724](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-724)
-- Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746)
+- Avoid potential crashes from NullReferenceException in FireStateChangeNotifications.
+- Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746).
 
 ## [1.8.2] - 2024-04-29
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -9,6 +9,9 @@ Due to package verification, the latest version below is the unpublished version
 however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
+
+### Fixed
+- Avoid potential crashes from NullReferenceException in FireStateChangeNotifications [ISXB-724](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-724)
 - Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746)
 
 ## [1.8.2] - 2024-04-29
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ Due to package verification, the latest version below is the unpublished version
 however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
+
+### Fixed
+- Avoid potential crashes from NullReferenceException in FireStateChangeNotifications [ISXB-724](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-724)
```
ekcoh: This URL doesn't work. It seems the ticket is internal and not available via public tracker? In that case remove the hyperlink.

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
```diff
@@ -351,6 +351,11 @@ internal unsafe void FireStateChangeNotifications(int deviceIndex, double intern
         {
             Debug.Assert(m_StateChangeMonitors != null);
             Debug.Assert(m_StateChangeMonitors.Length > deviceIndex);
+            Debug.Assert(m_StateChangeMonitors[deviceIndex].listeners != null);
+            
+            if (m_StateChangeMonitors == null || m_StateChangeMonitors.Length <= deviceIndex ||
```
ekcoh: I think we need to reconsider this. I see the following issues (correct me if I am wrong about this....)

1) The checks in the if statement are basically the same as the Debug.Asserts (including the addition of checking listeners). It seems at first glance that the previous Debug.Asssert statements are there to catch "programmer errors" and not that its valid to call this function if there are no state change monitors or using an index out of bounds. I assume this is not happening. I am not certain that it's a good approach to just return if these conditions are not met since they likely indicate that something is logically wrong?
2) The next issue are these Debug.Assert methods in general which I believe is bad practise from an unsafe context since I believe it will crash the application/editor if not fulfilled right? Try with Debug.Assert(false) and see what happens from this method. Hence I believe we should remove the Debug.Asserts here and either throw exception if these fundamental conditions are not fulfilled or Debug.Assert them on a higher level in a non-unsafe context. It seems this method is being invoked from two places which are also in turn unsafe so it might be problematic to refactor this way even though internal....

I would propose (and let's discuss based on your opinion), that we convert those existing two Debug.Assert statements into Debug.Asserts on a higher level if possible or convert those into throwing if violated. In addition the listeners check you added here is new and it might actually be the offender here since I notice we set it to default (as mentioned in the comments of the ticket) when removing a state monitor. It might be possible that tis condition alone should be a condition for early return.

Packages/com.unity.inputsystem/CHANGELOG.md
graham-huws-unity: Thanks, fixed in 1c1c67b459aff7a28a55edf33f84e22648268d1b and I added a minor other fix too.

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
graham-huws-unity: For future reference, Håkan and I discussed this yesterday:
• Debug asserts can catch programmer errors, but in the couple of cases of crashes we've seen, it's been in release builds where they don't fire, and seemingly is caused by a timing issue so the problem doesn't happen in development builds. So, assert to catch obvious mistakes so people can improve their code, but be extra safe too, just in case.
• We believe the issue with Asserts in unsafe is specifically from the unit test setup, where we can use alternative assert functionality - standard Debug.Assert is fine to use in `unsafe` "normal" code.

graham-huws:
@stefanunity I realise there's nothing really to test here given that there's no repro steps for the original issue, but just wanted you aware of this regardless in case you had anything to add 😃

stefanunity (approved):
OK, given we don't have a repro for it.

graham-huws:
N.B. tvos failure is also happening on develop branch.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,7 +9,10 @@ Due to package verification, the latest version below is the unpublished version
 however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
-- Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746)
+
+### Fixed
+- Avoid potential crashes from NullReferenceException in FireStateChangeNotifications.
```
ekcoh: Minor: Use code formatting quotes, e.g. `NullReferenceException` in `FireStateChangeNotifications`.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,7 +9,10 @@ Due to package verification, the latest version below is the unpublished version
...
+- Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746).
```
ekcoh: I would suggest rewriting this to explain the actual problem in a more readable sentence than copying the subject of the ticket, e.g. 
"Fixed an issue where a composite binding would not be consecutively triggered after `ResetDevice()` has been called from the associated action handler." + link as currently.

ekcoh: Maybe not part of your PR but would be good to address. Also noticed fixe lines below lack full stop at the end.

Packages/com.unity.inputsystem/InputSystem/Devices/Gamepad.cs
```diff
@@ -86,7 +86,7 @@ public struct GamepadState : IInputStateTypeInfo
         internal const string ButtonSouthShortDisplayName = "Cross";
         internal const string ButtonNorthShortDisplayName = "Triangle";
         internal const string ButtonWestShortDisplayName = "Square";
-        internal const string ButtonEastShortDisplayName = "East";
+        internal const string ButtonEastShortDisplayName = "Circle";
```
ekcoh, line 89: Good catch, but realize this is solved in a very platform-dependent way in what is supposed to be platform-agnostic code. Will feed this into a doc I am creating on the subject.

Packages/com.unity.inputsystem/InputSystem/InputManagerStateMonitors.cs
```diff
@@ -349,14 +349,28 @@ private unsafe bool ProcessStateChangeMonitors(int deviceIndex, void* newStateFr
 
         internal unsafe void FireStateChangeNotifications(int deviceIndex, double internalTime, InputEvent* eventPtr)
         {
-            Debug.Assert(m_StateChangeMonitors != null);
-            Debug.Assert(m_StateChangeMonitors.Length > deviceIndex);
+            if (m_StateChangeMonitors == null)
```
ekcoh, line 352: I think this is better, good change.

ekcoh (approved):
LGTM

PR #1928, opened 2024-05-15
CHANGE: Remove redundant FastMouse current time call on InputState.Change (case ISX-1414)

AlexTyrer:
### Description

Reduce FastMouse calls to get the current time (one of them is redundant).

### Changes made

**FastMouse.OnNextUpdate**() makes two calls to **InputState.Change**() - one each for delta and scroll.

This results in two calls to **InputRuntime.s_Instance.currentTime** inside the **InputSystem.Change**() to pass as a parameter to the InputSystem **UpdateState**().

Instead query the current time and pass that as a parameter to the new templated **Change<>**() function which takes a time parameter. Events have their own internalTime which is used if present.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:
CHANGE: FastMouse.OnNextUpdate() now queries the current time and passes it as a parameter to InputState.Change() avoiding a redundant native call (case ISX-1414)

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -11,6 +11,7 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased] - yyyy-mm-dd
 - Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746)
 - Fixed resource designation for "d_InputControl" icon to address CI failure
+- Changed FastMouse.OnNextUpdate() to pass the current time as a parameter to InputState.Change() improving performance (case ISX-1414)
 
 ## [1.8.2] - 2024-04-29
 
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -11,7 +11,7 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased] - yyyy-mm-dd
 - Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746)
 - Fixed resource designation for "d_InputControl" icon to address CI failure
-- Changed FastMouse.OnNextUpdate() to pass the current time as a parameter to InputState.Change() improving performance (case ISX-1414)
+- Changed FastMouse.OnNextUpdate() to pass the current time as a parameter to InputState.Change() avoiding a redundant scripting to native call to query the time (case ISX-1414)
 
 ## [1.8.2] - 2024-04-29
 
--- a/Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
@@ -91,9 +91,9 @@ namespace UnityEngine.InputSystem.LowLevel
         /// device current.
         /// </remarks>
         public static void Change<TState>(InputControl control, TState state, InputUpdateType updateType = default, InputEventPtr eventPtr = default)
-                where TState : struct
+            where TState : struct
         {
-            double time = eventPtr.valid ? default : InputRuntime.s_Instance.currentTime; 
+            double time = eventPtr.valid ? default : InputRuntime.s_Instance.currentTime;
             Change(control, ref state, updateType, eventPtr, time);
         }
 
@@ -109,7 +109,7 @@ namespace UnityEngine.InputSystem.LowLevel
         /// device current.
         /// </remarks>
         public static unsafe void Change<TState>(InputControl control, TState state, double time, InputUpdateType updateType = default)
-                where TState : struct
+            where TState : struct
         {
             InputEventPtr eventPtr = new InputEventPtr(null);
             Change(control, ref state, updateType, eventPtr, time);
--- a/Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
@@ -90,10 +90,28 @@ namespace UnityEngine.InputSystem.LowLevel
         /// also performs related tasks such as checking state change monitors, flipping buffers, or making the respective
         /// device current.
         /// </remarks>
-        public static void Change<TState>(InputControl control, TState state, InputUpdateType updateType = default,
-            InputEventPtr eventPtr = default, double time = 0.0)
-            where TState : struct
+        public static void Change<TState>(InputControl control, TState state, InputUpdateType updateType = default, InputEventPtr eventPtr = default)
+                where TState : struct
         {
+            double time = eventPtr.valid ? default : InputRuntime.s_Instance.currentTime; 
+            Change(control, ref state, updateType, eventPtr, time);
+        }
+
+        /// <summary>
+        /// Perform one update of input state specifying the time
+        /// </summary>
+        /// <remarks>
+        /// Incorporates the given state and triggers all state change monitors as needed.
+        ///
+        /// Note that input state changes performed with this method will not be visible on remotes as they will bypass
+        /// event processing. It is effectively equivalent to directly writing into input state memory except that it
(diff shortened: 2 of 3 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased] - yyyy-mm-dd
 - Fixed Composite binding isn't triggered after ResetDevice() called during Action handler [ISXB-746](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-746)
 - Fixed resource designation for "d_InputControl" icon to address CI failure
+- Changed FastMouse.OnNextUpdate() to pass the current time as a parameter to InputState.Change() improving performance (case ISX-1414)
```
jfreire-unity: Do you know what is the performance impact? 
Should QA look at any performance tests to validate this claim?

AlexTyrer: It will be quite minor I think - it's just removing a redundant call from scripting to native and back to get the current accurate time.

AlexTyrer: I could change the changelog I suppose:

- Changed FastMouse.OnNextUpdate() to pass the current time as a parameter to InputState.Change() avoiding a redundant scripting to native call to query the time (case ISX-1414).

AlexTyrer:
Looks like I may need to fix-up some documentation after adding the default parameter.

This change also fails the APIVerificationTests.cs test:

o APIVerificationTests.API_MinorVersionsHaveNoBreakingChanges

Packages/com.unity.inputsystem/CHANGELOG.md
jfreire-unity: @AlexTyrer I see, thanks for clarifying. Yeah, maybe that sounds better. Fine for me 👍

jfreire-unity:
> Looks like I may need to fix-up some documentation after adding the default parameter.
> 
> This change also fails the APIVerificationTests.cs test:
> 
> o APIVerificationTests.API_MinorVersionsHaveNoBreakingChanges

Yeah.. maybe we need a overload method to not break the API, unfortunately.

Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
```diff
@@ -122,12 +122,20 @@ public static unsafe void Change(InputDevice device, InputEventPtr eventPtr, Inp
             var statePtr = UnsafeUtility.AddressOf(ref state);
             var stateOffset = control.stateBlock.byteOffset - device.stateBlock.byteOffset;
 
+            // Report any calls with no event and no specified time (relying on both default parameter values)
+            // This may not matter but would like to know about these during testing with more devices
+            //
+            if (!eventPtr.valid && time == 0.0)
```
ekcoh: Should we still proceed without return if this scenario occurs?

ekcoh (commented):
I noticed this is changing public API but with additive default values so I guess package verification tool categories this as a minor API break and not a major. It might require next release to be a minor revision though but not sure. @unity-cliff or @stefanunity might have insight into this?

Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
AlexTyrer: Good question - I thought it better to continue rather than just return here.

I haven't actually seen this case happen yet with my limited devices for testing.

What might be better is to simply set the time here in this case:

```
            if (!eventPtr.valid && time == default)
            {
                Debug.Log($"InputState.Change<>('{control.name}') called with null event and default time\n");
                time = InputRuntime.s_Instance.currentTime;
            }
```


I think this would preserve the existing behaviour where it would query the time if no event was provided.

AlexTyrer:
> > Looks like I may need to fix-up some documentation after adding the default parameter.
> > This change also fails the APIVerificationTests.cs test:
> > o APIVerificationTests.API_MinorVersionsHaveNoBreakingChanges
> 
> Yeah.. maybe we need a overload method to not break the API, unfortunately.

I've added a new Change<>() API that takes a non-optional time parameter - this leaves the existing API unchanged.

Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
AlexTyrer: I've refactored this so we no longer need this check.

There are two paths into this function from the two **Change<>()** templated functions.

The original **Change<>()** function always passes in an event (with it's own internal time) - if the event is not valid it passes the current time.

The other newly added **Change<>()** function always passes in a time value (it's a required parameter) along with a dummy event - this is the one used by the **FastMouse.OnNextUpdate()** function.

Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
```diff
@@ -90,11 +90,29 @@ public static unsafe void Change(InputDevice device, InputEventPtr eventPtr, Inp
         /// also performs related tasks such as checking state change monitors, flipping buffers, or making the respective
         /// device current.
         /// </remarks>
-        public static void Change<TState>(InputControl control, TState state, InputUpdateType updateType = default,
-            InputEventPtr eventPtr = default)
-            where TState : struct
+        public static void Change<TState>(InputControl control, TState state, InputUpdateType updateType = default, InputEventPtr eventPtr = default)
+                where TState : struct
+        {
+            double time = eventPtr.valid ? default : InputRuntime.s_Instance.currentTime;
```
lyndon-unity: Is this right - should it actually be set to default if the eventPtr is NOT valid ?

Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
```diff
@@ -90,11 +90,29 @@ public static unsafe void Change(InputDevice device, InputEventPtr eventPtr, Inp
...
+        /// <remarks>
+        /// Incorporates the given state and triggers all state change monitors as needed.
+        ///
+        /// Note that input state changes performed with this method will not be visible on remotes as they will bypass
+        /// event processing. It is effectively equivalent to directly writing into input state memory except that it
+        /// also performs related tasks such as checking state change monitors, flipping buffers, or making the respective
+        /// device current.
+        /// </remarks>
+        public static unsafe void Change<TState>(InputControl control, TState state, double time, InputUpdateType updateType = default)
+                where TState : struct
```
jfreire-unity: Maybe we could differentiate in the comments the need for `time` parameter vs the above method l? It's not immediately clear to me why we would need and it would help understand the need for this method.

jfreire-unity (approved):
Looks ok to me. 
Just asked some comment improvements in the "new" added `Change` method so that their need/use case is clearer. We will now have 4 overloaded `Change` methods so it helps to have some differentiation between them

Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
AlexTyrer: No, it's correct.

If the eventPtr **IS** valid then that event's internalTime will be used in the unsafe Change<>() function that's called (the time parameter can be default avoiding the call to get the current time).

If eventPtr is **NOT** valid then we need to look up the time here and pass that instead for the unsafe Change<>() function to use.

Packages/com.unity.inputsystem/InputSystem/State/InputState.cs
AlexTyrer: I only added this function as I first add a new default "time" parameter to the existing function but that caused the CI tooling complain that I've changed an API and this is not allowed without a version change.

I was just trying to avoid that restriction by not changing an existing API.

However I've still had to add the time parameter to the "guts" unsafe Change<>() function - it's not clear to me that the unsafe function shouldn't be an internal API... but I guess I can't change that now.

It all seems a bit of mess to me - I just want to add a new benign default parameter that we can optionally use to avoid a redundant call back to native the get the time.

AlexTyrer:
Closing as do not intend to land in trunk due to API change)

PR #1931, opened 2024-05-21
CHANGE: Update macos images to latest

stefanunity:
### Description

About a year ago we pinned the MacOS images for tvOS and iOS to a specific version because the new version broke our CI.
This is updating the image to the current latest.

### Changes made

Update image from MacOS 12 to latest MacOS 13 (4.50.0) which are used for tvOS and iOS. 
Also moved the image version into a variable in config.metadata.

Images for Mac Editor + Standalone have and will continue to use "latest".

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ekcoh (approved):
I think this adds value, but would be beneficial to align so that we test on similar images/versions as trunk. Have you checked this with relevant platform team(s)? It could be time consuming to test against another version and get issues not seen on trunk.

stefanunity:
> I think this adds value, but would be beneficial to align so that we test on similar images/versions as trunk. Have you checked this with relevant platform team(s)? It could be time consuming to test against another version and get issues not seen on trunk.

Trunk runs this https://images.bokken.prd.ds.unity3d.com/image/mobile/ios-macos-11-unity image for iOS + tvOS tests. 

We are using an image from a different pool of images (package-ci) https://images.bokken.prd.ds.unity3d.com/image/package-ci/macos-13.

PR #1934, opened 2024-05-24
NEW: Add check for PR titles prefixes in CI

jfreire-unity:
### Description

This PR adds a check for PR titles prefixes in our repo. It tries to enforce our strict PR title prefixes to make sure we have a consistent commit history when we squash-merge.

### Changes made

Added a GitHub Action, instead of Yamato, because we don't need any Unity infrastructure for this.

Runs a script to validate the PR title once it is created and edited. 

**Impact:**

If the PR has the wrong prefix, checks will fail. ❌
To fix it, just edit the PR title and the CI check will run again automatically.  ✅

### Notes

This can always be bypassed when we squash the commits and create the title for the new squashed commit in GitHubs UI. 

No Jira link. 
I did this on my own time since I wanted to have automated checks of things that usually are done during a PR review and now don't need to. We have one less thing to review :) 

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -1,0 +1,36 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on pull request creation and edit events but only for the "main" branch
+  pull_request:
+    types:
+      - edited
+      - opened
+      - synchronize
+    branches: [ "develop" ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  check-pr-title:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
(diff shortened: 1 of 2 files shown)


stefanunity (approved):

ekcoh (approved):
Good help with slap on the fingers when done inconsistently, does this guard in any way for editing again as part of merge stage though? Or is the possible to guard/check there as well? I believe that would be even better since it then could be corrected without Ci rerun. I often rewrite both title and commit message as part of the squash as one example.

jfreire-unity:
> Good help with slap on the fingers when done inconsistently, does this guard in any way for editing again as part of merge stage though? Or is the possible to guard/check there as well? I believe that would be even better since it then could be corrected without Ci rerun. I often rewrite both title and commit message as part of the squash as one example.

I didn't find a easy way to do that, unfortunately. It might be a Github UX limitation. But something to consider as an improvement for this.

The way that seems to allow that requires us to use Github Merge Queue. I tried using them in the past, but it will increase our merge to develop time.
 Using GH Merge Queues would create an intermediate "merged with `develop`" branch where all the CI checks would run again. So unless we want to have that, this extra "squash title prefix check" can't exist.

PR #1935, opened 2024-05-24
FIX: Add workaround for cases where button presses can be missed (ISXB-491)

graham-huws:
### Description

Fix for [ISXB-491](https://jira.unity3d.com/browse/ISXB-491).
The current design of the input system package relies on the following method for checking if a button has been pressed on the current frame: Check the current value of the button, and compare to the final value of the previous frame.
The issue with this approach is that presses can be missed if they happened very quickly, or there was some form of lag making the update take longer than intended. Whether a button is pressed is only properly evaluated when asked, so if we don't check every input coming in, some can be missed.

### Changes made

This PR seeks to solve this in a very simplistic way - check every value that affects ButtonControls and record their pressed state at the time the change comes in. Current belief is that a "clever" solution to this issue would require a fairly extensive rewrite of how the input system works.
[Performance comparison data can be viewed on Sheet 2 here](https://docs.google.com/spreadsheets/d/1mwc6MZ6dAoLvFUoi_jKOR8O9aLab04tifJWqQ9NWw-8/edit#gid=1572787835).

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
```diff
@@ -972,25 +976,27 @@ internal static TDevice Build<TDevice>(string layoutName = default, string layou
...
             else
             {
                 if (m_ControlTreeNodes[0].leftChildIndex != -1)
-                    WriteChangedControlStatesInternal(statePtr, stateSizeInBytes * 8,
-                        deviceStateBuffer, m_ControlTreeNodes[0], 0);
+                    WriteChangedControlStatesInternal(statePtr, deviceStateBuffer, m_ControlTreeNodes[0], 0);
             }
         }
 
-        private unsafe void WritePartialChangedControlStatesInternal(void* statePtr, uint stateSizeInBits,
```
lyndon-unity, line 995: Are these parameters removed simply because they were never used?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3650,6 +3650,15 @@ internal unsafe bool UpdateState(InputDevice device, InputEvent* eventPtr, Input
                     stateOffsetInDevice, statePtr, stateSize, flipped);
             }
 
+            if (makeDeviceCurrent)
```
lyndon-unity, line 3754: Why do we have this if check - shouldn't it always update the buttons ?


(also unrelated to this PR but the UpdateState function doesn't document the return info)

lyndon-unity (commented):
Would be good to get a summary of the performance impact. The spreadsheet has lots of data but doesn't make it very clear what the overall impact is.

My main concern is that this fixes lost events but we are hiding it behind a flag. I'd like to know the impact of enabling the flag by default.

Packages/com.unity.inputsystem/InputSystem/Devices/InputDevice.cs
graham-huws-unity: Yep, they just passed themselves into themselves with recursive function calls. Rider is super handy for spotting stuff like that :)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
graham-huws-unity: if makeDeviceCurrent is false, my understanding is that means that while there was a state change event, nothing actually happened that would be considered an input, hence the "noise" mask this is checked with. I was getting a lot of hits on this from mouse updates where none of the buttons were actually changing, so this saved a lot of unneeded processing.

Hopefully others can weigh in more if I'm wrong here!

Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -782,7 +782,7 @@ public void Performance_OptimizedControls_EvaluateStaleControlReadsWhenGamepadSt
 
 #if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
         // Disable the project wide actions actions to avoid performance impact.
-        InputSystem.actions.Disable();
+        InputSystem.actions?.Disable();
```
ekcoh, line 1001: Good (maybe unrelated) catch

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - Avoid potential crashes from `NullReferenceException` in `FireStateChangeNotifications`.
+- Fixed cases where `wasPressedThisFrame` would not return true if a press and release happened within the same frame before being queried (and vice versa for `wasReleasedThisFrame`). As this is an edge case which necessitates checking more data, enabling this fix requires the `"USE_READ_VALUE_CACHING"` feature flag to be enabled.
```
ekcoh: I am not very fond of tying this fundamental functional requirement to an optional hidden optimisation flag.

Assets/Tests/InputSystem/CoreTests_State.cs
```diff
@@ -478,8 +484,18 @@ public void State_PressingAndReleasingButtonInSameFrame_DoesNotShowStateChange()
         InputSystem.Update();
 
         Assert.That(gamepad.buttonEast.isPressed, Is.False);
-        Assert.That(gamepad.buttonEast.wasPressedThisFrame, Is.False);
-        Assert.That(gamepad.buttonEast.wasReleasedThisFrame, Is.False);
+        if (usesReadValueCaching)
```
ekcoh: I think its fine use the caching feature to provide fast-path evaluation if it provides additional value, but I would like us to fix this without being dependent on that feature. Wouldn't this be possible?

ekcoh: I do not see any reason to associate proper button press event functionality with this other half-hidden feature? If a flag is needed as some kind of "recent to incorrect legacy behavior" introduce another flag is my suggestion.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -1583,6 +1583,7 @@ public static DeviceBuilder Setup(this InputDevice device, int controlCount, int
             device.m_Device = device;
 
             device.m_ChildrenForEachControl = new InputControl[controlCount];
+            device.m_UpdatedButtons = new Dictionary<int, ButtonControl>();
```
ekcoh: Wouldn't we be able to initialize this once directly here since we should know how many buttons the device has? We basically know it's at least less than controlCount. I also couldn't see we actually use the key at any point, would a set be sufficient?

Assets/Tests/InputSystem/CoreTests_State.cs
```diff
@@ -462,11 +462,17 @@ public void State_CanDetectWhetherButtonStateHasChangedThisFrame()
...
     // The way we keep state does not allow observing the state change on the final
-    // state of the button. However, actions will still see the change.
+    // state of the button, unless the readValueCaching path is enabled.
+    // However, actions will still see the change.
     [Test]
     [Category("State")]
-    public void State_PressingAndReleasingButtonInSameFrame_DoesNotShowStateChange()
+    [TestCase(true)]
+    [TestCase(false)]
+    public void State_PressingAndReleasingButtonInSameFrame_ShowsStateChange_WithCaching(bool usesReadValueCaching)
```
ekcoh: I would suggest we extend to cover some combinations here as well, having 1 event, 2 events, 3 events to cover those permutations.

lyndon-unity:
Added Paulius to help testing performance on Android

Pauliusd01:
The original bug project crashes when built on my only Android testing phone (Xiaomi MI 8 Lite) so I used the QA mobile sensor testing project instead when idling and when interacting with the UI buttons. Not sure what to make of the results so I'm just gonna post the screenshot with analyzer data as well as profiler captures. [Capture data](https://drive.google.com/file/d/1kAylQIt6gChtpiGuC87UJ2WnwHz5MYOb/view?usp=sharing)

analyzer screenshots

(In the screenshot blue highlighted area is for simply idling the app and the right green area is me pressing UI buttons every 0.5~ sec)

lyndon-unity:
These captures don't show any significant slow down so I think this is good to land (pending reacting to feedback above in PR)

lyndon-unity:
Screenshot 2024-06-03 at 16 39 28
There is a slowdown on some frames for the PreUpdate.NewInputUpdate.
0.20 to 0.34 on one frame
Only 2 frames show slow down more and 0.1

Screenshot 2024-06-03 at 16 41 14
InputProcess has some slow down but its minimal 0.03 to 0.04

How many taps were made in this sample ?

Pauliusd01:
For the past tapping captures I just profiled for 15~ secs and tapped twice every second, the number of taps could be different from both versions. Here's a brand new capture of me idling for 5 secs, tapping once -> idling for 5 secs more and tapping again. Both of these should now just be 2 taps each

image
[ExtraCaptures.zip](https://github.com/user-attachments/files/15544081/ExtraCaptures.zip)

I used this project: https://github.cds.internal.unity3d.com/unity/InputSystem-TestProject

Pauliusd01:
Here's one of me just tapping non stop for around 15 secs

image
[ExtraCaptures2.zip](https://github.com/user-attachments/files/15544802/ExtraCaptures2.zip)

lyndon-unity:
> Here's one of me just tapping non stop for around 15 secs

Screenshot 2024-06-04 at 13 30 08
Sorting by time and looking at the slowest 100 frames (when there is active tap interation) there is a small slowdown in InputUpdate 0.01ms (0.06->0.07ms). But this seems reasonable for the reliability improvement of the edge detection from the fix.

Packages/com.unity.inputsystem/CHANGELOG.md
lyndon-unity: I would also like to see this decoupled from feature flag as this seems like a fundamental fix thats required to be on all the time

lyndon-unity (changes requested):
I'd like to see this decoupled from the feature flag as I see this as a critical fix that should always be enabled.

graham-huws:
Have updated to address above comments, plus some extra issues that cropped up. Have added new perf results to the second sheet of the gdoc.

Assets/Tests/InputSystem/CoreTests_State.cs
```diff
@@ -478,8 +484,33 @@ public void State_PressingAndReleasingButtonInSameFrame_DoesNotShowStateChange()
         InputSystem.Update();
 
         Assert.That(gamepad.buttonEast.isPressed, Is.False);
-        Assert.That(gamepad.buttonEast.wasPressedThisFrame, Is.False);
-        Assert.That(gamepad.buttonEast.wasReleasedThisFrame, Is.False);
+        Assert.That(gamepad.buttonEast.wasPressedThisFrame, Is.True);
```
lyndon-unity, line 495: Please to see this change! so we do detect the state change

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Fixed
 - Avoid potential crashes from `NullReferenceException` in `FireStateChangeNotifications`.
+- Fixed cases where `wasPressedThisFrame` would not return true if a press and release happened within the same frame before being queried (and vice versa for `wasReleasedThisFrame`).
```
lyndon-unity, line 20: Woohoo!

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
```diff
@@ -118,9 +132,50 @@ public new bool IsValueConsideredPressed(float value)
         /// </code>
         /// </example>
         /// </remarks>
-        public bool wasPressedThisFrame => device.wasUpdatedThisFrame && IsValueConsideredPressed(value) && !IsValueConsideredPressed(ReadValueFromPreviousFrame());
+        #if UNITY_EDITOR
+        public bool wasPressedThisFrame => InputUpdate.s_UpdateStepCount ==
```
lyndon-unity: I find it odd that we have a separate editor state - but that's outside the scope of this change.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3702,6 +3702,40 @@ internal unsafe bool UpdateState(InputDevice device, InputEvent* eventPtr, Input
...
+                        else
+#endif
+                        ((ButtonControl)device.allControls[button]).UpdateWasPressed();
+                    }
+                }
+                else
+                {
+                    foreach (var button in device.m_ChildrenThatAreButtonControls)
+                    {
+                        #if UNITY_EDITOR
```
lyndon-unity, line 3777: minor but different indenting to the #ifdef a few lines above

Assets/Tests/InputSystem/CoreTests_Controls.cs
```diff
@@ -488,29 +487,47 @@ public void Controls_ValueCachingWorksAcrossEntireDeviceMemoryRange()
         InputSystem.actions?.Disable();
 #endif
 
-        var keyboard = InputSystem.AddDevice<Keyboard>();
+        // Value caching is supported for all controls, but with changes to button press detection, ButtonControls will
```
ekcoh: Good catch, thanks for updating this.

Assets/Tests/InputSystem/CoreTests_State.cs
```diff
@@ -462,11 +462,17 @@ public void State_CanDetectWhetherButtonStateHasChangedThisFrame()
     }
 
     // The way we keep state does not allow observing the state change on the final
-    // state of the button. However, actions will still see the change.
+    // state of the button, unless the readValueCaching path is enabled.
```
ekcoh: Is this still valid?

ekcoh: It seems on test case now that the setting doesn't matter? (Desirable)

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
ekcoh: Agreed, there is actually a lot of work done to separate interaction with game view vs editor while in play mode in general in many places.

Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs
```diff
@@ -1583,6 +1582,8 @@ public static DeviceBuilder Setup(this InputDevice device, int controlCount, int
             device.m_Device = device;
 
             device.m_ChildrenForEachControl = new InputControl[controlCount];
+            device.m_UpdatedButtons = new HashSet<int>();
```
ekcoh: Do you think it would be relevant to initialize the capacity of this hash-set based off the number of buttons available on the device? This should be fixed so its basically a "constant" based off the device descriptor.

ekcoh: Code below here basically seems to count the number of buttons anyway

Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceBuilder.cs
```diff
@@ -52,6 +52,8 @@ internal struct InputDeviceBuilder : IDisposable
             InstantiateLayout(layout, variants, new InternedString(), null);
             FinalizeControlHierarchy();
 
+            m_Device.m_UpdatedButtons = new HashSet<int>();
```
ekcoh: Similar comment as before.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3739,7 +3773,8 @@ internal unsafe bool UpdateState(InputDevice device, InputEvent* eventPtr, Input
                     deviceStateSize);
             }
 
-            if (InputSettings.readValueCachingFeatureEnabled)
+            // Keyboard has enough ButtonControls that it's faster to find out which have actually changed here.
+            if (InputSettings.readValueCachingFeatureEnabled || m_Devices[deviceIndex] is Keyboard)
```
ekcoh: I wonder if there is also a synergy with knowing the button count also here? It sounds like the benefit here is actually more related to the number of buttons on the device rather than it being a Keyboard? Hence maybe this should be a threshold based off number of buttons? Its fine as is but something like that would make it more future proof?

Assets/Tests/InputSystem/CoreTests_Controls.cs
ekcoh: However, this makes me wonder, will this evaluation work for detecting "button behavior presses" for controls that are not buttons? E.g. transition of axis under the button behavior (magnitude based) within a sub-frame? Basically this is not unique to buttons in any way, it's all about avoiding aliasing in the sampling process. Not saying these changes not already carry a lot of value, but trying to understand what problems still exist. Hence I wonder if we should setup an additional test case similar to the one done for wasPresseed/wasReleased this frame w.r.t. e.g. axis, or maybe you already can say what we guarantee for that use case? (We could just say, use actions for something like that)

ekcoh (approved):
I am going to approve this one now from my perspective and not taking other reviewer comments into account. I left a few questions and minor comments though for consideration. Great to see this being adressed, thanks @graham-huws!

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
graham-huws-unity: Fair point, I'll fix - unfortunately there seems to be not much standardisation for #if alignment even within this one file, let alone the rest of the codebase 😅

ekcoh:
@lyndon-unity Could you help summarise what additional updates are needed to transition this from "Request Changes"?

lyndon-unity (approved):
I've going to approve this, but ideally we'd follow up to adjust it to limit the performance impact by only applying the extra checks when there is code that needs it. (I..e I think only if wasPressedThisFrame or wasReleasedThisFrame are used by the user). Ideally we'd defer the work to the call site.

I would like to hear the XR team perspective of the performance impact as there is a risk this is unexceptable in their use case.
Ideally they would have the option to opt out of this performance cost if they don't need if for specific actions/bindings.

Assets/Tests/InputSystem/CoreTests_Controls.cs
graham-huws-unity: I believe what you mention will be detected just fine, eg. gamepad has a ButtonControl for "left stick up"?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
graham-huws-unity: Updated - with some testing with the new changes, it seems that having about 45 buttons that have pressed/releasedThisFrame usage is where it becomes more efficient to take the cache value, so we now swap over at that point.

graham-huws:
Now updated to a new version which only does processing for ButtonControls which have had wasPressedThisFrame/wasReleasedThisFrame called, saving a lot of performance when not many are being utilised.

Added V2 benchmarks to the spreadsheet, and new tests which try varying amounts of buttons to test to see the impact.

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
```diff
@@ -132,17 +173,49 @@ public new bool IsValueConsideredPressed(float value)
...
+                {
+                    var currentlyPressed = IsValueConsideredPressed(value);
+                    var pressedLastFrame = IsValueConsideredPressed(ReadValueFromPreviousFrame());
+                    BeginTestingForFramePresses(currentlyPressed, pressedLastFrame);
 
-        public bool wasReleasedThisFrame => InputUpdate.s_UpdateStepCount ==
-        (InputUpdate.s_LatestUpdateType.IsEditorUpdate() ? m_UpdateCountLastReleasedEditor : m_UpdateCountLastReleased);
-        #else
-        public bool wasPressedThisFrame => InputUpdate.s_UpdateStepCount == m_UpdateCountLastPressed;
+                    return device.wasUpdatedThisFrame && currentlyPressed && !pressedLastFrame;
```
lyndon-unity, line 192: This line is much easier to read than before which is great.

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
```diff
@@ -172,8 +202,11 @@ public bool wasReleasedThisFrame
                 // Take the old path if this is the first time calling.
                 if (!needsToCheckFramePress)
                 {
-                    needsToCheckFramePress = true;
-                    return device.wasUpdatedThisFrame && !IsValueConsideredPressed(value) && IsValueConsideredPressed(ReadValueFromPreviousFrame());
+                    var currentlyPressed = IsValueConsideredPressed(value);
+                    var pressedLastFrame = IsValueConsideredPressed(ReadValueFromPreviousFrame());
+                    BeginTestingForFramePresses(currentlyPressed, pressedLastFrame);
+
+                    return device.wasUpdatedThisFrame && !currentlyPressed && pressedLastFrame;
```
lyndon-unity, line 214: Again this is easier to read which is a good change

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
lyndon-unity: So if I'm understanding correctly, this old flow can still miss some rapid in frame press/release on the first check, since we haven't run though all the transitions.

Its good we have the more expensive flow opted in, so not always paying the cost if never used, but I'd like to see a comment above to state that this can miss a transition on the first ever call.

lyndon-unity (approved):
As the performance is improved for the cases with less calls to wasPressedThisFrame with this update then I'm comfortable with it landing as it fixes the key issue and limits the performance impact to the flows where the checks are needed.

Packages/com.unity.inputsystem/InputSystem/Controls/ButtonControl.cs
ekcoh: Agree on the comment. Then if you care for the first call you may at least call this once up front to get it triggered.

ekcoh (approved):
Looks good, but agree to Lyndons comment. 

Did you check consistency with wasPerformedThisFrame? Since that one is only available on actions I think it could have value to have a test that sets up an action bound to button and then coevaluates wasPerformedThisFrame with outcome of wasPressedThisFrame and wasReleasedThisFrame.

PR #1941, opened 2024-05-27
FIX: Scheme Name in Control Scheme editor menu that gets reset when editing devices (ISXB-763)

bmalrat:
### Description

Fixed Scheme Name in Control Scheme editor menu that gets reset when editing devices

### Changes made

In ControlSchemesView
* Fixed RedrawUI to use the control scheme name or new/renamed one currently in edition
* Fixed control scheme name text box change callback to now support :
1. Trim the name to prevent having space at the begin/end of the name
2. Prevent renaming the scheme if you change the name and put back the original name in the same edition without saving
3. Write back in the UI if a unique name has been generated

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
@@ -35,7 +35,7 @@ namespace UnityEngine.InputSystem.Editor
                     var newName = ((TextField)evt.currentTarget).value.Trim();
                     if (string.IsNullOrEmpty(newName) || String.Compare(newName, state.selectedControlScheme.name) == 0)
                     {
-                        m_NewName = "";
+                        m_NewName = String.Empty;
                         // write back the value to the text field if the name was empty
                         ((TextField)evt.currentTarget).value = state.selectedControlScheme.name;
                     }
@@ -146,7 +146,7 @@ namespace UnityEngine.InputSystem.Editor
             // the changes retained. However, if a different ControlScheme is selected or the Asset
             // Editor window is closed, then the changes are lost.
 
-            m_NewName = "";
+            m_NewName = string.Empty;
             OnClosing?.Invoke(this);
         }
 
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
@@ -33,9 +33,11 @@ namespace UnityEngine.InputSystem.Editor
                 {
                     // If the name is the same as the current name, don't change it
                     var newName = ((TextField)evt.currentTarget).value.Trim();
-                    if (String.Compare(newName, state.selectedControlScheme.name) == 0)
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
```diff
@@ -31,8 +31,21 @@ public ControlSchemesView(VisualElement root, StateContainer stateContainer, boo
             {
                 Dispatch((in InputActionsEditorState state) =>
                 {
-                    m_NewName = ControlSchemeCommands.MakeUniqueControlSchemeName(state,
-                        ((TextField)evt.currentTarget).value);
+                    // If the name is the same as the current name, don't change it
+                    var newName = ((TextField)evt.currentTarget).value.Trim();
+                    if (string.IsNullOrEmpty(newName) || String.Compare(newName, state.selectedControlScheme.name) == 0)
+                    {
+                        m_NewName = "";
```
ekcoh: Minor, why not string.Empty?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
```diff
@@ -31,8 +31,21 @@ public ControlSchemesView(VisualElement root, StateContainer stateContainer, boo
             {
                 Dispatch((in InputActionsEditorState state) =>
                 {
-                    m_NewName = ControlSchemeCommands.MakeUniqueControlSchemeName(state,
-                        ((TextField)evt.currentTarget).value);
+                    // If the name is the same as the current name, don't change it
+                    var newName = ((TextField)evt.currentTarget).value.Trim();
+                    if (string.IsNullOrEmpty(newName) || String.Compare(newName, state.selectedControlScheme.name) == 0)
```
ekcoh, line 36: Minor: I think you can remove the string.IsNullOrEmpty since it cannot be null from Trim() right?
In addition String.Compare will be 0 if newName is empty.

ekcoh (approved):
Basically this fix looks good to me, good work, but I had some minor tweaks I think could be adressed. Minor though so approving already now but let you consider if it makes sense to update based on comments.

Pauliusd01 (approved):
LGTM, checked changing scheme names when making a new one/editing and adding new devices at various times.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
bmalrat: copy pasted from CloseView()

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
bmalrat: yes it can't be null but IsEmpty doesn't exist.
String.Compare() will not be 0 if the string is empty, except if both are empty

PR #1944, opened 2024-05-31
DOCS: ISXB-621 input system UI input module enables referenced UI action map

smnwttbr:
### Description

Addresses confusion when default action map is enabled, and the UI action map is also enabled via the InputSystemUIInputModule. [ISXB-621](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-621)

### Changes made

Documentaiion added to UISupport.md

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/Documentation~/UISupport.md
```diff
@@ -57,7 +57,7 @@ You can also reset the UI action map to its default bindings by selecting **Rese
 ### Input System Module
 
 > **Note:**
-> If you have an instance of the [Input System UI Input Module](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html) component in your scene, the settings on that component takes priority and are used instead of the UI settings in your project-wide actions.
+> If you have an instance of the [Input System UI Input Module](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html) component in your scene, the settings on that component takes priority and are used instead of the UI settings in your project-wide actions. Also, The UI action map will be enabled, along with the default action map specified on any InputSystem component in the scene.
```
duckets: When you say "InputSystem component" in the added sentence, do you mean the UI Input Module component?

ekcoh (approved):

ekcoh (commented):
PR is all good, but I wonder if this is the expected behavior. Did you compare to behavior with Input System prior to introduction of Project wide Input actions and UI for Input? If not I would recommend this is not a regression relating to the auto-enable functionality of Project-wide input actions.

simonwittber:
I have not tested myself, however comments in the Jira ticket indicated this was not a regression.

I believe this should be expected behaviour, IMO it makes sense based on how the InputSystemUIInputModule works and it's us-case.

Packages/com.unity.inputsystem/Documentation~/UISupport.md
duckets: I've gone ahead and made the change

duckets (approved):
approved with minor amendment

ekcoh:
Adding @Pauliusd01 for QA assessment w.r.t. above mentioned auto-enabling

Pauliusd01:
Help me clear up a few things, for instance
>If you have an instance of the [Input System UI Input Module](../api/UnityEngine.InputSystem.UI.InputSystemUIInputModule.html) component in your scene, the settings on that component takes priority and are used instead of the UI settings in your project-wide actions

What settings in the PWA asset are being referred here? Bindings? Interactions/Processors? Isn't that inferred naturally as you don't see that asset selected in this picker:
image

And for this part:
> Also, The UI action map will be enabled, along with the default action map specified on any UI Input Module component in the scene.

This seems to be true, both maps show up enabled in the debugger. But my question is why? If only one is actually being used, why do we keep the PWA one enabled? If there's a good reason then we should maybe mention that here? If not then change it? (not in this PR of course)

Pauliusd01:
I played around a bit more to see what is loaded in and when 

For UITK UI it goes like this:

No PWA asset and UI Module exists/doesn't exist - DefaultInputActions is loaded and used
PWA asset and UI Module doesn't exist - only PWA is loaded and used
PWA asset and UI Module exists - PWA and DefaultInputActions loaded in, but DefaultInputActions are used

For IMGUI UI the UI Module has to exist at all times for the UI to work and the PWA asset does not affect it in anyway yet it is loaded in. So I guess the new wording is correct yet the whole thing is a bit confusing still to me, mostly the reasoning behind PWA UI map being loaded in even when not needed

Pauliusd01 (approved):

PR #1946, opened 2024-06-04
FIX: ISXB-704 scroll wheel support windows raw input value

benoitalain:
### Description

Fixes ISXB-704. Based on recently landed [trunk PR](https://github.cds.internal.unity3d.com/unity/unity/pull/50445).

Details of what was changed and the thought process can be found in this Google Document:
https://docs.google.com/document/d/1XCC6GqaWzKf1M1jm-UmKRn6sIlmvF7rvF1C8Ws5Yxa4/edit?usp=sharing

This PR introduces a new Settings option for the InputSystem package. See screenshot:
image

When the "Keep Platform Specific Input Range" option is selected, InputSystem actions that result from a Scroll Wheel hardware input will be expressed in the OS native range when retrieved with the `InputAction.CallbackContext.ReadValue<Vector2>()` method. On Windows, the values should range between -120 to 120, whereas on other platforms it will remain between -1 and 1 in general.

The new setting's default option is "Uniform Across All Platforms". When that option is selected, InputSystem actions that result from a Scroll Wheel will always be expressed in the -1 to 1 range, regardless of platform.

### Changes made

### Notes

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1946) <br/>All committers have signed the CLA.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -281,6 +281,25 @@ private void OnFocusChanged(bool focus)
         public Vector2 screenSize => new Vector2(Screen.width, Screen.height);
         public ScreenOrientation screenOrientation => Screen.orientation;
 
+        public bool normalizeScrollWheelDelta
+        {
+#if UNITY_6000_0_OR_NEWER
+            get => NativeInputSystem.normalizeScrollWheelDelta;
+            set => NativeInputSystem.normalizeScrollWheelDelta = value;
+#else
+            get; set;
```
ekcoh: Not sure I see what role these getter setters play when earlier Unity version? Wouldn't it make sense to remove the property completely instead?

ekcoh: (or would have expected hard coded behavior, but the answer is - it depends)

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -281,6 +281,25 @@ private void OnFocusChanged(bool focus)
...
+#endif
+        }
+
+        public float scrollWheelDeltaPerTick
+        {
+#if UNITY_6000_0_OR_NEWER
+            get { return NativeInputSystem.GetScrollWheelDeltaPerTick(); }
```
ekcoh: Recommend being consistent with getter below and use the short-hand version with =>

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -529,8 +529,9 @@ void OnClickPerformed(InputAction.CallbackContext ctx, EventSource eventSource,
 
         void OnScrollWheelPerformed(InputAction.CallbackContext ctx)
         {
-            var scrollDelta = ctx.ReadValue<Vector2>();
-            if (scrollDelta.sqrMagnitude < k_SmallestReportedMovementSqrDist)
+            // ISXB-704: convert input value to uniform ticks before sending them to UI.
+            var scrollTicks = ctx.ReadValue<Vector2>() / InputSystem.scrollWheelDeltaPerTick;
```
ekcoh, line 533: Is InputSystem.scrollWheelDeltaPerTick guaranteed to never be zero (to make sure we never can run into division by zero)

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -529,8 +529,9 @@ void OnClickPerformed(InputAction.CallbackContext ctx, EventSource eventSource,
...
+            if (scrollTicks.sqrMagnitude < k_SmallestReportedMovementSqrDist)
```
ekcoh, line 534: Nice to avoid the sqr root

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -281,6 +281,25 @@ private void OnFocusChanged(bool focus)
         public Vector2 screenSize => new Vector2(Screen.width, Screen.height);
         public ScreenOrientation screenOrientation => Screen.orientation;
 
+        public bool normalizeScrollWheelDelta
+        {
+#if UNITY_6000_0_OR_NEWER
```
ekcoh: I guess we run into issues with UNITY 6 revision prior to trunk change landing but maybe ok if/when that revision is current or is there opportunity for a even more specific define in the asmdef to detect the exact expected version once the trunk change lands maybe?

ekcoh (commented):
Commenting for now until aligned on potential graceful version handling.

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -489,6 +491,28 @@ public void UIActionScroll_FiresUIScrollEvents_FromInputMouse(bool useProjectWid
         });
     }
 
+    [Category(kTestCategory)]
+    [TestCase(1.0f)]
+    [TestCase(120.0f)]
+    public void UIActionScroll_ReceivesNormalizedScrollWheelDelta(float scrollWheelDeltaPerTick)
```
ekcoh, line 498: Thanks for adding these tests to cover scaling.

ekcoh:
Looks like a branch update and CI rerun might be in order.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
benoitalain: I added an Assert on the native side in the trunk PR, is that enough? I'm not a fan of adding more asserts on all the stages where we propagate that value, do you recommend adding one here too? Only here?

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
benoitalain: Not sure what the general solution to this problem is. There are other places in the Input System package code where we have a similar problem, so if we find a solution for this we should apply it everywhere (at once) probably.

ekcoh: What I was after above was whether UNITY_6000_0_OR_NEWER was specific enough or whether to define a custom asmdef symbol based on version value to define a symbol more specifically reflecting the change as introduced into UNITY_6000 version. There are other examples here https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef but I noticed they very not specific on a "sub-version" level so probably ok.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputSystemProvider.cs
ekcoh: If trunk version guarantees that its never zero I believe its fine to not check it in package. Agree that its better to define the API contract on lower level.

ekcoh (approved):
Approving this now, I saw some minor comments remain but ok with me to leave them as-is if you consider it to be preferable to have it as it currently is.

duckets (approved):
Approved with amendments to tooltip and API docs

Pauliusd01 (changes requested):
Couple of concerns from me: 

- I did not need these changes to fix the issue, simply opening the Editor version that has your latest changes is enough to fix it
- The ifdefs for the code are not granular enough and I still get errors about missing Editor code on 6.0.6f1 Editor, should that be changed?

Also, please take a look at the slack discussion about this change in our input channel

benoitalain:
> Couple of concerns from me:
> 
> * I did not need these changes to fix the issue, simply opening the Editor version that has your latest changes is enough to fix it
> * The ifdefs for the code are not granular enough and I still get errors about missing Editor code on 6.0.6f1 Editor, should that be changed?
> 
> Also, please take a look at the slack discussion about this change in our input channel

You say you only need the Editor changes, are you sure there's no copy of my package fixes that survived in your Library folder? Otherwise I can't see how the InputSystemUIInputModule fixes could come from the Editor only... makes no sense :-(

About the ifdefs not being granular enough, I'll comment in the slack discussion but I'm not sure there's a clean solution.

Pauliusd01:
> > Couple of concerns from me:
> > 
> > * I did not need these changes to fix the issue, simply opening the Editor version that has your latest changes is enough to fix it
> > * The ifdefs for the code are not granular enough and I still get errors about missing Editor code on 6.0.6f1 Editor, should that be changed?
> > 
> > Also, please take a look at the slack discussion about this change in our input channel
> 
> You say you only need the Editor changes, are you sure there's no copy of my package fixes that survived in your Library folder? Otherwise I can't see how the InputSystemUIInputModule fixes could come from the Editor only... makes no sense :-(

It's the user project from ISXB-704 with no library and 1.8.2 input system

benoitalain:
> > > Couple of concerns from me:
> > > 
> > > * I did not need these changes to fix the issue, simply opening the Editor version that has your latest changes is enough to fix it
> > > * The ifdefs for the code are not granular enough and I still get errors about missing Editor code on 6.0.6f1 Editor, should that be changed?
> > > 
> > > Also, please take a look at the slack discussion about this change in our input channel
> > 
> > 
> > You say you only need the Editor changes, are you sure there's no copy of my package fixes that survived in your Library folder? Otherwise I can't see how the InputSystemUIInputModule fixes could come from the Editor only... makes no sense :-(
> 
> It's the user project from ISXB-704 with no library and 1.8.2 input system

Hi! I've updated the PR description to emphasize the changes that this PR contains. It's not observable scroll range fixes in any of the UI systems, but just the option to allow InputSystem actions returning values in the -120 to 120 range.

Pauliusd01 (approved):
Thanks for the extra info, lgtm. Do check the CI failures though, there's more than the usual culprit ones

PR #1951, opened 2024-06-20
FIX: Fixed a performance issue with many objects using multiple action maps [ISXB-573]

AlexTyrer:
### Description

Avoid costly (and pointless) iterations over all messages / objects / maps when we don't need to.

Issue: https://jira.unity3d.com/browse/ISXB-573

This codepath can be very expensive when there are a lot of messages to send to lots of objects with maps - much of the cost is in dereferencing **handle.Target** in **DeferredResolutionOfBindings**() which involves taking a mutex lock.

Further, once the lock is taken for each object, **ResolveBindingsIfNecessary**() is called for each of that object's **InputActionMaps** which in the general case does nothing (when there is nothing to do).

In the supplied (ridiculous to show the point) test project with 900 objects, each with 6 maps receiving 4500 mouse move messages this leads to:

- over 21 million calls to **ResolveBindingsIfNecessary**() that do nothing
- over 4 million mutex locks in resolving **handle.Target** in **DeferredResolutionOfBindings**()
- frame times over 1 second!

### Changes made

This change adds a static **InputActionMap.s_NeedToResolveBindings** that is set when we know that at least one InputActionMap has changes that need resolving. This is cleared when a resolve pass has been completed.

With this change the frame time in the example only increases by about 2-3ms when the mouse is moving.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -24,6 +24,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where `InputActionAsset.FindAction(string, bool)` would throw `System.NullReferenceException` instead of returning `null` if searching for a non-existent action with an explicit action path and using `throwIfNotFound: false`, e.g. searching for "Map/Action" when `InputActionMap` "Map" exists but no `InputAction` named "Action" exists within that map [ISXB-895](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-895).
 - Fixed an issue where adding a `OnScreenButton` or `OnScreenStick` to a regular GameObject would lead to exception in editor.
 - Fixed an issue where adding a `OnScreenStick` to a regular GameObject and entering play-mode would lead to exceptions being generated.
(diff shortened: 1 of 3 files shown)


lyndon-unity (approved):
I've seen this fixed the performance issue and also still passes all the tests 

(2 tests failed if the triggering call itself was commented out in ProcessControlStateChange : https://github.com/Unity-Technologies/InputSystem/blob/f7cc21e2faa15d1e999cbfe1fc441e16e5b45538/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs#L1416 )

ekcoh (approved):

ekcoh:
@AlexTyrer Seems to be conflicts in changelog that needs to be resolved before merge.

Pauliusd01 (approved):

AlexTyrer:
> @AlexTyrer Seems to be conflicts in changelog that needs to be resolved before merge.

Thanks - I've resolved this now!

PR #1954, opened 2024-06-21
FIX: ISXB-925 Fixed an issue where changing InputSettings instance fo…

ekcoh:
…r the Input System wouldn't affect feature flags.

### Description

The optimization flags within InputSettings were stored as static fields for performance reasons. This makes little sense from API perspective where these are set and evaluated using a non-static API. This also leads to switching InputSettings instance having no impact on current settings. As long as these are exposed via this API they should not be static. However, I added extraction of the settings into InputManager which is represented by a static and used direct access from where its used so expecting little if any impact.

Adressing:
- Feature flags set and checked via InputSettings non-static interface but backed by static variables.

### Changes made

- Changed flags to be extracted in Input Manager instead when settings change.
- Removed custom handling in InputSettings which allow IsFeatureEnabled to work.

Note that this fix also includes https://github.com/Unity-Technologies/InputSystem/pull/1955 since the test case would otherwise fail with debug error log on non Windows platforms.

### Testing

Added 1 test case for scenario `Settings_ShouldStoreSettingsAndFeatureFlags(string featureName)`. I didn't add a new test category at this point since none of the existing fits well so I left the test case uncategorised.

### Risk

Potential risk is features not being enabled correctly for some corner-case scenarios.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


graham-huws (approved):

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2116,6 +2116,12 @@ internal struct AvailableDevice
         internal IInputRuntime m_Runtime;
         internal InputMetrics m_Metrics;
         internal InputSettings m_Settings;
+
+        // Extract as booleans (from m_Settings) because feature check is in the hot path
+        internal bool m_OptimizedControlsFeatureEnabled;
+        internal bool m_ReadValueCachingFeatureEnabled;
+        internal bool m_ParanoidReadValueCachingChecksEnabled;
```
timkeo: Could you please change these to private fields and add an internal read-only accessor property?

When refactoring for CoreCLR support, internal fields were very frustrating because I had to check the entire code base for any usages.

timkeo (approved):
From [my FEPM refactoring worksheet,](https://docs.google.com/spreadsheets/d/1FHIpLELBkFgMURChojRbHsPA1jKEVlT0EaFbvwQLYUQ/edit?gid=1391290404#gid=1391290404) I didn't touch these static fields, since determined it wasn't necessary for CoreCLR support. However, there may still be conflicts.

I recommend checking these changes against [my FEPM refactor branch](https://github.com/Unity-Technologies/InputSystem/pull/1915) to make sure isn't anything too complicated, otherwise porting this to Neutron will be a pain.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: This have been adressed now in https://github.com/Unity-Technologies/InputSystem/pull/1954/commits/c93e565d0b9260c0911f2af882871dd8551fe80a

ekcoh: Regarding FWPM refactor I think this is straightforward since ApplySettings seem intact.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2116,6 +2117,33 @@ internal struct AvailableDevice
         internal IInputRuntime m_Runtime;
         internal InputMetrics m_Metrics;
         internal InputSettings m_Settings;
+
+        // Extract as booleans (from m_Settings) because feature check is in the hot path
```
ritamerkl, line 2121: nice!

ritamerkl (approved):
Good change!

stefanunity (approved):

PR #1957, opened 2024-06-25
FIX: UI generation of custom interactions of action properties when it rely on OnGUI callback (ISXB-886) 

bmalrat:
### Description

The action editor was not displaying the content of custom interactions when it's done in IMGUI. [ISXB-886](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-886)
jira : https://jira.unity3d.com/browse/ISXB-886

When no custom editor exist the generated UI for enum doesn't work well. The value is lost when save + error in the console.

### Changes made

Added an IMGUIContainer to handle IMGUI UI.
Before the change :
image

After the change :
image

Also added some early exit when IMGUI is not the default to prevent to UI to be drawn twice.

Also fixed the generation of UI for enum which loosed the value when saved. Now it use the index instead of trying to convert the string to the value.

### Testing

Tested the repro project which contain IMGUI UI and buildin interaction in UITK + custom interactions without editor 

### Risk

Only in editor, should be safe

### Checklist

Before review:

- [X ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

- [X ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Pauliusd01:
Checking

Pauliusd01:
It seems like I lose focus every time I interact with these fields from the custom interactions, is this a bug or something the user script lacks? I'm leaning towards a bug 😅 
https://github.com/Unity-Technologies/InputSystem/assets/54306142/e39154cd-d191-45cb-8b94-78fc31b37cc7

Also, added @ekcoh for code review, feel free to add someone else if you're busy

Pauliusd01:
For some reason the Axis Deadzone processor is duplicated with your changes. This seems to be the only processor/interaction affected (All settings are completely default, all you have to do is create a new input actions asset and add the processor)
image

Pauliusd01 (changes requested):
Updating status

bmalrat:
> ract with these fields from the custom interactions, is this a bug or something the user script lacks? I'm leaning towards a bug 😅

good catch I missed those ones, processor editor are also InputParameterEditor, I only checked Interaction.
They also have 2 UI Implementation.

bmalrat:
> It seems like I lose focus every time I interact with these fields from the custom interactions, is this a bug or something the user script lacks? I'm leaning towards a bug 😅 https://github.com/Unity-Technologies/InputSystem/assets/54306142/e39154cd-d191-45cb-8b94-78fc31b37cc7
> 
> Also, added @ekcoh for code review, feel free to add someone else if you're busy
thanks I will take a look

ekcoh (approved):
Code changes LGTM. Do I understand it correctly that the focus issues are still an open issue? It sounds like that might need some attention. I am approving this now given that the code changes looks good but there might be side effects as called out by Paulius?

Pauliusd01:
> Code changes LGTM. Do I understand it correctly that the focus issues are still an open issue? It sounds like that might need some attention. I am approving this now given that the code changes looks good but there might be side effects as called out by Paulius?

We've discussed this with Benoit and decided to make a separate bug for that issue. I've since noticed it also affects some non custom interactions such as the multi tap one, but it is not as a big of a problem there as the custom ones could be. (Bug link: https://jira.unity3d.com/browse/ISX-2074)

Pauliusd01 (approved):
LGTM, checked that extisting interactions/processors function normally and that the custom ones show up correctly as well.

PR #1958, opened 2024-07-02
FIX: Errors are thrown when assigning Input Actions Asset that contains a custom InputProcessor or InputBindingComposite as Project-wide Input Actions (ISXB-856)

bmalrat:
### Description

The static registration of custom processor, interaction, BindingComposite could lead to undefined behavior for project action if the static initiation doesn't occur in the right order which is by design hard to control.

### Changes made

Add reflection to load custom InputProcessor, IInputInteraction and InputBindingComposite during the InputManage initialisation.
Added a profiler marker for this step.

### Testing

local test + bug project

### Risk

It add some time in the initialisation and register all derived type which was not the case before.
We may encounter some name collision and he doesn't have a way to disable the automatic discovery.

### Checklist

Before review:

- [ X] Changelog entry added.
    - Performance tests.
    - Integration tests.
- [ X] Docs for new/changed API's.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Actions/IInputInteraction.cs
```diff
@@ -80,7 +80,7 @@ namespace UnityEngine.InputSystem
     /// </code>
     /// </example>
     ///
-    /// Custom interactions can be registered using <see cref="InputSystem.RegisterInteraction"/>. This can be
+    /// Custom interactions are automatically registed by reflexion but it can also be manually registered using <see cref="InputSystem.RegisterInteraction"/>. This can be
```
chris-massie:
```suggestion
    /// Custom interactions are automatically registered by reflection but it can also be manually registered using <see cref="InputSystem.RegisterInteraction"/>. This can be
```

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1965,6 +1966,61 @@ internal void InitializeData()
             composites.AddTypeRegistration("ButtonWithTwoModifiers", typeof(ButtonWithTwoModifiers));
             composites.AddTypeRegistration("OneModifier", typeof(OneModifierComposite));
             composites.AddTypeRegistration("TwoModifiers", typeof(TwoModifiersComposite));
+
+            // Register custom types by reflection
+            RegisterCustomTypes();
+        }
+
+        void RegisterCustomTypes(Type[] types)
```
chris-massie, line 1975: Can we make sure to add a `ProfilerMarker` to track how long this takes?

bmalrat: good idea I will add that

ekcoh (approved):
Good improvement, LGTM!

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: yeah, that sounds good. @chris-massie are you concerned with startup times with this change?

jfreire-unity (approved):

PR #1808, opened 2023-12-18
CHANGE: Input Action Editor Session, Component Editor, Build/Settings Analytics Support + Refactor (ISX-1546)

ekcoh:
### Description

This PR introduces Unity Editor analytics support for getting insight into Input Action editor engagement and usage. Analytics have been designed to enable differentiation between an editor embedded into Project Settings (Project-wide actions) as well as a free-floating asset based editor, but otherwise be comparable. The analytics currently only track engagement on a high-level without being specific about specific user actions. Defined metrics are tracked via associated DASD ticket, see ISX-1546 for reference.

Note that this PR has been marked as a CHANGE instead of NEW since it doesn't introduce new user features.

### Changes made

Added new class `InputActionsEditorSessionAnalytic` acting as a proxy object around `InputActionsEditorSessionData` to provide abstraction around counting operations to prevent direct access and possible related mistakes in code providing metrics. The purpose of this class is to track and report analytics related to .inputactions asset editor windows.

Added new classes to track editor (Inspector) editor changes for the following components:  
- `PlayerInputEditorAnalytic` tracking selected changes to configuration of `PlayerInput`.
- `PlayerInputManagerEditorAnalytic` tracking selected changes to configuration of `PlayerInputManager`.
- `OnScreenStickEditorAnalytic` tracking selected changes to configuration of `OnScreenStick`.
- `VirtualMouseInput` tracking selected changes to configuration of `VirtualMouseInput`

Added new class `InputComponentEditorAnalytic` tracking statistics of how often an Inspector with the associated component editor is displayed for the following components:
- `PlayerInput`
- `PlayerInputManager`
- `OnScreenStick`
- `OnScreenButton`
- `VirtualMouseInput`
- `TouchSimulation`
- `StandaloneInputModule`
- `InputSystemUIInputModule`

Added new class `InputBuildAnalytic` that reports statistics on settings when a project is being built. Currently all InputSettings are being logged, see associated spreadsheet (via internal ticket) for details.

Other changes
- Modified new analytics code to conditionally utilise `UnityEngine.Analytics.IAnalytic.IData` when compiled on Unity 2023.2 or later and otherwise use legacy setup for analytics. This is achieved by conditionally defining the new IInputAnalytic to either derive from the new IAnalytic or to define a similar interface. This simplifies test and usage code.
- Refactored existing analytics collection code and reduced number of allocations by using arrays directly instead of lists.
- Refactored existing analytics constants into C# constants.
- Refactored analytics tests based on above changed.
[...]

PR #1971, opened 2024-07-23
FIX: Input System 1.9.0 does not compile on Console Platforms (ISXB-963)

jfreire-unity:
### Description

Fix build issues on console platforms that do not support analytics.

### Changes made

Guards code with `ENABLE_CLOUD_SERVICES_ANALYTICS` which was raising compilation errors on console platforms.

From [documentation](https://docs.unity3d.com/ScriptReference/Analytics.Analytics.html):
> Important: Always make calls to the Analytics API inside the scope of an #if ENABLE_CLOUD_SERVICES_ANALYTICS statement to prevent compilation errors on platforms that do not support Analytics:

Possible there's some things we can do to improve our `UNITY_ANALYTICS` defines vs `ENABLE_CLOUD_SERVICES_ANALYTICS` use. I raised the question on what's the difference between them. cc @ekcoh for further discussions.

For now, this will likely unblock the console platform but we should revisit this in the future IMO. Raised task here

### Testing

I tested data data was sent from the macOS editor and that there apparently there's no regression in the data that is being sent. But maybe @ekcoh can validate 100% once he's back from vacation. @Pauliusd01 if you can check that there's no regression in the data sent to the backend, that would be great (I mentioning you because you approved the PR with analytics #1808 ).

Also validated that at least one console platform the build error was fixed. @lewish-unity please make sure this branch fixes the error you reported.

### Risk

I validated that the Editor analytics were sent with Unity 6, but couldn't validate with 2022.3f as the analytics debugger is not available for that version. 

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
@@ -393,22 +393,24 @@ namespace UnityEngine.InputSystem.LowLevel
 
         public void SendAnalytic(InputAnalytics.IInputAnalytic analytic)
         {
-        #if (UNITY_EDITOR && ENABLE_CLOUD_SERVICES_ANALYTICS)
-            #if (UNITY_2023_2_OR_NEWER)
+        #if ENABLE_CLOUD_SERVICES_ANALYTICS
+            #if (UNITY_EDITOR)
+                #if (UNITY_2023_2_OR_NEWER)
             EditorAnalytics.SendAnalytic(analytic);
-            #else
+                #else
             var info = analytic.info;
             EditorAnalytics.RegisterEventWithLimit(info.Name, info.MaxEventsPerHour, info.MaxNumberOfElements, InputAnalytics.kVendorKey);
             EditorAnalytics.SendEventWithLimit(info.Name, analytic);
-            #endif // UNITY_2023_2_OR_NEWER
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -394,22 +393,22 @@ public Action onProjectChange
 
         public void SendAnalytic(InputAnalytics.IInputAnalytic analytic)
         {
-            #if (UNITY_EDITOR)
+        #if (UNITY_EDITOR && ENABLE_CLOUD_SERVICES_ANALYTICS)
```
ekcoh: Looks alright, sorry for overlooking this. But why not just add a single #if ENABLE_CLOUD_SERVICES_ANALYTICS outside the existing editor checks since all of it should be excluded?

ekcoh (approved):

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
jfreire-unity: Thanks for the suggestion, improved it now based on what you said

lewish-unity (approved):
Tested on affected platforms and the error no longer appears. Can create a build successfully with this version installed.
CI is passing.

PR #1977, opened 2024-08-02
NEW: Added tests for Input Action Editor UI for managing action maps (ISX-2087)

bmalrat:
### Description

Added UI Toolkits test for the Input Action Editor. https://jira.unity3d.com/browse/ISX-2087
It tests list, creation, rename and delete.

### Changes made

Fixed Action Maps name edition which could be inconsistent in Input Action Editor UI. The time used used was not the editor one and the UI was destroyed and recreated on each change.

### Testing

local test on windows + CI

### Risk

limited to Input Action Editor UI.

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -30,7 +30,8 @@ static InputActionsEditorWindow()
         }
 
         static readonly Vector2 k_MinWindowSize = new Vector2(650, 450);
-
+        // For UI testing purpose
+        internal InputActionAsset currentAssetInEdition => m_AssetObjectForEditing;
```
lyndon-unity: Is this name a typo?
currentAssetInEdition 

Should it be ...
currentAssetInEditior

lyndon-unity (commented):
Overall looks like a reasonable test.
Small naming query added, but happy to approve.

lyndon-unity (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
bmalrat: I changed the name, thanks

Assets/Tests/InputSystem.Editor/InputActionsEditorTests.cs
```diff
@@ -0,0 +1,261 @@
...
+    public void TearDown()
+    {
+        m_Window?.Close();
+    }
+
+    #endregion
+
+    #region Helper methods
+
+    void Click(VisualElement ve)
```
ekcoh: Awesome to see this being added but I wonder whether it would make sense to move these helper methods into a separate utility class to make it more accessible across any test classes that would utilise this kind of "UI automation" actions? Would also be great with a short comment on what each one does and maybe some preconditions asserts to check that the passed element is applicable to avoid mistakes? 

Also would suggest renaming "ve" to "target" since it represents the target of the automated action to be carried out?

Assets/Tests/InputSystem.Editor/InputActionsEditorTests.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        Event evtu = new Event();
+        evtu.type = EventType.MouseUp;
+        evtu.mousePosition = ve.worldBound.center;
+        evtu.clickCount = 1;
+        using var pue = PointerUpEvent.GetPooled(evtu);
+        ve.SendEvent(pue);
+    }
+
+    void SendText(VisualElement ve, string text, bool sendReturn = true)
```
ekcoh: Similarly, think it could be good with some docs on these utility functions, is sendReturn used to configure whether the pretended user is hitting Return/Enter key after entering the text?

Assets/Tests/InputSystem.Editor/InputActionsEditorTests.cs
ekcoh: Encapsulating it might further reduce visibility of some of these methods where SendXXX is likely internal while a utility class for testing might contain "actions" like, Click, EnterText etc.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionMapsTreeViewItem.cs
```diff
@@ -87,7 +101,7 @@ public void FocusOnRenameTextField()
 
             //a bit hacky - e.StopImmediatePropagation() for events does not work like expected on ListViewItems or TreeViewItems because
             //the listView/treeView reclaims the focus - this is a workaround with less overhead than rewriting the events
-            DelayCall();
+            schedule.Execute(() => renameTextfield.Q<TextField>().Focus()).StartingIn(120);
```
ekcoh, line 104: Thank you for getting rid of that unreliable DelayCall

ekcoh: Guess its still suffers from not being fully deterministic though...

ekcoh (approved):
Awesome to see some UI automation happening for tests, good job. Some minor comments.

Assets/Tests/InputSystem.Editor/InputActionsEditorTests.cs
bmalrat: thanks I will make some name adjustment.
Most of the methods are copied from https://internaldocs.unity.com/editor_and_runtime_development_guide/TestAutomation/TestFrameworks/UITestFramework/CodeSamples/code-samples/ and will be replaced by the helpers on neutron

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionMapsTreeViewItem.cs
bmalrat: yes this delay could create some instabilities, it was hard to find somethings reliable 🤞

Assets/Tests/InputSystem.Editor/InputActionsEditorTests.cs
bmalrat: yes the sendReturn is used to define if we send return at the end of the text

Assets/Tests/InputSystem.Editor/InputActionsEditorTests.cs
bmalrat: For the ve => target is unfortunately not as simple. For clicking it may work because it took the positon and if the target it on top the click will be dispatched to it.
For all other command even if you call visualelement.sendevent it will be dispatched to the panel which will search for te focused element.
I make some change to make it clearer

ekcoh (approved):

PR #1862, opened 2024-02-29
FIX: ISXB-543 GamepadButton property drawer fix

ekcoh:
### Description

FIX: ISXB-543 GamepadButton property drawer now has consistent behaviour for which aliased enum value is selected and drop-down shows alias mapping. https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-543 

Original contribution by @ppandi-rythmos in https://github.com/Unity-Technologies/InputSystem/pull/1860 routed via this PR.

### Changes made

Introducing a custom aliased enum property drawer and a specialised version for GamepadButton enum type.

### Notes

Easiest to test by creating a script with a GamepadButton public property and check in Inspector.

### Checklist

Before review:

- [x] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #1983, opened 2024-08-13
CHANGE: Removed aliased values in Gamepad enum. (ISXB-543)

adrian-koretski-unity3d:
### Description

Fixing confusing menu when selecting face buttons due to aliased enum values. 

### Changes made

Removed aliased enum values. 

### Testing

Manual testing. 

### Risk

None. 

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=1983) <br/>All committers have signed the CLA.

Packages/com.unity.inputsystem/InputSystem/Devices/Gamepad.cs
```diff
@@ -326,64 +326,6 @@ public enum GamepadButton
         /// The right trigger button on a gamepad.
         /// </summary>
         RightTrigger = 33,
-
```
ekcoh: We cannot remove these from this enum without requiring a major version revision since it would break public API for existing customers (and according to semver), hence this is not a viable way forward.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Use `ProfilerMarker` instead of `Profiler.BeginSample` and `Profiler.EndSample` when appropriate to enable recording of profiling data.
+- Removed aliased values in GamepadButton enum.
```
ekcoh: We do not want to release a major revision due to this change.

ekcoh (changes requested):

ekcoh:
I would recommend considering a fix based off this one but with inlined/combined options for aliased entries? https://github.com/Unity-Technologies/InputSystem/pull/1862

ekcoh:
Seems like @Billreyn  was also considering that unless opinion has shifted, see https://github.com/Unity-Technologies/InputSystem/pull/1862#issuecomment-1971097297

Packages/com.unity.inputsystem/InputSystem/Devices/Gamepad.cs
adrian-koretski-unity3d: I hadn't thought of this. I re-added the aliased enum values (although I did not reset them in the other files that used them to move towards north/south/east/west). I also added a slightly modified version of the aliased enum drawer from [#1862](https://github.com/Unity-Technologies/InputSystem/pull/1862) where we can remove values from the drawer.

adrian-koretski-unity3d:
> Seems like @Billreyn was also considering that unless opinion has shifted, see [#1862 (comment)](https://github.com/Unity-Technologies/InputSystem/pull/1862#issuecomment-1971097297)

I believe so. Based on our conversation, we want to display only North/South/East/West

Packages/com.unity.inputsystem/CHANGELOG.md
adrian-koretski-unity3d: Should we plan to do so down the line? Perhaps mark the aliased values as depricated (I'm not sure how we go about doing this).

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -4661,7 +4661,7 @@ public void Actions_PressingAndReleasingButtonInSameUpdate_StillTriggersAction()
             ctx => { ++receivedCalls; };
         action.Enable();
 
-        var firstState = new GamepadState {buttons = 1 << (int)GamepadButton.B};
```
ekcoh: The fix shouldn't have any impact on existing tests?

ekcoh: Hence I wouldn't expect any changes to any tests.

Assets/Tests/InputSystem/Plugins/UserTests.cs
```diff
@@ -942,7 +942,7 @@ public void Users_CanDetectUseOfUnpairedDevice()
 
         ++InputUser.listenForUnpairedDeviceActivity;
 
-        InputSystem.QueueStateEvent(gamepad, new GamepadState { leftStick = new Vector2(1, 0)}.WithButton(GamepadButton.A));
```
ekcoh: This shouldn't need to change either.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: We basically shouldn't modify the enum at all, we should aim for a completely editor-based "fix" here. There is no functional issue with the current solution, its only confusing users from UX perspective.

ekcoh:
> > Seems like @Billreyn was also considering that unless opinion has shifted, see [#1862 (comment)](https://github.com/Unity-Technologies/InputSystem/pull/1862#issuecomment-1971097297)
> 
> I believe so. Based on our conversation, we want to display only North/South/East/West

Exactly, with emphasis on "display", hence it should be property drawer only changes as I see it.

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: It makes sense though to file an internal ticket to consider removing aliases for control enums down the line (in a major future release). I have now filed it.

Packages/com.unity.inputsystem/InputSystem/Devices/Gamepad.cs
```diff
@@ -255,31 +255,31 @@ public enum GamepadButton
         /// The upper action button on a gamepad.
         /// </summary>
         /// <remarks>
-        /// Identical to <see cref="Y"/> and <see cref="Triangle"/> which are the Xbox and PlayStation controller names for this button.
+        /// Identical to <see cref="North"/> and <see cref="North"/> which are the Xbox and PlayStation controller names for this button.
```
ekcoh: This seems incorrect? Referencing to the same thing

Assets/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -103,7 +103,7 @@ public void Events_CanListenForEvents()
         allButtonPresses.Clear();
 
         Release(gamepad.buttonSouth);
-        InputSystem.QueueStateEvent(gamepad, new GamepadState(GamepadButton.A, GamepadButton.B));
```
ekcoh: Not expecting any changes needed to this file

Assets/Tests/InputSystem/CoreTests_State.cs
```diff
@@ -424,7 +424,7 @@ public void State_CanUpdateButtonStateUsingEvent()
 
         Assert.That(gamepad.buttonEast.isPressed, Is.False);
 
-        var newState = new GamepadState {buttons = 1 << (int)GamepadButton.B};
```
ekcoh: Not expecting any changes needed to this file

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AliasedEnumPropertyDrawer.cs
```diff
@@ -0,0 +1,72 @@
+#if UNITY_EDITOR
```
ekcoh: Screenshot 2024-09-13 at 11 09 14
Screenshot 2024-09-13 at 11 11 23
I am afraid this is not doing what its supposed to.....

The following things need to be addressed:
- Aliased values should be part of the same entry when the inspector is displayed, e.g. "South / X / A"
- Drop down need to allow all values to be picked, e.g. either "South / X / A" or each one individually would be fine from my perspective.
- At the moment when selecting e.g. South, the value displayed is "North", so something with mapping is incorrect

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AliasedEnumPropertyDrawer.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            ProcessDisplayNamesForAliasedEnums();
+            return base.CreatePropertyGUI(property);
+        }
+
+        protected abstract bool TryGetNonAliasedNames(string enumName, string displayName, out string outputName);
+
+        private void ProcessDisplayNamesForAliasedEnums()
+        {
+            var enumNamesAndValues = new Dictionary<string, int>();
```
ekcoh: Since the enum is small and we want consistency, maybe skip this base class and just move into specific GamepadButtonPropertyDrawer class and just use a big switch statement to get it right? That would be quite straightforward. Otherwise there needs to be some smartness to get the order of each aliased combination separated by slashes consistent.

ekcoh (changes requested):
Needs a bit more work I believe

Assets/Tests/InputSystem/Plugins/UserTests.cs
adrian-koretski-unity3d: I believe this is due to my first pass where I removed the aliased values. Will revert

Assets/Tests/InputSystem/CoreTests_Actions.cs
adrian-koretski-unity3d: I believe this is due to my first pass where I removed the aliased values. Will revert

Packages/com.unity.inputsystem/InputSystem/Devices/Gamepad.cs
adrian-koretski-unity3d: I believe this is due to my first pass where I removed the aliased values. Will revert

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/AliasedEnumPropertyDrawer.cs
adrian-koretski-unity3d: I'll go over it and make sure it behaves and displays correctly

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/GpadButtonPropertyDrawer.cs
```diff
@@ -0,0 +1,82 @@
...
+using UnityEditor;
+using UnityEngine.UIElements;
+
+namespace UnityEngine.InputSystem.Editor
+{
+    /// <summary>
+    /// Property drawer for <see cref = "GamepadButton" />
+    /// </summary >
+    [CustomPropertyDrawer(typeof(GamepadButton))]
+    internal class GpadButtonPropertyDrawer : PropertyDrawer
```
lyndon-unity: Minor but can you use a full name here rather than abbreviation 
GpadButtonPropertyDrawer ->
GamepadButtonPropertyDrawer

lyndon-unity (approved):
Code looks reasonable, one minor recommendation, but I haven't tested it in action.
Suggest adding QA or design for a manual verification

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ however, it has to be formatted properly to pass verification tests.
...
```
ekcoh: This should be reverted as well, suggest to change to something like e.g. "Changed representation of GamepadButton enum values in Inspector to display aliased enum values as a single item to avoid confusion around selection and aliased value display when multiple enum items map to the same numerical value."

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/GpadButtonPropertyDrawer.cs
ekcoh: Agree

ekcoh: Which also means renaming file and .meta

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/GpadButtonPropertyDrawer.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public override VisualElement CreatePropertyGUI(SerializedProperty property)
+        {
+            var enumNamesAndValues = new Dictionary<string, int>();
+            var enumDisplayNames = Enum.GetNames(typeof(GamepadButton));
+            var enumValues = Enum.GetValues(typeof(GamepadButton)).Cast<GamepadButton>().ToArray();
+
+            for (var i = 0; i < enumDisplayNames.Length; ++i)
+            {
+                string enumName = enumDisplayNames[i];
```
ekcoh: This assignment could be on line 40 right since overwritten for the special cases?

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/GpadButtonPropertyDrawer.cs
```diff
@@ -0,0 +1,82 @@
...
+                    case nameof(GamepadButton.West):
+                        enumName = "West, X, Square, Y";
+                        break;
+                    default:
+                        break;
+                }
+
+                if (enumName != null)
+                {
+                    enumNamesAndValues.Add(enumName, (int)enumValues.GetValue(i));
```
ekcoh: Could be moved up to the place where it is currently assigned as null?

ekcoh (approved):
Approving this to not block this PR in any way while I have limited accessability. I didn't run it on this review round so recommend checking functionality in QA pass.

Pauliusd01:
Will check this today or tomorrow morning at the latest

Pauliusd01 (changes requested):
Entering play mode with one of these new values selected breaks inspector and throws errors. Also, the enum value does not match the one in the inspector (See debug log in screenshot, whether that's expected I dunno) 

1726666305 IN-45621_-_Scene_-_Windows,_Mac,_Linux_-_Unity_6_P

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/GpadButtonPropertyDrawer.cs
adrian-koretski-unity3d: I changed it up a little bit. For the cases where we do not want to enter the enum into the list, I continue the for loop, eliminating the need for null/null check. The switch statement sets an string value which is then entered into the list.

adrian-koretski-unity3d:
> Entering play mode with one of these new values selected breaks inspector and throws errors. Also, the enum value does not match the one in the inspector (See debug log in screenshot, whether that's expected I dunno)
> 
> 1726666305 IN-45621_-Scene-_Windows,_Mac,Linux-_Unity_6_P

The nullref issue should be fixed now. The value being displayed will be one of the aliased enums for that value so this is to be expected.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Use `ProfilerMarker` instead of `Profiler.BeginSample` and `Profiler.EndSample` when appropriate to enable recording of profiling data.
+- Changed representation of GamepadButton enum values in Inspector to display aliased enum values as a single item to avoid confusion around selection and aliased value display when multiple enum items map to the same numerical value.
```
Pauliusd01: Changelog entry in the wrong version and no bug issue link

Pauliusd01 (approved):

Pauliusd01:
@adrian-koretski-unity3d I approved too hastily, the link has to be to a public issue tracker not our internal jira lol

PR #1967, opened 2024-07-18
FIX: tree state in treeview should be preserved when saving (ISXB-966)

smnwttbr:
### Description

When the input action asset is saved, the selection state of the trees in the editor will occasionally be reset.

### Changes made

The TreeStateView has now been setup correctly.

### Testing

Manual testing only.

### Risk

Risk is 0.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - JIRA ticket linked, example ([case %<ID>%](https://issuetracker.unity3d.com/product/unity/issues/guid/<ID>)). If it is a private issue, just add the case ID without a link.
    - Jira port for the next release set as "Resolved".
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

- [ ] Create forward/backward port if needed. If you are blocked from creating a forward port now please add a task to ISX-1444.

PR #1984, opened 2024-08-14
FIX: ISXB-1013-prevent-ui-update-loop

smnwttbr:
### Description

The asset editor could go into a UI update loop in some circumstances.

### Changes made

Removed an else clause in the code which was executed when selecting an empty action map.

### Testing

Manual and CI tests are successfuly.

### Risk

0

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -13,6 +13,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 - Fixed memory allocation on every frame when using UIDocument without EventSystem. [ISXB-953](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-953)
 - Fixed Action Maps name edition which could be inconsistent in Input Action Editor UI.
+- Fixed an update loop in the asset editor.
 
 ### Added
 - Added Hinge Angle sensor support for foldable devices.
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -13,7 +13,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 - Fixed memory allocation on every frame when using UIDocument without EventSystem. [ISXB-953](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-953)
 - Fixed Action Maps name edition which could be inconsistent in Input Action Editor UI.
-- Fixed an update loop in the asset editor.
+- Fixed an update loop in the asset editor that occurs when selecint an Action Map that has no Actions.
 
 ### Added
 - Added Hinge Angle sensor support for foldable devices.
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 - Fixed memory allocation on every frame when using UIDocument without EventSystem. [ISXB-953](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-953)
 - Fixed Action Maps name edition which could be inconsistent in Input Action Editor UI.
+- Fixed an update loop in the asset editor.
```
ekcoh: Would suggest describing the symptom and the actual issue happening that was fixed here instead or what do you think?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -135,11 +135,6 @@ public ActionsTreeView(VisualElement root, StateContainer stateContainer)
                     var item = m_ActionsTreeView.GetItemDataForIndex<ActionOrBindingData>(m_ActionsTreeView.selectedIndex);
                     Dispatch(item.isAction ? Commands.SelectAction(item.name) : Commands.SelectBinding(item.bindingIndex));
                 }
```
ekcoh: I guess this is safe, the problem here is that there seem to be two related but separate concepts of selection, one in the UI model part of the UITK controls and one in the custom model part of this repo.

ekcoh: My only worry is that when the number of actions goes down to zero, as in deleting the last action form a map, the UI might be without selection but the State model might have a selection so they are out of sync. I suspect the loop is due to the below commands when being handled by the UI code do not skip further processing if the UI model and this custom selection model are actually equal. E.g. that else should maybe be an else if checking if the selections are not in sync.

ekcoh (approved):

Pauliusd01:
Will check this today

Pauliusd01:
https://github.com/user-attachments/assets/fc302023-56c2-4120-bd1f-25e4ee267701
Noticed there's still "ghost" selections that I can interact with and refresh the UI with by just pressing/holding delete after my first deletion. This also happens without the PR so I don't think it needs addressing but just had to mention it

Pauliusd01 (approved):
LGTM, checked whether the bug still reproduces or not with custom input action window/PWA. Checked whether this could cause performance issues or other problems with selections when deleting/undoing/redoing/duplicating etc.

bmalrat (approved):

Pauliusd01:
For context a fix has already landed for this accidentally by Simon here: https://github.com/Unity-Technologies/InputSystem/pull/1967#event-13925624234 I don't know if this can cause conflicts or just needs a changelog entry only, etc, just mentioning it

PR #1986, opened 2024-08-19
FIX: JsonParser / ParseStringValue now creates unescaped strings

AlexTyrer:
### Description

Comparing escaped and non-escaped strings is very slow - this happens when devices are deactivated/reactivated such as when the Quest2 headset and controllers enter and awake from sleep mode.

The high cost comes from having to compare each string character-by-character (skipping '\' escapes) each time entering the C# runtime to perform System.Globalization.ToUpperInvariant.

### Changes made

This change to TryMatchDisconnectedDevice() means that any parsed strings that contain escape characters are converted to non-escaped strings for much quicker comparison.

This saves approximately 0.5ms / frame when an Oculus headset or controller is reactivated.

### Testing

Local testing with the Quest2 headset and controllers using Superluminal to capture the cost of device awakening.

### Risk

We allocate a temporary string where before we did not.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -19,6 +19,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Use `ProfilerMarker` instead of `Profiler.BeginSample` and `Profiler.EndSample` when appropriate to enable recording of profiling data.
+- Improved performance of disconnected device activation (ISX-1450)
 
 ### Added
 -  Added tests for Input Action Editor UI for managing action maps (List, create, rename, delete) (ISX-2087)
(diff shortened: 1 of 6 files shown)


Packages/com.unity.inputsystem/InputSystem/Utilities/PrimitiveValue.cs
```diff
@@ -450,7 +450,7 @@ public bool ToBoolean(IFormatProvider provider = null)
                 case TypeCode.Single:
                     return !Mathf.Approximately(m_FloatValue, default);
                 case TypeCode.Double:
-                    return NumberHelpers.Approximately(m_DoubleValue, default);
+                    return !NumberHelpers.Approximately(m_DoubleValue, default);
```
ekcoh, line 453: Bonus bug fix?

ekcoh (approved):
LGTM, looks like a good improvement @AlexTyrer

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Use `ProfilerMarker` instead of `Profiler.BeginSample` and `Profiler.EndSample` when appropriate to enable recording of profiling data.
+- Improved performance of disconnected device activation (ISX-1450)
```
ekcoh: I think this could live under ### Fixed since it's not really a changed behavior, right?

ekcoh (approved):

Packages/com.unity.inputsystem/InputSystem/Utilities/PrimitiveValue.cs
AlexTyrer: Yes, I spotted a bug we didn't even know we had!

PR #1990, opened 2024-08-23
NEW: Added InputDeviceMatcher.WithManufacturerContains(string noRegexMatch) to improve DualShockSupport initialization performance

AlexTyrer:
### Description

The **InputDeviceMatcher.WithManufacturer**() can perform a regex match even for simple patterns (eg "Sony.+Entertainment" and "PLAYSTATION(R)3 Controller" are both treated as Regexes) which is surprisingly costly on initialization.

This PR addresses that by adding an API to perform a simple string keyword match on the device manufacturer string,

### Changes made

The new **WithManufacturerContains**(string **noRegexMatch**) API performs a String.**Contains**() check instead looking for the simple pattern in the manufacturer string.

Changed the DualShockSupport for **DualShock4Gamepad** and **DualShock3Gamepad** to avoid regex matching.

These changes save 25ms in **DualShockSupport.Initialize** during player start-up (26ms -> 1ms)

### Testing

Local testing using SONY controllers and Quest VR devices.

### Risk

Previously a controller that reported the manufacturer as "SONY Entertainment" would match - with this change a hypothetical device that reported the manufacturer as "Entertainment Not SONY" would also match. I think this is a risk worth taking.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
@@ -378,7 +378,7 @@ namespace UnityEngine.InputSystem.Layouts
         {
             // String match.
             if (pattern is string str)
-                return str.Contains(value, StringComparison.OrdinalIgnoreCase);
+                return value.Contains(str, StringComparison.OrdinalIgnoreCase);
 
             return false;
         }
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
@@ -150,7 +150,7 @@ namespace UnityEngine.InputSystem.Layouts
         /// <seealso cref="InputDeviceDescription.manufacturer"/>
         public InputDeviceMatcher WithManufacturerContains(string noRegExPattern)
         {
-            return With(kManufacturerContainsKey, noRegExPattern, supportRegex:false);
+            return With(kManufacturerContainsKey, noRegExPattern, supportRegex: false);
         }
 
         /// <summary>
@@ -323,7 +323,7 @@ namespace UnityEngine.InputSystem.Layouts
                 else if (key == kManufacturerContainsKey)
                 {
                     if (string.IsNullOrEmpty(deviceDescription.manufacturer)
-                          || !MatchSinglePropertyContains(pattern, deviceDescription.manufacturer))
+                        || !MatchSinglePropertyContains(pattern, deviceDescription.manufacturer))
                         return 0;
                 }
                 else if (key == kProductKey)
--- a/Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
@@ -145,12 +145,12 @@ namespace UnityEngine.InputSystem.Layouts
         /// <summary>
         /// Add a pattern (simple string) to <see cref="patterns"/> to match a <see cref="InputDeviceDescription.manufacturer"/>.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
```diff
@@ -142,6 +142,17 @@ public InputDeviceMatcher WithManufacturer(string pattern, bool supportRegex = t
...
 
+        /// <summary>
+        /// Add a pattern (simple string) to <see cref="patterns"/> to match a <see cref="InputDeviceDescription.manufacturer"/>.
+        /// </summary>
+        /// <param name="noRegExPattern">String to match - simple keyword search in a device manufacturer string (eg "SONY").</param>
+        /// <returns>The modified device matcher with the added pattern.</returns>
+        /// <seealso cref="InputDeviceDescription.manufacturer"/>
+        public InputDeviceMatcher WithManufacturerContains(string noRegExPattern)
+        {
+            return With(kManufacturerContainsKey, noRegExPattern, supportRegex:false);
```
lyndon-unity: Given that supportRegex:false is passed, do you reallly need kManufacturerContainsKey.
It seems you could just use kManufacturerKey since the object type will remain a string

(but I may have missed something)

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockSupport.cs
```diff
@@ -65,8 +66,9 @@ public static void Initialize()
             InputSystem.RegisterLayoutMatcher<DualShock3GamepadHID>(
                 new InputDeviceMatcher()
                     .WithInterface("HID")
-                    .WithManufacturer("Sony.+Entertainment")
-                    .WithProduct("PLAYSTATION(R)3 Controller"));
+                    .WithManufacturerContains("Sony")
```
lyndon-unity, line 69: I'm wondering it this could be just WithManufacturer("Sony", supportRegex:false) ?

There is a small chance this matches wider now - E.g. Entertainment Sony would match now, but I don't think this is likely to be any false positives.

lyndon-unity (approved):
Generally I approve the change and the savings. 
I've asked a couple of questions though as it looks like a slightly smaller change could give the same result.

Might also be worth adding Lewis Hammond to this PR from the PlayStation team

Packages/com.unity.inputsystem/InputSystem/Devices/InputDeviceMatcher.cs
AlexTyrer: The different key is needed to take the different path through MatchPercentage().

kManufacturerKey -> MatchSingleProperty -> string.Compare(...) == 0  // string/regex needs exactly match
kManufacturerContainsKey -> MatchSinglePropertyContains -> string.Contains(...) // string just needs to be in there somewhere

Packages/com.unity.inputsystem/InputSystem/Plugins/DualShock/DualShockSupport.cs
AlexTyrer: I believe it needs to be an exact match as the code does:

`string.Compare(...) == 0 `

... so we would need to provide the entire manufacturer string - and there seem to be two flavours of Sony manufacturer string:

- "Sony Interactive Entertainment"
- "Sony Computer Entertainment"

... hence the regex in the original code.

Just checked this.

`                    .WithManufacturer("Sony", supportRegex:false)   // test
`
... and some of the tests fail.

This is the reason why I added the 'Contains' thing - an explicit non-regex pattern that just has to be in the manufacturer string.

Issue #392, opened 2019-01-26
Inputmanager when Join behavior is set "When Join Action Is Trigger" pair every device connected to only first user. 


GilbertoBitt:
After adding the InputSystem i add PlayerInput and Create a Prefab with the PlayerInput script then when i press start (Join action) or Keyboard Enter (Join action) my player is instanciate but on input debugger the user has both keyboard and mouse and joystick has pair devices and becouse of that i can't add more users becouse InputManager add all devices to user1.

so in resume i have 3 devices a keyboard, mouse and joystick(xbox one/Xinput), and when keyboard join i want to add mouse to this user but left joystick unpaired, but all devices are being connected to the only one user no matter if i press join on keyboard first or joystick.

**OBS: this only occur when Join Behavior is set to "Join Players When Join Action  Is Trigger"**

[Chiisaii.zip](https://github.com/Unity-Technologies/InputSystem/files/2798482/Chiisaii.zip)

Here a simple video trying to explain the better the problem: [video file](https://drive.google.com/file/d/1a-cIzOwnDVi_9H5IHZY9D2kumu_O80h0/view?usp=sharing)

GilbertoBitt:
The problem only occur when player input is using Invoke Unity Events

hadesfury:
It is happening again : https://forum.unity.com/threads/players-join-when-join-action-is-triggered-not-working-when-using-a-reference-input-action.795537/

Vladyslav25:
Same here! Its still happening, when you use a Reference. When you create a custom Action or custom Actios (without a Ref) it works.

FancyFurret:
I have the exact same problem!

I have found two workarounds for this:
    1) Clone the action and tell the input manager to use the cloned action for the join action:
    ```_playerInputManager.joinAction = new InputActionProperty(_playerInputManager.joinAction.action.Clone());```
    2) Use a completely separate InputActions asset for actions that shouldn't be tied to players. This is what I've done. I've named that asset "GlobalInputActions" and included actions for player joining, opening my debug console, pausing and ui controls.

Hopefully this helps you guys until Unity fixes it!

PR #1991, opened 2024-08-29
FIX: Narrowing scope of ENABLE_CLOUD_SERVICES_ANALYTICS

ekcoh:
### Description

An attempt to revisit https://github.com/Unity-Technologies/InputSystem/pull/1971 narrowing the introduced conditional compile guard to allow player analytics to pass through (like previous to 1.9.0) but gate editor analytics to  when ENABLE_CLOUD_SERVICES_ANALYTICS is defined. See internal ticket ISX-2092 for reasoning. Still confused why this would make a difference for the target platform in the linked issue though, so still uncertain whether this is a correct fix.

### Changes made

Narrowed scope of define introduced to not attempt to send editor analytics when platform not having it enabled.

### Testing

Would need assistance, @lewish-unity if this tweak still resolves the build issue on consoles or if it reintroduces it. 

### Risk

_Please describe the potential risks of your changes for the reviewers._

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -393,8 +393,7 @@ public Action onProjectChange
 
         public void SendAnalytic(InputAnalytics.IInputAnalytic analytic)
         {
-        #if ENABLE_CLOUD_SERVICES_ANALYTICS
-            #if (UNITY_EDITOR)
+            #if (UNITY_EDITOR && ENABLE_CLOUD_SERVICES_ANALYTICS)
```
lyndon-unity: I'm not 100% sure on this

This documentation about  Unity**Engine**.Analytics
https://docs.unity3d.com/ScriptReference/Analytics.Analytics.html 
Say to use ENABLE_CLOUD_SERVICES_ANALYTICS

But the  **Editor**Analytics API doesn't mention it
https://docs.unity3d.com/ScriptReference/EditorAnalytics.html
https://docs.unity3d.com/ScriptReference/EditorAnalytics.RegisterEventWithLimit.html

https://docs.unity3d.com/ScriptReference/EditorAnalytics.html

lyndon-unity: I think this might be around the wrong section. 

I.e. needed for this part #elif (UNITY_ANALYTICS) // Implicitly: !UNITY_EDITOR

lyndon-unity (commented):
Needs testing to verify as I'm not convinced this is correct

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
lyndon-unity: See also https://jira.unity3d.com/browse/ISX-2092

jfreire-unity: Have you tested this for a console platform @ekcoh ?  EDIT: Switch to a console platform in the build settings of a Unity project.
I believe I tried this logic in the past and it didn't work, as the error was raised in the Editor. Because if you're in the Editor and the platform cannot compile Analytics ( ENABLE_CLOUD_SERVICES_ANALYTICS is false), it would still try to call `EditorAnalytics.SendAnalytic(analytic);` and error for projects targeting a Console platform.

lewish-unity (commented):
I don't see the errors that I was getting before the original fix and our Input Automated test suites pass.

However, I would like Consoles QA to have a proper test of this before it is merged as they have better test projects than me.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -405,15 +404,14 @@ public void SendAnalytic(InputAnalytics.IInputAnalytic analytic)
             EditorAnalytics.SendEventWithLimit(info.Name, analytic);
                     #endif // UNITY_INPUT_SYSTEM_ENABLE_ANALYTICS || UNITY_2023_1_OR_NEWER
                 #endif // UNITY_2023_2_OR_NEWER
-            #elif (UNITY_ANALYTICS) // Implicitly: !UNITY_EDITOR
+            #elif (ENABLE_CLOUD_SERVICES_ANALYTICS) // Implicitly: !UNITY_EDITOR && UNITY_ANALYTICS
```
ekcoh, line 407: Follow-up, logged into a console VM and built a basically empty project for target hardware.

Interestingly enough I got these results in editor (build profile set to console profile):

UNITY_EDITOR:True, UNITY_ANALYTICS:False, ENABLE_CLOUD_SERVICES_ANALYTICS:False
UnityEngine.Debug:Log (object)
Symbols:Start () (at Assets/Symbols.cs:26)

And on target device (player):

UNITY_EDITOR:False, UNITY_ANALYTICS:False, ENABLE_CLOUD_SERVICES_ANALYTICS:False
UnityEngine.Debug:Log (object)
Symbols:Start () (at Assets/Symbols.cs:26)

Additionally, I removed the ENABLE_CLOUD_SERVICES_ANALYTICS check completely from NativeInputRuntime.cs  and I could still build for the console without issues, both in editor and player based on how it's currently defined. This confused me since the original fix states this was not possible?

I then removed all checks/code but the Analytics.XXX part and that fails to build in editor in this case. Hence, guarding this with ENABLE_CLOUD_SERVICES_ANALYTICS makes sense.
Note: Previously guarded with UNITY_ANALYTICS which also happen to be false and hence why observation in paragraph above worked to compile as-is.

I then removed all but the EditorAnalytics part. This compiles/runs fine in editor and is excluded in player since guarded with UNITY_EDITOR.

Hence it boils down to ENABLE_CLOUD_SERVICES_ANALYTICS should be used for access to Analytics, i.e. the analytics module. However, whether UNITY_ANALYTICS should be checked or not remains an enigma to me. Probably it should since we had it there previously?

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -393,7 +393,6 @@ public Action onProjectChange
```
ekcoh, line 393: Question remains whether UNITY_ANALYTICS should be a precondition here...

ekcoh: On line #392

ekcoh: Based on git history I would just assume it should, e.g. before I started to edit this it looked like this (dates back to 2018 when it was introduced) https://github.com/Unity-Technologies/InputSystem/commit/6df6e1178809c515f5cdac0e5416d067db200206:

```
        public void RegisterAnalyticsEvent(string name, int maxPerHour, int maxPropertiesPerEvent)
        {
            #if UNITY_ANALYTICS
            const string vendorKey = "unity.input";
            #if UNITY_EDITOR
            EditorAnalytics.RegisterEventWithLimit(name, maxPerHour, maxPropertiesPerEvent, vendorKey);
            #else
            Analytics.Analytics.RegisterEvent(name, maxPerHour, maxPropertiesPerEvent, vendorKey);
            #endif // UNITY_EDITOR
            #endif // UNITY_ANALYTICS
        }

        public void SendAnalyticsEvent(string name, object data)
        {
            #if UNITY_ANALYTICS
            #if UNITY_EDITOR
            EditorAnalytics.SendEventWithLimit(name, data);
            #else
            Analytics.Analytics.SendEvent(name, data);
            #endif // UNITY_EDITOR
            #endif // UNITY_ANALYTICS
        }
```

lyndon-unity (approved):
Looks simpler and tidier and what I would expect.

Not sure the CHANGELOG should be fix in the same commit but I don't have a problem with it either.

ekcoh:
@lewish-unity do you have some recommendations for console QA to verify the updated fix? I am pretty confident this is how it should be now.

jfreire-unity (approved):

Pauliusd01:
Hi @cristian-murarescu could you help us with the playstation platform testing? Just need to see if there's any issues building/running its player

cristian-murarescu (approved):
Tested PS4 & PS5 player builds & runs successfully.

Pauliusd01 (approved):
LGTM, here's some things I checked:
-  tried deploying to Xbox VM without issues
-  checked that Analytics debugger shows same events as before when interacting with input related things (although I have to note that they seem kinda sparse in both versions, I don't know if I needed to enable anything extra but there wasn't anything for setting bindings, changing control schemes, etc. Things that were logged were context menu actions, interaction/processor adding and file saving)
- checked basic actions window functionality and testing scenes on Windows

ekcoh:
> LGTM, here's some things I checked:
> 
> * tried deploying to Xbox VM without issues
> * checked that Analytics debugger shows same events as before when interacting with input related things (although I have to note that they seem kinda sparse in both versions, I don't know if I needed to enable anything extra but there wasn't anything for setting bindings, changing control schemes, etc. Things that were logged were context menu actions, interaction/processor adding and file saving)
> * checked basic actions window functionality and testing scenes on Windows

Hmmm that sounds odd, I will do a final check myself on desktop before merging then.

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
ekcoh: Sorry that I did not respond to this promptly, the current version of this I tested on one console platform in editor and player builds.

ekcoh:
@Pauliusd01 I checked this branch with Analytics debugger but couldn't see anything that looked odd to me.

PR #2003, opened 2024-09-12
NEW: ISX-2099 InputAction code-setup (authoring) analytics

ekcoh:
### Description

This PR adds play-mode editor analytics for tracking input action code-setup paths where users are using code/APIs to setup actions instead of assets.

### Changes made

Added basic per-function analytics that also get aggregated into a simple boolean expression whether code setup paths are being used or not.

I personally think this is quite brittle and may generate false positives since it require manual suppression to avoid it. I see no current way around it. Notably in `OnScreenStick` and in `TrackedPoseDriver`. All other (current) references to this API surface seems to be from test cases where the analytics are anyway stubbed away.

### Testing

Added test case to track analytics being recorded similar to existing analytics unit tests.

Tested with Analytics SDK locally. Also will be confirmed with analytics team in backend before merged.

### Risk

Small, main risk is recording false positives indirectly through internal calls (in the future). Might be that this is heavily used on XR side? If it is it might require suppression there.

### Checklist

Before review:

- [X] Changelog entry added.
    - Functional tests `Analytics_ShouldReportCodeAuthoringAnalytic` added `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Assets/Tests/InputSystem/CoreTests_Analytics.cs
```diff
@@ -663,6 +663,55 @@ public void Analytics_ShouldReportPlayerInputManagerData()
...
 
+    [Test]
+    [Category("Analytics")]
+    public void Analytics_ShouldReportCodeAuthoringAnalytic()
+    {
+        CollectAnalytics(InputActionCodeAuthoringAnalytic.kEventName);
+
+        // NOTE: We do not want to trigger entering/exiting play-mode for this small data-sanity check
+        //       so just stick to triggering it explicitly. A better test would have been an editor test
+        //       going in and out of play-mode for real but not clear if this is really possible.
```
lyndon-unity, line 674: Yes this is possible - there are other tests that do this E.g. texture streaming tests in trunk

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputActionCodeAuthoringAnalytic.cs
```diff
@@ -0,0 +1,168 @@
...
+            ChangeCompositeBinding = 3,
+            Rename = 4,
+            AddControlScheme = 5,
+            RemoveControlScheme = 6,
+            ControlSchemeWithBindingGroup = 7,
+            ControlSchemeWithDevice = 8,
+            ControlSchemeWithRequiredDevice = 9,
+            ControlSchemeWithOptionalDevice = 10,
+            ControlSchemeOrWithRequiredDevice = 11,
+            ControlSchemeOrWithOptionalDevice = 12
```
lyndon-unity: Some of these seem more granular than needed but I don't have an issue with them being present.

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputBuildAnalytic.cs
```diff
@@ -31,7 +31,7 @@ public InputBuildAnalytic(BuildReport buildReport)
         public InputAnalytics.InputAnalyticInfo info =>
             new InputAnalytics.InputAnalyticInfo(kEventName, kMaxEventsPerHour, kMaxNumberOfElements);
 
-#if UNITY_EDITOR && UNITY_2023_2_OR_NEWER
+#if UNITY_2023_2_OR_NEWER
```
lyndon-unity, line 34: I see UNITY_EDITOR check is higher up the file so this looks fine an easier to read

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -91,20 +91,32 @@ private void Start()
                     if (m_PointerDownAction == null)
                         m_PointerDownAction = new InputAction();
 
+                    #if UNITY_EDITOR
+                    InputActionCodeAuthoringAnalytic.suppress = true;
```
lyndon-unity: I think this makes sense to exclude so looks good

lyndon-unity (approved):
Looks good to me.

Assets/Tests/InputSystem/CoreTests_Analytics.cs
ekcoh: I would love to learn how to do that properly. I tried it but when I programmatically exited play-mode the test aborted. We state its possible in the docs, but couldn't see any reference to how: https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/edit-mode-vs-play-mode-tests.html 

`Note: You can also control entering and exiting Play Mode from your Edit Mode test. This allow your test to make changes before entering Play Mode.`

EditorApplication.playMode doesn't seem to do the job for me....

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputActionCodeAuthoringAnalytic.cs
ekcoh: Yes they are, but while I was going at it I had to carefully make sure each one was properly mapped through all the overloads and calls between them so I thought it made sense to keep the detail and throw it away in the end instead in case we need it later it would be easy to switch.

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputBuildAnalytic.cs
ekcoh: Yes it was unrelated but something I stumbled on so fixed it

Assets/Tests/InputSystem/CoreTests_Analytics.cs
lyndon-unity: In and edit and play mode test You should be able to use code like this:

yield return new EnterPlayMode();
yield return new ExitPlayMode();


See this for more details internally:
https://github.cds.internal.unity3d.com/unity/com.unity.memoryprofiler/blob/46687aa966bea49d093041968fb6d40cc8fd39f4/com.unity.memoryprofiler/Tests/Editor/SceneRootsAndAssetBundles/ReferenceToTests.cs#L58

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputActionCodeAuthoringAnalytic.cs
```diff
@@ -0,0 +1,168 @@
...
+        private static readonly Action<PlayModeStateChange> PlayModeChanged = OnPlayModeStateChange;
+
+        // Note: Internal visibility to simplify unit testing
+        internal static void OnPlayModeStateChange(PlayModeStateChange change)
+        {
+            Debug.Log("Play mode state change: " + change);
+            if (change == PlayModeStateChange.ExitingEditMode)
+            {
+                // Reset all counters when exiting edit mode
+                for (var i = 0; i < m_Counters.Length; ++i)
```
ritamerkl: Array.Clear(m_Counters, 0, m_Counters.Length)?

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputActionCodeAuthoringAnalytic.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            try
+            {
+                // Determine aggregated perspective, i.e. was "any" code-authoring API used
+                var usedCodeAuthoringDuringPlayMode = false;
+                for (var i = 0; i < m_Counters.Length; ++i)
+                {
+                    if (m_Counters[i] > 0)
+                    {
+                        usedCodeAuthoringDuringPlayMode = true;
```
ritamerkl: setting this bool in Register() would save the cost of going through the array here.

ritamerkl (approved):
Nice, only minor comments.

Pauliusd01:
Will check this today/tomorrow morning

Pauliusd01 (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputActionCodeAuthoringAnalytic.cs
ekcoh: Of course, thanks. Now updated.

Packages/com.unity.inputsystem/InputSystem/Editor/Analytics/InputActionCodeAuthoringAnalytic.cs
ekcoh: Yes that is a valid point. However, I created the counters to allow tracking in more detail at a later point. Hence I intend to leave this as-is for now. If I would change it, it would suggest removing the counters, which would suggest all registrations should not be specific etc. This should not have much cost to bother about?

ekcoh: (One reason for that decision was that it was a very manual process to go through all method calls and stacked method calls to established the detailed information about what is being used).

ritamerkl: Okay, that makes sense

Assets/Tests/InputSystem/CoreTests_Analytics.cs
ekcoh: I will land this as-is and file a separate task on this. if ok @lyndon-unity ? There are many places to change to that / correct triggers.

ekcoh: A ticket have now been filed.

PR #2007, opened 2024-09-18
CHANGE: Added message to notify the user if a PlayerInputComponent fails to bind a control scheme (ISXB-1020)

lyndon-unity:
### Description

Added message to notify the user if a PlayerInputComponent fails to bind a control scheme (ISXB-1020)
This can happen if there are multiple game objects with PlayerInputComponent's and not enough devices to support them

### Changes made

Added an error message in the case when the list of available control schemes has been exhausted

### Testing

Ran on the test case in the bug report and it fires a message on PC when only a keyboard/mouse plugged in.
(Does not fire of a touchscreen is also connected)

Ran all the unit tests and they all pass

### Risk

Potentially there is a use case I missed which doesn't warrant the error message

### Checklist

Before review:

- [X] Changelog entry added.
- Ran existing tests
- No new docs 

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -16,6 +16,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Changed
 - Renamed editor Resources directories to PackageResources to fix package validation warnings.
+- PlayerInput component now warns if the system cannot find matching control scheme, which can occur if all control schemes already paired (e.g. to other game objects with PlayerInput components) [ISXB-1020](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1020)
 
 ## [1.11.0] - 2024-09-10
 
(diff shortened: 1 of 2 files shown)


Pauliusd01 (changes requested):
I'm still reproducing it with no error showing up, do I need to do anything extra or is this a proper issue?

https://github.com/user-attachments/assets/7a033914-274f-4eea-8c55-04baa3ce0729

lyndon-unity:
> I'm still reproducing it with no error showing up, do I need to do anything extra or is this a proper issue?
> 
>  1726740435.inputTest_-_GameplayScene_-_Windows._Mac._Linux_-_.mp4

That's the expected result when touchscreen and keyboard present, the error only shows when just keyboard present (i.e. only one valid control scheme that can be bound)

Pauliusd01:
Can you change it so it says "(all control schemes are already paired to matching devices)" or "(all control schemes are already paired to available devices)" to make it a bit more clear on what the issue is? Since I saw plenty of schemes to go around and it wasn't immediately obvious to me why I'd get this error as a user

lyndon-unity:
> Can you change it so it says "(all control schemes are already paired to matching devices)" or "(all control schemes are already paired to available devices)" to make it a bit more clear on what the issue is? Since I saw plenty of schemes to go around and it wasn't immediately obvious to me why I'd get this error as a user

Done.
I've also switched from error to warning to match feedback from Hakan on a port of this change.

Pauliusd01 (approved):

ekcoh (approved):

PR #2012, opened 2024-09-24
FIX: The button is not clickable when using Windows touch simulation inputs (ISXB-483)

jfreire-unity:
### Description

Fixes [ISXB-483](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-483)

Simulated touches were not processed using Touch simulation (Input Debug > Options > Simulate Touch Input) with the PlayerInput component. This PR addresses that.

image

### Changes made

~Made sure the simulated touchscreen is paired with the PlayerInput users' devices.~
TouchSimulation queues input events instead of changing state. This triggers device pairing and control scheme switching of PlayerInput.
And also, makes the events show up in Input Debugger.

https://github.com/user-attachments/assets/2862bf46-f824-4c4d-82e0-921742b7d3fe

### Testing

Tested:
- Toggling the Simulate Touch Input option in PlayMode
- Removing the simulated touchscreen device.
- Toggling the PlayerInput component in the project scene

**Careful** :warning: : For the specific project of the bug report, the actions that have touch bindings are not set to the Touch control scheme. So the simulated touchscreen/real touchscreen will not work out of the box. The InputActionAsset actions for touchscreen devices need to be changed to included "Touch" control scheme:
image

### Risk

Minimal regression risk. I validated that queuing these simulated events still occurs in the same frame where they were generated from Mouse/Pen.

### Checklist

Before review:

- [x] Changelog entry added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed "MissingReferenceException" errors when closing an in-game dropdown field [ISXB-1081](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1081).
 - Fixed potential crash on Mac when using stale references to deleted InputDevice objects [ISXB-606](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-606).
 - Fixed conditional compilation for non-editor analytics on platforms not enabling analytics.
+- Fixed simulated touch input not working with PlayerInput component [ISXB-483](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-483)
```
ekcoh: Minor: Missing period at end.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs
```diff
@@ -264,6 +273,28 @@ protected void OnEnable()
 
             InputSystem.onDeviceChange += m_OnDeviceChange;
             InputSystem.onEvent += m_OnEvent;
+            
+            // In case there's a PlayerInput component in the scene, we want to pair the simulated touchscreen to
+            // the PlayerInput user. The touchscreen device does not queue events so we need to pair it to the user.
+            m_PlayerInput = PlayerInput.s_AllActivePlayersCount > 0 ? PlayerInput.s_AllActivePlayers[0] : null;
```
ekcoh: I find it interesting in general that the existing TouchSimulation is modifying state memory instead of acting as a proper touch device and enqueuing events. Did you by any chance see anything in implementation why that would be the case?

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs
```diff
@@ -264,6 +273,28 @@ protected void OnEnable()
...
+            // In case there's a PlayerInput component in the scene, we want to pair the simulated touchscreen to
+            // the PlayerInput user. The touchscreen device does not queue events so we need to pair it to the user.
+            m_PlayerInput = PlayerInput.s_AllActivePlayersCount > 0 ? PlayerInput.s_AllActivePlayers[0] : null; 
+            if (m_PlayerInput)
+            {
+                TryPerformUserPairingOfSimulatedTouchscreen();
+                InputUser.onChange += PairSimulatedTouchscreenOnChange;
+            }
+        }
+        private void PairSimulatedTouchscreenOnChange(InputUser user, InputUserChange change, InputDevice device)
```
ekcoh: Is the reported issue also an issue when not using control schemes or does it only surface when the control schemes are present? Basically what would the intended behavior be in case a user was using a real mouse and a real touch screen at the same time? I would expect ping ponging effects on pointer. So maybe this is different since one want to avoid that and the current system do not support reasoning about sources of synthesised events...

ekcoh (commented):
Code changes look good, I think I need to check this again to figure out how this is intended to work. I think you need to run formatter on the code.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs
jfreire-unity: There are comments with regards [to that](https://github.cds.internal.unity3d.com/unity/neutron/blob/f5445a31e8f52f9b0e49c0e0825f9c9806bda95f/Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs#L15-L24), but there's a point that [suggests there might be extra work](https://github.cds.internal.unity3d.com/unity/neutron/blob/f5445a31e8f52f9b0e49c0e0825f9c9806bda95f/Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs#L132-L133) to deal with queueing events from within `InputSystem.onEvent`.

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs
jfreire-unity: >Is the reported issue also an issue when not using control schemes or does it only surface when the control schemes are present?

I didn't test this without control schemes, as they essentially control the PlayerInput use device pairing mechanism (which uses input events). And the problem is pairing mechanism was not taking into account the simulated touchscreen, because it doesn't generate input events.

> Basically what would the intended behavior be in case a user was using a real mouse and a real touch screen at the same time? I would expect ping ponging effects on pointer

Yes that would be the behavior. But for this case, mouse (or pen) doesn't exist as its driving the simulated touchscreen. Also the ping pong only ocurrs if "Auto-switch" in PlayerInput is selected. Unless I'm misunderstanding something.

I'm confused with what you mean with this. Can you clarify?
> So maybe this is different since one want to avoid that and the current system do not support reasoning about sources of synthesised events...

Packages/com.unity.inputsystem/InputSystem/Plugins/EnhancedTouch/TouchSimulation.cs
jfreire-unity: I think a follow-up task to make TouchSimulated to use input events could be added to the backlog as it would make it much more "real" than it currently is. I'll deal with that.

ekcoh (approved):
LGTM. Side-note: Its extremely odd that this primaryTouch is not part of the array, but that is unrelated to your PR

Pauliusd01:
@jfreire-unity is this ready to look at? There's a lot of CI failures

jfreire-unity:
> @jfreire-unity is this ready to look at? There's a lot of CI failures

Not yet. I'll make sure to ping you when it's done. Dealing with CI failures first

Pauliusd01:
Will check today

Pauliusd01 (changes requested):
Updating status, had an issue with simulated multi touch events staying active. More details in DM

Pauliusd01 (approved):
LGTM, checked real and simulated touches on Windows and Android in a couple of touch samples as well as the user's project (after fixing their input actions) Did not test pen support though as I do not have the device for it

PR #2017, opened 2024-09-27
FIX: Wrong mapping of Xbox Series S|X and Xbox One wireless controllers "View" button on macOS

jfreire-unity:
### Description

Fixes [ISXB-385](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-385)

The problem in a nutshell:
image

### Changes made

Set the correct bit field that maps to the equivalent "Select" button of Xbox controllers, the "View" button.

### Testing

I tested with both Xbox wireless (Bluetooth capable) controllers. Both the Xbox One and Xbox Series S|X models.
It feels like the previous value didn't work for Xbox One wireless controllers so I'm not sure why that value was set in the first place. But maybe I just couldn't test with that device.
If there's a Xbox wireless controller that now doesn't work due to this change, let me know and I'll create a new class and state struct for it.

### Risk

Untested Xbox wireless controllers stop working. 

### Checklist

Before review:

- [x] Changelog entry added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
@@ -111,7 +111,7 @@ namespace UnityEngine.InputSystem.XInput.LowLevel
         public enum Button
         {
             Start = 11,
-            Select = 16,
+            Select = 10,
             LeftThumbstickPress = 13,
             RightThumbstickPress = 14,
             LeftShoulder = 6,
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -14,6 +14,9 @@ however, it has to be formatted properly to pass verification tests.
 - Added the display of the device flag `CanRunInBackground` in device debug view.
 - Added analytics for programmatic `InputAction` setup via `InputActionSetupExtensions` when exiting play-mode.
 
+### Fixed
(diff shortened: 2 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
```diff
@@ -111,7 +111,7 @@ internal struct XInputControllerWirelessOSXState : IInputStateTypeInfo
         public enum Button
         {
             Start = 11,
-            Select = 16,
+            Select = 10,
```
lyndon-unity, line 114: Looks like a very targetted fix, just needs some QA to verify.

lyndon-unity (approved):
If QA confirm its working I'm happy to approve

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
lyndon-unity: I did wonder - Should we still expose the 'share' button too?
Not needed to fix the issue but seems a little odd we don't expose it.

bmalrat: According to https://github.com/mdqinc/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt#L1094 it seems that it depends on the VID/PID of the device.
we might do some mapping in the native part or in c# if we can switch with the pid/vid

Pauliusd01 (approved):
LGTM, checked Series X controller on Mac and & Windows for good measure in Editor and Playmode. The select button and others work correctly. Though I only had the newer xbox controller and didn't check xbox one

PR #2019, opened 2024-09-30
FIX: MakeEscapedJsonString now null-checks inputs. Empty strings also now bypass an allocation.

AlexTyrer:
### Description

Most device decriptor fields are null-checked in ComparePropertyToDeviceDescriptor, however the addition of MakeEscapedJsonString meant that 'capabilities' was not being checked anymore, and so could throw a null-ref exception if a device didn't fill this field.

### Changes made

This change adds a null/empty string check to MakeEscapedJsonString to mirror the old behaviour.
Additionally this'll also now mean any devices with an empty non-null string can skip a minor StringBuilder allocation.

### Testing

Local testing with gamepads to see that the resulting comparisons in ComparePropertyToDeviceDescriptor still work correctly. This was tested between null, empty, and valid json string inputs.

A new unit test has been added too to exercise the null-input path.
Unit tests were also ran.

### Risk

Minor difference in construction of a default JsonString (with a non-null text field).

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -4162,13 +4162,40 @@ partial class CoreTests
         recorder.CollectFromAllThreads();
 #endif
 
-        // We expect a single allocation for each call to ReportNewInputDevice when there is one disconnected device
-        //
-        int numberOfRepeats = 2;
-        int numberOfDisconnectedDevices = 1;
-        int numberOfCallsToReportNewInputDevicePerRun = 2;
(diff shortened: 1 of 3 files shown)


AlexTyrer:
Thanks to @stuartbrown-unity for the fix!

lyndon-unity (approved):
Looks safe - good to see it fixed the allocation count part too.

PR #2027, opened 2024-10-10
NEW: Add two performance tests measuring frame times

stefanunity:
### Description

Add 2 new performance tests measuring frame times of all profiler markers found in the package.

### Changes made

- one test where nothing at all happens
- one test with keyboard and 4kHz mouse events

### Testing

New tests run and seem stable.

### Risk

None, just new tests

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -1100,4 +1102,106 @@ public void Performance_OptimizedControls_ReadingPose4kTimes(OptimizationTestTyp
...
+                PressAndRelease(keyboard.sKey, queueEventOnly: true);
+                PressAndRelease(keyboard.dKey, queueEventOnly: true);
+
+                PressAndRelease(keyboard.leftShiftKey, queueEventOnly: true);
+
+                PressAndRelease(keyboard.spaceKey, queueEventOnly: true);
+
+                Click(mouse.leftButton, queueEventOnly: true);
+
+                //mouse movements for higher polling mice (66*60 ~ 4kHz)
```
ekcoh: I think the for loop here should match the achievable event rate at the OS. So 66 here like you are using means 66 Hz. I think 500-1000 Hz is common as hardware sampling frequency. The maximum I have measured on my machine with a 8 kHZ mouse have been 5700 mouse events per second. Which means 95 mouse events per frame @60 Hz. So that would map to 95 here in this for. The outer for loop just means, lest simulate 500 frames.

Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -1100,4 +1102,106 @@ public void Performance_OptimizedControls_ReadingPose4kTimes(OptimizationTestTyp
...
+
+        sprintAction.performed += context => {
+            performedCallCount++;
+        };
+
+        using (Measure.ProfilerMarkers(markers))
+        {
+            for (int i = 0; i < 500; ++i)
+            {
+                PressAndRelease(keyboard.wKey, queueEventOnly: true);
```
ekcoh: On the other hand (compared to comment below), pressing W, A, S and D as well as shift and space and mouse button at a constant rate of e.g. 60 times per second is not very realistic? I would suggest to use a modulo operation here with a switch to do different input for each frame depending on a modulo computation, e.g. if (i % 60 == 0) PressSomething() 
would mean I press something every 60th frame, which means once a second if playing at 60 Hz. If benchmark is targeting something specific (microbenchmark) it may make sense to stick to large data amount though.

Assets/Tests/InputSystem/CorePerformanceTests.cs
stefanunity: https://github.com/Unity-Technologies/InputSystem/pull/2027/commits/79525e045620aef6636af2f751375e40b02f6862

Assets/Tests/InputSystem/CorePerformanceTests.cs
stefanunity: https://github.com/Unity-Technologies/InputSystem/pull/2027/commits/79525e045620aef6636af2f751375e40b02f6862

Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -1100,4 +1102,110 @@ public void Performance_OptimizedControls_ReadingPose4kTimes(OptimizationTestTyp
...
 #endif
+
+    #if UNITY_2022_3_OR_NEWER
+    [PrebuildSetup(typeof(ProjectWideActionsBuildSetup))]
+    [PostBuildCleanup(typeof(ProjectWideActionsBuildSetup))]
+    [UnityTest, Performance]
+    [Category("Performance")]
+    public IEnumerator Performance_MeasureInputSystemFrameTimeWithProfilerMarkers_FPS()
+    {
+        string[] markers =
```
lyndon-unity: Perhaps move this to global so can share between the two tests

lyndon-unity (approved):

ekcoh (approved):

PR #2033, opened 2024-10-18
FIX: PointerEventData.pointerId is the same when simultaneously releasing and then pressing with another finger (ISXB-845)

jfreire-unity:
### Description

Fix for https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-845

A new pointer was not being created when we released and pressed a finger in the same frame, even though a new touchId was being produced by the native side.

This PR fixes that by looking into pointerIds instead of cached pointer controls. The cached pointer controls wouldn't reflect the control with the correct touchId when we release and press a touch control in the same frame.

### Testing status & QA

Please test the bug project with and without this PR on a mobile device (Android or iOS)

For example, to reproduce the issue (more instructions in the bug report):
1. Press and hold with one finger on the left panel (left panel is green)
2. Quickly release the finger and press it with another on the right panel
3. The right will not stay green. This means OnPointerDown wasn't called for this pointer ❌

With the PR, in this example, step 3 should show the right panel green ✅

### Overall Product Risks

Low

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.cs
@@ -1719,7 +1719,6 @@ internal partial class UITests : CoreTestsFixture
 
         scene.leftChildReceiver.events.Clear();
 
-        Assert.That(scene.eventSystem.IsPointerOverGameObject(), Is.False);
         Assert.That(scene.eventSystem.IsPointerOverGameObject(1), Is.False);
         Assert.That(scene.eventSystem.IsPointerOverGameObject(2), Is.False);
 
@@ -1730,8 +1729,8 @@ internal partial class UITests : CoreTestsFixture
 
         var pointerIdTouch1 = ExtendedPointerEventData.MakePointerIdForTouch(touchScreen.deviceId, 1);
         var pointerIdTouch2 = ExtendedPointerEventData.MakePointerIdForTouch(touchScreen.deviceId, 2);
+        var pointerIdTouch3 = ExtendedPointerEventData.MakePointerIdForTouch(touchScreen.deviceId, 3);
 
-        Assert.That(scene.eventSystem.IsPointerOverGameObject(), Is.True);
         Assert.That(scene.eventSystem.IsPointerOverGameObject(1), Is.True);
         Assert.That(scene.eventSystem.IsPointerOverGameObject(2), Is.False);
 
@@ -1749,7 +1748,9 @@ internal partial class UITests : CoreTestsFixture
                 .Matches((UICallbackReceiver.Event e) => e.pointerData.pointerId == pointerIdTouch1).And
                 .Matches((UICallbackReceiver.Event e) => e.pointerData.pointerType == UIPointerType.Touch).And
                 .Matches((UICallbackReceiver.Event e) => e.pointerData.position == firstPosition));
+
         Assert.That(scene.rightChildReceiver.events, Is.Empty);
+        Assert.That(scene.uiModule.m_PointerStates.length, Is.EqualTo(1));
 
         scene.leftChildReceiver.events.Clear();
         scene.rightChildReceiver.events.Clear();
@@ -1758,12 +1759,15 @@ internal partial class UITests : CoreTestsFixture
         var secondPosition = scene.From640x480ToScreen(350, 200);
         EndTouch(1, firstPosition);
         BeginTouch(2, secondPosition);
+        BeginTouch(3, secondPosition);
+        MoveTouch(2, secondPosition);
         yield return null;
 
-        Assert.That(scene.eventSystem.IsPointerOverGameObject(), Is.True);
         Assert.That(scene.eventSystem.IsPointerOverGameObject(1), Is.True);
         Assert.That(scene.eventSystem.IsPointerOverGameObject(2), Is.True);
+        Assert.That(scene.eventSystem.IsPointerOverGameObject(3), Is.True);
 
+        // Pointer 1
         Assert.That(scene.leftChildReceiver.events,
             Has.Exactly(1).With.Property("type").EqualTo(EventType.PointerUp).And
                 .Matches((UICallbackReceiver.Event e) => e.pointerData.device == touchScreen).And
@@ -1772,6 +1776,7 @@ internal partial class UITests : CoreTestsFixture
                 .Matches((UICallbackReceiver.Event e) => e.pointerData.pointerType == UIPointerType.Touch).And
                 .Matches((UICallbackReceiver.Event e) => e.pointerData.position == firstPosition));
 
+        // Pointer 2
         Assert.That(scene.rightChildReceiver.events,
             Has.Exactly(1).With.Property("type").EqualTo(EventType.PointerEnter).And
                 .Matches((UICallbackReceiver.Event e) => e.pointerData.device == touchScreen).And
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1729,16 +1729,33 @@ private int GetPointerStateIndexFor(InputControl control, bool createIfNotExists
...
+                // To accomodate for cases when touch is released and pressed in the same frame, the pointer
+                // state touchId needs to be checked against the cached touch control touchId (in m_PointerTouchControls)
+                // This is because the pointer state of the released touch (unpressed) still exists.
+                // If touchIDs are different, a new pointer should be allocated. Otherwise, a cached reference will
+                // be used.
+                var pointerTouchControl = m_PointerTouchControls[touchControlIndex];
+                ref var pointerState = ref GetPointerStateForIndex(touchControlIndex);
+
+                if (!(pointerTouchControl is TouchControl) ||
+                    (pointerTouchControl is TouchControl &&
```
bmalrat: since `!(pointerTouchControl is TouchControl)` is false `pointerTouchControl is TouchControl` is always true at that stage an doesn't need to be retested.

you can use `pointerTouchControl is TouchControl touchControl)` form to limit the number of casting

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed "AnalyticsResult" errors on consoles [ISXB-1107]
 - Fixed wrong `Display Index` value for touchscreen events.[ISXB-1101](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1101)
 - Fixed event handling when using Fixed Update processing where WasPressedThisFrame could appear to true for consecutive frames [ISXB-1006](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1006)
+- Fixed pointerId staying the same when simultaneously releasing and then pressing in the same frame on mobile using touch [ISXB-1006](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-845)
```
ritamerkl: Missing point at the end.

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -1693,6 +1693,153 @@ public IEnumerator UI_CanDriveUIFromMultipleTouches()
         Assert.That(scene.leftChildReceiver.events, Is.Empty);
     }
 
+    // https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-845
+    // This tests that we can release and press touches on the same frame with the expected events and touchIds.
+    [UnityTest]
+    [Category("UI")]
+    public IEnumerator UI_CanReleaseAndPressTouchesOnSameFrame()
```
ritamerkl, line 1700: Nice!

ritamerkl (approved):
I didn't test it, looks good from what I see

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1729,16 +1729,33 @@ private int GetPointerStateIndexFor(InputControl control, bool createIfNotExists
...
+                     pointerState.eventData.touchId == ((TouchControl)pointerTouchControl).touchId.value))
```
bmalrat: should we search in m_PointerTouchControls the touchid to prevent creating a new one?

Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -1693,6 +1693,153 @@ public IEnumerator UI_CanDriveUIFromMultipleTouches()
...
+        scene.uiModule.leftClick = InputActionReference.Create(leftClickAction);
+
+        map.Enable();
+
+        yield return null;
+
+        scene.leftChildReceiver.events.Clear();
+
+        Assert.That(scene.eventSystem.IsPointerOverGameObject(), Is.False);
+        Assert.That(scene.eventSystem.IsPointerOverGameObject(touchScreen.deviceId), Is.False);
```
bmalrat: not sure that this assert is valid `touchScreen.deviceId` doesn't represent a pointer id.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jfreire-unity: Ah good point, thanks for the simplification 👍

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jfreire-unity: I think I didn't understand what you mean. Do you mean prevent creating a `pointerTouchControl` ? 
We also don't create touchIDs here, the native side provides them.
Not sure if any of this helps but feel free to clarify.

Assets/Tests/InputSystem/Plugins/UITests.cs
jfreire-unity: Removed. It was copied from previous tests

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
bmalrat: if the new constrain doesn't match, the function will call `AllocatePointer` for a new pointer with the same `control` but it will be never accessible since the previous one will always be hit before.
You might tune/add a new function for `m_PointerTouchControls.IndexOfReference(controlParent);` to search for the right one.

jfreire-unity: >  the function will call AllocatePointer for a new pointer with the same control but it will be never accessible since the previous one will always be hit before.

This is what we want: to allocate a new pointer when the touchId on the control and in the `pointerState.eventData` are different. 
An example: When we release and press a touch in the same frame this happens:
- Release: m_PointerTouchControls[0] touchId = 1, pointerState.eventData.touchId = 1;
- Press: m_PointerTouchControls[0] touchId = 2, pointerState.evenData.touchId = 1;

Ideally, this check should return -1 ` m_PointerTouchControls.IndexOfReference(controlParent);` but it doesn't since we're not storing copies of InputControl objects (it seems). 
The reason seems to be that the control reference in m_PointerTouchControls[0] is always referencing the "touch0" control which always seems to have [the same position in memory](https://github.com/Unity-Technologies/InputSystem/blob/2125c295976ad25cdc69bd67cd4276936fd13510/Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs#L523) and its content gets updated.

If we don't allocate a new pointer, basically we will assume there's no other pointer and that we're still "updating" the one with touchId = 1 which is not correct and generates the bug this PR is fixing.

Allocating a new pointer will create a new entry in `m_PointerTouchControls` with the same touchControl for 1 frame, as when we call `RemovePointerAtIndex` in the next frame, the entry for touchId = 1 will be removed.

bmalrat: If you change a lit bit the test with a third finger on the second button you will have some remaining pointers and at.
```
`       // Release left object and Touch right object on the same frame.
        var secondPosition = scene.From640x480ToScreen(350, 200);
        EndTouch(1, firstPosition);
        BeginTouch(2, secondPosition);
        BeginTouch(3, secondPosition);
        MoveTouch(2, secondPosition);`
```
The interleave of 2/3/2 is important to create extra pointer not 'saved' by
https://github.com/Unity-Technologies/InputSystem/pull/2033/files#diff-3b5d28ba317b2b6240ae00166a94786980b2126bde8a6d830b6c962581b44583R1786 
```
        // End second touch.
EndTouch(2, secondPosition);
EndTouch(3, secondPosition);
yield return null;
```

stefanunity (approved):
checked with 22LTS + Pixel 6

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jfreire-unity: Now I understand the impact, thanks!

image

On the MoveTouch(2),  `m_PointerTouchControls.IndexOfReference(controlParent);`  will return index 0, which has the wrong PointerId later on, and will allocate an irrelevant pointer. Good catch 👍 
I'll fix this

jfreire-unity: Fixed

bmalrat (approved):

PR #2038, opened 2024-10-25
NEW: Input System Profiler Module

ekcoh:
### Description

Made a draft implementation of a Profiler Module for the Input System based on lack of observability to work as an aid in understanding system mechanics. Hopefully could work as an embryo of something more refined if considered a good idea.

Screenshot 2024-10-25 at 13 36 23

### Testing status & QA

Just a prototype at this point. Try it out and suggest or contribute to this PR to make it more useful.

### Overall Product Risks

Some more work likely need to be done to InputManager.OnUpdate to make that method cleaner and this work more reliably. 

Might not be possible to merge unless minimum Unity version requirement is changed to 2020.1. (Profiler package dependency)

- Complexity: Small
- Halo Effect: Medium

### Comments to reviewers

At this point mainly looking for feedback and suggestions since this was unplanned work.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputSystemProfilerModule.cs
```diff
@@ -0,0 +1,148 @@
...
+                var eventProcessingTime = ProfilerDriver.GetFormattedCounterValue(selectedFrameIndex,
+                    InputStatistics.Category.Name, InputStatistics.EventProcessingTimeName);
+                var stateBufferSizeBytes = ProfilerDriver.GetFormattedCounterValue(selectedFrameIndex,
+                    InputStatistics.Category.Name, InputStatistics.StateBufferSizeBytesName);
+                var deviceCount = ProfilerDriver.GetFormattedCounterValue(selectedFrameIndex,
+                    InputStatistics.Category.Name, InputStatistics.DeviceCountName);
+                var controlCount = ProfilerDriver.GetFormattedCounterValue(selectedFrameIndex,
+                    InputStatistics.Category.Name, InputStatistics.ControlCountName);
+
+                m_UpdateCountLabel.text = $"{InputStatistics.UpdateCountName}: {updateCount}";
```
ekcoh, line 107: This is just a basic label list at the moment. Is it possible to pass any extra information via the profiler API? Or extract sample data to e.g. create a distribution within detail window?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3072,8 +3079,25 @@ internal bool ShouldRunUpdate(InputUpdateType updateType)
         /// which buffers we activate in the update and write the event data into.
         /// </remarks>
         /// <exception cref="InvalidOperationException">Thrown if OnUpdate is called recursively.</exception>
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1809:AvoidExcessiveLocals", Justification = "TODO: Refactor later.")]
         private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer eventBuffer)
+        {
+            try
+            {
+                DoUpdate(updateType, ref eventBuffer);
```
ekcoh, line 3147: This needs more work to extract some stuff from DoUpdate below

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3621,11 +3653,22 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+                m_Metrics.totalEventProcessingTime += eventProcessingTime;
                 m_Metrics.totalEventLagTime += totalEventLag;
 
-                ResetCurrentProcessedEventBytesForDevices();
+                // Profiler counters
+                InputStatistics.EventCount.Value += totalEventCount;
+                InputStatistics.EventSize.Value += totalEventSizeBytes;
+                InputStatistics.AverageLatency.Value += ((totalEventLag / totalEventCount) * 1e9);
+                InputStatistics.MaxLatency.Value += (maxEventLag * 1e9);
+                InputStatistics.EventProcessingTime.Value += eventProcessingTime * 1e9; // TODO Possible to replace Stopwatch with marker somehow?
```
ekcoh, line 3733: I guess this last line here is an anti pattern, ideally I would like to include the Update profiler marker into the profiler module, is this possible?

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputSystemProfilerModule.cs
surfnerd: There is a [ProfilerRecorder](https://docs.unity3d.com/ScriptReference/Unity.Profiling.ProfilerRecorder.html) that you can use to gather samples, which then allows you to do more statistical computations.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
surfnerd: During hackweek I used StopWatch.  It was preallocated and reset and reused every frame and accumulated in a [ProfilerCounter](https://docs.unity3d.com/Packages/com.unity.profiling.core@0.2/manual/profilercounter-guide.html)

surfnerd: which is what it looks like you're doing here :)

surfnerd:
My first thought looking at the window is the mix of types being presented.  I see Counters, Memory Size, and Times.  It looks a little bit busy and hard to know what I'm looking at exactly.  Overall I think this is a great idea.  I wonder if there's a way we can group like values together within one window.

ekcoh:
> My first thought looking at the window is the mix of types being presented. I see Counters, Memory Size, and Times. It looks a little bit busy and hard to know what I'm looking at exactly. Overall I think this is a great idea. I wonder if there's a way we can group like values together within one window.

Agreed, I just saw other modules were doing that and replicated it. I think there is opportunity to slip things into separate modules to handle it, e.g. Latency and Input System Counters. I wonder if execution time should even be there since its covered by the ProfilerMarkers anyway.....

Packages/com.unity.inputsystem/InputSystem/Utilities/InputStatistics.cs
```diff
@@ -0,0 +1,115 @@
+using Unity.Profiling;
+
+namespace UnityEngine.InputSystem
+{
+    /// <summary>
+    /// Input Statistics for Unity Profiler integration.
+    /// </summary>
+    internal static class InputStatistics
```
jfreire-unity, line 8: I'd create a `Editor/Internal/Profiler` folder and but everything in there. I find odd to have this class next to a bunch of "Utilities" classes.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputSystemProfilerModule.cs
jfreire-unity: @ekcoh it seems to be possible to add more complex UITK elements. I think a list can be fine for now, and then we could add a nicer UI using UITK (there's more info [here](https://docs.unity3d.com/Manual/profiler-customizing-details-view.html) , which you probably have seen; and a more more complex UI example in [this package](https://github.cds.internal.unity3d.com/unity/com.unity.adaptiveperformance/blob/be81b29e9031944bfb7b5d1b4b2c21fa4d36aa7b/Editor/Profiler/AdaptivePerformanceDetailsViewController.cs#L36-L36) where this is the result: image

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -2,7 +2,8 @@
     "name": "Unity.InputSystem",
     "rootNamespace": "",
     "references": [
-        "Unity.ugui"
+        "Unity.ugui",
+        "Unity.Profiling.Core"
```
jfreire-unity, line 6: Wondering if this might create some overhead? Probably we could set a define like `INPUTSYSTEM_USE_CUSTOMPROFILER` or something similar, in case users don't want to have this code in their release builds.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3055,6 +3058,10 @@ internal bool ShouldRunUpdate(InputUpdateType updateType)
             return (updateType & mask) != 0;
         }
 
+        struct UpdateMetrics
```
jfreire-unity, line 3122: To remove?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3465,9 +3492,14 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
-                    ++m_Metrics.totalEventCount;
-                    m_Metrics.totalEventBytes += (int)currentEventReadPtr->sizeInBytes;
+                    {
+                        var lag = currentTime - currentEventTimeInternal;
+                        totalEventLag += lag;
+                        if (lag > maxEventLag)
+                            maxEventLag = lag;
+                    }
+                    ++totalEventCount;
+                    totalEventSizeBytes += (int)currentEventReadPtr->sizeInBytes;
```
jfreire-unity, line 3564: This a bit more complex so I would extract this to a method, as `OnUpdate` is already big enoug.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3621,11 +3653,22 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+                // Update metrics (exposed via analytics and debugger)
+                var eventProcessingTime =
                     ((double)(Stopwatch.GetTimestamp() - processingStartTime)) / Stopwatch.Frequency;
+                m_Metrics.totalEventCount += totalEventCount;
+                m_Metrics.totalEventBytes += totalEventSizeBytes;
+                m_Metrics.totalEventProcessingTime += eventProcessingTime;
                 m_Metrics.totalEventLagTime += totalEventLag;
 
-                ResetCurrentProcessedEventBytesForDevices();
+                // Profiler counters
```
jfreire-unity, line 3728: I'd add a comment that these counters are PerFrame, as we can call update more than once?
Or Maybe a prefix of something like `InputStatistics.EventCountPerFrame.Value` or `PerUpdate`, something like that.

Also if could we use `ProfileRecorder` for this?

Packages/manifest.json
```diff
@@ -5,12 +5,12 @@
     "com.unity.coding": "0.1.0-preview.24",
     "com.unity.ide.rider": "1.2.1",
     "com.unity.ide.visualstudio": "2.0.15",
+    "com.unity.profiling.core": "1.0.2",
     "com.unity.settings-manager": "1.0.3",
     "com.unity.test-framework": "1.4.5",
     "com.unity.test-framework.build": "0.0.1-preview.12",
     "com.unity.test-framework.performance": "3.0.3",
     "com.unity.test-framework.utp-reporter": "1.1.0-preview",
-    "com.unity.textmeshpro": "2.1.4",
```
jfreire-unity: Any reason why this is removed?

jfreire-unity (commented):
Great work in adding this functionality.
I think this could easily land in an experimental kind of namespace, almost as is. Just with small tweaks IMO.

The only issue I have is from a UI perspective. Similar to what @surfnerd mentioned, the Module UI is a bit confusing with all the Counts, Bytes, Timings.

I think we could put the Bytes and Timings statistics only in the panel label for now. In my opinion the count statistics are better suited for the "line" UI type.

Ideally, the best would be to have some kind of "Timeline" similar to what CPU has (see img below). But we don't need to be perfect for this to land as I don't think it adds a big risk. I think this would already provide benefit and we could just improve the parts that are confusing a little bit. But of course, we should ask ourselves what are trying to solve with this and see if this 
image

I will try to play around a bit with this this sprint and see if I can improve anything.

ritamerkl (commented):
I tested the Profiler in the editor, and I think this is a great initiative! Thanks for setting it up Håkan. 

First of all I think the values in the InputSystem profiler make a lot of sense. I agree that some of them could leave the Highlights window and the graph and just show up in the detailed view underneath (I lack the terms - but I mean the window where all values appear in text form with their current values) since they don't really give more insights in a graph form but are more confusing than helpful.

Thinking about some debugging cases I had already I think it would be interesting to be able to get more insights in the kind of events fired and more information on the connected devices and controls. 
Something like a Hierarchy view for all fired events (see picture). 
Screenshot 2025-01-03 at 15 38 34

For the devices/contorls I could imagine a table like this: 
Screenshot 2025-01-03 at 15 39 33
To visualize all devices with the controls (maybe even which ones were acuated).

Last but not least: A general variable which indicates the total storage needed and used by the inputsystem would be great, I am not sure if Total state buffer size represents this completely - maybe it is possible to split this up in several aspects below in the detail area as well.

I did not take a look at the code yet, but only tried it in the editor.

Packages/com.unity.inputsystem/InputSystem/Editor/Internal/InputSystemProfilerModule.cs
ekcoh: Yes it's possible, but IMHO it would make most sense to align on the data first and keep it simple (as most other modules do). It could be improved after that point if there is value in it.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
ekcoh: Not sure I see how a define helps you omit a reference?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Makes sense

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: makes sense, I haven't looked into the ProfileRecorder yet so I am not sure

Packages/manifest.json
ekcoh: Can't recall so probably a mistake

PR #1874, opened 2024-03-05
FIX: Canceling ControlScheme edits resets the UI to the previously saved values (ISX-1892)

timkeo:
### Description

The issue occurred because, although the changes weren't persisted to the SerializedProperty, the UI was still holding changes, and if the ControlScheme Edit dialog is re-opened (without changing ControlSchemes) the old values are displayed. We must explicitly reset the `InputControlScheme` in the ViewState to the last saved values.

### Changes made

- Created `ControlSchemeCommands.ResetSelectedControlScheme()` command to reload `InputControlScheme` from the SerializedProperty (called during Cancel)
- Minor refactoring of "Close" methods in `ControlSchemesView`
- Added comments explaining current functionality

### Notes

I don't know if the current functionality is what we want or not, and so I added verbose comments explaining how it works now. We can change it later if not to everyone's liking.

The new `ResetSelectedControlScheme()` command is nearly identical to `SelectControlScheme`. I decided this was the safer route for now, but maybe there's a better way.

I only did the most basic of testing before opening this PR.

### Checklist

Before review:

- [x] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [x] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

PR #2044, opened 2024-11-06
FIX: The "Add Control Scheme..." popup now requires explicit Save/Close user action (case ISXB-1131)

AlexTyrer:
### Description

Previously clicking outside the popup window (anywhere in the editor) would close the popup window.

This would leave a new blank (unnamed, orphaned) control scheme populating the UI - this was confusing.

Now this popup will persist until the user explicitly chooses to Save or Cancel.

### Testing status & QA

Manually testing the UI using the reproduction steps provided.
Local and automated testing.

### Overall Product Risks

- Complexity: minimal - it's single line change (deletion)
- Halo Effect: minimal - just affects the behaviour of this popup window

### Comments to reviewers

This issue manifested when the Input System started using the newer UIToolkit implementation of its UI.

It was a result of fixing [another issue](https://jira.unity3d.com/browse/ISX-1892) in [this PR](https://github.com/Unity-Technologies/InputSystem/pull/1874).

[Discussions ](https://unity.slack.com/archives/C04RDRXRJ1L/p1730823258485409)with @Pauliusd01 and @lyndon-unity led to this choice of behaviour change of this popup window:

- it's the least surprising for the user
- it avoids losing any unsaved changes without the user explicitly making a choice.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/.yamato/config.metadata
+++ b/.yamato/config.metadata
@@ -1,5 +1,5 @@
 editors:
-  - version: 2021.3
+  - version: 2021.3.45f
   - version: 2022.3
     disable_tvos_run: true
   - version: 6000.0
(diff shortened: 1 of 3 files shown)


lyndon-unity (approved):
Testing this on Alex's machine shows the issue is resolved, whist the other related issue is not regressed. It also seems a more intuitive UX flow.

jfreire-unity:
Adding QA

Pauliusd01 (approved):
LGTM. Checked: 
- trying to interact with things within the asset while the control scheme popup is open (you can't, It's fully static. Would be great if it blinked when you click outside it just like on Windows, but I don't think that's a thing we can use in UITK)
- entering play mode while it is open (it auto closes and refreshes the window, which I think is fine)
- closing the input window while the popup is still there (same as the play mode scenario) 
- basic functionality of adding/removing/editing schemes

PR #2047, opened 2024-11-13
FIX: Check for valid index in current Control Scheme (ISXB-1129)

tspiller:
### Description

The **ShouldBindingBeVisible()** function was not checking that deviceIndex passed in was a valid index in the current Control Scheme's deviceRequirements array. This could lead to ArgumentOutOfRangeException error when accessing the element, particularly in the case of an empty array.

### Testing status & QA

Local testing / automated tests

### Overall Product Risks

Simple additional check for valid index

Complexity: 0
Halo Effect: 0

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
- [X] Docs for new/changed API's.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


ekcoh (approved):
Fix looks legit to circumvent the problem, but why would the function the called with an invalid deviceIndex in the first place? What I am after is that there must be an earlier point where we should have detected that the index is invalid?

PR #2058, opened 2024-11-22
NEW: Add Packageworks setup to selfserve publishing

stefanunity:
### Description

We need to move package pack, verification and publishing to PackageWorks so we can publish ourselves moving forward.

- installed Wrench tooling
- customized jobs to fit our package
- generated yml files

### Testing status & QA

- All the new jobs are green
- packed package has the same contents as if packed with our old job

### Overall Product Risks

Very little risk, all the old jobs still exist and we can release with them until RM stops us.

- Complexity: low
- Halo Effect: none

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Tools/CI/onboard.py
```diff
@@ -0,0 +1,149 @@
+# Description: This script is used to onboard a project onto Wrench
```
jfreire-unity: Out of curioisity: when is this Python script called in the process?

.yamato/wrench/wrench_config.json
```diff
@@ -0,0 +1,37 @@
...
+  ],
+  "jobs_to_monitor": {
+    "com.unity.inputsystem": [
+      ".yamato/wrench/api-validation-jobs.yml#api_validation_-_inputsystem_-_2019_4_-_windows",
+      ".yamato/wrench/preview-a-p-v.yml#all_preview_apv_jobs",
+      ".yamato/wrench/promotion-jobs.yml#publish_dry_run_inputsystem"
+    ]
+  },
+  "publishing_job": ".yamato/wrench/promotion-jobs.yml#publish_inputsystem",
+  "branch_pattern": "ReleaseSlash",
```
jfreire-unity, line 37: Do you know if this is true? Our releases have always been tagged commits in the `stable` branch. Unless we are changing this now or this is some other "branch pattern." I'm fine with either; I just want to surface this in case it becomes a problem.

jfreire-unity (approved):
Thanks for doing this!
If releasing works, maybe a intro about this to team would be great. Or just pointing in the right direction of internal documentation :)

ritamerkl (approved):

Tools/CI/onboard.py
stefanunity: it is called at the very beginning and you can then delete it

.yamato/wrench/wrench_config.json
stefanunity: I made a new branch called release/1.X and we need to follow this pattern says RM.
I'll update the release manual and it shouldn't be a big change for us.

PR #2064, opened 2024-12-02
DOCS: Improved script API docs of InputAction.CallbackContext

ekcoh:
### Description

Improved script API Doscs of InputAction.CallbackContext:
- Added example
- Extended description with links for struct type.

### Testing status & QA

Reviewed generated XML docs in IDE inspector (Rider)
Generated docs via package docs tool.

### Overall Product Risks

- Complexity: very small
- Halo Effect:  very small

### Comments to reviewers

Check language, compliance and example.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


ekcoh:
I noticed this one has malformed xmldoc so need to revisit that

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1791,7 +1791,80 @@ internal int BindingIndexOnMapToBindingIndexOnAction(int indexOfBindingOnMap)
...
+        ///          /// Receive notifications when move or fire actions are performed
+        ///          move.action.performed += MovePerformed;
+        ///          fire.action.performed += FirePerformed;
+        ///      }
+        ///
+        ///      void OnEnable()
+        ///      {
+        ///          /// Enable actions as part of enabling this behavior.
+        ///          move.action?.Enable();
+        ///          move.action?.Enable();
```
lyndon-unity: should be fire.action here

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1791,7 +1791,80 @@ internal int BindingIndexOnMapToBindingIndexOnAction(int indexOfBindingOnMap)
...
+        ///      }
+        ///
+        ///      void OnDisable()
+        ///      {
+        ///          /// Disable actions as part of disabling this behavior.
+        ///          move.action?.Disable();
+        ///          move.action?.Disable();
```
lyndon-unity: should be fire.action here

lyndon-unity (changes requested):
Small correction needed in the example code

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1791,7 +1791,80 @@ internal int BindingIndexOnMapToBindingIndexOnAction(int indexOfBindingOnMap)
...
         /// </summary>
         /// <remarks>
+        /// The callback context represents the current state of an <see cref="action"/> associated with the callback
+        /// and provides information associated with the bound <see cref="control"/>, its value as well as its
+        /// <see cref="phase"/>.
+        ///
+        /// The callback context provides means to consume events (push-based input) as part of an update when using
+        /// input action callback notifications, e.g. <see cref="InputAction.started"/>,
+        /// <see cref="InputAction.performed"/>, <see cref="InputAction.canceled"/> rather than relying on
+        /// pull-based reading.
```
ritamerkl: Is there a piece of documentation to link to (push/pull based reading) to clarify this statement?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -2096,6 +2186,44 @@ public unsafe void ReadValue(void* buffer, int bufferSize)
...
+            ///
+            ///     void Awake()
+            ///     {
+            ///         if (move.action != null)
+            ///         {
+            ///             move.action.performed += context =>
+            ///             {
+            ///                 // Note: Assumes the underlying value type is Vector2.
+            ///                 Vector2 value = context.ReadValue&lt;Vector2&gt;();
+            ///                 Debug.Log($"Value is: {value}");
```
ritamerkl: redundant line in my opinion add something up to a vector (eg the player position)?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -2119,6 +2247,40 @@ public TValue ReadValue<TValue>()
...
+            ///
+            ///     void Awake()
+            ///     {
+            ///         if (fire.action != null)
+            ///         {
+            ///             fire.action.performed += context =>
+            ///             {
+            ///                 // ReadValueAsButton attempts to interpret the value as a button.
+            ///                 bool value = context.ReadValueAsButton();
+            ///                 Debug.Log($"Button state is: {value}");
```
ritamerkl: same as above comment. Is it usual to use prints in the docs examples?

ritamerkl (approved):
LGTM, mainly minor comments

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: Good catch, fixed in https://github.com/Unity-Technologies/InputSystem/pull/2064/commits/fe9c8c07a66b20b95edc9824e36b1fe15a2b7542

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: Good catch, fixed in https://github.com/Unity-Technologies/InputSystem/pull/2064/commits/fe9c8c07a66b20b95edc9824e36b1fe15a2b7542

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: It's just referencing the general concept of push vs pull architecture, i.e. callbacks vs polling in this case. Would it be better to switch to those?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: Fair, I originally had the read value inside the log call but it was more difficult to read. Also I am torn around what to do with the value since the code should compile and ideally do something useful if copied over into a script and entering playmode was my though. Not sure what the suggestion is? To have a public Vector3 direction so its visible in the Inspector?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: Honestly I am not sure what is recommended, other suggestions to do something useful that indicates that it works if copied into a .cs file without adding much more code to the sample?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ritamerkl: I was just wondering if there is a documentation about the push/pull architecture, differences and advantages in the InputSystem, then this would be a good place to link it, but if there isn't it is fine I think.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ritamerkl: That, or maybe modifiying the ridgid body of the gameobject directly?
eg ridgidBody.velocity = value*speed; (where ridgidbody is a public or serialized field that needs to be assigned in the editor and speed can be a private float field?

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ritamerkl: I think it would be enough even to call an empty function like if(value){Jump();} -> Jump() { //insert player jump logic here OR Debug.Log("Jump"); } for example

ritamerkl: Anyway, I think it's complaining on a high level and is okay to handle like that, too.

lyndon-unity (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: I would vote for the transform in this case otherwise, if the object has a mesh renderer, otherwise there is nothing to be seen. If rigidBody is used, then we need to add additional detail about that they must reconfigure the system to use FixedUpdate etc or sample value and use additional FixedUpdate to use sampled values. It seems @duckets was fine using Debug.Log at least

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: I think about it an update once CI / package tool issues are resolved

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ritamerkl: I see, the only reason why I proposed using the ridgid body was that translating the transform directly is a rather bad practice in general using physics. So I am not sure about how to show case it without messing up with best practices from other areas. Maybe in that case a log would be the safest option.

ekcoh: Maybe I am slow but why do we need to use physics?

ritamerkl: Sorry, I need to rephrase that: for a game using physics translating a player directly using the transform component is a bad idea, since it will not align with the physics of the game (the gravity, wind and other forces). That's why I am wondering if showing that in an example is a bad idea.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
duckets: Yes there is, here:

https://docs.unity3d.com/Packages/com.unity.inputsystem@1.11/manual/RespondingToActions.html

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
duckets: there's a few examples of using Debug.Log in our internal docs guide here, so I think this is the best solution:
https://docs.unity.com/internal/content-ops/en-us/documentation/content-model/style-formatting-code-examples

duckets (approved):

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: Thanks, I added the link as an hyperlink.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
ekcoh: I am still confused, sorry. I do not see where we use rigid body in examples here? If we were this example would still be valid but only if InputSystem update mode was set to FixedUpdate. If Debug.Log was instead affecting a ridigbody that is. Transform is the way to go when using DynamicUpdate AFAIK, unless sampling in the callback and applying later via FixedUpdate script method.

ekcoh: I will mark this resolved know unless someone may help me understand :)

PR #2071, opened 2024-12-03
Fix issue: New Input System - All mouse buttons have the same pointer Id

gapspt:
### Description

Fixes the following bug: https://unity3d.atlassian.net/servicedesk/customer/portal/2/IN-90349

When using the IPointerDownHandler (OnPointerDown method) with the New Input System, the eventData.pointerId for any mouse button is always the same, corresponding to the mouse's deviceId.

According to the [documentation for Unity 2018](https://docs.unity3d.com/2018.1/Documentation/ScriptReference/EventSystems.PointerEventData-pointerId.html):
> When using a mouse the pointerId returns -1, -2, or -3. These are the left, right and center mouse buttons respectively.

I mention the 2018 because the [documentation on the standalone Unity UI package](https://docs.unity3d.com/Packages/com.unity.ugui@3.0/api/UnityEngine.EventSystems.PointerEventData.html) is lacking the values for the mouse buttons.

### Testing status & QA

Tested locally.

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers

Note that `m_CurrentPointerId` is set and checked against the value `-1`, which is the same value that should be used for the left mouse button, but in the case of the `m_CurrentPointerId` it means "none". This is not a problem in the current fix, as it only changes the `eventData` object being passed to the event listeners, so the `m_CurrentPointerId` will still contain the original `pointerId`, which would be the mouse's `deviceId` .

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -343,6 +343,8 @@ namespace UnityEngine.InputSystem.UI
 
             ////REVIEW: for touch, we only need the left button; should we skip right and middle button processing? then we also don't need to copy to/from the event
 
+            var pointerId = eventData.pointerId;
+
             // Left mouse button. Movement and scrolling is processed with event set left button.
             eventData.button = PointerEventData.InputButton.Left;
             state.leftButton.CopyPressStateTo(eventData);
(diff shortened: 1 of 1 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=2071) <br/>All committers have signed the CLA.

PR #2090, opened 2024-12-16
FIX: Fixed multiple `OnScreenStick` Components that does not work together when using them simultaneously in isolation mode (ISXB-813)

bmalrat:
### Description

When multiple OnScreenStick is used at the same time, in isolation mode, it doesn't distinguish which finger drive the components and will be not functional. 

To fix that I added, when it come from a touch event a track of which finger triggered it and only use information from it.
I also switched the pointerdown tracking to bypass to be able to track any pointer down/up instead of only the First down and last Up.

### Testing status & QA

Added tests
tested on android with the customer project.

### Overall Product Risks

- Complexity: 
- Halo Effect: 

### Comments to reviewers

### Checklist

Before review:

- [X ] Changelog entry added.
- [ X] Tests added/changed, if applicable.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
@@ -243,7 +243,7 @@ namespace UnityEngine.InputSystem.OnScreen
             {
                 screenPosition = pointer.position.ReadValue();
             }
-            
+
             m_PointerEventData.position = screenPosition;
             EventSystem.current.RaycastAll(m_PointerEventData, m_RaycastResults);
             if (m_RaycastResults.Count == 0)
@@ -267,11 +267,11 @@ namespace UnityEngine.InputSystem.OnScreen
                 m_TouchControl = touchControl;
                 m_PointerMoveAction.ApplyBindingOverride($"{touchControl.path}/position", path: "<Touchscreen>/touch*/position");
             }
-           
+
             m_PointerMoveAction.performed += OnPointerMove;
             m_IsIsolationActive = true;
         }
-        
+
         private void OnPointerChanged(InputAction.CallbackContext ctx)
         {
             if (ctx.control.IsPressed())
@@ -285,8 +285,8 @@ namespace UnityEngine.InputSystem.OnScreen
             // only pointer devices are allowed
             Debug.Assert(ctx.control?.device is Pointer);
             Vector2 screenPosition;
-           
-            // If it's a finger take the value from the finger that initiated the change            
+
+            // If it's a finger take the value from the finger that initiated the change
             if (m_TouchControl != null)
             {
                 // if the finger is up ignore the move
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -227,11 +229,20 @@ private void EndInteraction()
 
         private void OnPointerDown(InputAction.CallbackContext ctx)
         {
+            if (m_IsIsolationActive) { return; }
             Debug.Assert(EventSystem.current != null);
 
             var screenPosition = Vector2.zero;
-            if (ctx.control?.device is Pointer pointer)
+            TouchControl touchControl = null;
+            if (ctx.control?.parent is TouchControl touch)
```
ekcoh, line 237: Unfortunately this is the design, generally this is a perfect fit for a device-agnostic pointer interface (just a side note)

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -251,23 +262,63 @@ private void OnPointerDown(InputAction.CallbackContext ctx)
...
         }
 
         private void OnPointerMove(InputAction.CallbackContext ctx)
         {
             // only pointer devices are allowed
             Debug.Assert(ctx.control?.device is Pointer);
+            Vector2 screenPosition;
 
-            var screenPosition = ((Pointer)ctx.control.device).position.ReadValue();
+            // If it's a finger take the value from the finger that initiated the change
```
ekcoh, line 289: More a general conceptual thought around the scalability of this than the particular fix, what if the on screen control is interacted with a pen?

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
```diff
@@ -430,6 +481,10 @@ public bool useIsolatedInputActions
         private List<RaycastResult> m_RaycastResults;
         [NonSerialized]
         private PointerEventData m_PointerEventData;
+        [NonSerialized]
+        private TouchControl m_TouchControl;
```
ekcoh, line 485: Not really related to this fix, but more to the original design, this implies that a member would need to be added for any kind of pointer device right? Or is this particularly due to TouchControl not being a derivative of pointer in your opinion @bmalrat ?

ekcoh (approved):

Pauliusd01 (approved):
Checked touch controls using the user project as well as the standard touch control project we use for release testing, did not notice any issues

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
bmalrat: good question I don't have a pen/touch setup to test.
But we can fix the pen case after if needed. Here it only fix when it comes from a touch device

Packages/com.unity.inputsystem/InputSystem/Plugins/OnScreen/OnScreenStick.cs
bmalrat: Normally we don't need it for each type, only for touch to identify which one was touched in the <Touchscreen>/touch*/press. I'm not aware of an other type that could aggregate multiple source into one binding.

PR #2091, opened 2024-12-17
FIX: Avoid problems with CI and reimports

ekcoh:
### Description

FIX: Solves problem with CI and reimports issues by `EditorApplication.delayCall` which leads to CI build errors "Error: Scripts are compiling" when attempting to run a batch mode project via UTR when source code associated with an .inputactions asset is out of sync due to asset or code generator update.

### Testing status & QA

Have tested various ways to go around the problem `EditorApplication.update`, `AssetDatabase.ImportAsset` etc but went with `AssetDatabase.ImportAsset` which is a solution that goes straight against what previous inline comment suggested. Discussed solution with Asset Database team and no solution is good since this should use a source generator (Roslyn) for this kind of feature, but cannot be reliably setup in current Unity version. Hence this solution.

Recommendation for testing is to also ensure this solution works with regular editor use (non-batch mode). Also proven by the following PR which has intentionally outdated test source asset and sample source asset: https://github.com/Unity-Technologies/InputSystem/pull/2060 

The linked PR should be closed if/when this lands.

### Overall Product Risks

- Complexity: Small to medium
- Halo Effect: Medium due to many corner cases and unknown side effects

### Comments to reviewers

Focus on corner cases and potential race conditions.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -34,6 +34,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue with default device selection when adding new Control Scheme.
 - Fixed an issue where action map delegates were not updated when the asset already assigned to the PlayerInput component were changed [ISXB-711](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-711).
 - Fixed Action properties edition in the UI Toolkit version of the Input Actions Asset editor. [ISXB-1277](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1277)
+- Fixed an issue where batch jobs would fail with "Error: Error building Player because scripts are compiling" if a source generated .inputactions asset is out of sync with its generated source code. (ISXB-1300).
 
 ### Changed
 - Added back the InputManager to InputSystem project-wide asset migration code with performance improvement (ISX-2086).
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -34,7 +34,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue with default device selection when adding new Control Scheme.
 - Fixed an issue where action map delegates were not updated when the asset already assigned to the PlayerInput component were changed [ISXB-711](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-711).
 - Fixed Action properties edition in the UI Toolkit version of the Input Actions Asset editor. [ISXB-1277](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1277)
-- Fixed an issue where batch jobs would fail with "Error: Error building Player because scripts are compiling" if a source generated .inputactions asset is out of sync with its generated source code. (ISXB-1300).
+- Fixed an issue where batch jobs would fail with "Error: Error building Player because scripts are compiling" if a source generated .inputactions asset is out of sync with its generated source code (ISXB-1300).
 
 ### Changed
 - Added back the InputManager to InputSystem project-wide asset migration code with performance improvement (ISX-2086).
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
@@ -256,22 +256,12 @@ namespace UnityEngine.InputSystem.Editor
 
             if (InputActionCodeGenerator.GenerateWrapperCode(wrapperFilePath, asset, options))
             {
-                // When we generate the wrapper code cs file during asset import, we cannot call ImportAsset on that directly because
-                // script assets have to be imported before all other assets, and are not allowed to be added to the import queue during
-                // asset import. So instead we register a callback to trigger a delayed asset refresh which should then pick up the
-                // changed/added script, and trigger a new import.
-                if (++_importedAssetsSinceLastRefresh == 1)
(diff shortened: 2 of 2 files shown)


Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue with default device selection when adding new Control Scheme.
 - Fixed an issue where action map delegates were not updated when the asset already assigned to the PlayerInput component were changed [ISXB-711](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-711).
 - Fixed Action properties edition in the UI Toolkit version of the Input Actions Asset editor. [ISXB-1277](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1277)
+- Fixed an issue where batch jobs would fail with "Error: Error building Player because scripts are compiling" if a source generated .inputactions asset is out of sync with its generated source code. (ISXB-1300).
```
ritamerkl: One sentence point too much

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -256,11 +256,11 @@ private static void GenerateWrapperCode(AssetImportContext ctx, InputActionAsset
 
             if (InputActionCodeGenerator.GenerateWrapperCode(wrapperFilePath, asset, options))
             {
-                // When we generate the wrapper code cs file during asset import, we cannot call ImportAsset on that directly because
-                // script assets have to be imported before all other assets, and are not allowed to be added to the import queue during
-                // asset import. So instead we register a callback to trigger a delayed asset refresh which should then pick up the
-                // changed/added script, and trigger a new import.
-                EditorApplication.delayCall += AssetDatabase.Refresh;
```
ritamerkl, line 263: > use a source generator (Roslyn) for this kind of feature, but cannot be reliably setup in current Unity version. 

Why? And how is that handled in similar cases among the editor then?
It sounds like you put in quite an effort on researching the details around it, it would be interesting to hear more about it.

ritamerkl (approved):
Provisory approving, when the testing is done and no issues found

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
ekcoh: I consulted the asset database team for the best way forward with this one. It seems its not recommended/support to generate c-sharp code assets from another assets. This is what we do.

Roslyn generators have been supported for some time https://docs.unity3d.com/6000.0/Documentation/Manual/roslyn-analyzers.html and it works well, I wrote a couple of generators myself to explore previously and its used by e.g. DOTS, but it doesn't support mapping assets to them. So they are usable for attributed code typically. Also UI Toolkit uses them I believe with the e.g. UxmlElement tags etc.

Regarding similar cases I am not aware of any other case where source files (assets) are generated from another asset. Are you aware of any?

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Now fixed in https://github.com/Unity-Technologies/InputSystem/pull/2091/commits/e64855201417fa1484498e1da4857a12a91339e9

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
ritamerkl: Thank you for elaborating.

That's weird, is the inputactionasset the only asset that is suspect to be imported then? Or how come this case is so unique? I guess what I am after is, why do we generate it from another asset, and why can't we solve this like the import of any other asset?

ritamerkl: I am just asking the five why's to understand the root of the problem.

ekcoh: Importing isn't unique at all, but the usual import is just that we parse JSON and create ScriptableObjects (InputActionAsset). The problematic part is the writing .cs files when source generation option is checked since code needs to be imported before assets.

ekcoh: Its not possible to customly import scripts

ritamerkl: Okay I think I get the gist of it, it sounds a bit like a devils cycle to write it, It's a bummer that there is no legitimate way to do that properly. But as long as it works it works I guess. Thanks for explaining.

ekcoh: Will mark this resolved

Pauliusd01 (approved):
LGTM, things I tried checking:

- Running most of our testing scenes in the main repo (player/playmode)
- Making changes to these scenes and then running them again (player/playmode)
- Doing above while also throwing in a reimport before running them
- Doing the above again while having their input actions assets generate a C# file option on

I did not notice any issues with the testing scenes or using/saving my changes whether I had the C# option on or whether the assets were reimported recently

PR #2093, opened 2024-12-18
FIX: Input State Window is now force-closed when the device it is inspecting is disconnected

Secticide:
### Description

Fixes [ISXB-658](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-658).

The `InputStateWindow` currently caches a bunch of state. When a device is disconnected, then reconnected, this state is reused. It is possible - likely based on the specifics of the device - that this will cause the window to `memcpy` from or into memory no longer owned by the device or the window leading to an editor crash.

To resolve this, I've used the same strategy used for the Input debugger window; when the device is disconnected - the window is force-closed.

### Testing status & QA

I tested this locally using a Stadia controller, but it would be great to get some wider testing of devices on this before merging. Although I certainly don't see this as high risk.

### Overall Product Risks

- Complexity: 0
- Halo Effect: 0

### Comments to reviewers

The changes in `DrawHexDump` were to resolve some spammy null reference exception errors caused by the window. I believe it to be now unlikely that we'd run into a case where null reference exceptions are thrown during the drawing of the window, but there is no harm in doing these checks.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -34,6 +34,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue with default device selection when adding new Control Scheme.
 - Fixed an issue where action map delegates were not updated when the asset already assigned to the PlayerInput component were changed [ISXB-711](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-711).
 - Fixed Action properties edition in the UI Toolkit version of the Input Actions Asset editor. [ISXB-1277](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1277)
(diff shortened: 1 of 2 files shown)


ekcoh (approved):
LGTM, makes sense to force close due to how the underlying memory is managed. Adding QA to PR.

stefanunity (approved):
Test with DS4, DualSense, Xbox One & Switch Pro controller.
Editor didn't crash, window closed on disconnect.

ekcoh:
@Secticide I resolved the PR conflict in https://github.com/Unity-Technologies/InputSystem/pull/2093/commits/1204c35b61b2dea45e0a44c0b8a987a20d76d4c7, saw a CI failure on samples. Lets see if there is a difference this run.

PR #2096, opened 2024-12-19
FIX: collapsed and expanded treeview state is preserved when saving

smnwttbr:
### Description

Fix for [ISXB 1164](https://jira.unity3d.com/browse/ISXB-1164)

I've removed (unused) methods related to saving the expanded state of the tree. These might have been an artifact from an earlier implementation, any any case, these were unused.

I changed the guidToId caching mechanism to use guid.GetHashCode() so that we get a deterministic guid->int mapping.

### Testing status & QA

Ran local package tests.

### Overall Product Risks

- Complexity: 0 
- Halo Effect: 0

### Comments to reviewers

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Tests/InputSystem.Editor/SelectorsTests.cs
+++ b/Assets/Tests/InputSystem.Editor/SelectorsTests.cs
@@ -24,7 +24,7 @@ internal class SelectorsTests
         var actionTwo = actionMap.AddAction("Action2", binding: "<Keyboard>/d");
 
 
-        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate());
+        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate(), new Dictionary<Guid, int>());
 
 
         Assert.That(treeViewData.Count, Is.EqualTo(2));
--- a/Assets/Tests/InputSystem.Editor/SelectorsTests.cs
+++ b/Assets/Tests/InputSystem.Editor/SelectorsTests.cs
@@ -24,7 +24,7 @@ internal class SelectorsTests
         var actionTwo = actionMap.AddAction("Action2", binding: "<Keyboard>/d");
 
 
-        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate(), new Dictionary<Guid, int>());
+        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate());
 
 
         Assert.That(treeViewData.Count, Is.EqualTo(2));
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
@@ -26,6 +26,9 @@ namespace UnityEngine.InputSystem.Editor
         private bool m_RenameOnActionAdded;
         private readonly CollectionViewSelectionChangeFilter m_ActionsTreeViewSelectionChangeFilter;
 
+        //save TreeView element id's of individual input actions and bindings to ensure saving of expanded state
+        private Dictionary<Guid, int> m_GuidToTreeViewId;
+
         public ActionsTreeView(VisualElement root, StateContainer stateContainer)
             : base(root, stateContainer)
         {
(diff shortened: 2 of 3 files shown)


Assets/Tests/InputSystem.Editor/SelectorsTests.cs
```diff
@@ -24,7 +24,7 @@ public void GetActionsAsTreeViewData_ReturnsActionsAndBindingsAsTreeViewData()
         var actionTwo = actionMap.AddAction("Action2", binding: "<Keyboard>/d");
 
 
-        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate(), new Dictionary<Guid, int>());
+        var treeViewData = Selectors.GetActionsAsTreeViewData(TestData.EditorStateWithAsset(asset).Generate());
```
ekcoh: Makes sense to get rid of default constructed object

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -619,7 +616,7 @@ public static List<TreeViewItemData<ActionOrBindingData>> GetActionsAsTreeViewDa
                                 if (isVisible)
                                 {
                                     var name = GetHumanReadableCompositeName(nextBinding, state.selectedControlScheme, controlSchemes);
-                                    compositeItems.Add(new TreeViewItemData<ActionOrBindingData>(GetIdForGuid(new Guid(nextBinding.id), idDictionary),
+                                    compositeItems.Add(new TreeViewItemData<ActionOrBindingData>(treeviewItemIDCounter++,
```
ekcoh: Not sure I understand this binding counter, I though items where mapped to the asset GUIDs for stability?

ekcoh (commented):
Look trough the code (didn't run it) and in general I must say I have love PRs that remove more lines than they add :) Not sure I fully understand how this rather compact PR solves the problem though.

ekcoh (commented):
Code changes LGTM if it works. I noticed you have changed to a index based id it seems so would recommend covering expansion state robustness in the tests, e.g. what happens if you expand two branches and then insert a new branch between them? So just want to sort this out before approving.

ekcoh (dismissed):
Unfortunately UI tests are failing based on this change it seems, see CI jobs.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
smnwttbr: This deterministic counter replaces the idDictionary, which appeared to be non-deterministic. I cannot work out why this is the case, but simplifying the code solved the issue. The idDictionary is not used anywhere else.

ritamerkl: I think this was written this way to maintain the same Id's for actions/bindings over rebuilds. This condition is not met with this change. I don't remember the particularites why it is solved this way, it might be that there is no urge to do that anymore, but I believe there was a reason to maintain it (at least back then).

ritamerkl (commented):
I would recommend checking if the id's need to be consistent between rebuilds - some functionality may use the id's where the information might get lost assigning new ids.

smnwttbr:
> I would recommend checking if the id's need to be consistent between rebuilds - some functionality may use the id's where the information might get lost assigning new ids.

You are right, I've changed this back to the original behaviour, but with a consistent id that does not depend on order of insertion.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ActionsTreeView.cs
```diff
@@ -659,9 +659,12 @@ public static List<TreeViewItemData<ActionOrBindingData>> GetActionsAsTreeViewDa
 
         private static int GetIdForGuid(Guid guid, Dictionary<Guid, int> idDictionary)
         {
+            //This method is used to ensure that the same Guid always gets the same id
+            //We use getHashCode instead of a counter, as we cannot guarantee that the same Guid will always be added in the same order
+            //There is a tiny chance of a collision, but it is it does happen it will only affect the expanded state of the tree view
             if (!idDictionary.TryGetValue(guid, out var id))
             {
-                id = idDictionary.Values.Count > 0 ? idDictionary.Values.Max() + 1 : 0;
+                id = guid.GetHashCode(); //idDictionary.Values.Count > 0 ? idDictionary.Values.Max() + 1 : 0;
```
ekcoh: Thanks for pointing out the chance of collision here. Maybe I am overlooking something but why don't we just use Guid as ID instead of converting to an int (with risk of collision?

Would recommend inserting space above after the // in the inline comments and also to remove the commented out code.

smnwttbr: The reason for the guid -> int id mapping is that that the UIElements TreeViewItem requires an int id.

ritamerkl (approved):
LGTM now, good to get rid of unused code

Pauliusd01 (approved):
LGTM, I checked expanding/collapsing actions, cut/copy/paste workflow, adding new actions and bindings and binding actual keys to them. The expansion states no longer reset and flash as well selection is maintained properly when using the binding dropdown

PR #2097, opened 2024-12-20
FIX: Set macOS Xbox wireless gamepad button mappings based on PID/VID 

jfreire-unity:
### Description

This PR fixes https://jira.unity3d.com/browse/ISXB-1264 by mapping  "old" Xbox One wireless controllers based on their PID and VID.

Only for macOS since Windows uses XInput.

I also added the Version field to the Input Debugger device view. There might be some devices that get their firmware changed without a PID and VID change, so this could help us track down those devices.

### Testing status & QA
Tested with a Xbox One wireless device which maps to a struct equivalent to Xbox Series.

More devices should be tested with, especially the ones that caused the ISS bug cc @ieuanrees .

### Overall Product Risks

Unclear what other combinations of PID/VID we are missing.

- Complexity: Minimal
- Halo Effect: Low

### Comments to reviewers

**In general, I don't like this approach since I'm creating a lot of duplication.** But I didn't find a better approach with the logic we have. 

Also, we could probably expand the coverage by [using this database](https://github.com/mdqinc/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt#L1094). But I prefer for users to report their devices to us as we don't know if all those devices are still up to date.

A better alternative would be to stop using HID and rely on Apple Game Controller framework. But that requires platform code changes.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
@@ -132,6 +132,8 @@ namespace UnityEngine.InputSystem.Editor
                 EditorGUILayout.LabelField("Product", m_Device.description.product);
             if (!string.IsNullOrEmpty(m_Device.description.manufacturer))
                 EditorGUILayout.LabelField("Manufacturer", m_Device.description.manufacturer);
+            if (!string.IsNullOrEmpty(m_Device.description.version))
+                EditorGUILayout.LabelField("Version", m_Device.description.version);
             if (!string.IsNullOrEmpty(m_Device.description.serial))
                 EditorGUILayout.LabelField("Serial Number", m_Device.description.serial);
             EditorGUILayout.LabelField("Device ID", m_DeviceIdString);
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
@@ -111,7 +111,7 @@ namespace UnityEngine.InputSystem.XInput.LowLevel
         public enum Button
         {
             Start = 11,
-            Select = 10,
+            Select = 16,
             LeftThumbstickPress = 13,
             RightThumbstickPress = 14,
             LeftShoulder = 6,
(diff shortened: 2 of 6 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
```diff
@@ -194,6 +194,98 @@ public XInputControllerWirelessOSXState WithDpad(byte value)
...
+        [InputControl(name = "buttonNorth", bit = (uint)Button.Y, displayName = "Y")]
+
+        [FieldOffset(14)]
+        public uint buttons;
+
+        public FourCC format => kFormat;
+
+        public XInputControllerWirelessOSXStateV2 WithButton(Button button)
+        {
+            Debug.Assert((int)button < 32, $"Expected button < 32, so we fit into the 32 bit wide bitmask");
```
ritamerkl: This message is a bit odd and misses the unit "button < 32," here.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
```diff
@@ -194,6 +194,98 @@ public XInputControllerWirelessOSXState WithDpad(byte value)
             leftStickY = 32767
         };
     }
+
+    [StructLayout(LayoutKind.Explicit)]
+    internal struct XInputControllerWirelessOSXStateV2 : IInputStateTypeInfo
```
ritamerkl, line 199: It is irretating that we need a whole new layout because buttons are messed up, but I also don't see another way.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
```diff
@@ -132,6 +132,8 @@ internal void OnGUI()
                 EditorGUILayout.LabelField("Product", m_Device.description.product);
             if (!string.IsNullOrEmpty(m_Device.description.manufacturer))
                 EditorGUILayout.LabelField("Manufacturer", m_Device.description.manufacturer);
+            if (!string.IsNullOrEmpty(m_Device.description.version))
+                EditorGUILayout.LabelField("Version", m_Device.description.version);
```
ritamerkl, line 136: 👍🏻

ritamerkl (approved):
Looks good to me - just one minor comment about the Assert message.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
ekcoh: Agree, "A maximum of 32 buttons is supported for this layout". Which anyway can only happen if casting an integer (Out of valid range) into button and calling this function? So basically its an ArgumentOutOfRange exception.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
```diff
@@ -194,6 +194,98 @@ public XInputControllerWirelessOSXState WithDpad(byte value)
...
+        public FourCC format => kFormat;
+
+        public XInputControllerWirelessOSXStateV2 WithButton(Button button)
+        {
+            Debug.Assert((int)button < 32, $"Expected button < 32, so we fit into the 32 bit wide bitmask");
+            buttons |= 1U << (int)button;
+            return this;
+        }
+
+        public XInputControllerWirelessOSXStateV2 WithDpad(byte value)
```
ekcoh, line 275: If doing Debug.Assert range checking maybe also have it here to check integrity of value?

ekcoh (approved):
Changes look good to me and thanks for expanding the test coverage. Its a bit unfortunate we need to add a whole class to fix something like this though.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
ekcoh: Agree, if we had a device database being able to query a device based on a multi-variant key to identify it and just fetch button and value remapping we could do this through data only I believe but the current "layout system" is causing issues with that since it has an explicit field definition strongly coupled with preprocessing. If those would be separated it would be easier to achieve. It would require larger changes though.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
jfreire-unity: Ah good point. This was essentially copy pasted. 
I changed the messages but left it as Debug.Assert as I think this will be something to be caught in Debug builds.

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
jfreire-unity: I'm essentially duplicating existing behavior and will maintain this code for now.

stefanunity (approved):
@Pauliusd01 says: Tried: Xbox, PS5, PS4 controllers on Win11 and MacOS player/playmode visualiser as well as the debugger. Seemed fine to me.

Also checked with Xbox pad I have and looked good.

PR #2102, opened 2025-01-09
FIX: An infinite loop in `InputSystemUIInputModule` was causing an editor / player hang

Secticide:
### Description

Fixes [ISXB-1258](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1258)

This was caused by the assumption that `RemovePointerAtIndex` would _always_ successfully remove the pointer, which is not the case with touch based pointers.

The following code highlights the issue:
```cs
private void PurgeStalePointers()
{
    for (var i = 0; i < m_PointerStates.length; ++i)
    {
        ref var state = ref GetPointerStateForIndex(i);
        var device = state.eventData.device;
        if (!device.added || // Check if device was removed altogether.
            (!HaveControlForDevice(device, point) &&
             !HaveControlForDevice(device, trackedDevicePosition) &&
             !HaveControlForDevice(device, trackedDeviceOrientation)))
        {
            SendPointerExitEventsAndRemovePointer(i);
            --i;
        }
    }
```

If `SendPointerExitEventsAndRemovePointer` fails to remove the pointer from the list; decrementing `i` is incorrect and will lead to attempting to remove the pointer at the same index again, which will fail and we'll loop ad infinitum.

### Testing status & QA

I've tested this locally on a Windows 11 touch-enabled machine (laptop), but it would be great to get some testing across mobile devices (including Switch) to ensure no additional issues were introduced.

### Overall Product Risks

- Complexity: 0
- Halo Effect: 1

### Comments to reviewers

The only remaining concern I have with this solution is with the `OnDisable` code. The first thing we call in `OnDisable` for the Input system UI module is `ResetPointers` - how important is it that the pointers are all 100% reset here? If it is an absolute must, we could always add a _force_ flag as a parameter to the `SendPointerExitEventsAndRemovePointer` method.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -1662,9 +1662,11 @@ namespace UnityEngine.InputSystem.UI
 
         private void ResetPointers()
         {
-            var numPointers = m_PointerStates.length;
-            for (var i = 0; i < numPointers; ++i)
-                SendPointerExitEventsAndRemovePointer(0);
+            for (var i = 0; i < m_PointerStates.length; ++i)
+            {
+                if (SendPointerExitEventsAndRemovePointer(i))
+                    --i;
+            }
 
             m_CurrentPointerId = -1;
             m_CurrentPointerIndex = -1;
@@ -2021,16 +2023,17 @@ namespace UnityEngine.InputSystem.UI
             return m_PointerStates.AppendWithCapacity(new PointerModel(eventData));
         }
 
-        private void SendPointerExitEventsAndRemovePointer(int index)
+        // Returns true if the pointer was successfully removed (ISXB-1258)
+        private bool SendPointerExitEventsAndRemovePointer(int index)
         {
             var eventData = m_PointerStates[index].eventData;
             if (eventData.pointerEnter != null)
                 ProcessPointerMovement(eventData, null);
 
-            RemovePointerAtIndex(index);
+            return RemovePointerAtIndex(index);
         }
 
-        private void RemovePointerAtIndex(int index)
+        private bool RemovePointerAtIndex(int index)
         {
             Debug.Assert(m_PointerStates[index].eventData.pointerEnter == null, "Pointer should have exited all objects before being removed");
 
@@ -2038,7 +2041,8 @@ namespace UnityEngine.InputSystem.UI
             ref var state = ref GetPointerStateForIndex(index);
             if (state.pointerType == UIPointerType.Touch && (state.leftButton.isPressed || state.leftButton.wasReleasedThisFrame))
             {
-                return;
+                // The pointer was not removed
+                return false;
             }
 
             // Retain event data so that we can reuse the event the next time we allocate a PointerModel record.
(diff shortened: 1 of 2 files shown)


jfreire-unity:
Good catch! I can have a look today or Monday 👍

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1662,9 +1662,11 @@ protected override void OnDisable()
...
         private void ResetPointers()
         {
-            var numPointers = m_PointerStates.length;
-            for (var i = 0; i < numPointers; ++i)
-                SendPointerExitEventsAndRemovePointer(0);
+            for (var i = 0; i < m_PointerStates.length; ++i)
+            {
+                if (SendPointerExitEventsAndRemovePointer(i))
+                    --i;
+            }
```
jfreire-unity, line 1669: Do you think this needs to change? It feels like the infinite loop issue can only occur in the other changes you make, so maybe we could leave it as is here.

jfreire-unity (commented):
Looks ok, good find!
Just added a question before approving. Also added QA to test this and make sure Touch functionality has no regression.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
Secticide: I believe it does need to change, while it won't cause the infinite loop we're seeing elsewhere; if the pointer isn't removed from the list we'll just call `SendPointerExitEventsAndRemovePointer` on it several times rather than moving forward in the list.

This loop _expects_ the call to `SendPointerExitEventsAndRemovePointer` to succeed and for the pointer to be removed, essentially calling `pop_front`. Does that make sense?

Pauliusd01:
Checking

Pauliusd01:
Hi @MahumKhan , could you help with checking the switch platform?

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
jfreire-unity: I see, thanks for clarifying!

jfreire-unity (approved):

stefanunity (approved):
Helped with some testing:
Can reproduce the bug and it looks fixed with this PR.
TouchSamples project also still works.

Tested on: 
- Win11 touchscreen
- Pixel 6a
- Unity 22.3 + 6.0

Pauliusd01 (approved):
LGTM, checked touch + PS5/Xbox input on the user repro project as well as the internal Touch Event/Sensor project on Mi 8 Lite, iPhone 8 and did not notice any unexpected issues or crashes

cristian-murarescu (approved):
Tested this on Switch with and without the fix, can confirm that without installing this fixed package, the project freezes. With it installed, it doesn't freeze. 

Tested by spamming A and touching the white square on the screen repeatedly.

PR #2107, opened 2025-01-15
FIX: Delayed input backend prompt to make sure asset import is not interrupted (ISXB-608)

ritamerkl:
### Description

Like described in this ticket [ISXB-608](https://jira.unity3d.com/browse/ISXB-608) the prompt to activate the InputSystem backend is delayed to after assets are imported. This means the prompt will appear once the editor launched, and in case of the package being imported after all assets loaded in the current domain reload.

FYI: The delayCall is called AFTER domain reloads when the UI inspectors update.

The dependency to the InputSystem (for affected assets eg Enemies) should be updated to the latest version after this change.

### Testing status & QA

I tested this for editor launches and importing the package in newly created projects. Testing this in combination with importing assets is cumbersome, since the import happens based on the version define of the asset dependency. 

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Pauliusd01:
One thing that I'm not sure is expected is this scenario
- I made a new project -> Removed the default IS package and set the backend to Input Manager -> Installed Input System back (your PR) -> Did not get the prompt at all
- Rebooted editor -> now I get the prompt
Was this an expected compromise?

Pauliusd01 (approved):
LGTM, here's what I've checked:

- Adding just the Input System package while the Editor is open/closed
- Bundling the Input System package with some large, valid assets to import (Editor open) - Pass - packages/assets were imported first -> only then I got the dialog
- Bundling the Input System package with some large, valid assets to import (Editor closed) - Pass - packages/assets were imported first -> Editor opened and I got the dialog
- Bundling the Input System package with some erroring out assets (Editor open) - Pass
- Bundling the Input System package with some erroring out assets (Editor closed) - Pass - safe mode did not interfere with the dialog order
- Reimport All option (Input Manager backend) - Pass - got the prompt on reimport
- Reimport All option (Input System backend) - Pass - no unnecessary prompt or backend resets

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3599,14 +3599,29 @@ internal static void InitializeInEditor(IInputRuntime runtime = null)
             }
 
             Debug.Assert(settings != null);
-            #if UNITY_EDITOR
             Debug.Assert(EditorUtility.InstanceIDToObject(settings.GetInstanceID()) != null,
```
ekcoh, line 3602: Why was the guards around this removed? EditorUtility would not be available outside of editor? Was there an outer guard so it was a duplicate?

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3599,14 +3599,29 @@ internal static void InitializeInEditor(IInputRuntime runtime = null)
...
                 "InputSettings has lost its native object");
-            #endif
 
             // If native backends for new input system aren't enabled, ask user whether we should
             // enable them (requires restart). We only ask once per session and don't ask when
             // running in batch mode.
+            // The warning is delayed to delay call (called a short while after the Asset are loaded, on Inspector update) to make sure it doesn't pop up while the editor is still loading or assets are not fully loaded -
+            // this would cancel the import of large assets that are dependent on the InputSystem package and import it as a dependency.
+            EditorApplication.delayCall += ShowRestartWarning;
```
ekcoh, line 3610: Does this guarantee that it will execute after asset import? I would have expected this to be triggered by an [AssetP](https://docs.unity3d.com/6000.0/Documentation/ScriptReference/AssetPostprocessor.OnPostprocessAllAssets.html) since this one will simply but the callback on the editor event queue? 
If it works its great, but felt I was not seeing how this is robust.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3599,14 +3599,29 @@ internal static void InitializeInEditor(IInputRuntime runtime = null)
...
+
+#if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
```
ekcoh, line 3612: Would also be great to understand why this was moved out of its previous place?

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ritamerkl: yes, it was a duplicate. The function call InitializeInEditor is surrounded by defines.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ritamerkl: The AssetPostProcessor was an other option I considered, but it requires another class to inherit from AssetPostprocessor, so I decided for delayedCall instead. delayCall is called after domain reloads (which include importing of assets at the end) when the inspector windows refresh, which I think is suitable for a prompt. I don't see why it's not robust, in which way do you think it couldn't be?

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ritamerkl: I put it into a private function of the same class to make it more readable, I think having big chunks of code in single functions do no good for the readability some times. (technically still in the same place though)

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: OK, great to get rid of it then

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: I am not sure I see the function you are referring to? What I meant was basically that previously EnableActions(), RunInitialUpdate(), was executed from within ShowRestartWarning() as part of the deferred ShowRestartWarning. Now it executes directly in InitializeInEditor (before) code in ShowRestartWarning(). Hence why I asked if it still provides the same behavior or what the reason was for relocating it / changing the execution order.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ekcoh: I see, yeah we have some classes doing that as part of InputActionImporter. If delayCall runs after domain reloads it's good, but to me it wasn't clear it runs after asset import (which might not trigger a domain reload if no code changed?). At least it's not explicit from https://docs.unity3d.com/6000.0/Documentation/ScriptReference/EditorApplication-delayCall.html, so it seems to relate to inspectors but not necessarily asset imports? Maybe you know something I do not in this area, but to me it wasn't clear that e.g. delayCall could be executed between two imports.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ritamerkl: It all is still in the same order, it may look a bit confusing just looking at the diffs, but the only thing I changed is outsourcing the logic of the prompt to a new private function called ShowRestartWarning()

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
ritamerkl: Since InitializeInEditor is called from the static constructor it's only executed when the editor launches or the package gets added into the project (afaik - correct me if I forgot a case), in both cases there is a domain reload happening. 
Here a short explanation why I think the call is reliable: 
The callbacks in delayCall are invoked in the subsequent editor update.
Editor updates do not happen during a domain reload (which includes import of assets - no matter how many), Editor updates resume only after the domain reload is complete.
One reason for this is that Editor updates depend on stable managed objects and assemblies. Since these objects are being reloaded, updates during this period could result in crashes or undefined behavior.

ekcoh (approved):

PR #2119, opened 2025-01-30
FIX: ISXB-1221 input action binding discarded

smnwttbr:
### Description

Fix for [ISXB-1221](https://jira.unity3d.com/browse/ISXB-1221) where the chose input action binding can be discarded and not saved with the asset in different circumstances.

The root cause of this issue is the two-asset-solution we have, where we edit a copy of the asset using a serializedObject, then write that out to disk, then reload a new copy of the asset into a new serializedObject to continue editing.

This causes our UITK editor to be rebuilt and rebound to the new asset on every user interaction, which causes this issue. The workaround involves storing a reference in a static variable which is updated whenever the editor is rebuilt.

### Testing status & QA

### Overall Product Risks

- Complexity: 0
- Halo Effect: 0

### Comments to reviewers

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Pauliusd01:
@smnwttbr seeing a lot of CI failures, is this one's ready?

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
```diff
@@ -89,10 +90,10 @@ public void OnGUI()
             EditorGUILayout.EndHorizontal();
         }
 
+        //TODO: on next major version remove property argument.
```
ekcoh, line 94: I would recommend filing a JIRA ticket on this if/when this lands, otherwise it would be forgotten.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
```diff
@@ -200,7 +195,16 @@ private void SetExpectedControlLayoutFromAttribute(SerializedProperty property)
                 SetExpectedControlLayout(attribute.layout);
         }
 
-        public SerializedProperty pathProperty { get; }
+        // This static variable is a hack. Because the editor is rebuilt at unpredictable times with a new serializedObject, we need to keep updating
+        // this variable with most up to date serializedProperty, so that the picker dropdown can access the correct serializedProperty.
+        // The picker dropdown is a separate window and does not have access to the changed serializedObject reference.
+        // This could be removed if the InputControlPathEditor is converted to UITK with a stable, persistent serializedObject backing this editor.
+        // This property will be shared among multiple asset editor windows.
+        private static SerializedProperty s_pathProperty { get; set; }
```
ekcoh, line 204: This static SerializedProperty, wouldn't it leak memory? Its referencing a SerializedObject that was dynamically allocated and may/would be disposed from outside this editor?

ekcoh: Additionally this assumes there can only exist one InputControlPathEditor instance right?

ekcoh (commented):
Thanks for looking into this. I think you are right about the underling root cause and this may (or may not) be a bit of duck taping on top of the underlying problem. It's great if this resolves the issue but I am still thinking it may have other side-effects but maybe I am wrong on this. I added some comments / questions as a start on the review process.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
smnwttbr: Yes, there will be a dangling reference tht may hang around. I'll add some code to clean it up when the editor is closed.

Yes, only one instance of the InputControlPathEditor will work at a time. I think this is ok, as it is a popup window, and disappears when it loses focus.

smnwttbr: https://jira.unity3d.com/browse/ISX-2257

ekcoh (approved):

Pauliusd01 (approved):

PR #2123, opened 2025-02-03
DOCF-6396 - Processor content model re-org

siobhan-unity:
Re-org for Processor documentation.


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=2123) <br/>All committers have signed the CLA.

Packages/com.unity.inputsystem/Documentation~/introduction-to-processors.md
```diff
@@ -0,0 +1,37 @@
...
+    "invert,normalize(min=0,max=10)"
+```
+
+## Processors on bindings and actions
+
+When you create bindings for your [actions](actions.md), you can choose to add processors to the Bindings. These process the values from the controls they bind to, before the system applies them to the Action value. For example, you could invert the `Vector2` values from the controls along the Y axis before passing the values to the associated Action. 
+
+Processors on Actions work the same way, but affect all bindings on an Action. 
+
+If there are processors on both the Binding and the Action, the Input System processes the processors from the binding first.
```
jo-unity: Mixture of B(b) and A(a) for bindings and actions

Packages/com.unity.inputsystem/Documentation~/built-in-processors.md
```diff
@@ -1,25 +1,30 @@
-# Predefined Processors
+# Built-in Processors
```
jo-unity, line 1: Do we even need this page? it's repeating information that's already in the API docs.

I'd be tempted to bin it and add something to the intro doc that's like:

```
## Built in processors

The Input System package comes with a set of built-in processors, defined in the `Processors` class, which you can use with [bindings](ActionBindings.md), [actions](actions.md) and [controls](controls.md). For a list of processors, refer to the [`Processors` API documentation](../api/UnityEngine.InputSystem.Processors.html)

```

@duckets what do you think?

Packages/com.unity.inputsystem/Documentation~/write-custom-processors.md
```diff
@@ -1,8 +1,10 @@
-# Writing custom Processors
+# Writing custom processors
```
jo-unity: In the toc this is "Write" rather than "Writing"

Packages/com.unity.inputsystem/Documentation~/write-custom-processors.md
```diff
@@ -1,8 +1,10 @@
...
+You can write custom processors to use use with [bindings](ActionBindings.md), [actions](actions.md) and [controls](controls.md) in your Project. Custom processors are available in the UI and code in the same way as the [built-in processors](built-in-processors.md). 
 
->__IMPORTANT__: Processors must be __stateless__. This means you cannot store local state in a processor that will change depending on the input being processed. The reason for this is because processors are not part of the [input state](./Controls.md#control-state) that the Input System keeps.
+To create a custom processor:
+
+**1.** Add a class derived from [`InputProcessor<TValue>`](../api/UnityEngine.InputSystem.InputProcessor-1.html), and implement the [`Process`](../api/UnityEngine.InputSystem.InputProcessor-1.html#UnityEngine_InputSystem_InputProcessor_1_Process__0_UnityEngine_InputSystem_InputControl_) method:
```
jo-unity, line 13: To get around the awkward formatting with this, I'd suggest doing something like creating a brief bullet point list after "to create a custom processor:" and then adding H2s before each step, which you could link to in the bullet list.

I think you _can_ make this a "proper" list, but you'd probably have to play around with a lot of indenting to get all the code examples to play ball

Packages/com.unity.inputsystem/Documentation~/add-processors-controls.md
```diff
@@ -0,0 +1,38 @@
...
+    // smaller then 0.2, as our Control does not have a stable
+    // resting position.
+    [InputControl(layout = "Axis", processors = "AxisDeadzone(min=0.2)")]
+    public short axis;
+}
+```
+
+If you [create a layout from JSON](layouts.md#layout-from-json), you can specify processors on your Controls like this:
+
+```CSharp
```
jo-unity:
```suggestion
```json
```

Packages/com.unity.inputsystem/Documentation~/TableOfContents.md
```diff
@@ -46,7 +46,12 @@
     * [Controls schemes](control-schemes.md)
     * [Interactions](interactions.md)
     * [Processors](processors.md)
-    * [Configure sctions](configure-actions.md)
+      * [Introduction to processors](introduction-to-processors.md)
+      * [Built-in processors](built-in-processors.md)
+      * [Write custom processors](write-custom-processors.md)
+      * [Add processors to bindings and actions](add-processors-bindings-actions.md)
+      * [Add processors to controls](add-processors-controls.md)
+    * [Configure actions](configure-actions.md)
```
jo-unity: This markdown file doesn't seem to be in this PR

siobhan-unity: I didn't add that line, just the stuff under the Processors section - I think this might be a placeholder for another section (on a different jira ticket).

Packages/com.unity.inputsystem/Documentation~/write-custom-processors.md
siobhan-unity: Ah, great solution, thank you!

Packages/com.unity.inputsystem/Documentation~/built-in-processors.md
duckets: I think it would be useful to have a much smaller simpler page with a list or table of the built-in processors, each with a single line describing what they are and a link to the API.

Also need to check that the API description is at last as good as the description from the manual, and replace if not - for example i looked at "Invert Vector 3" and the manual says:

_Inverts the values from a Control (that is, multiplies the values by -1). Inverts the x axis of the vector if invertX is true, the y axis if invertY is true, and the z axis if invertZ is true._

Whereas the API says:

_Inverts the x and/or y and/or z channel of a Vector3._

The manual content is better, so we should replace the API text with that.

duckets (changes requested):
One change, which is to simplify the built-in processors page

Packages/com.unity.inputsystem/Documentation~/built-in-processors.md
siobhan-unity: Ben's suggestions works for me.

There is some information in this page that isn't in the API docs, and I"m not sure if it's more appropriate for the user manual or the API docs - specifically I'm talking about the paragraphs in lines 112 and 113. What would you suggest for those?

PR #2124, opened 2025-02-05
FIX: ISXB-1343 show dialog for unsaved changes

smnwttbr:
### Description

A check for an existing and dirty asset was added to the OpenWindow(asset) method. If a dirty asset is loaded, a dialog is presented allowing the user to save changes, cancel the operation or discard changes and continue.

### Testing status & QA

Package tests.

### Overall Product Risks

- Complexity: 0
- Halo Effect: 0

### Comments to reviewers

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
@@ -100,6 +100,28 @@ namespace UnityEngine.InputSystem.Editor
             }
 
             var window = GetWindow<InputActionsEditorWindow>();
+            if (window.m_IsDirty)
+            {
+                var assetPath = AssetDatabase.GUIDToAssetPath(window.m_AssetGUID);
+                if (!string.IsNullOrEmpty(assetPath))
+                {
+                    // Prompt user with a dialog
+                    var result = Dialog.InputActionAsset.ShowSaveChanges(assetPath);
+                    switch (result)
+                    {
+                        case Dialog.Result.Save:
+                            window.Save(isAutoSave: false);
+                            break;
+                        case Dialog.Result.Cancel:
+                            return window;
+                        case Dialog.Result.Discard:
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException(nameof(result));
+                    }
+                }
+            }
+
             window.m_IsDirty = false;
             window.minSize = k_MinWindowSize;
             window.SetAsset(asset, actionToSelect, actionMapToSelect);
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -100,6 +100,28 @@ private static InputActionsEditorWindow OpenWindow(InputActionAsset asset, strin
             }
 
             var window = GetWindow<InputActionsEditorWindow>();
+            if (window.m_IsDirty)
```
ritamerkl, line 103: Don't we have such a logic somewhere else to reuse already (showing the dialog with three different answer options)?

smnwttbr: There is a nice self contained method to do this in the old editor which returns a bool, however the new editor does not use this, it has to handle the 3 different options specifically. (Save, Discard and Cancel) This requires the switch statement to handle the 3 options, and cannot be simplified with booleans.

The reasons for this:

1. The UITK EditorWindow does not use the built in SaveChanges/DiscardChanges methods and hasUnsavedChanges property which are part of the EditorWindow API because of the double-asset stuff we are doing.

2. This means we cannot use the built-in dialog which is shown when the window is closed.

3. Instead, we use the OnDestroy method to detect dirty changes and show a dialog. However, once this method has been called, there is no way to prevent the window from being closed... So if the user decides to cancel the dialog, our code will actually relaunch a new editor with the unsaved changes.

In the case of OnLoadAsset, we don't need to relaunch the editor, we can simply avoid loading a new asset into the window.

ritamerkl: Okay, I see! Thanks for explaining

ritamerkl (approved):

Pauliusd01 (approved):
There were some CI failures so might want to check those out, otherwise lgtm. Checked opening a different action window with both manual save/auto save options after changing schemes,maps,actions,bindings, etc.

Secticide:
I noticed my previous change #2149 added a changelog entry to the previous release, fixed up as part of 2677b68a5ffa6394bffe868574d20dfe63185b43.

PR #2129, opened 2025-02-07
FIX: "ArgumentNullException: Value cannot be null." when deleting a newly created input action (ISXB-1405)

jfreire-unity:
### Description

Fixes https://jira.unity3d.com/browse/ISXB-1405.

The problem ocurred when creating a new action and deleting it before finishing "renaming it". The delete triggers a focus change which calls the event to "rename" the action. 
Since at this stage, the action is in "renaming mode" (`m_IsEditing` == `true`) and the action was deleted, the search for that action serialized property would return "null". That case was not being considered before.

### Testing status & QA

Local testing on macOS trying to replicate the problem.
Creating new actions with a name and removing them also works.

### Overall Product Risks

- Complexity: 0
- Halo Effect: 0

### Comments to reviewers

I'd only ask to do some more exploratory work around renaming and creating actions in both Project Settings > Input System and a standalone Asset Editor view.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -10,6 +10,9 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
 
+### Fixed
+- Fixed an issue where removing a newly created action in the Asset Editor would cause an exception. [UUM-95693](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-95693)
+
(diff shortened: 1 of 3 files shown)


ritamerkl (approved):
Looks good! Thanks

Pauliusd01 (approved):

PR #2134, opened 2025-02-12
NEW: add touch test measuring input system profiler markers

stefanunity:
### Description

- see title. 
- similar test to MeasureInputSystemFrameTimeWithProfilerMarkers_FPS() just with touch instead of mouse&keyboard input

### Testing status & QA

- test runs
- data submitted to hoarder

### Overall Product Risks
- none, internal tests

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


jfreire-unity (commented):
Looks good to me.
Maybe add some top level comments to explain what it does for this particular test?

One suggestion will to be clear on of what we want to measure this against to. I know we already do test doing nothing, doing keyboard and mouse input in `Performance_MeasureInputSystemFrameTimeWithProfilerMarkers_FPS` but it contains actions so it's not a 1 to 1 comparison. Should this test also contain some actions setup?

Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -1207,5 +1207,58 @@ public IEnumerator Performance_MeasureInputSystemFrameTimeWithProfilerMarkers_Do
...
+
+        clickAction.performed += context => {
+            performedCallCount++;
+        };
+
+        pointAction.performed += context => {
+            performedCallCount++;
+        };
+
+        using (Measure.ProfilerMarkers(allInputSystemProfilerMarkers))
```
ekcoh, line 1236: I guess changes look ok here but I wonder what scenario we are trying to replicate here? It seems that there is not many touch points per frame here, essentially looks like one touch update per frame. That is fine given a e.g. 30 or 60 Hz touch sensor , but I believe most these days generate higher frequency of touch. It might be meaningful to measure a decently modern touch screen and take the actual achievable frequency. I wouldn't trust just reading specs since the HW frequency might be higher than what you get out of the driver.

ekcoh (approved):
Approving this as is but I think we should consider relevance of scenario.

Assets/Tests/InputSystem/CorePerformanceTests.cs
ekcoh: It be good to outline the intended scenario as clearly as possible in the test description or inline (Seems like @jfreire-unity suggested something similar.

ekcoh:
Is this still work in progress or why Draft status?

PR #2140, opened 2025-02-19
FIX: Xbox Controllers on macOS don't detect up-left DPAD presses (ISXB-810)

jfreire-unity:
### Description

Fixes https://jira.unity3d.com/browse/ISXB-810.

Dpad presses for up-left were not being detected with `IsPressed()` function because [we always compare it against their default state first](https://github.com/Unity-Technologies/InputSystem/blob/refs/heads/develop/Packages/com.unity.inputsystem/InputSystem/Controls/InputControlExtensions.cs#L96) to avoid calling the `EvaluateMagnitude()` virtual method.

On Xbox Controllers that I tested with (Xbox One Conteoller, Model 1708, Bluetooth), their default state is 0 for Dpad values. And byte value 8 was being considered as such before, which in fact corresponds to Dpad up-left presses.  This PR fixes that.
 

### Testing status & QA

@stefanunity and @Pauliusd01 could you test this issue all the Xbox controllers you can find, with both Wired and Wireless connections? Just to make sure there are no devices with a different default state for DPAD presses.

### Overall Product Risks

- Complexity: 0
- Halo Effect: 0

### Comments to reviewers

There's a chance that at some point the default DPAD state was 8 but I can't really tell as I haven't used that much Xbox controllers on macOS. 
I did validate that Dualshock 4 uses value 8 as the default state for DPAD.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
@@ -149,7 +149,7 @@ namespace UnityEngine.InputSystem.XInput.LowLevel
         [InputControl(name = "rightTrigger", format = "USHT", parameters = "normalize,normalizeMin=0,normalizeMax=0.01560998")]
         [FieldOffset(11)] public ushort rightTrigger;
 
-        [InputControl(name = "dpad", format = "BIT", layout = "Dpad", sizeInBits = 4, defaultState = 8)]
+        [InputControl(name = "dpad", format = "BIT", layout = "Dpad", sizeInBits = 4)]
         [InputControl(name = "dpad/up", format = "BIT", layout = "DiscreteButton", parameters = "minValue=8,maxValue=2,nullValue=0,wrapAtValue=9", bit = 0, sizeInBits = 4)]
         [InputControl(name = "dpad/right", format = "BIT", layout = "DiscreteButton", parameters = "minValue=2,maxValue=4", bit = 0, sizeInBits = 4)]
         [InputControl(name = "dpad/down", format = "BIT", layout = "DiscreteButton", parameters = "minValue=4,maxValue=6", bit = 0, sizeInBits = 4)]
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
```diff
@@ -149,7 +149,7 @@ public enum Button
         [InputControl(name = "rightTrigger", format = "USHT", parameters = "normalize,normalizeMin=0,normalizeMax=0.01560998")]
         [FieldOffset(11)] public ushort rightTrigger;
 
-        [InputControl(name = "dpad", format = "BIT", layout = "Dpad", sizeInBits = 4, defaultState = 8)]
```
ritamerkl, line 152: Is there a history of this (defaultState = 8)? Why was this implemented like this in the first place? I think the change makes totally sense if this is just derived from another control.

ekcoh (approved):
Changes look good to me, I wonder though if this changed with some Xbox firmware update or if the backend changed behavior or if this has always been broken?

Pauliusd01 (approved):
LGTM, all controls worked fine in play mode and player. Although I only had 1 controller that I checked with (1914) and only on bluetooth as wired isn't supported there natively

ritamerkl (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/XInput/XboxGamepadMacOS.cs
jfreire-unity: I didn't find any history in this.
I'm not sure this was a copy paste issue from controls such as the DualShock controllers who actualkly have default state = 8. I guess we need to wait and see if there are any Xbox controllers with such default state...

jfreire-unity:
> Changes look good to me, I wonder though if this changed with some Xbox firmware update or if the backend changed behavior or if this has always been broken?

I cannot answer if it was a firmware update and didn't find any history regarding this. But there's a chance that have might been the case. Let's monitor this.
But I also wouldn't completely disregard this as always being broken, as it only happens for `IsPressed()` calls. But I'm just guessing here.

PR #2142, opened 2025-02-19
NEW: Added achievable frequency and latency (processing delay) to Input Debugger (ISX-2254)

ekcoh:
### Description

Adds achievable sample/event frequency (as well as sensor frequency and global polling frequency) to Input Debugger device window. These are just basic calculations over periods of one second at the moment similar to FPS counter.
Adds input processing delay (average, min, max) to Input Debugger device window.

The diagnostic values are similar to FPS updated once a second to show primarily the average. This is intended to help with diagnosing the behaviour of the system and devices and help spot problems.

Added 11 new test cases to cover the calculators as part of editor test assembly which required modification to accept unsafe code in the test assembly which I believe should be ok to change.

Target frequency shown is based on IOCTL sample frequency query response which I believe is only supported by sensors currently. Polling frequency is as we know global at the moment, but showing it since it relates to devices with polling backends which we currently cannot determine from the system. If both of these could be capability queried in the future we can make a better message here.

Screenshot 2025-02-20 at 10 57 35

### Testing status & QA

Tested manually with mouse, keyboard and gamepad on macOS during development.
Added unit tests to verify that calculations are working as expected.

### Overall Product Risks

Very small. Has no impact on production or player code.

- Complexity: 0
- Halo Effect: 0

### Comments to reviewers

If possible to cover sensors in review/QA it would be great.

Note that both label fields have tooltip that also needs review.

### Checklist

Before review:

- [x] Changelog entry added.
    - Functional tests: 11 new test cases added to `InputLatencyCalculatorTests` and `SampleFrequencyCalculatorTests`.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
```diff
@@ -35,7 +37,7 @@ namespace UnityEngine.InputSystem.Editor
 {
     // Shows status and activity of a single input device in a separate window.
     // Can also be used to alter the state of a device by making up state events.
-    internal sealed class InputDeviceDebuggerWindow : EditorWindow, ISerializationCallbackReceiver, IDisposable
+    internal sealed partial class InputDeviceDebuggerWindow : EditorWindow, ISerializationCallbackReceiver, IDisposable
```
ritamerkl: Where is the class splitted? I only see it here, is this a leftover of a previous change?

ritamerkl (commented):
It's a great to have, thanks for adding this! 
I only did a quick run through but it looks good from my POV. I need to take another closer look and try it myself.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
ekcoh: Thanks, yeah I think Rider added it when I extracted calculators into their own files, I will remove this diff.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
```diff
@@ -143,6 +146,18 @@ internal void OnGUI()
...
+                "The Polling Frequency indicates system polling target frequency.";
+            if (!string.IsNullOrEmpty(m_DeviceFrequencyString))
+                EditorGUILayout.LabelField(new GUIContent("Sample Frequency", sampleFrequencyTooltip), new GUIContent(m_DeviceFrequencyString), EditorStyles.label);
+            const string inputSystemLatencyTooltip =
+                "Displays the average/minimum/maximum observed input processing latency " +
+                "(Excluding OS, driver, firmware or transport latency) for data reported for this device. " +
+                "Note that additional experienced latency (a.k.a. input lag) will be introduced if any deferred " +
+                "processing and output delay (rendering, display refresh latency etc).";
+            if (!string.IsNullOrEmpty(m_DeviceLatencyString))
+                EditorGUILayout.LabelField(new GUIContent("Input Latency", inputSystemLatencyTooltip), new GUIContent(m_DeviceLatencyString), EditorStyles.label);
```
ekcoh: I consider renaming this from "Input Latency" to "Processing Delay" to more accurately specify its the latency between first seeing the sample/event in native to actually receiving it. Thoughts?

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/SampleFrequencyCalculator.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    /// <summary>
+    /// A utility for computing sample frequency based on input events.
+    /// </summary>
+    internal struct SampleFrequencyCalculator
+    {
+        private double m_LastUpdateTime;
+        private int m_SampleCount;
+
+        public SampleFrequencyCalculator(float targetFrequency, double realtimeSinceStartup)
```
ekcoh, line 13: Target frequency is currently only a "convient storage" here. Hower I would suggest keeping it inside this type since an evolved implementation will likely use it to compute error compared to target and create a CDF or PDF of it to illustrate the applications or architectures  ability to promptly react to stimuli.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
ritamerkl: Agreed, I think ProcessingDelay or InputProcessingDelay would be a better fit.

ekcoh: Maybe "Processing Delay" will do then since we are already inside "Input Debugger"

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
ekcoh: This has been adressed in https://github.com/Unity-Technologies/InputSystem/pull/2142/commits/40948dbb4837c44a989fa41a0a95ea8950499f8b

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
ekcoh: This has been addressed in https://github.com/Unity-Technologies/InputSystem/pull/2142/commits/8af0ad199b0974844ca8ad4cdd59dc041427f681

ekcoh: Will mark this resolved now

surfnerd:
I think this is great step toward transparency of processing of input events.  I think it should be made clear that this doesn't measure the latency between the creation of an input event to when it shows up on the screen, but from the creation of the event to when it's processed by the input system.

ekcoh:
> I think this is great step toward transparency of processing of input events. I think it should be made clear that this doesn't measure the latency between the creation of an input event to when it shows up on the screen, but from the creation of the event to when it's processed by the input system.

Agreed, that's a much simpler/better way to say what I am trying to say in that tool tip. Also why I renamed the label to "Processing Delay". Will try to improve the tooltip based your description here. Thanks! :)

surfnerd (approved):

surfnerd:
Perhaps the [manual](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.13/manual/Debugging.html) could be updated with the new screenshot, and a little blurb about what value this adds could be helpful as well.

Pauliusd01:
Is it expected that I'm seeing processing delay this high when plugging in an android device? Is it because It's a "remotely" connected one?

27 02 2025 - Unity 138
27 02 2025 - Unity 139

Pauliusd01:
Also, the duelsense controllers have a really high sample frequency, but that might be due to the them being very noisy?
27 02 2025 - Unity 141

ekcoh:
> Is it expected that I'm seeing processing delay this high when plugging in an android device? Is it because It's a "remotely" connected one?
> 
> 27 02 2025 - Unity 138 27 02 2025 - Unity 139

You probably found another issue with remote devices having some unusable timestamps. This is probably a root of other bad things to happen. Great find. Did you use Unity remote to get this happening?

I decided to cap the "age" at 1000.0 ms and this is what you see. However this is only based on current time - event timestamp. If event timestamp comes from another clock (remote) device, this doesn't make sense, so it probably means remote devices are not re-timestamped when coming in to the host or the remote protocol do not allow sending the remote device time to allow computing the true latency (including network latency). I think its unrelated to the PR, but we might to either filter remote ones out for now until we have figured that part out or do something similar.... I didn't try remote when doing this.

Pauliusd01:
>Did you use Unity remote to get this happening?

The steps to repro are: Connect an android device and build with the development build type  -> In the Input Debugger click on the Remote Devices dropdown and pick your android device -> in the android devices list double click any device -> Processing Delay is always above 1k ms

ekcoh:
> Also, the duelsense controllers have a really high sample frequency, but that might be due to the them being very noisy?

Not really, on Windows which was the platform you tested on, DualSense is processed over HID. The current HID implementation do not have any rate-adoption (apart from max read requests in flight I believe) so it will run as fast as the underlying HID driver allows it to do. It's a good observation since this shows processing differences in platform-dependent code that could be up for discussion. However, you are correct that the noise is part of it as well since the noise will allow events/samples to be forwarded since the state "changes". If there was no noise or if signal was pre filtered it would reduce the frequency while the device is idle.

ekcoh:
E.g. on macOS where I am currently working with another DualSense bug, the achieved frequency is around 65 Hz. So 1/8 compared to Win at the moment.

Pauliusd01 (approved):
Checked debugger window for these devices: touch, keyboard, mouse, PS and Xbox controllers and commonly used android sensors.

ekcoh:
Thanks for the rigorous testing of this debug feature @Pauliusd01

ritamerkl (approved):
Approving to unblock this, will add more thoughts next week !

ritamerkl:
I tested the newly added Input Debugger features myself now, and think it is priceless. The only wish I would have is to name target/ Polling Frequency with a bit more clarity or adding an explanation with a small text underneath, or a hover textbox? For me the meaning wasn't clear initially (your live demo on friday was very helpful) - also for the restrictions eg when n/a shows up and what this means.

ekcoh:
> I tested the newly added Input Debugger features myself now, and think it is priceless. The only wish I would have is to name target/ Polling Frequency with a bit more clarity or adding an explanation with a small text underneath, or a hover textbox? For me the meaning wasn't clear initially (your live demo on friday was very helpful) - also for the restrictions eg when n/a shows up and what this means.

I agree, the UI layout isn't very helpful, maybe I should try revisit the IMGUI design here to at least make it a bit better (I think the debugger window is overpopulated to begin with, contains a mix of static and dynamic data and should likely be split into multiple windows and/or perspectives). 

Regarding the N/A and Target polling rate I would propose the following:
- Add a new IOCTL command allowing to query whether a device is subject for polling or not. If it's not, there is no point in showing Polling Frequency. This would be trivial if done on module level and could even be a flag. Will investigate.
- Only show target if the device responds that it has a device-specific target frequency. This will only show something for sensors currently. In the future I hope all devices can be device-specific. 

Taking both above into account maybe we should do this before landing this:
- Dynamically change "Sampling Frequency" label into "Sample Frequency" if it's a polling device and "Event Frequency" if it's not a polling device. Only for Sampling Frequency would we show a "Target" and the target could mean device-specific target or global polling frequency setting dependencies on device. Then the info is more contextual.

What do you think?

Adding the IOCTL could be done on this branch in managed and separately in native. It just needs a good fallback when we cannot tell. When we cannot tell, maybe we should just hide Target since it doesn't add any info anyways.

ritamerkl:
> > I tested the newly added Input Debugger features myself now, and think it is priceless. The only wish I would have is to name target/ Polling Frequency with a bit more clarity or adding an explanation with a small text underneath, or a hover textbox? For me the meaning wasn't clear initially (your live demo on friday was very helpful) - also for the restrictions eg when n/a shows up and what this means.
> 
> I agree, the UI layout isn't very helpful, maybe I should try revisit the IMGUI design here to at least make it a bit better (I think the debugger window is overpopulated to begin with, contains a mix of static and dynamic data and should likely be split into multiple windows and/or perspectives).
> 
> Regarding the N/A and Target polling rate I would propose the following:
> 
> * Add a new IOCTL command allowing to query whether a device is subject for polling or not. If it's not, there is no point in showing Polling Frequency. This would be trivial if done on module level and could even be a flag. Will investigate.
> * Only show target if the device responds that it has a device-specific target frequency. This will only show something for sensors currently. In the future I hope all devices can be device-specific.
> 
> Taking both above into account maybe we should do this before landing this:
> 
> * Dynamically change "Sampling Frequency" label into "Sample Frequency" if it's a polling device and "Event Frequency" if it's not a polling device. Only for Sampling Frequency would we show a "Target" and the target could mean device-specific target or global polling frequency setting dependencies on device. Then the info is more contextual.
> 
> What do you think?
> 
> Adding the IOCTL could be done on this branch in managed and separately in native. It just needs a good fallback when we cannot tell. When we cannot tell, maybe we should just hide Target since it doesn't add any info anyways.

Sounds good to me!

Packages/com.unity.inputsystem/Documentation~/Debugging.md
```diff
@@ -42,7 +42,11 @@ In the Input Debugger window, navigate to the __Devices__ list and double-click
 
 ![Device in Input Debugger](Images/DeviceInDebugger.png)
 
-The top of the Device window displays general information about the specific Device, such as name, manufacturer, and serial number.
+The top of the Device window displays general information about the specific Device, such as name, manufacturer, associated layout, device flags, device ID and serial number.  In addition, this section also display the current __sample frequency__ and __processing delay__ of the deivce.
+
+__Sample frequency__ indicates the current frequency of samples or events the Input System is currently processing at given in Hertz (Hz). For devices receiving events this reflects the current flow of events received by the system, and for device receiving periodic readings this reflects the achievable sample rate of the system. The latter may be compared to the globally configured target sampling frequency, while achievable event frequency is uncorrelated to the sample polling frequency setting.
```
chrstphrsxtn:
```suggestion
__Sample frequency__ indicates the frequency in Hertz (Hz) at which the Input System is currently processing samples or events. For devices receiving events this reflects the current flow of events received by the system. For devices receiving periodic readings this reflects the achievable sample rate of the system. The latter may be compared to the globally configured target sampling frequency, while achievable event frequency is uncorrelated to the sample polling frequency setting.
```

Packages/com.unity.inputsystem/Documentation~/Debugging.md
```diff
@@ -42,7 +42,11 @@ In the Input Debugger window, navigate to the __Devices__ list and double-click
 
 ![Device in Input Debugger](Images/DeviceInDebugger.png)
 
-The top of the Device window displays general information about the specific Device, such as name, manufacturer, and serial number.
+The top of the Device window displays general information about the specific Device, such as name, manufacturer, associated layout, device flags, device ID and serial number.  In addition, this section also display the current __sample frequency__ and __processing delay__ of the deivce.
```
chrstphrsxtn, line 45:
```suggestion
The top of the Device window displays general information about the specific Device, such as name, manufacturer, associated layout, device flags, device ID and serial number. This section also displays the current __sample frequency__ and __processing delay__ of the device.
```

Packages/com.unity.inputsystem/Documentation~/Debugging.md
```diff
@@ -42,7 +42,11 @@ In the Input Debugger window, navigate to the __Devices__ list and double-click
...
+
+__Sample frequency__ indicates the current frequency of samples or events the Input System is currently processing at given in Hertz (Hz). For devices receiving events this reflects the current flow of events received by the system, and for device receiving periodic readings this reflects the achievable sample rate of the system. The latter may be compared to the globally configured target sampling frequency, while achievable event frequency is uncorrelated to the sample polling frequency setting.
+
+__Processing delay__ indicates the average, minimum and maximum latency contribution from creating an input event or reading until the Input System has processed the same input event. Note that this excludes any additional input delay caused by OS, drivers or device communication. Also note that this excludes any additional output latency that may be caused by additional processing, rendering, GPU swap-chains or display refresh rate.
```
chrstphrsxtn, line 49:
```suggestion
__Processing delay__ indicates the average, minimum, and maximum time delay from when an input event or reading is created until the Input System has processed it. Note that this excludes any additional input delay caused by OS, drivers, or device communication. Also note that this excludes any additional output latency that may be caused by additional processing, rendering, GPU swap-chains, or display refresh rate.
```

chrstphrsxtn (approved):
Reviewing on behalf of docs. Approving so as not to block, but left some suggestions which should be applied before merging.

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
```diff
@@ -143,6 +146,19 @@ internal void OnGUI()
...
             if (m_Device is Keyboard)
                 EditorGUILayout.LabelField("Keyboard Layout", ((Keyboard)m_Device).keyboardLayout);
+            const string sampleFrequencyTooltip = "Displays the current event or sample frequency of this device in Hertz (Hz) averaged over measurement period of 1 second. " +
+                "The target frequency is device and backend dependent and may not be supported by all devices nor backends. " +
+                "The Polling Frequency indicates system polling target frequency.";
+            if (!string.IsNullOrEmpty(m_DeviceFrequencyString))
+                EditorGUILayout.LabelField(new GUIContent("Sample Frequency", sampleFrequencyTooltip), new GUIContent(m_DeviceFrequencyString), EditorStyles.label);
+            const string processingDelayTooltip =
+                "Displays the average, minimum and maximum observed input processing delay. This shows the time from " +
+                "when an input event is first created within Unity until its processed by the Input System. " +
```
chrstphrsxtn, line 156:
```suggestion
                "when an input event is first created within Unity until it's processed by the Input System. " +
```

Packages/com.unity.inputsystem/InputSystem/Editor/Debugger/InputDeviceDebuggerWindow.cs
```diff
@@ -143,6 +146,19 @@ internal void OnGUI()
...
+                "Note that this hence excludes additional input latency introduced by OS, driver or device communication. " +
```
chrstphrsxtn:
```suggestion
                "Note that this excludes additional input latency introduced by OS, driver or device communication. " +
```

Packages/com.unity.inputsystem/Documentation~/Debugging.md
ekcoh: Thanks @chrstphrsxtn for the suggestions, its much better.

PR #2144, opened 2025-02-21
FIX: Project wide actions enabling actions overrides PlayerInput (ISXB-920)

ritamerkl:
### Description

This PR fixes this issue: [ISXB-920](https://jira.unity3d.com/browse/ISXB-920), where the root cause is that the project wide actions enable all action maps and therefore override the PlayerInput settings. 

Major changes:
- all action maps other than the current one are disabled in PlayerInput at the time of initialization (this resets the activated action maps)
- the action maps only get enabled if there is not already enabled action in the action map for the project wide action asset, this avoids losing configurations and fixes setting the action maps enabled after PlayerInput initialized (which is a timing issue and seemed to have changed after [dc2ce6c875130e0fb8e9077eb48eabd01da75b99](https://github.com/Unity-Technologies/InputSystem/commit/dc2ce6c875130e0fb8e9077eb48eabd01da75b99). 

### Testing status & QA

Testing with the client project.

### Overall Product Risks

- Complexity: Low
- Halo Effect: Medium

### Comments to reviewers

To QA: please also test if using the project wide actions works like before, and if the PlayerInput with using different action maps (and in best case also multiple action maps enabled at once) works as expected.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


ekcoh (approved):
Changes/fix looks good to me @ritamerkl. I think auto-enabling action maps for Project-wide input actions might cause more issues than it solves generally. For this fix, would it be possible to add a test case to make sure we have fixed the bug and avoid future regression?

Pauliusd01 (changes requested):
Seems like PWA are enabled whether they're being used or not, is it expected?
https://github.com/user-attachments/assets/3e177cdf-3d46-4dc6-9f13-684a6d1cbe98
In the video I first show off this PR, PWA actions are enabled in the debugger alongside the player input actions I'm using and then show the same thing on current develop

ritamerkl:
I changed the approach to fix to duplicate the action asset always, since we do not want to operate directly on it, but on a copy for PlayerInput. This is the case for project wide action assets as well as for assets that are used for the UIInputModule or elsewhere. This should fix the issues you mentioned @Pauliusd01.
I would appreciate a second round of review with the latest changes @ekcoh.

ritamerkl:
Hold back with testing / reviews for now, WIP

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -118,6 +117,8 @@ public void PlayerInput_CanLinkSpecificDeviceToUI()
 
         var instance = PlayerInput.Instantiate(prefab, pairWithDevices: gamepad);
 
+        ui.actionsAsset = instance.actions; // need to link to instance, since prefac action asset will be copied per instance
```
ekcoh: typo: "prefab"

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -149,11 +150,11 @@ public void PlayerInput_CanUseSameActionsForUIInputModule()
         eventSystemGO.SetActive(true);
         playerGO.SetActive(true);
 
-        Assert.That(actions.FindActionMap("Gameplay").enabled, Is.True);
-        Assert.That(actions.FindActionMap("UI").enabled, Is.True);
-        Assert.That(actions["UI/Navigate"].controls, Is.Empty);
-        Assert.That(actions["UI/Point"].controls, Is.EquivalentTo(new[] { mouse.position }));
-        Assert.That(actions["UI/Click"].controls, Is.EquivalentTo(new[] { mouse.leftButton }));
+        Assert.That(player.actions.FindActionMap("Gameplay").enabled, Is.True);
```
ekcoh, line 152: Maybe insert a new line after this one and add a comment why there is a split now?

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -391,12 +392,13 @@ public void PlayerInput_CanAssignActionsToPlayer()
         var actions = InputActionAsset.FromJson(kActions);
         playerInput.actions = actions;
 
-        Assert.That(playerInput.actions, Is.SameAs(actions));
+        Assert.That(playerInput.actions, !Is.SameAs(actions)); // Should have created a copy.
```
ekcoh: This is dangerous since it would call Equals or any overload, what you really want to check here is ReferenceEquals to make sure its a different instance (some types compare equal even if they are different instances)

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -391,12 +392,13 @@ public void PlayerInput_CanAssignActionsToPlayer()
...
+        Assert.That(playerInput.actions.actionMaps.Count, !Is.SameAs(actions.actionMaps.Count));
```
ekcoh: I'm confused around this assert, what is it proving?

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
```diff
@@ -391,12 +392,13 @@ public void PlayerInput_CanAssignActionsToPlayer()
...
     }
 
     [Test]
     [Category("PlayerInput")]
-    public void PlayerInput_AssigningNewActionsToPlayer_DisablesExistingActions()
+    public void PlayerInput_AssigningNewActionsToPlayer_LeavingOriginalAssetsUntouched()
```
ekcoh: Related to this, what happens when the original asset is modified after being cloned? Is the PlayerInput updated to reflect or just fixed to what it was when instantiated? I am not sure if it intends to monitor that or not...

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -1368,21 +1361,18 @@ private void InitializeActions()
...
-                if (s_AllActivePlayers[i].m_Actions == m_Actions && s_AllActivePlayers[i] != this)
-                {
-                    var oldActions = m_Actions;
-                    m_Actions = Instantiate(m_Actions);
-                    for (var actionMap = 0; actionMap < oldActions.actionMaps.Count; actionMap++)
-                    {
-                        for (var binding = 0; binding < oldActions.actionMaps[actionMap].bindings.Count; binding++)
-                            m_Actions.actionMaps[actionMap].ApplyBindingOverride(binding, oldActions.actionMaps[actionMap].bindings[binding]);
-                    }
+            // duplicate action asset to not operate on the original(as it might be used outside - eg project wide action asset or UIInputModule)
```
ekcoh: IMO this is much cleaner to not let it depend on some condition whether there is a clone or not.

ekcoh (approved):
Changes seem to be in line with what we discussed, some minor questions I hope we can sort out but approving anyway.

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
ritamerkl: If the original changes it wouldn't reflect in the copies of the PlayerInput actions component. They could only be changed through the playerInput.actions property for a single PlayerInput. 
Thinking about it I am less certain if it wouldn't introduce more regression seeing this comment:
// Make sure that no cloning of actions happened on the prefab.
// https://fogbugz.unity3d.com/f/cases/1319756/
Thus I wonder if that is no problem for the other PlayerInput instances, since it will be cloned if more than one player jumps in.

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
ritamerkl: outdated with new commits (comment removed)

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
ritamerkl: outdated with new commits, replaced with (playerInput.actions.actionMaps[0].name, Is.EqualTo(actions.actionMaps[0].name)

Assets/Tests/InputSystem/Plugins/PlayerInputTests.cs
ritamerkl: outdated with new commits, replaced with (playerInput.actions.actionMaps[0].name, Is.EqualTo(actions.actionMaps[0].name)

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -342,8 +342,14 @@ public InputActionAsset actions
                         UninitializeActions();
                 }
 
+                var didChange = m_Actions != null;
```
ekcoh, line 345: I guess this is related to your question around comparing InputActionAssets? It's not going to be cheap to compare, but a comparator would be useful. Additionally that comparator could compute a hash to make the comparison cheaper. However, this could also be fine since it will very likely be different if it's another instance. Otherwise the project itself is oddly designed.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
```diff
@@ -1430,6 +1429,18 @@ private void InitializeActions()
             m_ActionsInitialized = true;
         }
 
+        private void CopyActionAsset()
```
ekcoh: Maybe rename this CopyActionAssetAndApplyBindingOverrides() since it attempts to preserve binding overrides? I wonder how it actually works though since the new asset might be completely different.... (not new on this PR)

ekcoh (approved):
New changes looks good to me.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
ritamerkl: Yes, all this is around avoiding copying the asset multiple times, which works as far as I tested and for the cases I could come up with, but like you said ideally it would be great to be able to compare an copy of asset to it's original to determine that it's a clone of the asset.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInput.cs
ritamerkl: addressed in [ce0a9a5](https://github.com/Unity-Technologies/InputSystem/pull/2144/commits/ce0a9a5b4b6578e2ee3ded2918cd80c4cf539a74)

Pauliusd01:
Still seeing some odd things with project wide actions getting duplicated.
https://github.com/user-attachments/assets/ac39d5c9-6710-4902-ae19-610f93f98965
In the video I'm using project wide actions for player input with your PR and develop. First showing your PR, it shows project wide actions active in the debugger twice while the develop one does not. Also, I noticed that the actions asset gets the "cloned" tag as well in your version

Pauliusd01 (changes requested):
Updating status

ritamerkl:
> Still seeing some odd things with project wide actions getting duplicated. https://github.com/user-attachments/assets/ac39d5c9-6710-4902-ae19-610f93f98965 In the video I'm using project wide actions for player input with your PR and develop. First showing your PR, it shows project wide actions active in the debugger twice while the develop one does not. 

Yes, that is intended. The project wide actions will be enabled, if the PlayerInput uses the same action asset like the project wide actions, the PlayerInput clone of the asset will be enabled at the same time. 

Also, I noticed that the actions asset gets the "cloned" tag as well in your version

Yes, that is due to the fact that the PlayerInput asset is duplicated for the first player now too. But this will reset to the original asset exiting the play mode.

Pauliusd01 (approved):

PR #2150, opened 2025-03-13
FIX: Prevent errors on certain platforms

jamesmcgill:
### Description
https://jira.unity3d.com/browse/ISXB-1465

Add some platforms to the exclusion list for not requiring an extension package.

### Testing status & QA

Manually tested fix on latest Editor, see ticket for reproduction steps.

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


aceunity (approved):
Thanks for adding us back in to remove the error.

ritamerkl (approved):

ekcoh (approved):
I am going to approve this since very small and tidy, but I just want to emphasise that its very odd to me that the Input System contains such a list at all. Shouldn't we strive to get rid of this and let the platform itself report this as a capability via IOCTL?

ritamerkl:
> I am going to approve this since very small and tidy, but I just want to emphasise that its very odd to me that the Input System contains such a list at all. Shouldn't we strive to get rid of this and let the platform itself report this as a capability via IOCTL?

Agreed, the intent behind was to get this in quickly with the deprecation of the InputManager and change to InputSystem for templates, since platforms with the need of an extra extension would just silently fail. I think to implement this into the platforms would be the better, cleaner way - but like I said, with the pressing need at that time it wasn't considered an option.

ekcoh:
> > I am going to approve this since very small and tidy, but I just want to emphasise that its very odd to me that the Input System contains such a list at all. Shouldn't we strive to get rid of this and let the platform itself report this as a capability via IOCTL?
> 
> Agreed, the intent behind was to get this in quickly with the deprecation of the InputManager and change to InputSystem for templates, since platforms with the need of an extra extension would just silently fail. I think to implement this into the platforms would be the better, cleaner way - but like I said, with the pressing need at that time it wasn't considered an option.

Aha, I see. I thought this list was something older and I wasn't aware of it being something we added lately, guess I have suppressed that memory. Regardless, since this is internal we can get it removed. Lets get a ticket filed on this and discuss how we best get it in place.

PR #2156, opened 2025-03-27
FIX: Use the invariant culture for capitalization when generating code

K-Tone:
### Description

A customer with the Turkish locale set, noticed that when we're generating function names for Input Actions, we're using the wrong function. This way, the "info" input action results in Onİnfo method name (notice the dot above the I, it's apparently how the capital for i looks like in Turkish). 

### Testing status & QA

Pending a QA pass, local testing by dev, a new regression test added

### Overall Product Risks

- Complexity: Medium
- Halo Effect: Medium

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Editor/AddScenesToBuild.cs
+++ b/Assets/Editor/AddScenesToBuild.cs
@@ -66,7 +66,7 @@ public class AddScenesToBuild : EditorWindow
         // Check if the path or any part of it contains any of the excluded folder names
         foreach (string folder in excludedFolders)
         {
-            if (path.ToLower().Contains(folder.ToLower()))
+            if (path.Contains(folder, StringComparison.InvariantCultureIgnoreCase))
             {
                 return true;
             }
--- a/Assets/Editor/AddScenesToBuild.cs
+++ b/Assets/Editor/AddScenesToBuild.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using UnityEditor;
 using UnityEditor.SceneManagement;
 using UnityEngine;
+using System;
 
 [InitializeOnLoad]
 public class AddScenesToBuild : EditorWindow
--- a/Assets/Tests/InputSystem.Editor/InputActionAssetManagerTests.cs
+++ b/Assets/Tests/InputSystem.Editor/InputActionAssetManagerTests.cs
@@ -142,6 +142,7 @@ class InputActionAssetManagerEditorTests
 
     [Test]
     [Category("Editor")]
+    [Description("A regression test for ISXB-1406")]
     public void Editor_InputActionAssetManager_ActionsCodeGeneration_TypeNamesAreNotAffectedByCultureChange()
     {
         var culture = Thread.CurrentThread.CurrentCulture;
--- a/Assets/Tests/InputSystem.Editor/InputActionAssetManagerTests.cs
+++ b/Assets/Tests/InputSystem.Editor/InputActionAssetManagerTests.cs
@@ -1,12 +1,15 @@
 #if UNITY_EDITOR
 
 using System;
+using System.Globalization;
 using System.IO;
+using System.Threading;
 using NUnit.Framework;
 using UnityEditor;
 using UnityEngine;
 using UnityEngine.InputSystem;
 using UnityEngine.InputSystem.Editor;
+using UnityEngine.InputSystem.Utilities;
 
 class InputActionAssetManagerEditorTests
 {
@@ -135,6 +138,20 @@ class InputActionAssetManagerEditorTests
             // Expecting SaveChangesToAsset to throw when asset no longer exist
             Assert.Throws<Exception>(() => inputActionAssetManager.SaveChangesToAsset());
         }
+    }
+
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputActionAssetManager_ActionsCodeGeneration_TypeNamesAreNotAffectedByCultureChange()
+    {
+        var culture = Thread.CurrentThread.CurrentCulture;
+        Thread.CurrentThread.CurrentCulture = CultureInfo.GetCultureInfoByIetfLanguageTag("tr-TR");
+        
+        var name = CSharpCodeHelpers.MakeTypeName("info");
+
+        Thread.CurrentThread.CurrentCulture = culture;
+
+        Assert.AreEqual('I', name[0], "Unexpected first letter in a type name.");
     }
 }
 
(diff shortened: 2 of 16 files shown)


ekcoh:
Glanced at the draft PR and I would just like to suggest to (if possible) get some unit test in that could catch any problems with culture info.

K-Tone:
Also I see that some tests are failing due to compilation issues apparently, fixing that in a moment.

ritamerkl (approved):
The changes look good to me. Many thanks for going through all the cases!

The PR is still a draft, so this is probably still in work, but just a changelog and a link to a Jira ticket is missing in my POV.

K-Tone:
Moving this out of the draft stage:
* Added a new editor regression test precisely for the user issue (Hope Håkan is ok with it)
* Looked through the current coverage on this branch and made sure that most of the changed lines in this PR are covered by existing tests

Assets/Editor/AddScenesToBuild.cs
```diff
@@ -66,7 +67,7 @@ private static bool IsPathInExcludedFolder(string path)
         // Check if the path or any part of it contains any of the excluded folder names
         foreach (string folder in excludedFolders)
         {
-            if (path.ToLower().Contains(folder.ToLower()))
+            if (path.Contains(folder, StringComparison.InvariantCultureIgnoreCase))
```
ekcoh, line 70: Great to fix case issues here, but I wonder if this wasn't incorrect to begin with. Shouldn't this be using https://learn.microsoft.com/en-us/dotnet/api/system.io.path?view=net-9.0 which should take platform-specific behavior of underlying file systems into account instead? Its probably fine, but it is a bit of a warning flag in my eyes when code tries to "outsmart" the filesystem implementation. Its also interesting previous code was doing ToLower() on a constant that is already in lowercase (its not an arbitrary input). The point I am after is that on some OS you may have both a XR, Xr, xR and xr directory and they are all different. (e.g. linux)

Assets/Tests/InputSystem.Editor/InputActionAssetManagerTests.cs
```diff
@@ -136,6 +139,20 @@ public void Editor_InputActionAssetManager_CanDeleteAssetOnDisk()
             Assert.Throws<Exception>(() => inputActionAssetManager.SaveChangesToAsset());
         }
     }
+
+    [Test]
+    [Category("Editor")]
+    public void Editor_InputActionAssetManager_ActionsCodeGeneration_TypeNamesAreNotAffectedByCultureChange()
```
ekcoh, line 146: Awesome you added this test!

ekcoh: Since this is a very slim test case targeting a previous regression maybe add [Description("See <url-to-issue>")] for future reference?

Packages/com.unity.inputsystem/InputSystem/Editor/InputSystemPluginControl.cs
```diff
@@ -73,10 +73,10 @@ public static void RegisterPlatform(BuildTarget target)
         private static bool IsPluginInstalled()
         {
             var registeredPackages = UnityEditor.PackageManager.PackageInfo.GetAllRegisteredPackages();
-            var plugInName = PlugInName + EditorUserBuildSettings.activeBuildTarget.ToString().ToLower();
+            var plugInName = PlugInName + EditorUserBuildSettings.activeBuildTarget.ToString();
```
ekcoh, line 76: It looks like you fixed a separate issue here right? Or are active build targets package names ToUpper? I am confused around how this previously was supposed to work. The new code looks solid though, effectively ignoring case.

Packages/com.unity.inputsystem/InputSystem/Utilities/InternedString.cs
```diff
@@ -171,26 +171,22 @@ public override string ToString()
...
-            return string.Compare(a.m_StringLowerCase, b.ToLower(CultureInfo.InvariantCulture),
-                StringComparison.InvariantCultureIgnoreCase) != 0;
+            return string.Compare(a.m_StringLowerCase, b, StringComparison.InvariantCultureIgnoreCase) != 0;
         }
 
         public static bool operator==(string a, InternedString b)
         {
-            return string.Compare(a.ToLower(CultureInfo.InvariantCulture), b.m_StringLowerCase,
-                StringComparison.InvariantCultureIgnoreCase) == 0;
+            return string.Compare(a, b.m_StringLowerCase, StringComparison.InvariantCultureIgnoreCase) == 0;
```
ekcoh, line 184: Minor: One could call the other operator, e.g. return (b == a) to avoid the redundant code

Packages/com.unity.inputsystem/InputSystem/Utilities/InternedString.cs
```diff
@@ -171,26 +171,22 @@ public override string ToString()
...
         }
 
         public static bool operator!=(string a, InternedString b)
         {
-            return string.Compare(a.ToLower(CultureInfo.InvariantCulture), b.m_StringLowerCase,
-                StringComparison.InvariantCultureIgnoreCase) != 0;
+            return string.Compare(a, b.m_StringLowerCase, StringComparison.InvariantCultureIgnoreCase) != 0;
```
ekcoh, line 189: Minor: One could call the other operator, e.g. return (b != a)

ekcoh (approved):
Great to get this fix in, I am approving but had some minor comments I let you decide whether they needs to be addressed or not.

Assets/Editor/AddScenesToBuild.cs
K-Tone: I guess for paths, we should probably use https://github.com/lucasmeijer/NiceIO because it makes pathing so easily and cool. For this instance, I would agree it might have been somewhat incorrect on case-sensitive file systems - which are normally Linux and the default Mac. Won't target changing behaviour here though.

Assets/Tests/InputSystem.Editor/InputActionAssetManagerTests.cs
K-Tone: Yes, good point. I didn't add the case because it's apparently a private one, as I've just noticed. That also makes my changelog entry incorrect. This case refers to a public case, but the public one is actually a UUM case. Hmm, I'll check with Paulius on my options here.

Pauliusd01:
Did not forget about this one, will check today/tomorrow

Assets/Tests/InputSystem.Editor/InputActionAssetManagerTests.cs
K-Tone: Actually addressed this one.

Pauliusd01 (approved):
LGTM, I checked generating and using input actions with non latin characters (Chinese, Russian, Japanese, Arabic, Hebrew, Korean, Thai, Vietnamese, Turkish) on Windows and Mac with Chinese, English and Turkish locales.  

But I have to note that I wasn't able to repro the original turkish i character issue on my end. The latin letters i and I stayed the same after generating them on a different locale

PR #2158, opened 2025-03-28
FIX: Fixed the on hover behaviour of the two plus buttons in the Input Actions Editor window

K-Tone:
### Description

This aims to react to a user report where they complain about the two `plus` buttons actually not looking like buttons at all. Since they have no visible border and no background, they don't react to hovering, and neither they react to presses. 

Here it's proposed to correct that by removing the transparent background while keeping the zero border config in, as that is deemed to look a bit better in this particular case.

b0d463d68c07f0b3ab1e2b0d2d09a4b4

### Testing status & QA

Only affects UI, no behaviour change at the moment. Tested locally, check the gif above.

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Pauliusd01 (approved):
LGTM, checked light/dark mode themes in project wide actions window as well as custom ones. Now I don't know how design people feel about this plus button looking different from the others but to me it looks fine and changing all of them would probably be worse than not

ekcoh (approved):
If it's a button I rather have users understanding the UI and then we can reiterate with UX representation if needed to make it look "nicer". Good fix with less customisation.

However I am afraid CHANGELOG.md has a conflict that needs to be resolved. Approving anyway assuming its fixed before landing this.

K-Tone:
I fixed the merge conflict, that was easy. Re-ran one instability, but the other one proves to be tricky - `inputsystem - 6000.2 - macos` won't rerun because of some unrelated editor installation issues:

`unity-downloader-cli -u 6000.2 -c Editor --fast` -- failed.

PR #2159, opened 2025-04-01
FIX: Fixed Input Actions code generation overwriting user files when the names happened to match.

K-Tone:
### Description

Apparently, if a user project happens to already have a script with the same name as was also chosen in the Input Actions importer as the target C# file, we would silently overwrite it with auto-generated stuff and so the user might loose data. 

In the comments to the case it was suggested to check if the file exists, and if it does - then only overwrite when we're sure that it's whatever we generated before. Otherwise, output an error. So that's precisely what I did here.

### Testing status & QA

Local testing, added the area QA.

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -247,6 +247,18 @@ private static void GenerateWrapperCode(AssetImportContext ctx, InputActionAsset
             if (!Directory.Exists(dir))
                 Directory.CreateDirectory(dir);
 
+            // Check for the case where the target file already exists.
+            // If it does, we don't want to overwrite it unless it's been generated by us before.
+            if (File.Exists(wrapperFilePath))
+            {
+                var text = File.ReadAllText(wrapperFilePath);
+                var autoGeneratedMarker = "This code was auto-generated by com.unity.inputsystem";
```
ritamerkl, line 255: I was wondering if we can utilise the string the auto generator uses as well, to avoid that this breaks in the case the string would change. I am not sure how much overhead this would entail, I am comfortable with this solution too.

ritamerkl (approved):
Just a small nitpick, otherwise looks good! Nice to have this!

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
K-Tone: Thanks for the fast review! 

This is a good point so I've checked code around just a little bit. Appears, the code generator template is declared as a constant inside a [static function](https://github.com/Unity-Technologies/InputSystem/blob/7c3ab2cece781ab2543c0f7151021caf75a0726e/Packages/com.unity.inputsystem/InputSystem/Utilities/CSharpCodeHelpers.cs#L93). It's used in just two spots, both of which specialise it with some extra data. Regardless of the specialization, it's going to include this particular string that we have right now on the PR. It feels immanent to the code logics that we would include the package name as part of it always. 

Looking at it, it feels unlikely it's going to be broken any time soon. We would need to add another code generator. I guess, it's indeed more risk in the form of extra code churn if we were to use (portions) of the actual template used by generator.

ekcoh (approved):
Fix makes sense and seems to match the description.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
ritamerkl: Thanks for checking, that is fine for me!

ekcoh:
Restarted CI run since I believe the failure was just noise, CHANGELOG conflict needs to be fixed though.

Pauliusd01 (approved):
LGTM, here's what I checked:

- User repro, when moving the input actions asset to a new folder and renaming
- Whether error still shows up when regenerating the class (uncheck recheck the option)
- Input actions asset itself is preserved correctly when failing to generate the code
- Whether code generation works when it finds auto generated files with the same name like the error message describes. (I copied over info from a different asset with different names and it did overwrite it)
- Does generated code update correctly when updating the Input Actions asset

PR #2160, opened 2025-04-03
FIX: ISXB-1306 Fixed a bug within InputActionAsset.cs that prevented actions containing slashes to be found by name.

ekcoh:
### Description

Suggested fix for https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1306. Essentially we had a bug preventing actions with names containing a "/" to be found by name since logic couldn't handle both map syntax and this case. 

### Testing status & QA

Added unit tests that cover the specified behavior (see diff).

### Overall Product Risks

- Complexity: Small
- Halo Effect: Small

### Comments to reviewers

Repro case on ticket should be sufficient as a check for manual testing (Also covered by automated existing and new tests) so I believe chance of regression is minimal. Good to look for potential mistakes with names vs path-syntax I might have overlooked.

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Pauliusd01 (approved):
LGTM, tried finding actions with 1/multiple slashes, special symbols, non latin characters

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -471,7 +471,7 @@ public static InputActionAsset FromJson(string json)
         /// in the asset.
         /// </summary>
         /// <param name="actionNameOrId">Name of the action as either a "map/action" combination (e.g. "gameplay/fire") or
-        /// a simple name. In the former case, the name is split at the '/' slash and the first part is used to find
+        /// a simple name e.g. "fire". In the former case, the name is split at the '/' slash and the first part is used to find
```
K-Tone: This would better read as (e.g. "fire") to be uniform with the above sentence, or otherwise write in full: 'a simple name such as "fire" for example').

K-Tone:
```suggestion
        /// a simple name (e.g. "fire"). In the former case, the name is split at the '/' slash and the first part is used to find
```

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -8127,6 +8137,26 @@ public void Actions_CanLookUpActionInAssetByName()
         }, Is.Not.AllocatingGCMemory());
     }
 
+    // https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1306
```
K-Tone: Alternatively, links could be put into a Description attribute:

[Description("See https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1306")]

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -491,6 +491,10 @@ public static InputActionAsset FromJson(string json)
         /// An exception is if, of the multiple actions with the same name, some are enabled and some are disabled. In
         /// this case, the first action that is enabled is returned.
         ///
+        /// If an action name contains a slash "/", e.g. "yaw/pitch" and there is also a map called "yaw" which
+        /// contains an action "pitch", the action "pitch" within the map "jump" will be returned instead of the
```
K-Tone: I think I have some problems with text search today because it tells me `jump` is mentioned here for the first time.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -583,6 +570,23 @@ public InputAction FindAction(string actionNameOrId, bool throwIfNotFound = fals
                         break;
                     }
                 }
+
+                // It's just a simple action name. Return either first enabled action or, if
+                // none are enabled, first action with the given name.
```
K-Tone: Obviously, I have no idea how this function is supposed to work, but the comment looks inaccurate to me. In fact,  it appears to return either the first enabled action, or the first action that matched id (can be a disabled one in this case), or can be both disabled and with the wrong id (when the return clause wasn't hit at line 583). This looks a bit weird to my uneducated glance.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: Thanks for the suggestions, accepted it so will resolve this one since outdated.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: It should say "yaw"

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: The comment predates the PR. I didn't pay too much attention since internal and didn't change that part (only moved)

ekcoh: Yeah the Find has this weird quirk that it prioritises enabled actions which I find odd.

jfreire-unity (approved):

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: This is a fair change to this test since specific to reported use-case

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2160/commits/b78188bc492ef186ddfc4a9ec4d6e81cb547d813

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2160/commits/b78188bc492ef186ddfc4a9ec4d6e81cb547d813

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: Attempted to clarify the previous comment here: https://github.com/Unity-Technologies/InputSystem/pull/2160/commits/19887019366d8b35e47244de8b8fecf2abcf7945, it seems inline with xmldoc above?

K-Tone (approved):
lgtm, thanks for working on the suggestions -- and have a great weekend

PR #2164, opened 2025-04-08
FIX: Custom processors serialises enum by index rather than by value.

AswinRajGopal:
### Description

Bug: https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1474
Port 1.13.1

The issue is that the DropdownField expects a selected index (starting at 0), but the enum’s underlying values (10, 20, 30..) don't match these indices. Instead of passing the enum’s integer value directly, I have mapped the value to the correct index in the list of enum options.

An automated test added which validates that a custom input processor with an enumerated parameter integrates correctly with both the Unity Input System’s asset serialization and its UI.

The issue is that 

### Testing status & QA

Verified manually in windows machine with the provided repro project. 
Authored an automated test to cover the fix. 

Observer details of the test run: 

### Overall Product Risks

- Complexity:  0
- Halo Effect: 0

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=2164) <br/>All committers have signed the CLA.

ekcoh:
@AswinRajGopal Its generally recommended to add one developer on the team and one QA as reviewers on the PR when published. This will help making sure you receive feedback on it.

Looks like there are CI failures for non-required tests and that branch is outdated. I would recommend updating the branch (which will automatically re-run CI).

ekcoh:
@AswinRajGopal I notice you have published internal URLs in this public-facing PR. This should be avoided. Instead use public URLs to issue tickets when available and otherwise only use issue number (without link). I will remove the URLs for now.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Collections;
+using System.Linq;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.Editor;
+using UnityEngine.TestTools;
+using UnityEngine.UIElements;
+
+public enum SomeEnum
```
ekcoh: Do this enum have to be public?

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,113 @@
+#if UNITY_EDITOR && UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS && UNITY_6000_0_OR_NEWER
```
ekcoh, line 1: Is this fix only valid for Unity 6000.0 or newer?

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,113 @@
...
+    OptionA = 10,
+    OptionB = 20,
+    OptionC = 50,
+    OptionD = 100
+}
+
+#if UNITY_EDITOR
+[InitializeOnLoad]
+#endif
+public class CustomProcessor : InputProcessor<float>
```
ekcoh: This can be internal to test assembly right?

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,113 @@
...
+        InputSystem.RegisterProcessor<CustomProcessor>();
+    }
+
+    public override float Process(float value, InputControl control)
+    {
+        return value;
+    }
+}
+
+internal class CustomProcessorEnumTest : UIToolkitBaseTestWindow<InputActionsEditorWindow>
```
ekcoh, line 47: Awesome you added a proper UI-automated test case!

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public override IEnumerator UnitySetup()
+    {
+        m_Window = InputActionsEditorWindow.OpenEditor(m_Asset);
+        yield return base.UnitySetup();
+    }
+    
+    [UnityTest]
+    public IEnumerator CustomProcessorEnum_SerialiseTheValue_InAsset()
```
ekcoh: I would suggest you name the test case according to the requirement or alternatively explain it in a Description attribute, e.g. `ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset`

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,113 @@
...
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.Editor;
+using UnityEngine.TestTools;
+using UnityEngine.UIElements;
+
+public enum SomeEnum
+{
+    OptionA = 10,
```
ekcoh, line 16: Do we need OptionC and OptionD? I cannot see them being referenced.

ekcoh: (It makes sense to keep OptionA since default)

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,113 @@
...
+        var dropdown = dropdownList.First();
+
+        var newIndex = dropdown.choices.IndexOf("OptionB"); 
+        var oldValue = dropdown.value;
+        var newValue = dropdown.choices[newIndex];
+
+        dropdown.Focus();
+        dropdown.value = newValue;
+
+        var changeEvent = ChangeEvent<Enum>.GetPooled(SomeEnum.OptionA, SomeEnum.OptionC);
```
ekcoh: It would be great to write a one-line comment for each parapgraph to summarise the scenario step, e.g. "User clicks Save button to serialise the asset to disc"

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed Inspector Window being refreshed all the time when a PlayerInput component is present with Invoke Unity Events nofication mode chosen [ISXB-1448](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1448)
 - Fixed an issue where an action with a name containing a slash "/" could not be found via `InputActionAsset.FindAction(string,bool)`. [ISXB-1306](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1306).
 - Fixed Gamepad stick up/down inputs that were not recognized in WebGL. [ISXB-1090](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1090)
+- Fixed Action with custom processor serialises enums by index rather than by value. [ISXB-1474](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1474)
```
ekcoh: This URL is public, great. Would generally recommend explaining what fix does and previous symptom, e.g. "Fixed an issue that caused input processors with enum properties to incorrectly serialise by index instead of by value."

ekcoh (changes requested):
Thanks for looking into this @AswinRajGopal! Left some comments on things to address.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -0,0 +1,112 @@
...
+        var json = m_Window.currentAssetInEditor.ToJson();
+        
+        Assert.That(json.Contains("Custom(SomeEnum=10)"), Is.True,
+            "Serialized JSON does not contain the expected custom processor string for OptionA.");
+
+        // Query the dropdown with exactly two enum choices and check that the drop down is present in the UI
+        var dropdownList = m_Window.rootVisualElement.Query<DropdownField>().Where(d => d.choices.Count == 2).ToList();
+        Assume.That(dropdownList.Count > 0, Is.True, "Enum parameter dropdown not found in the UI.");
+
+        // Determine the new value to be set in the dropdown, focus the dropdown before dispatching the change
```
ekcoh: Thanks for adding this, increases readability significantly

ekcoh (approved):
Thanks for addressing comments. Changes look good to me now.

Pauliusd01 (commented):
One concern I noticed is that the enum field is blank after upgrading to the new version, requiring the user to repick and save. Real problem? Any way to fix it?

AswinRajGopal:
> One concern I noticed is that the enum field is blank after upgrading to the new version, requiring the user to repick and save. Real problem? Any way to fix it?

Hi @Pauliusd01 could you please give me the steps to repro this issue? which version should I upgrade? Thanks.

Pauliusd01:
> > One concern I noticed is that the enum field is blank after upgrading to the new version, requiring the user to repick and save. Real problem? Any way to fix it?
> 
> Hi @Pauliusd01 could you please give me the steps to repro this issue? which version should I upgrade? Thanks.

Your version. Steps are: Open the user project -> Upgrade the version to yours (package manager -> add from disk -> point to the version with your changes) -> observe the user's enum field is blank

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/ParameterListView.cs
```diff
@@ -278,11 +278,26 @@ void OnEditEnd()
...
+                    var field = new DropdownField(label.text, names, selectedIndex)
+                    {
+                        tooltip = label.tooltip
+                    };
+
+                    field.RegisterValueChangedCallback(evt =>
+                    {
+                        var newBackingValue = parameter.enumValues[field.index];
+                        parameter.value.value = PrimitiveValue.FromObject(newBackingValue).ConvertTo(parameter.value.type);
+                        m_Parameters[closedIndex] = parameter;
```
ekcoh, line 296: Probably safe (you know this better), but `closedIndex` is captured as part of the lambda which implies this callback might execute when m_Parameters have changed in length? It could be safer to check closedIndex is within valid bounds? If already safe due to some other reason maybe adda comment why?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -121,17 +126,58 @@ public override void OnImportAsset(AssetImportContext ctx)
...
-            var asset = CreateFromJson(ctx);
+            var asset = CreateFromJson(ctx); 
             if (asset == null)
                 return;
 
             if (m_GenerateWrapperCode)
                 GenerateWrapperCode(ctx, asset, m_WrapperCodeNamespace, m_WrapperClassName, m_WrapperCodePath);
         }
+        
+        static string MigrateJson(string originalJson)
```
ekcoh: If we need to do this much work to work around the serialisation bug here I think it would be nice to document this internal function explaining what it does and why. IMO this function will be a bit of an enigma if we come back to this again in 2 years to understand why it exists.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -413,6 +459,37 @@ private static void CheckAndRenameJsonNameIfDifferent(string assetPath)
...
+                    {
+                        continue;
+                    }
+
+                    var patched = InputActionImporter.MigrateJson(original);
+                    if (patched == original)
+                        continue;
+
+                    File.WriteAllText(diskPath, patched);
+                    AssetDatabase.ImportAsset(assetPath, ImportAssetOptions.ForceUpdate);
```
ekcoh: Is the explicit/forced reimport here really needed?

ekcoh (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/ParameterListView.cs
AswinRajGopal: Thanks for pointing this out, parameters can be added, removed or re-ordered, I'll make changes accordingly.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
AswinRajGopal: Yes, had to do this much work for this bug, I'll write documentation above this function with case id.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
AswinRajGopal: You are correct, the force update is not necessary here. Thanks for pointing it out.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -413,6 +463,36 @@ private static void CheckAndRenameJsonNameIfDifferent(string assetPath)
...
+
+                    var diskPath = EditorHelpers.GetPhysicalPath(assetPath);
+                    string original;
+                    try
+                    {
+                        original = File.ReadAllText(diskPath);
+                    }
+                    catch
+                    {
+                        continue;
```
ekcoh: What is the intention for suppressing failure to read the file in the migration step? IMO a console warning that migration could not be done would make sense?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -133,6 +136,53 @@ public override void OnImportAsset(AssetImportContext ctx)
...
+        // Case https://jira.unity3d.com/browse/ISXB-1474
+        // Converts any Custom(EnumName=index) processor entries in the JSON to use the enum’s actual underlying value.
+        // Inserts a one-time "__upgradeVersion" field if at least one conversion ran.
+        // Leaves files untouched if there’s nothing to convert or already stamped.
+        static string MigrateJson(string originalJson)
+        {
+            if (originalJson.Contains($"\"{k_VersionFieldName}\""))
+                return originalJson;
+
+            var regex = new Regex("\"processors\"\\s*:\\s*\"Custom\\((\\w+)=(\\d+)\\)\"", RegexOptions.Compiled);
```
ekcoh: At least to me this regex is non-trivial. Would you mind adding a comment explaining what conditions are checked for?

ekcoh (approved):
Changes look ok to me, added some suggestions to motivate complex code parts for increased readability. Would also recommend checking the codecov bot suggestions for logic not being tested.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -413,6 +473,37 @@ private static void CheckAndRenameJsonNameIfDifferent(string assetPath)
                 }
             }
         }
+
+        private class InputActionsMigrationPostprocessor : AssetPostprocessor
+        {
+            static void OnPostprocessAllAssets(string[] importedAssets, string[] deletedAssets, string[] movedAssets, string[] movedFromAssetPaths)
```
LeoUnity: This is not ideal, this will get invoked everytime an esset gets imported by unity, this won't scale well with big projects... even if we are trying to return on line 485, it will still have to iterate through all assets being processed and do a string comparison...

I'm not convinced this is the way to tackle the conversion

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
LeoUnity: I don't think patching the JSON via regex is a viable long term solution. This is very fragile, if we change an unrelated detail on how our JSON is serialized it could break this.

LeoUnity: We have an importer that is reading the data from disk and converting to an asset, this is the perfect time to patch anything, I don't see the need to patch things at the JSON level.
Here is what I have in mind:
 - The importer kicks in and calls "asset.LoadFromJson(content);" this is the perfect entry point for us to "upgrade" anything we want
 - asset.LoadFromJson(content); ends up calling ReadFileJson.ToAsset, this code translates the parsed JSON to a Unity asset, here you could look for a "version" number, and if the version is below the new version number apply any data transformation needed.
 - This will only affect the "imported" asset, we have 2 choices here:
   - we keep the changes only to the imported object, which means the file on disk will not be upgraded yet, only after the user changes something else on the file
   - We mark the asset as dirty, which should prompt unity to flush the changes to disk the next time save is done  (i'm not 100% on this since this is an imported asset, so some tests might be required)

LeoUnity (changes requested):
I think the PR as it is increases the tech debt in a way that is not worth the fix.
I believe there are better ways to achieve the desired outcome, and if not I would rather take the fix without the migration.
Take a look at migrating the data during importing, let me know if you have any questions

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
LeoUnity: Take a look at ReadActionJson.ToAction on InputSystem/Actions/InputActionMap.cs, it seems like there is some code there related to migrating data from old format, this sounds like a promising place to handle the upgrading

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -29,6 +30,8 @@ namespace UnityEngine.InputSystem.Editor
     internal class InputActionImporter : ScriptedImporter
     {
         private const int kVersion = 13;
+        // Bump this whenever you make a breaking change to the on-disk JSON format
```
ekcoh: How would this be different to kVersion? I am confused by introducing another version value? To me this looks undesirable, isn't kVersion supposed to be reflecting the compatibility of json already?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -67,6 +70,19 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
                 // Attempt to parse JSON
                 asset.LoadFromJson(content);
 
+                // If this JSON was authored before we switched to enum-by-value, migrate it now
```
ekcoh: Should we really migrate and modify the file like this? It feels like it would be sufficient to do this when the asset is saved and just convert the json in-memory before parsing here instead? It would simplify this?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -117,6 +133,76 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
...
+                            rebuilt.Add(e);
+                        }
+                    }
+
+                    action.m_Processors = string.Join(";", rebuilt);
+                }
+            }
+        }
+
+        static string PatchEnumArgs(Type procType, string args)
```
ekcoh, line 179: This can be private or internal?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
AswinRajGopal: My bad I was testing the flow missed to refactor. Removed it now.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
AswinRajGopal: Hi @ekcoh For existing .inputactions assets, our importer migrates all processor‐enum fields in memory so that the editor UI is upgraded accordingly. However, the on-disk JSON still remains ‘enum-by-index’ until the user explicitly saves (or auto-save kicks in). In order to complete the upgrade, adding "version": 13 and replacing index-based enum entries with their numeric equivalents we need this. Please correct me if I am missing something here.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -946,10 +947,12 @@ private void OnDestroy()
         [NonSerialized] internal InputActionRebindingExtensions.ParameterOverride[] m_ParameterOverrides;
 
         [NonSerialized] internal InputActionMap.DeviceArray m_Devices;
+        [SerializeField] internal int m_Version;
```
ekcoh: The introduced "version" seems related to the the file format and not the asset object right? Hence I would recommend removing this field. I.e. in a code-only project version doesn't play any role, hence its misplaced IMO and we would better not associate it with the asset.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -946,10 +947,12 @@ private void OnDestroy()
...
 
         [Serializable]
         internal struct WriteFileJson
         {
+            public int version;
```
ekcoh, line 974: It's fine to have version here since related to the file. At the same time it introduces some technical weight to carry forward, but it makes sense to have a version defined unless it can be determined based on structure.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -965,12 +968,14 @@ internal struct WriteFileJsonNoName
         [Serializable]
         internal struct ReadFileJson
         {
+            public int version;
```
ekcoh, line 990: It's fine to have version here as well. However I believe it should default to version before this PR and not version 0?

ekcoh: I realise my previous comment on importer vs json version might have been unjustified since I now realise that this is a file format and somewhat uncorrelated to the importer attribute version. 

If this is introduced I would like us document here inline where we parse JSON what changed with each file format version.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetEditor/ParameterListView.cs
```diff
@@ -278,11 +278,26 @@ void OnEditEnd()
 
                 if (parameter.isEnum)
                 {
-                    var intValue = parameter.value.value.ToInt32();
-                    var field = new DropdownField(label.text, parameter.enumNames.Select(x => x.text).ToList(), intValue);
-                    field.tooltip = label.tooltip;
-                    field.RegisterValueChangedCallback(evt => OnValueChanged(ref parameter, field.index, closedIndex));
-                    field.RegisterCallback<BlurEvent>(_ => OnEditEnd());
+                    var names = parameter.enumNames.Select(c => c.text).ToList();
+                    var rawValue = parameter.value.value.ToInt32();
```
ekcoh, line 282: Minor: It seems the only place this rawValue is used is on the following line so why not pass it directly?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -67,6 +68,19 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
                 // Attempt to parse JSON
                 asset.LoadFromJson(content);
 
+                // If this JSON was authored before we switched to enum-by-value, migrate it now
+                if (asset.m_Version < kVersion)
```
ekcoh: If we remove the version field from the asset as I suggested this cannot be checked here. However this could be moved inside the LoadFromJson call. This would even be necessary when loading an asset directly from file in a player (not using the importer). I think that makes more sense as well since the migration/supporting multiple formats is a problem for the parser.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -67,6 +68,19 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
...
+                {
+                    MigrateAllEnumParams(asset);
+                    asset.m_Version = kVersion;
+
+                    if (!EditorHelpers.WriteAsset(context.assetPath, asset.ToJson()))
+                        context.LogImportError($"Could not write migrated JSON to '{context.assetPath}'");
+
+                    EditorUtility.SetDirty(asset);
```
ekcoh: If you really need to do this, wouldn't it be possible to use an internal overloaded LoadJson returning e.g. "isDirty" flag from the load to indicate it should be marked dirty by editor code?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -67,6 +68,19 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
                 // Attempt to parse JSON
                 asset.LoadFromJson(content);
 
+                // If this JSON was authored before we switched to enum-by-value, migrate it now
+                if (asset.m_Version < kVersion)
+                {
+                    MigrateAllEnumParams(asset);
+                    asset.m_Version = kVersion;
+
+                    if (!EditorHelpers.WriteAsset(context.assetPath, asset.ToJson()))
```
ekcoh: Shouldn't this if statement return after logging an importer error if its fatal?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -117,6 +131,76 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
             return asset;
         }
 
+        internal static void MigrateAllEnumParams(InputActionAsset asset)
```
ekcoh: This is mutating the initially parsed InputActionAsset. I would expect this to be inside InputActioNAsset since its part of Parsing the asset and is not only edit-mode.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -117,6 +131,76 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
...
+                            rebuilt.Add(e);
+                        }
+                    }
+
+                    action.m_Processors = string.Join(";", rebuilt);
+                }
+            }
+        }
+
+        internal static string PatchEnumArgs(Type procType, string args)
```
ekcoh: This should move to same place as InputActionAsset LoadFromJson for same resons as previous comment

ekcoh (changes requested):
Marking this Request Changes based on code being in impropriate places given that this is handling the migration in-memory (See comments). 

While this PR now handles multi-version parsing I struggle with two main issues:
- The parsing is inside the importer which is fine in one sense, but it implies this won't work (produce incorrect results), for a JSON loaded in a player via LoadFromJson directly.
- The migration code is heavily modifying strings which is unfortunate but maybe not possible to avoid since the processor definitions are actually strings to begin with. So as long as we mark dirty the user is likely not experiencing this much anyway if existing assets are migrated in edit-mode.

It might be that migrating directly in https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionMap.cs#L1609 reduces the complexity since this is where an action is parsed. However doing this might require the version to be parsed first prior to parsing the asset to be able to reason about the version.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
LeoUnity: You probably don't want to check against kVersion, as that constant will always be the latest version of the importer.
You only want to execute this code if the asset version is less or equal than the version before we introduced the fix (13), and that means we probably also want to increase the kVersion to 14. So we know that when asset version 13 needs to be upgraded to a higher version they need to run the enum migration code.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
LeoUnity: As Hakan mentioned above, I'm not sure about writing to the file while importing, I'm also wondering if just fixing the problem on the in-memory representation would be good enough, the next time the user changes something in the asset via the tool then it will be flushed to disk.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -117,6 +131,76 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
...
         }
 
+        internal static void MigrateAllEnumParams(InputActionAsset asset)
+        {
+            foreach (var map in asset.actionMaps)
+            {
+                foreach (var action in map.actions)
+                {
+                    var raw = action.processors;
+                    if (string.IsNullOrEmpty(raw))
```
LeoUnity: I'm not fond of doing this parsing here manually, Isn't there some function that does that for the processor that we can reuse?
I believe that is done by NameAndParameters.ParseMultiple(namesAndParameters, ref m_Parameters).
You can see being used in some tests here: Assets/Tests/InputSystem/Utilities/ParameterValueTests.cs

And I believe you can use NameAndParameters.ToString to convert back into string, this should reduce the amount of lines and reuse existing code.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -946,10 +956,12 @@ private void OnDestroy()
         [NonSerialized] internal InputActionRebindingExtensions.ParameterOverride[] m_ParameterOverrides;
 
         [NonSerialized] internal InputActionMap.DeviceArray m_Devices;
+        [SerializeField] internal int m_Version;
 
         [Serializable]
         internal struct WriteFileJson
         {
+            public int? version;
```
ekcoh: I would prefer that this is an int (not nullable), it could default to version 0 (file format version)

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: I am unresolving this since I do not see why we need this. In a project without assets version has no purpose. Only the file format is (needs to) be versioned?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -379,6 +383,12 @@ public void LoadFromJson(string json)
                 throw new ArgumentNullException(nameof(json));
 
             var parsedJson = JsonUtility.FromJson<ReadFileJson>(json);
+
+            m_Version = parsedJson.version;
+            if (m_Version <= 13)
```
ekcoh: I would propose introducing documented constants for the JSON versions introduced by this PR, e.g.

```cs
private static struct JsonVersion
{
   /// <summary>The original JSON version format for InputActionAsset.</summary>
   public static const int Version0 = 0,
   
   /// <summary>Updated JSON version format for InputActionAsset.</summary>
   /// <remarks>Changes representation of parameter values from being serialized by value to being serialized by value.</remarks>
   public static const int Version1 = 1;
   
   /// <summary>The current version.</summary>
   public static const int Current = Version1;
}
```
To create a natural place to call out all changes in-between those formats and remove "magic values". Hence that constant could be checked against here.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -298,6 +301,7 @@ public string ToJson()
         {
             return JsonUtility.ToJson(new WriteFileJson
             {
+                version = (m_Version >= 13) ? (int?)m_Version : null,
```
ekcoh: Shouldn't this always be the latest (current version)? I see no need to generate obsolete formats?

`version = JsonVersion.Current;`

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -965,12 +977,14 @@ internal struct WriteFileJsonNoName
...
         internal struct ReadFileJson
         {
+            public int version;
             public string name;
             public InputActionMap.ReadMapJson[] maps;
             public InputControlScheme.SchemeJson[] controlSchemes;
 
             public void ToAsset(InputActionAsset asset)
             {
+                asset.m_Version = version;
```
ekcoh: Why do we need to store this? I think it should be removed.

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -28,7 +28,7 @@ namespace UnityEngine.InputSystem.Editor
     [ScriptedImporter(kVersion, InputActionAsset.Extension)]
     internal class InputActionImporter : ScriptedImporter
     {
-        private const int kVersion = 13;
+        internal const int kVersion = 14;
```
ekcoh: Makes sense to bump this one as you have done to trigger re-import to ensure parameters have been correctly parsed, see https://docs.unity3d.com/ScriptReference/AssetImporters.ScriptedImporterAttribute-ctor.html

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
```diff
@@ -67,6 +67,13 @@ private static InputActionAsset CreateFromJson(AssetImportContext context)
                 // Attempt to parse JSON
                 asset.LoadFromJson(content);
 
+                // If this JSON was authored before we switched to enum-by-value, migrate it now
```
ekcoh: This is handled by LoadFromJson (above) now and should no longer be part of importer right?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +995,74 @@ public void ToAsset(InputActionAsset asset)
                         map.m_Asset = asset;
             }
         }
+        internal static void MigrateAllEnumParams(InputActionAsset asset)
```
ekcoh: Nitpick: missing white space before method. Have you run the formatter lately?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +995,74 @@ public void ToAsset(InputActionAsset asset)
...
+                    var raw = action.processors;
+
+                    List<NameAndParameters> parsedList = null;
+                    NameAndParameters.ParseMultiple(raw, ref parsedList);
+
+                    if (parsedList == null || parsedList.Count == 0)
+                        continue;
+
+                    var rebuilt = new List<string>(parsedList.Count);
+
```
ekcoh: Nitpick: Remove empty line

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +995,74 @@ public void ToAsset(InputActionAsset asset)
...
+                        if (!anyChanged)
+                        {
+                            rebuilt.Add(nap.ToString());
+                        }
+                        else
+                        {
+                            var paramString = string.Join(",", dict.Select(kv => $"{kv.Key}={kv.Value}"));
+                            var newNamedValues = NamedValue.ParseMultiple(paramString);
+
```
ekcoh: Nitpick: remove empty line

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +995,74 @@ public void ToAsset(InputActionAsset asset)
...
+                            rebuilt.Add(nap.ToString());
+                            continue;
+                        }
+
+                        var procType = InputSystem.TryGetProcessor(nap.name);
+
+                        var dict = nap.parameters.ToDictionary(pv => pv.name, pv => pv.value.ToString());
+
+                        bool anyChanged = false;
+
```
ekcoh, line 1028: Nitpick: Remove empty line

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +995,74 @@ public void ToAsset(InputActionAsset asset)
...
+                    foreach (var nap in parsedList)
+                    {
+                        if (nap.parameters.Count == 0 || InputSystem.TryGetProcessor(nap.name) == null)
+                        {
+                            rebuilt.Add(nap.ToString());
+                            continue;
+                        }
+
+                        var procType = InputSystem.TryGetProcessor(nap.name);
+
```
ekcoh: Nitpick: Remove empty line

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +995,74 @@ public void ToAsset(InputActionAsset asset)
...
             }
         }
+        internal static void MigrateAllEnumParams(InputActionAsset asset)
+        {
+            foreach (var map in asset.actionMaps)
+            {
+                foreach (var action in map.actions)
+                {
+                    var raw = action.processors;
+
```
ekcoh: Nitpick: Remove empty line

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporter.cs
ekcoh: I think it should be private since uncorrelated to JSON version. Sorry for saying there should be only one earlier. That was incorrect.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: I might have incorrectly pushed against matching this with importer version previously but not sure that was the correct move.

ekcoh (changes requested):
Thanks for updating. Some additional (hopefully final) batch of feedback.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -379,6 +398,11 @@ public void LoadFromJson(string json)
                 throw new ArgumentNullException(nameof(json));
 
             var parsedJson = JsonUtility.FromJson<ReadFileJson>(json);
+            if ((parsedJson.maps?.Length ?? 0) > 0 && (parsedJson.version ?? 0) < JsonVersion.Current)
```
ekcoh: `JsonVersion.Current` here should probably be `JsonVersion.Version1` since if we move on to version 2 later we do not need to do this migration step for version 1 json files?

ekcoh: Additionally this may be skipped if the action:
- Do not contain actions
- Do not contain processors

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +1007,72 @@ public void ToAsset(InputActionAsset asset)
...
+                        }
+                    }
+
+                    actionJson.processors = string.Join(";", rebuilt);
+                    mapJson.actions[ai] = actionJson;
+                }
+                parsedJson.maps[mi] = mapJson;
+            }
+            // Bump the version so we never re-migrate
+            parsedJson.version = JsonVersion.Current;
```
ekcoh: This should be assigned JsonVersion.Version1 right? Current might be Version 2 or higher in the future which would make this incorrect then

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +1007,72 @@ public void ToAsset(InputActionAsset asset)
...
+
+        /// <summary>
+        /// If parsedJson.version is older than Current, rewrite every
+        /// action.processors entry to replace “enumName(Ordinal=…)” with
+        /// “enumName(Value=…)” and bump parsedJson.version.
+        /// </summary>
+        internal void MigrateJson(ref ReadFileJson parsedJson)
+        {
+            var existing = parsedJson.version ?? JsonVersion.Version0;
+            if (existing >= JsonVersion.Current)
```
ekcoh: This should compare to JsonVersion.Version1 instead of Current right?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: Minor: but you might as well move these "optimization" checks into the Migrate Json to remove the complexity from here and call it unconditionally, e.g. "MigrateJsonFromVersion0ToVersion1IfNeeded(ref parsedJson)"

ekcoh (changes requested):
This is a smaller now which I think is good and easier to read. I think the version used in comparison in a few places needs replacing but apart from that looks good to me. But it seems the assets needs updating to pass CI.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: Yes right, my bad I mis aligned the enum values. I'll modify.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: Yes we should skip migration in case the input action doesn't contain action and processors.

ekcoh (approved):
Changes look good to me, looks like Changelog is in conflict which needs to be fixed so the branch can be updated and CI rerun to avoid instabilities already fixed on develop.

ekcoh:
I fixed the conflict and updated the branch based on develop. Hopefully CI rerun is all green now.

Pauliusd01 (changes requested):
The value is no longer blank after updating to the new version but it still different, is that expected? Example steps are: Open the user project -> observe that that the enum is set to Option D -> update the package to your version -> observe that the enum is set to Option A.

AswinRajGopal:
> The value is no longer blank after updating to the new version but it still different, is that expected? Example steps are: Open the user project -> observe that that the enum is set to Option D -> update the package to your version -> observe that the enum is set to Option A.

@Pauliusd01 I've uploaded the repro, it works for me, let me know if Im missing something here. 

https://github.com/user-attachments/assets/7a395e33-f5a3-4dad-a14a-97560a48528a

AswinRajGopal:
> The value is no longer blank after updating to the new version but it still different, is that expected? Example steps are: Open the user project -> observe that that the enum is set to Option D -> update the package to your version -> observe that the enum is set to Option A.

@Pauliusd01 works for me again. this repro is loading a new extracted user project and upgrading the package. 
https://github.com/user-attachments/assets/43370f74-713e-4b3e-b9ee-39f1d636cdef

Pauliusd01 (changes requested):
Having the asset window already open might be what was different for you, try to close it while doing the repro steps. Bug's still there for me

https://github.com/user-attachments/assets/47982e35-6263-4694-bc1c-14b2f6555dab

AswinRajGopal:
> Having the asset window already open might be what was different for you, try to close it while doing the repro steps. Bug's still there for me
> 
>  13.06.2025.-.Unity.200.mp4

@Pauliusd01 it couldnt be the reason as the code changes regardless of window state, looks for the asset during its import time and perform the migration, it works for me here is the recording. 


https://github.com/user-attachments/assets/0e3764d1-51dc-40ca-8127-2868e553b5b5

Pauliusd01 (approved):
LGTM. So after quite a bit of back and forth the latest issue seems to only reproduce on first open and only in the first few minutes. Likely a library rebuild issue and it is probably not something worth worrying about as even on a rare case it occurs to the user, it would only be visual and the value returns to normal on reopening the project.

Things checked: making sure the enum values are preserved after updating the package, changing the values with manual and auto saves, changing the values in runtime, checking whether the field still works correctly in the inspector dropdown

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +1004,75 @@ public void ToAsset(InputActionAsset asset)
...
+
+        /// <summary>
+        /// If parsedJson.version is older than Current, rewrite every
+        /// action.processors entry to replace “enumName(Ordinal=…)” with
+        /// “enumName(Value=…)” and bump parsedJson.version.
+        /// </summary>
+        internal void MigrateJson(ref ReadFileJson parsedJson)
+        {
+            var existing = parsedJson.version ?? JsonVersion.Version0;
+            if (existing >= JsonVersion.Version1)
```
LeoUnity: Shouldn't this be "existing >= JsonVersion.Current"

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -950,6 +971,7 @@ private void OnDestroy()
         [Serializable]
         internal struct WriteFileJson
         {
+            public int? version;
```
LeoUnity: Do we need the nullable here? the default int value is 0, so not writing to this member would get us the correct value.
Having to remember to check the null state everywhere where this is used usually makes the code a bit harder to read.

LeoUnity: Same in the ReadFileJson

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -981,5 +1004,75 @@ public void ToAsset(InputActionAsset asset)
...
+                        }
+
+                        actionJson.processors = string.Join(";", rebuilt);
+                        mapJson.actions[ai] = actionJson;
+                    }
+                    parsedJson.maps[mi] = mapJson;
+                }
+            }
+            // Bump the version so we never re-migrate
+            parsedJson.version = JsonVersion.Version1;
```
LeoUnity, line 1073: This should probably be JSonVersion.Current

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: Hi @LeoUnity We are targeting exactly the format bump that we just added(Version 1), for this enum migration 
JsonVersion.Current == Version1, however in the future Current may advance to version2, so we need the migration exactly run once.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: Hi @LeoUnity True, there are few tests failing because of this we need nullable version to avoid those failures where we omit version check for empty Json or action asset is created with minimal Json from LoadFromJson versioning will be added unnecessarily.

ekcoh: I think I commented on the same thing previously. Are the tests failing due to the parser not being capable of handling non existing fields? If the nullable allows the parser to deal with non existing fields in file and still parse without errors (keeping the default value) I think it's ok to have. If we do not need it to handle parser limitations I also think we should not use nullable due to boxing.

What was the failures again caused when not having nullable?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
ekcoh: In my perspective this looks correct if this migration step only handles v0 to v1. It would need to be followed by a v1 to v2 if Current was e.g. 2

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: image
The minimal json in the test doesnt contain version: 0 so had to do this null check. 

 static string MinimalJson(string name = null)
    {
        if (name != null)
            return "{\n    \"name\": \"" + name + "\",\n    \"maps\": [],\n    \"controlSchemes\": []\n}";
        return "{\n    \"maps\": [],\n    \"controlSchemes\": []\n}";
    }

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
LeoUnity: I was assuming that this first check would be the checking if any migration needs to happen at all, and then below there would be specific checks for each migration, because you do the check again in the line below:
if ((parsedJson.maps?.Length ?? 0) > 0 && (parsedJson.version ?? 0) < JsonVersion.Version1)

So either this is checking for the specific upgrade needed, and then we remove this check on the line below, or this is a generic check if a migration is needed, and we keep the check ont he line below

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
LeoUnity: Yeah I guess I can see that, I was expecting that after running this function we would always end up on the latest version anyways, and we would always need to remember to update the version here... but its a nitpick

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
LeoUnity: Ah ok, the nullable is to make the field optional to serialize... I understand it now...
I'm wondering if we should just update those tests, from this version onwards the version field will always exist...

Is the same problem for the ReadFileJson? If you try to deserialize a json without the version field it would cause problems?

AswinRajGopal: @LeoUnity  Yes, same issues with ReadFileJson as well for other set of tests, it requires patching the assets generated manually, as we only migrate in memory. So had to do this.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: 1. If we’re already at or beyond the Current format, early out. This means parsedJson.version will never be re-migrated.
2. Now, version-specific migration. Only assets with at least one map and older than Version1 need the enum-by-value migration.

AswinRajGopal:
Hi @ekcoh @LeoUnity All the yamato failures are because of regenerated DefaultInputActionAsset.cs. Please review. Thanks.

Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/DefaultInputActions.cs
```diff
@@ -1,47 +1,93 @@
...
+// </auto-generated>
+//------------------------------------------------------------------------------
+
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using UnityEngine.InputSystem;
 using UnityEngine.InputSystem.Utilities;
 
-// NOTE: This is mostly auto-generated from DefaultInput.actions. To regenerate, temporarily toggle on "Generate C# Class" and
```
LeoUnity: Yeah this diff is a bit scary, I don't have the historical context to advise here, my gut feeling is to just do the minimal changes needed manually, but I'll defer to @ekcoh here

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
LeoUnity: We never check against Current on this function, we actually never check against Current at all, Current is only used a single time to handle empty data, in this function we are only checking against the Version1 Specifically.
I guess it depends on what is the goal of this function, is this function meant to migrate the format to version1 only, or overtime will it be expanded to always migrate to Current as new versions are needed?

The comment at the top says:
        /// If parsedJson.version is older than Current, ...
But we never check against Current, we only check against Version1 directly, so if Current becomes different than Version1, this comment will be wrong.

Technically there is nothing wrong with this function today, it will work, it's more about communicating intent and preparing for future readability and modifications.

The name of the function is simply MigrateJson, which leads me to believe it should handle all migrations in the future as well, but the code inside it is very specific to Version1 migration, when a new version becomes available the first check and the version assignment would need to be updated creating an opportunity for mistakes.

But I think we spent too much time on this issue already, this is good enough for now, let's move on.

LeoUnity (approved):

ekcoh (approved):

ekcoh:
I fixed the conflicts on this branch

PR #2165, opened 2025-04-09
FIX: Gamepad stick inputs are not recognized in WebGL Player (ISXB-1090)

jfreire-unity:
### Description

Fixes issue reported [here](https://jira.unity3d.com/browse/ISXB-1090) due to WebGLGamepad's not having the stick up/down control values.

To fix the issue, the clamping range for those controls was changed and the test now asserts for the correct values.

Moreover,  the UI test was improved so that the actions set-up are `DefaultInputActions` that come by default with UIInputModule. Previously, the test didn't account for a composite binding that are used for `left/right/up/down` controls.
These changes to the UITest would have failed with the fix (I confirmed it).

### Testing status & QA

Tested in the WebGL player and it fixes the reported issue.

### Overall Product Risks

- Complexity: 0
- Halo Effect: 1

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Assets/Tests/InputSystem/Plugins/UITests.cs
```diff
@@ -2871,26 +2871,29 @@ public IEnumerator UI_CanOperateMultiplayerUILocallyUsingGamepads()
...
-        var gamepad = InputSystem.AddDevice<Gamepad>();
+        var gamepad = (Gamepad)InputSystem.AddDevice(deviceLayout);
 
         var scene = CreateTestUI();
 
-        var asset = ScriptableObject.CreateInstance<InputActionAsset>();
-        var map = asset.AddActionMap("map");
-        var moveAction = map.AddAction("move", type: InputActionType.PassThrough, binding: "<Gamepad>/*stick");
-        var submitAction = map.AddAction("submit", type: InputActionType.PassThrough, binding: "<Gamepad>/buttonSouth");
-        var cancelAction = map.AddAction("cancel", type: InputActionType.PassThrough, binding: "<Gamepad>/buttonEast");
```
jfreire-unity, line 2884: This doesn't reflect ou UI default actions setup. That's why I used the ones we use by default.

ritamerkl (approved):

ritamerkl:
Good to have the testcase for WEBGL now, thanks for adding

Pauliusd01:
Did not forget about this one, will try to check today/tomorrow

Pauliusd01 (changes requested):
Things checked: different testing scenes on firefox and chrome builds with xbox, ps4 and ps5 controllers both wired and on bluetooth. Everything worked as expected on chrome while on firefox the ps5 controller throws this error below on both wired and bluetooth. And only right side buttons work while analog sticks do not. I don't know if this is for us to fix or something on firefox people so let me know if I should change this to approve instead. (The button behaviour before these changes on firefox is the same except there's no error there)

image

Pauliusd01 (approved):
Discussed the firefox issue with Joao and these changes shouldn't have caused it. Will file it separately

PR #2166, opened 2025-04-10
FIX: PlayerInput component is redrawn all the time when the notification mode is set to Invoke Unity events

K-Tone:
### Description

A customer noted that if you change the notification mode on a PlayerInput component, that forces refresh of the inspector window every frame (can be observed by watching at the blinking dot at the left top corner).

The reason this happens is `PlayerInputEditor` component has lazy initilization logic in `OnInspectorGUI`

```           
            if (EditorGUI.EndChangeCheck() || !m_ActionAssetInitialized || CheckIfActionAssetChanged())
            {
                OnActionAssetChange();
                actionsWereChanged = true;
            }
```

When we reach this block, `m_ActionAssetInitialized` is always false -- since it's false by default, and it will keep being reset every frame, as you'll see in a bit.

So, we fall inside, set `actionsWereChanged`, and dive into `OnActionAssetChange` where we do all sorts of things, but most importantly, we set `m_ActionAssetInitialized` (unconditionally, in the beginning of the function). 

However, if we keep reading `OnInspectorGUI` further, we see this block

```
            if (EditorGUI.EndChangeCheck() || actionsWereChanged || !m_NotificationBehaviorInitialized)
                OnNotificationBehaviorChange();
```

Recall that `actionsWereChanged` is set in the first block and hasn't been changed yet, so we fall into the if block and call `OnNotificationBehaviorChange` where we have this wonderful final block:

```
case PlayerNotifications.InvokeUnityEvents:
                {
                    var playerInput = (PlayerInput)target;
                    if (playerInput.m_DeviceLostEvent == null)
                        playerInput.m_DeviceLostEvent = new PlayerInput.DeviceLostEvent();
                    if (playerInput.m_DeviceRegainedEvent == null)
                        playerInput.m_DeviceRegainedEvent = new PlayerInput.DeviceRegainedEvent();
                    if (playerInput.m_ControlsChangedEvent == null)
                        playerInput.m_ControlsChangedEvent = new PlayerInput.ControlsChangedEvent();
                    serializedObject.Update();

                    // Force action refresh.
                    m_ActionAssetInitialized = false;
                    Refresh();
```

You can see that we're resetting `m_ActionAssetInitialized` and updating the serializedObject as well. This is precisely what triggers the inspector window update, since the next update we will surely be following exactly the same steps.  

The patch I'm applying aims at minimal impact by not resetting `m_ActionAssetInitialized` when none of the device events need to be recreated any more. This looks safe. 

### Testing status & QA

Local testing, waiting on a QA.

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
@@ -417,17 +417,21 @@ namespace UnityEngine.InputSystem.Editor
                 case PlayerNotifications.InvokeUnityEvents:
                 {
                     var playerInput = (PlayerInput)target;
-                    if (playerInput.m_DeviceLostEvent == null)
-                        playerInput.m_DeviceLostEvent = new PlayerInput.DeviceLostEvent();
-                    if (playerInput.m_DeviceRegainedEvent == null)
-                        playerInput.m_DeviceRegainedEvent = new PlayerInput.DeviceRegainedEvent();
-                    if (playerInput.m_ControlsChangedEvent == null)
-                        playerInput.m_ControlsChangedEvent = new PlayerInput.ControlsChangedEvent();
-                    serializedObject.Update();
 
-                    // Force action refresh.
-                    m_ActionAssetInitialized = false;
-                    Refresh();
+                    bool areEventsDirty = (playerInput.m_DeviceLostEvent == null) || (playerInput.m_DeviceRegainedEvent == null) || (playerInput.m_ControlsChangedEvent == null);
+
+                    playerInput.m_DeviceLostEvent ??= new PlayerInput.DeviceLostEvent();
+                    playerInput.m_DeviceRegainedEvent ??= new PlayerInput.DeviceRegainedEvent();
+                    playerInput.m_ControlsChangedEvent ??= new PlayerInput.ControlsChangedEvent();
+
+                    if (areEventsDirty)
+                    {
+                        serializedObject.Update();
+
+                        // Force action refresh.
+                        m_ActionAssetInitialized = false;
+                        Refresh();
+                    }
                     break;
                 }
             }
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -19,6 +19,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed Input Actions code generation overwriting user files when the names happened to match. [ISXB-1257](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1257)
 - Fixed Input Actions code generation using locale-dependent rules when lowercasing and uppercasing strings. [ISXB-1406]
 - Fixed an issue when providing JoinPlayer with a specific split screen index. [ISXB-897](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-897)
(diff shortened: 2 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/PlayerInput/PlayerInputEditor.cs
```diff
@@ -417,17 +417,21 @@ private void OnNotificationBehaviorChange()
...
-                    m_ActionAssetInitialized = false;
-                    Refresh();
+
+                    bool areEventsDirty = (playerInput.m_DeviceLostEvent == null) || (playerInput.m_DeviceRegainedEvent == null) || (playerInput.m_ControlsChangedEvent == null);
+
+                    playerInput.m_DeviceLostEvent ??= new PlayerInput.DeviceLostEvent();
+                    playerInput.m_DeviceRegainedEvent ??= new PlayerInput.DeviceRegainedEvent();
+                    playerInput.m_ControlsChangedEvent ??= new PlayerInput.ControlsChangedEvent();
+
+                    if (areEventsDirty)
```
K-Tone, line 427: Unless I'm missing something, recreating is not required in the case that we changed nothing here actually. Worse, if we set m_ActionAssetInitialized to false, it will trigger another reset the next update frame (as explained in the description section).

K-Tone: (Also for a style-aware reviewer, the null-coalescing operation looks safe in this context because none of the events derive from Unity's Object)

ekcoh: Re-iinitialization: I find it very confusing in general that the this implementation (also prior to your changes) assigns these events from the editor onto the playerInput instance. It seems the method that includes this code (`OnNotificationBehaviorChange`) is only intended to be executed whenever there is a reconfiguration of what notification mechanism to use. So from what I can digest from a quick glance at this it seems like you have identified something that is not behaving like its intended if it leads to a loop of updates to the SO.

Re-style: Agreed (`UnityEventBase`), but it may be less convenient to debug (depending on the debugger). I do not have a strong opinion about this though so less code might be nicer.

K-Tone: Agreed, I find this confusing too. Just acknowledging that the traffic in this file is fairly light, seems it's ok to target for lower impact that minimises the risk of regression in this fragile code corner. Good question what is lower risk though. Blaming this file, I can see that there was this change that probably caused the refresh issue: https://github.com/Unity-Technologies/InputSystem/commit/3660631fd9b8eddc5ac5b2410cb0fc0b1fb22db8. I don't know, maybe we could only set actionsWereChanged when they were really changed and not when we were in fact trying to initialize the thing for the first (and consecutive times). That's maybe more risk to be honest. What do you think?

ekcoh: I think the fix looks fine (unless I am overlooking something). I think it makes sense to just manually check that we do not regress https://github.com/Unity-Technologies/InputSystem/commit/3660631fd9b8eddc5ac5b2410cb0fc0b1fb22db8 as part of QA/testing this fix.

ekcoh (approved):

Pauliusd01 (approved):
LGTM. Tried changing different player input settings looking for updating issues, checked that ISXB-286 is still fixed, used/modified player input in our samples

PR #2175, opened 2025-04-30
FIX: Reenabling the VirtualMouseInput component may lead to NullReferenceException

K-Tone:
### Description

Fixed an exception that happens when you toggle enable/disable a few times on the VirtualMouseInput component. 

This exception is occasionally raised from within `InputActionState.ApplyProcessors` as it tries to read from the processors array that is null in this case. By the normal flow of things, we shouldn't have tried to index into a null array since just a few lines above there is a check that ensures there are more than zero processors before doing this. However, when no bindings are present , `processorCount` will actually be a random value. Let me explain how. 

```        
        internal TValue ApplyProcessors<TValue>(int bindingIndex, TValue value, InputControl<TValue> controlOfType = null)
            where TValue : struct
        {
            var processorCount = bindingStates[bindingIndex].processorCount;
            if (processorCount > 0)
            {
                var processorStartIndex = bindingStates[bindingIndex].processorStartIndex;
                for (var i = 0; i < processorCount; ++i)
                {
                    if (processors[processorStartIndex + i] is InputProcessor<TValue> processor)
                        value = processor.Process(value, controlOfType);
                }
            }

            return value;
        }
```

So in `InputActionState.ApplyProcessors`, we read `processorCount` from the `bindingStates` array using `bindingIndex`. However, when there are no bindings present, `bindingIndex` has a chance to be a fairly large random value if you disable and enable the `VirtualMouseInput` component. Note that `bindingStates` is an unsafe array so nothing prevents from reading out of bounds. How can this`bindingIndex` go wrong?

Moving a bit up the call stack, in `InputAction.ReadValue` we see that `bindingIndex` is taken from `InputActionState.TriggerState`'s bindingIndex field. 

```
        public unsafe TValue ReadValue<TValue>()
            where TValue : struct
        {
            var state = GetOrCreateActionMap().m_State;
            if (state == null)
                return default(TValue);

            var actionStatePtr = &state.actionStates[m_ActionIndexInState];
            return actionStatePtr->phase.IsInProgress()
                ? state.ReadValue<TValue>(actionStatePtr->bindingIndex, actionStatePtr->controlIndex)
                : state.ApplyProcessors(actionStatePtr->bindingIndex, default(TValue));
        }
```

This field is normally correct, however when we disable the `VirtualMouseInput` component there is some interesting logic that tries to restore the binding index to what is stored in the corresponding arrays. It is part of `InputActionState.ResetActionState` method, and the problem here arises from `memory.actionBindingIndices` being an invalid pointer when no bindings are present. This is an unsafe pointer, and it is not null in case the array contains no elements by design of the `UnmanagedMemory` class. Reading from this pointer will get random data that is used after you Enable the `VirtualMouseInput` component. 

```
            actionState->phase = toPhase;
             actionState->controlIndex = kInvalidIndex;
             actionState->bindingIndex = memory.actionBindingIndices[memory.actionBindingIndicesAndCounts[actionIndex]];
             actionState->interactionIndex = kInvalidIndex;
             actionState->startTime = 0;
             actionState->time = 0;
```

Let me pull up some code from `UnmanagedMemory` that I find a bit problematic at the moment:

```
            public int sizeInBytes =>
                mapCount * sizeof(ActionMapIndices) + // mapIndices
                actionCount * sizeof(TriggerState) + // actionStates
                bindingCount * sizeof(BindingState) + // bindingStates
                interactionCount * sizeof(InteractionState) + // interactionStates
                controlCount * sizeof(float) + // controlMagnitudes
                compositeCount * sizeof(float) + // compositeMagnitudes
                controlCount * sizeof(int) + // controlIndexToBindingIndex
                controlCount * sizeof(ushort) * 2 + // controlGrouping
                actionCount * sizeof(ushort) * 2 + // actionBindingIndicesAndCounts
                bindingCount * sizeof(ushort) + // actionBindingIndices
                (controlCount + 31) / 32 * sizeof(int); // enabledControlsArray

               public void Allocate(int mapCount, int actionCount, int bindingCount, int controlCount, int interactionCount, int compositeCount)
             {
                 Debug.Assert(basePtr == null, "Memory already allocated! Free first!");
                 // NOTE: This depends on the individual structs being sufficiently aligned in order to not
                 //       cause any misalignment here. TriggerState, InteractionState, and BindingState all
                 //       contain doubles so put them first in memory to make sure they get proper alignment.
                 actionStates = (TriggerState*)ptr; ptr += actionCount * sizeof(TriggerState);
                 interactionStates = (InteractionState*)ptr; ptr += interactionCount * sizeof(InteractionState);
                 bindingStates = (BindingState*)ptr; ptr += bindingCount * sizeof(BindingState);
                 mapIndices = (ActionMapIndices*)ptr; ptr += mapCount * sizeof(ActionMapIndices);
                 controlMagnitudes = (float*)ptr; ptr += controlCount * sizeof(float);
                 compositeMagnitudes = (float*)ptr; ptr += compositeCount * sizeof(float);
                 controlIndexToBindingIndex = (int*)ptr; ptr += controlCount * sizeof(int);
                 controlGroupingAndComplexity = (ushort*)ptr; ptr += controlCount * sizeof(ushort) * 2;
                 actionBindingIndicesAndCounts = (ushort*)ptr; ptr += actionCount * sizeof(ushort) * 2;
                 actionBindingIndices = (ushort*)ptr; ptr += bindingCount * sizeof(ushort);
                 enabledControls = (int*)ptr; ptr += (controlCount + 31) / 32 * sizeof(int);
             }
```

So essentially this thing makes one allocation to host all the smaller bits of data including the aggregated arrays. However, the problem is that if some array has zero length,  we don't set its pointer to null. Instead, we just shift the current pointer in the blob by how much memory we needed for the previous array. Because of that, some array pointers can be equal (imagine a few empty arrays going in a sequence). Also, especially problematic are the last pointers in the blob -- because if they are zero sized, then they are set to invalid pointer, pointing to the first byte right after the blob. That's precisely the reason why we had random bindingIndex - note how `actionBindingindices` will be set to point to the first byte after the blob in case bindingCount is zero. 

I believe there are a few ways we could attack this problem. 

One, that would aim for minimal footprint is to change just one line in `InputActionState.ApplyProcessors` to also include a check for an empty processors array, like this:

```
if (processorCount > 0 && processors != null)
```

However, in my mind that is slightly unfortunate because it leaves the core issue of having bad bindingIndices around, and it also still leaves us with invalid pointers being around. It's only until next time someone doesn't check for an empty array in the right way. When debugging, it's somewhat uneasy to understand the pointers that look somewhat reasonable are actually pointers to unallocated space. 

Because of that, I thought to propose a change that assigns null to any pointers in the `UnmanagedMemory` class that reflect zero-sized arrays. That way, if someone tries to access them, we get an instant exception that is easy to diagnose. Please advise what you think. 

### Testing status & QA

Local testing

### Overall Product Risks

- Complexity: Medium
- Halo Effect: Medium

### Comments to reviewers

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -872,7 +872,9 @@ namespace UnityEngine.InputSystem
             // Wipe state.
             actionState->phase = toPhase;
             actionState->controlIndex = kInvalidIndex;
-            actionState->bindingIndex = memory.actionBindingIndices[memory.actionBindingIndicesAndCounts[actionIndex]];
+            var idx = memory.actionBindingIndicesAndCounts[actionIndex];
+            var bindingCount = memory.actionBindingIndicesAndCounts[actionIndex] + 1;
+            actionState->bindingIndex = memory.actionBindingIndices != null ? memory.actionBindingIndices[idx] : 0;
             actionState->interactionIndex = kInvalidIndex;
             actionState->startTime = 0;
             actionState->time = 0;
@@ -2879,6 +2881,9 @@ namespace UnityEngine.InputSystem
         internal TValue ApplyProcessors<TValue>(int bindingIndex, TValue value, InputControl<TValue> controlOfType = null)
             where TValue : struct
         {
+            if (totalBindingCount == 0)
+                return value;
+
             var processorCount = bindingStates[bindingIndex].processorCount;
             if (processorCount > 0)
             {
@@ -4139,6 +4144,19 @@ namespace UnityEngine.InputSystem
 
             public ActionMapIndices* mapIndices;
 
+            // This is Anthony's version of block allocation from a blob
+            private byte* AllocFromBlob(ref byte* top, int size)
+            {
+                if (size == 0)
+                    return null;
+                
+                var allocation = top;
+
+                top += size;
+
+                return allocation;
+            }
+
             public void Allocate(int mapCount, int actionCount, int bindingCount, int controlCount, int interactionCount, int compositeCount)
             {
                 Debug.Assert(basePtr == null, "Memory already allocated! Free first!");
@@ -4164,17 +4182,17 @@ namespace UnityEngine.InputSystem
                 // NOTE: This depends on the individual structs being sufficiently aligned in order to not
                 //       cause any misalignment here. TriggerState, InteractionState, and BindingState all
                 //       contain doubles so put them first in memory to make sure they get proper alignment.
-                actionStates = (TriggerState*)ptr; ptr += actionCount * sizeof(TriggerState);
-                interactionStates = (InteractionState*)ptr; ptr += interactionCount * sizeof(InteractionState);
-                bindingStates = (BindingState*)ptr; ptr += bindingCount * sizeof(BindingState);
-                mapIndices = (ActionMapIndices*)ptr; ptr += mapCount * sizeof(ActionMapIndices);
-                controlMagnitudes = (float*)ptr; ptr += controlCount * sizeof(float);
-                compositeMagnitudes = (float*)ptr; ptr += compositeCount * sizeof(float);
-                controlIndexToBindingIndex = (int*)ptr; ptr += controlCount * sizeof(int);
-                controlGroupingAndComplexity = (ushort*)ptr; ptr += controlCount * sizeof(ushort) * 2;
-                actionBindingIndicesAndCounts = (ushort*)ptr; ptr += actionCount * sizeof(ushort) * 2;
-                actionBindingIndices = (ushort*)ptr; ptr += bindingCount * sizeof(ushort);
-                enabledControls = (int*)ptr; ptr += (controlCount + 31) / 32 * sizeof(int);
+                actionStates = (TriggerState*)AllocFromBlob(ref ptr, actionCount * sizeof(TriggerState));
+                interactionStates = (InteractionState*)AllocFromBlob(ref ptr, interactionCount * sizeof(InteractionState));
+                bindingStates = (BindingState*)AllocFromBlob(ref ptr, bindingCount * sizeof(BindingState));
+                mapIndices = (ActionMapIndices*)AllocFromBlob(ref ptr, mapCount * sizeof(ActionMapIndices));
+                controlMagnitudes = (float*)AllocFromBlob(ref ptr, controlCount * sizeof(float));
+                compositeMagnitudes = (float*)AllocFromBlob(ref ptr, compositeCount * sizeof(float));
+                controlIndexToBindingIndex = (int*)AllocFromBlob(ref ptr, controlCount * sizeof(int));
+                controlGroupingAndComplexity = (ushort*)AllocFromBlob(ref ptr, controlCount * sizeof(ushort) * 2);
+                actionBindingIndicesAndCounts = (ushort*)AllocFromBlob(ref ptr, actionCount * sizeof(ushort) * 2);
+                actionBindingIndices = (ushort*)AllocFromBlob(ref ptr, bindingCount * sizeof(ushort));
+                enabledControls = (int*)AllocFromBlob(ref ptr, (controlCount + 31) / 32 * sizeof(int));
             }
 
             public void Dispose()
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -593,7 +593,8 @@ namespace UnityEngine.InputSystem
                 if (newBindingState.isComposite)
                 {
                     var compositeIndex = newBindingState.compositeOrCompositeBindingIndex;
-                    memory.compositeMagnitudes[compositeIndex] = oldState.compositeMagnitudes[compositeIndex];
+                    if (oldState.compositeMagnitudes != null)
+                        memory.compositeMagnitudes[compositeIndex] = oldState.compositeMagnitudes[compositeIndex];
                 }
 
                 var actionIndex = newBindingState.actionIndex;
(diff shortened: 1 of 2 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -4139,6 +4143,19 @@ public struct UnmanagedMemory : IDisposable
 
             public ActionMapIndices* mapIndices;
 
+            // This is Anthony's version of block allocation from a blob
+            private byte* AllocFromBlob(ref byte* top, int size)
+            {
+                if (size == 0)
+                    return null;
```
K-Tone, line 4150: The largest diff here is this size check, which would guarantee that a zero-size array would be null by design. Priorly zero-sized data would cause invalid non-null pointers. So code that wasn't accurately written to check for bounds would access random memory. This way we might get some crashes more apparent, which is good as otherwise it's super tricky to debug code that uses garbage.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -872,7 +872,8 @@ public void ResetActionState(int actionIndex, InputActionPhase toPhase = InputAc
             // Wipe state.
             actionState->phase = toPhase;
             actionState->controlIndex = kInvalidIndex;
-            actionState->bindingIndex = memory.actionBindingIndices[memory.actionBindingIndicesAndCounts[actionIndex]];
+            var idx = memory.actionBindingIndicesAndCounts[actionIndex];
+            actionState->bindingIndex = memory.actionBindingIndices != null ? memory.actionBindingIndices[idx] : 0;
```
ekcoh, line 877: Would it even be smart to set bindingIndex to -1 to indicate it's not a valid index? 
It would allow doing (actionState->bindingIndex >= 0) as an indirect null-check, but if other indices are zero it makes sense to keep it at zero for consistency

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -4139,6 +4143,19 @@ public struct UnmanagedMemory : IDisposable
...
             public ActionMapIndices* mapIndices;
 
+            // This is Anthony's version of block allocation from a blob
+            private byte* AllocFromBlob(ref byte* top, int size)
+            {
+                if (size == 0)
+                    return null;
+                
+                var allocation = top;
+
```
ekcoh, line 4151: Minor: I think we can remove these blank lines in-between?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
ekcoh: Makes sense to have it in a function like this, looks like it could even be static?

ekcoh: Might it even be a local function within Allocate?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
K-Tone: This is a very good question to be honest. I'm surprised it's not been set to -1 before. I should see if that's safe to change in order not to break some stuff somewhere, this code is kinda interconnected quite a bit.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
K-Tone: Applied changes according to all suggestions in this function.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
ekcoh: Awesome, it was just a thought and not something I consider a must-have if it adds complexity

K-Tone:
Ok we have some failing tests apparently - Actions_CanAddCompositeBindingsToActions_AfterActionHasAlreadyResolvedControls and Actions_WhenShortcutsEnabled_CanConsumeInput, maybe a few others as well. Investigating.

ekcoh (approved):
Looks like you will consider the index (not set value), but changes LGTM to me already now so approving.

ekcoh (changes requested):
Just saw the failing tests so updating this to "Request Changes" since clearly more work is needed on figuring out why

K-Tone:
Dear reviewers! All automations are green now on this PR, may I ask for another review please? Thanks. :)

As a side note, I thought not to reset bindingIndex to -1 here, because I'm feared mixing two types of possible regressions in one PR: 
a) regressions coming from me setting null to zero-sized arrays in UnmanagedMemory thing already (the core of this PR)
b) a possible additional regression coming from -1 set to bindingIndex

Obviously, I agree that invalid bindingIndex should be -1 rather than 0, it's just a matter of doing it here or not doing it right here.

Pauliusd01 (approved):
It seemed fine to me both in the testing scenes and the debugger but I have to mention that I wasn't able to reproduce the original issue neither on the 'reproduced in' version or the newer ones

K-Tone:
Hey @Pauliusd01, I recalled that this case was also linked to this one: https://jira.unity3d.com/browse/ISXB-1283, could I ask you to try reproducing the problem there as well? I don't think I have the setup locally to do that, but the callstack looks exactly the same as here. Thanks!

Pauliusd01:
> Hey @Pauliusd01, I recalled that this case was also linked to this one: https://jira.unity3d.com/browse/ISXB-1283, could I ask you to try reproducing the problem there as well? I don't think I have the setup locally to do that, but the callstack looks exactly the same as here. Thanks!

Sorry I don't have any VR devices at all

ekcoh (approved):

PR #2183, opened 2025-05-16
CHANGE: Refactor class conditional guards

jdiehlUnity:
### Description

Moves around some conditional guards so that we don't unnecessarily wrap entire classes. This stops downstream packages from having to also wrap Input System objects with similar conditionals.

Only device class definitions that get replaced by specific packages should be entirely wrapped.

This PR is a copy of the internal PR created here: https://github.cds.internal.unity3d.com/unity/com.unity.inputsystem/pull/5

### Changes made
Removes  `#if UNITY_INPUT_SYSTEM_ENABLE_XR && (ENABLE_VR || UNITY_GAMECORE) && !PACKAGE_DOCS_GENERATION`
guards from wrapping entire class files under _InputSystem/Plugins/XR/Devices_ and _InputSystem/Plugins/XR/Haptics_.
This makes it so any packages that want to use these Input System package objects don't have to also wrap the use of those objects with `#if ENABLE_VR` guards. Classes that should ignore Package Generation have been added to the filter.yml file.

We now only wrap the use of device descriptors with `#if UNITY_INPUT_SYSTEM_ENABLE_XR` to prevent use when the XR Module package is not installed.

### Testing status & QA

Engines Supported: 6.2, 6.1, 6.0, 2022.3, 2021.3

**Test 1:** 

- 	Create project (or use an existing one)
- 	Install Input System package
- 	Install OpenXR package (this package uses the Input System but currently does not wrap any objects).
- 	Switch platform from Standalone or Android into tvOS or QNX (ENABLE_VR conditional wont be defined on these platforms)

 	_Expected result:_ No compiler errors

	
**Test 2:**

- 	Create project (or use an existing one)
- 	Install Input System package.
- 	Install the com.unity.xr.oculus, com.unity.xr.openvr, and com.unity.xr.windowsmr packages (these packages override certain device class definitions).

       _Expected result:_ No compiler errors

**Test 3:**

- 	Create project (or use an existing one)
- 	Install Input System package.
- 	Implement something that tries to leverage an XRHMD or XRController object
- 	Switch platform to tvOS or QNX (ENABLE_VR conditional wont be defined on these platforms)
- 	Execute code in Editor or in a build for the new platform.

	_Expected result:_ No runtime errors

### Overall Product Risks

- Complexity: 
- Halo Effect: 

### Comments to reviewers

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=2183) <br/>All committers have signed the CLA.

surfnerd (approved):
Approved the internal PR, approving this based on my review there. See link in description for discussion.

surfnerd:
should add @ekcoh

jfreire-unity:
@Pauliusd01 I'm adding you just to make sure there's no breakage. Going through the tests mentioned in the PR description should be enough

Packages/com.unity.inputsystem/Documentation~/filter.yml
```diff
@@ -37,3 +37,54 @@ apiRules:
   - exclude:
       uidRegex: ^UnityEngine\.InputSystem\.InputSystem\.runInBackground$
       type: Member
+  - exclude:
```
jfreire-unity, line 40: I'm understanding this correctly: this means that these API won't be published? Any reason why? I'm just out of context probably.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/Oculus.cs
```diff
@@ -1,6 +1,5 @@
-// ENABLE_VR is not defined on Game Core but the assembly is available with limited features when the XR module is enabled.
+#if !DISABLE_BUILTIN_INPUT_SYSTEM_OCULUS
 // Docs generation is skipped because these are intended to be replaced with the com.unity.xr.oculus package.
-#if UNITY_INPUT_SYSTEM_ENABLE_XR && (ENABLE_VR || UNITY_GAMECORE) && !DISABLE_BUILTIN_INPUT_SYSTEM_OCULUS && !UNITY_FORCE_INPUTSYSTEM_XR_OFF && !PACKAGE_DOCS_GENERATION
```
jfreire-unity, line 3: Thanks for removing this messy define. Are we ok with removing the UNITY_GAMECORE as well? I assume that was something Microsoft related?

jfreire-unity (approved):
Looks good. Just some non-blocking questions.
I also da Input QA to make a sanity check.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/Devices/Oculus.cs
jdiehlUnity: Yes we only added UNITY_GAMECORE to these places so that when you're on the gamecore platform these classes will be compiled. Since we are no longer wrapping entire classes they will always be compiled.

Packages/com.unity.inputsystem/Documentation~/filter.yml
jdiehlUnity: Hmm maybe I need to change this then. Besides the XRHMD devices, I started using this file to stop some xml errors I was getting from the yamato tests. I assumed the older define logic was hiding those same errors.

Pauliusd01 (approved):
I have no XR devices so only ran through non XR testing/sample scenes on windows playmode and player.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -1,52 +1,71 @@
...
+    /// <summary>
+    /// The base type of all XR head mounted displays.  This can help organize shared behaviour across all HMDs.
+    /// </summary>
+    ///
+    /// <remarks>
+    ///
+    /// To give your head tracking an extra update before rendering:
+    /// First, enable before render updates on your Device.
+    ///
+    ///     // JSON
```
ekcoh, line 17: I see this wasn't introduced on this PR (rather moved), but it's not clear what JSON this is referring to which might not be helpful for users consuming this information. I would recommend (while moving this) to add clarity around what JSON is expected to be modified accordingly.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -1,52 +1,71 @@
...
-        ///
-        /// Then, make sure you put extra `StateEvents` for your HMD on the queue right in time before rendering. Also, if your HMD is a combination of non-tracking and tracking controls, you can update just the tracking by sending a delta event instead of a full state event.
-        ///
-        /// </remarks>
-
         [InputControl(noisy = true)]
         public Vector3Control leftEyePosition { get; protected set; }
+
+        /// <summary>
+        /// Accessor for left eye rotation.
```
ekcoh, line 37: Not sure, but I suspect this might cause package violation due to length (short) and missing code example. Similar for properties below.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -72,22 +91,12 @@ public class XRController : TrackedDevice
         /// <remarks>
         /// If there is no left hand connected, this will be null.
         /// This also matches any currently tracked device that contains the 'LeftHand' device usage.
-        /// <example>
-        /// <code>
-        /// // To set up an Action to specifically target
-        /// // the left-hand XR controller:
-        ///
+        /// To set up an Action to specifically target
```
ekcoh, line 94: Any reason the example and code tags was removed from this?

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -97,10 +106,13 @@ public class XRController : TrackedDevice
         /// <remarks>If there is no left hand connected, this will be null. This also matches any currently tracked device that contains the 'RightHand' device usage.</remarks>
         public static XRController rightHand => InputSystem.GetDevice<XRController>(CommonUsages.RightHand);
 
+        /// <summary>
+        /// Override for FinishSetup().
```
ekcoh, line 110: Its clear from line 112 its an override for FinishSetup. I would suggest:

`<inheritdoc />`

ekcoh: ...since it doesn't (it cannot) change the contract of the API function it should just inherit the base documentation.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -119,11 +132,15 @@ protected override void FinishSetup()
     /// </summary>
     public class XRControllerWithRumble : XRController
     {
+        /// <summary>
+        /// Sends an impulse command with the given amplitude and duration.
+        /// </summary>
+        /// <param name="amplitude"> Amplitude of the impulse.</param>
```
ekcoh, line 138: I would recommend specifying expected or valid range for this amplitude to make it easier for users.

Packages/com.unity.inputsystem/InputSystem/Plugins/XR/GenericXRDevice.cs
```diff
@@ -119,11 +132,15 @@ protected override void FinishSetup()
...
+        /// <param name="duration"> Duration of the impulse.</param>
```
ekcoh, line 139: Similar here, what is acceptable, is zero ok? Is negative duration ok?

ekcoh: Looks like it mightn't be checked before allocating for command and sending down to native

ekcoh (commented):
Nice to see simplifications of compile time symbols which this code base already has too many off. I am not sure about the historical reason why XR defies are conditionally compiled at all since this is not done for e.g. Keyboards, Mice, Gamepads, Touchpads or other device types. Maybe you can remind me why we need to conditionally compile these at all? I suspect it is to remove them when no runtime backing them up is present, but generally I wonder if the API should be dynamic like this since it adds complexity and compile time burden.

jdiehlUnity:
> Nice to see simplifications of compile time symbols which this code base already has too many off. I am not sure about the historical reason why XR defies are conditionally compiled at all since this is not done for e.g. Keyboards, Mice, Gamepads, Touchpads or other device types. Maybe you can remind me why we need to conditionally compile these at all? I suspect it is to remove them when no runtime backing them up is present, but generally I wonder if the API should be dynamic like this since it adds complexity and compile time burden.

Thanks for your review. The problem is that we should have never wrapped entire classes conditionally because this causes every package that uses the Input System to also implement conditionals unnecessarily. We should only wrap necessary types that depend on the XR Module package or for calls that may try to invoke native functions from the XR Module packages.

Regarding the rest of your comments, all the comment changes in this PR were to stop Yamato XML errors that were previously hidden. At this point I would very much like this PR to get merged and the Input Team to come back and update any comments that aren't sufficient

ekcoh (approved):
I realise the main symbol define is dependent on the presence of the XR module being enabled (as stated in description). Since this is more of a refactor than changing anything (apart from simplifying) I think this is an improvement, thanks for doing it @jdiehlUnity. I will approve these changes but suspect that some documentation generating package verification errors might be present (See specific comments).

jamesmcgill:
@jdiehlUnity and maybe @Pauliusd01 
We have a blocking issue with the QNX release. I think it's related to this PR but still looking into it.
https://unity-ci.cds.internal.unity3d.com/job/54188255

In the PR description there is a test case section that mentions: `Switch platform from Standalone or Android into tvOS or QNX (ENABLE_VR conditional wont be defined on these platforms)`
It also mentions "expected results". But it's not clear if the tests were run and what the "actual results" were.

Do you know if the test cases were actually performed or if they were just planned test cases?

PR #2186, opened 2025-05-21
FIX: Make sure we respect the default button press point for triggers, same for Touchscreen's tap delay time and tap radius squared

K-Tone:
### Description

This is an attempt to fix the problem where in Editor we would really apply the default input settings correctly. Like, we would read them, but we wouldn't correctly update the derived static properties so whenever that happens, we have issues like the default button press point being incorrect.

Looking at code, I believe it's not just the default button press point that was incorrect, I think all of these folks were incorrect too (though the bug report won't mention them):

```            Touchscreen.s_TapTime = settings.defaultTapTime;
            Touchscreen.s_TapDelayTime = settings.multiTapDelayTime;
            Touchscreen.s_TapRadiusSquared = settings.tapRadius * settings.tapRadius;
            // Extra clamp here as we can't tell what we're getting from serialized data.
            ButtonControl.s_GlobalDefaultButtonPressPoint = Mathf.Clamp(settings.defaultButtonPressPoint, ButtonControl.kMinButtonPressPoint, float.MaxValue);
            ButtonControl.s_GlobalDefaultButtonReleaseThreshold = settings.buttonReleaseThreshold;
```

### Testing status & QA

Local testing, pending a QA pass. 

### Overall Product Risks

- Complexity:  Low
- Halo Effect: Low

### Comments to reviewers

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -23,7 +23,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where an action with a name containing a slash "/" could not be found via `InputActionAsset.FindAction(string,bool)`. [ISXB-1306](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1306).
 - Fixed Gamepad stick up/down inputs that were not recognized in WebGL. [ISXB-1090](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1090)
 - Fixed reenabling the VirtualMouseInput component may sometimes lead to NullReferenceException. [ISXB-1096](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1096)
-- Fixed the default button press point not being respected in Editor (as well as some other Touchsreen properties). [ISXB-1152](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1152)
+- Fixed the default button press point not being respected in Editor (as well as some other Touchscreen properties). [ISXB-1152](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1152)
 - Fixed PlayerInput component automatically switching away from the default ActionMap set to 'None'.
 - Fixed a console error being shown when targeting visionOS builds in 2022.3.
 - Fixed a Tap Interaction issue with analog controls. The Tap interaction would keep re-starting after timeout. [ISXB-627](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-627)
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -4122,7 +4122,8 @@ internal void RestoreStateWithoutDevices(SerializedState state)
 
             if (m_Settings != null)
                 Object.DestroyImmediate(m_Settings);
-            m_Settings = state.settings;
+                
+            settings = state.settings;
```
K-Tone, line 4126: So the theory here is that using the property instead of a direct setter makes sure we update all the derived data (static caches) -- that happens in ApplySettings that is ignored in here.

K-Tone:
Just made this PR ready for review because the CI redness  on iOS and tvOS looks caused by xcodebuild timing out while waiting for some update or whatever. Not related to this change.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Ah, good catch, since the property setter triggers the apply settings.

Packages/com.unity.inputsystem/InputSystem/InputSystem.cs
```diff
@@ -3878,7 +3878,7 @@ private static void Reset(bool enableRemoting = false, IInputRuntime runtime = n
                 s_Manager.UninstallGlobals();
             }
 
-            // Create temporary settings. In the tests, this is all we need. But outside of tests,d
+            // Create temporary settings. In the tests, this is all we need. But outside of tests,
```
ekcoh, line 3881: Always good to fix a typo.

ekcoh (approved):
Approving this fix but it would be great if we could add the missing test coverage for this, which seems unrelated (as you say) to the specific setting but relates to the restoration of settings after e.g. a domain reload. I wonder if we could get it covered by exiting and then re-entering playmode from a test.....

Pauliusd01:
Did not forget this one, will try to review it today/tomorrow

Pauliusd01 (approved):
LGTM, only one note though. Should these UTR version changes go in with this PR? I think you added them to test if it solves your CI problems? I wouldn't mind but just checking

K-Tone:
In the end, I think I would send the UTR change with this PR because otherwise it's very painful to get a green automation run on this PR otherwise

K-Tone:
I didn't make a test here in the end, I think I just don't know how to make a meaningful one without making a new test project with the updated defaults

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed an issue where an action with a name containing a slash "/" could not be found via `InputActionAsset.FindAction(string,bool)`. [ISXB-1306](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1306).
 - Fixed Gamepad stick up/down inputs that were not recognized in WebGL. [ISXB-1090](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1090)
 - Fixed reenabling the VirtualMouseInput component may sometimes lead to NullReferenceException. [ISXB-1096](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1096)
+- Fixed the default button press point not being respected in Editor (as well as some other Touchsreen properties). [ISXB-1152](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1152)
```
K-Tone:
```suggestion
- Fixed the default button press point not being respected in Editor (as well as some other Touchscreen properties). [ISXB-1152](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1152)
```

PR #2192, opened 2025-06-05
FIX: disabling the InputSystemUIInputModule component resets all actions to None

K-Tone:
### Description

This PR fixes a problem where disabling the InputSystemUIInputModule component would reset all the actions, even if they had been priorly set by user. 

This started to happen after a fix for another regression landed, where we unassigned all actions from the components OnDisable. I believe this is wrong, since the only thing we really need to be resetting is the default actions asset that shouldn't normally survive re-reentering playmode. 

### Testing status & QA

Local testing, a new regression test, pending a QA pass

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/UITests.InputModuleTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.InputModuleTests.cs
@@ -157,6 +157,17 @@ internal partial class UITests
             Assert.IsTrue(callbackCheck.pointerData.fullyExited == true);
         }
 
+        [UnityTest]
+        [Description("Regression test for https://jira.unity3d.com/browse/ISXB-1493")]
+        public IEnumerator DisablingDoesNotResetActions()
+        {
+            m_InputModule.enabled = false;
+
+            yield return null;
+
+            Assert.IsNotNull(m_InputModule.cancel, "Disabling component shouldn't lose its data.");
+        }
+
         public class PointerExitCallbackCheck : MonoBehaviour, IPointerExitHandler
         {
             public PointerEventData pointerData { get; private set; }
--- a/Assets/Tests/InputSystem/Plugins/UITests.InputModuleTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/UITests.InputModuleTests.cs
@@ -159,13 +159,19 @@ internal partial class UITests
 
         [UnityTest]
         [Description("Regression test for https://jira.unity3d.com/browse/ISXB-1493")]
-        public IEnumerator DisablingDoesNotResetActions()
+        public IEnumerator DisablingDoesNotResetUserActions()
         {
+            var actions = new DefaultInputActions();
+            m_InputModule.actionsAsset = actions.asset;
+            m_InputModule.cancel = InputActionReference.Create(actions.UI.Cancel);
+
             m_InputModule.enabled = false;
 
             yield return null;
 
             Assert.IsNotNull(m_InputModule.cancel, "Disabling component shouldn't lose its data.");
+
+            actions.Dispose();
         }
 
         public class PointerExitCallbackCheck : MonoBehaviour, IPointerExitHandler
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -1677,6 +1677,12 @@ namespace UnityEngine.InputSystem.UI
             DisableAllActions();
             UnhookActions();
 
+            // In the case we've been initialialised with default actions, we want to unalloc them
+            if (defaultActions != null && defaultActions.asset == actionsAsset)
+            {
+                UnassignActions();
+            }
+
             base.OnDisable();
         }
 
--- a/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
@@ -1528,6 +1528,17 @@ namespace UnityEngine.InputSystem.UI
             set => SwapAction(ref m_TrackedDevicePositionAction, value, m_ActionsHooked, m_OnTrackedDevicePositionDelegate);
         }
 
+        // We should dispose the static default actions thing because otherwise it will survive domain reloads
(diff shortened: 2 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1665,7 +1676,6 @@ protected override void OnDisable()
             InputActionState.s_GlobalState.onActionControlsChanged.RemoveCallback(m_OnControlsChangedDelegate);
             DisableAllActions();
             UnhookActions();
-            UnassignActions();
```
K-Tone, line 1668: ^ This line is very wrong, it results in user data being lost on component deactivation.

K-Tone:
As to the catches by Copilot:

(1) This is fine. Historically, we've only had UnassignActions since the beginning of the year, and that was a mistake, so I think future maintainers don't need to know we made that mistake and corrected it quickly

(2) There is a short, but to the point comment actually. Not sure I'd like to expand it further.

Pauliusd01 (changes requested):
Found an issue with the UI default actions getting added in after a scene load. Steps are: Open our main repo as a unity project -> top dropdown QA Tools -> Add all core samples / open core scene -> import the textmeshpro essentials if needed -> make sure the event system is set to use a project wide actions asset if it isn't already -> enter play mode and observe the action count -> pick anything from the scene loader -> esc -> return to scene picker and observe the action count again. This does not reproduce in Develop

https://github.com/user-attachments/assets/bddb4d13-cabb-480f-bc79-cc757d0e1a4f

K-Tone:
Hey @Pauliusd01, could you try it again please? With this new patch, I tried:

(1) With actions asset set before play mode
- Disabling component doesn't lead to all actions lost any more
- Entering playmode and loading the project-wide test scene and navigating the menu doesn't duplicate actions

(2) With no actions asset set before play mode
- Entering playmode sets the default actions correctly
- Disabling domain reload and entering play mode twice doesn't result in exception

Note that when you enter play mode with InputSystemUIInputModule that had no actions asset assigned it results in us creating some default actions asset behind the scenes and using it to assign all actions. Disabling this component results in us destroying the default asset, and thus the actions get unassigned as well, so you do get "None" in this case, which I guess is kinda by design (though unfortunate). Enabling it back will restore defaults. If I were to redesign this component from scratch, I would make an explicit checkbox in the component called "Use defaults", and in this case I would hide all the actions, so that the user doesn't get confused by what is going on behind the scenes. This would also simplify code a little bit.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1665,7 +1676,12 @@ protected override void OnDisable()
...
+
+            // In the case we've been initialialised with default actions, we want to unalloc them
```
K-Tone: We want to do this because if we don't dealloc them, then no one would

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1528,6 +1528,17 @@ public InputActionReference trackedDevicePosition
             set => SwapAction(ref m_TrackedDevicePositionAction, value, m_ActionsHooked, m_OnTrackedDevicePositionDelegate);
         }
 
+        // We should dispose the static default actions thing because otherwise it will survive domain reloads
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        static void ResetDefaultActions()
+        {
+            if (defaultActions != null)
+            {
+                defaultActions.Dispose();
```
K-Tone, line 1537: I guess, calling Dispose is just to stay on the safe side since the only data default actions stores is a ref to ScriptableObject that won't survive even with domain reload off. It will be in this intermediate state though, where its native data is gone but the instance is not null until the next gc run (however you'll see it as "null" in debugger, and ==null will be true as well, thanks to Unity's Equals overload for all Object descendants).

Assets/Tests/InputSystem/Plugins/UITests.InputModuleTests.cs
```diff
@@ -157,6 +157,23 @@ public IEnumerator PointerExitChildShouldFullyExit()
             Assert.IsTrue(callbackCheck.pointerData.fullyExited == true);
         }
 
+        [UnityTest]
+        [Description("Regression test for https://jira.unity3d.com/browse/ISXB-1493")]
```
ekcoh, line 161: Great you added a specific test for the regression

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
ekcoh: Does this comment imply that it should be `Object.ReferenceEquals(null, defaultActions)` in if statement above? If its disposable, yes it should be disposed, even if the implementation would do nothing IMO.

ekcoh (approved):
Nice and tidy, looks good to me, great you added a test for the regression. Only a single comment/question regarding the null check.

ekcoh:
@K-Tone Looks like you need to rerun the formatter on this branch before CI will be happier. Other failures look like instabilities.

Pauliusd01 (approved):

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
K-Tone: I think I wasn't clear, pardon. DefaultActions is fine to check for null like above since it's not destroyed natively. What is in the intermediate state happens to be aggregated in defaultActions (DefaultActions has a ref to ScriptableObject, so it's the ScriptableObject would have to be checked with ReferenceEquals if we needed to, but we don't).

PR #2194, opened 2025-06-13
FIX: DefaultActionMap dropdown set to <None> by default (ISXB-1559)

ritamerkl:
### Description

Like described in this [ticket](https://jira.unity3d.com/browse/ISXB-1559), the default ActionMap dropdown was set to <None> by default, this PR fixes this by setting the first ActionMap as a default option on changing/first adding the Action Asset. 

### Testing status & QA

Manual testing done. 

### Overall Product Risks

- Complexity: low
- Halo Effect: low

### Comments to reviewers

Reviewers:
I renamed the OnActionAssetChange() method since this also happens as soon as the component window needs to be re-initialised after navigating away from the view, since the dependent properties are not static. 
I edited the CheckIfActionAssetChanged() method, since it's not changing if the property just lost the reference when navigating away from the window. 

For testing:
- test changing action asset in the PlayerInput component and observe the pre-picked actionMap
- test adding a PlayerInput component and add a action asset and observe the pre-picked actionMap
- changing the actionMap to <None>/or a specific action map should last after navigating away from the gameobject in the hierachy
- changing the actionmap to <None>/or a specific action map should last after switching to PlayMode/EditMode

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Pauliusd01:
Can new automated tests be added for this?

Pauliusd01 (approved):
LGTM, correct map is picked and kept between play modes when adding fresh assets to player input, switching them out, modifying the maps on an already present one, etc. 

The only odd thing I noticed was that when a PWA asset exists and you create a player input component it pre-picks the PWA asset to it. Which is known and probably getting changed but the weird part is that it still pre-picks the None map. Not like it matters as that setting isn't respected with PWA assets at the moment but that was the only odd one out scenario

ritamerkl:
> The only odd thing I noticed was that when a PWA asset exists and you create a player input component it pre-picks the PWA asset to it. Which is known and probably getting changed but the weird part is that it still pre-picks the None map. Not like it matters as that setting isn't respected with PWA assets at the moment but that was the only odd one out scenario

I think that should be changed anyway since using the PWA asset for PlayerInput is not recommended - do we have a ticket to remove the automated prefill with the PWA asset?

jfreire-unity (approved):

PR #2199, opened 2025-06-30
FIX: All InputControls change when modifying one

K-Tone:
This reverts commit 2a37caac288ac09bc9122234339dc5df8d3a0ca6.

### Description

We received a regression from a fix for ISXB-1221 that makes it impossible to have more than one InputControl components at the same time. This is caused by the static variable used internally. We have to roll this back and return back to the drawing board in order to have a better fix for the original case. I'm closing the regression as fixed and reopening ISXB-1221 instead. 

### Testing status & QA

Local testing

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -31,6 +31,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed the TreeView compilation warnings when used with Unity 6.2 beta (ISX-2320)
 - Fixed actions being reset when disabling the InputSystemUIInputModule component [ISXB-1493](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1493)
 - Fixed a memory leak when disabling and enabling the InputSystemUIInputModule component at runtime [ISXB-1573](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1573)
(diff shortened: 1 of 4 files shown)


ekcoh (approved):
Revert looks correct to me. Guess the reverted case is reopened as part of landing this revert.

Pauliusd01:
I don't know if I agree that reverting this is the best option as the older bug will be noticed by everyone and the one being fixed here not as much. Maybe keep this one off from the upcoming release? Or someone can smack some sense into me I don't know, just feels like we're trading pinky pain to a headache

jfreire-unity:
> I don't know if I agree that reverting this is the best option as the older bug will be noticed by everyone and the one being fixed here not as much. Maybe keep this one off from the upcoming release? Or someone can smack some sense into me I don't know, just feels like we're trading pinky pain to a headache

Based on this, I'd prefer we don't land this in the upcoming release and revert+fix in the next one.

jfreire-unity:
My reasoning is that the fix we're reverting in this PR was a 6.1 release concern issue. So we should avoid reintroducing it.

Pauliusd01 (approved):
Update: I thought the bug with ISXB-1531 was purely in the UI which wouldn't affect as many users but it doesn't let you switch paths even via scripts making it severe. So it was indeed the right call to revert this.

I checked that this fixes the bug and did not notice any other problems besides the one we already know about.

PR #2207, opened 2025-07-24
NEW: MouseEvents for InputSystem - Sending data from package & Samples and Tests

ritamerkl:
### Description

This PR introduces sending pointer data (e.g. mouse, pen, touch...) to native for the OnMouse events driven by the InputSystem. It also adds tests and samples for the OnMouse events working with the InputSystem. The behaviour on native side is introduced with this [PR](https://github.cds.internal.unity3d.com/unity/unity/pull/73198) and both PR's should land together, or one (shortly) after the other.

### Testing status & QA

Only manual testing for the sample.

### Overall Product Risks

- Complexity: 1
- Halo Effect: 1

### Comments to reviewers

Please review together with https://github.cds.internal.unity3d.com/unity/unity/pull/73198.
This PR needs to adjust to the exact version in which the upper PR lands.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


jfreire-unity (changes requested):
Nice work @ritamerkl ! I like the approach and the sample is of reasonable size to test the main functionality. Also really nice to have plenty of tests.

However:
- This functionality needs to be guarded with defines in asmdef, as this will only be  available from 6.3 forward AFAIK. So if I try this PR without the trunk PR, I'll get compilation errors.
- I would like to see at least 1 test with Touch as well, particularly with more than 1 touch.  Mouse and Pen are pretty similar as they are a single pointer.

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,354 @@
...
+        var mouse = InputSystem.AddDevice<Mouse>();
+
+        var gameObject = GameObject.CreatePrimitive(PrimitiveType.Cube);
+        gameObject.AddComponent<OnMouseEventsTest>();
+        gameObject.transform.position = Vector3.zero;
+        var camera = new GameObject("MainCamera").AddComponent<Camera>();
+        camera.transform.position = new Vector3(0, 0, -2f);
+        camera.tag = "MainCamera";
+        var vec = Camera.main.WorldToScreenPoint(gameObject.transform.position);
+        SetMouse(mouse, new Vector2(vec.x, vec.y), 1f);
```
jfreire-unity: I find this boilerplate code being repeated amongst tests. Could we create a function for it or combine the tests together?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
 
-
+### Added
+- Added OnMouse events for the InputSystem, including samples and tests.
```
jfreire-unity: Maybe mention something about the requirement from the Unity version.

Assets/Samples/MouseEvents/OnMouseEventsTest.cs
```diff
@@ -0,0 +1,46 @@
...
+using Touch = UnityEngine.InputSystem.EnhancedTouch.Touch;
+
+public class OnMouseEventsTest : MonoBehaviour
+{
+    private Vector3 screenPoint;
+    private Vector3 offset;
+
+    private void Start()
+    {
+        EnhancedTouchSupport.Enable();
```
jfreire-unity: Do we need this? This is mostly if we use polling Touch APIs (see [reference](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.14/api/UnityEngine.InputSystem.EnhancedTouch.EnhancedTouchSupport.html)) and I don't see any calls for it in this script. Unless they are being called through `Pointer.current` ?

jfreire-unity:
> Nice work @ritamerkl ! I like the approach and the sample is of reasonable size to test the main functionality. Also really nice to have plenty of tests.
> 
> However:
> 
> * This functionality needs to be guarded with defines in asmdef, as this will only be  available from 6.3 forward AFAIK. So if I try this PR without the trunk PR, I'll get compilation errors.
> * I would like to see at least 1 test with Touch as well, particularly with more than 1 touch.  Mouse and Pen are pretty similar as they are a single pointer.

Also there's some CI failures that will need to be addressed. Not sure if it's only because of the lack of defines or if there's any impact on pointer devices functionality.

Packages/com.unity.inputsystem/CHANGELOG.md
ritamerkl: good point [bab6e35](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/bab6e35e6ef75aa4b08f71e52624a5371d4b5255)

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: Good idea, I did that here: [b0201e1](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/b0201e1c271628a0ef881c1dd162ea75ed3bf0c0)

ritamerkl:
WIP: working on adding touch samples (multi-touch) & parallel pointers (eg. touch&pen).

Assets/Samples/MouseEvents/OnMouseEventsTest.cs
ritamerkl: It was a leftover from previous work, removed it now: [c3e5dc3](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/c3e5dc32546590f18657b6c95901113070acd025)

Assets/Samples/MouseEvents/OnMouseEventsTest.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        screenPoint = Camera.main.WorldToScreenPoint(gameObject.transform.position);
+        float x;
+        float y;
+        if (Pointer.current != null)
+        {
+            x = Pointer.current.position.x.value;
+            y = Pointer.current.position.y.value;
+        }
+        else  // Fallback to InputManager if InputSystem is not available
```
LeoUnity: Do we need to support InputManager on this tests? I would assume they are covered by their original tests in trunk...

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,457 @@
...
+
+    #endregion
+
+    #region Pen
+
+    [UnityTest]
+    [Category("MouseEvents")]
+    public IEnumerator PenEvents_CanReceiveOnMouseDown()
+    {
+        var pen = InputSystem.AddDevice<Pen>();
```
LeoUnity: It seems like we are duplicating the exact same tests for both mouse and pen, I'm wondering if we can parameterize that and just have a single test...

untitled folder/Il2CppOutputProject/Source/il2cppOutput/il2cpp-compile.traceevents
```diff
@@ -0,0 +1,4215 @@
+{ "pid": 11157, "tid": -1, "ph": "M", "name": "process_name", "args": { "name": "il2cpp-compile" } },
```
LeoUnity: I'm guessing the untitled folder doesn't need to be included?

LeoUnity (approved):
This is looking great!
I'm happy to see us improving our feature parity! Also thanks for investing the time on the tests!
I've left a few comments, but I'll leave it up to you to decide which ones to address

Assets/Samples/MouseEvents/OnMouseEventsTest.cs
ekcoh: I see no reason why this would use InputManager similar to Leo, I would additionally recommend to rename this script to `OnMouseEvents`, or `MouseEvents` or similar. The `Test` suffix got me thinking I was going to see a NUnit test file or similar

ekcoh (changes requested):
I see 3 files under "untitled folder", the folder including those files should be removed from this PR right?

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ekcoh: Agree to previous comment, this isn't necessarily the case any longer either right? "That we have to queue into NativeInputRuntime"?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3646,6 +3647,9 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             ////       mess in the event buffer
             ////       same goes for events that someone may queue from a change monitor callback
             InvokeAfterUpdateCallback(updateType);
+            //send pointer data to backend for OnMouseEvents
+            if (Pointer.current != null && gameIsPlaying)
```
ekcoh: @ritamerkl Have you checked if Input Manager supports OnMouseXXX events from a script that is marked with [ExecuteInEditorMode]? That could be relevant here...

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3646,6 +3647,9 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+                NativeInputSystem.SetMouseEventsData(Pointer.current.press.isPressed, Pointer.current.press.wasPressedThisFrame, Pointer.current.position.x.value, Pointer.current.position.y.value);
```
ekcoh: Shouldn't this be routed via INativeInputRuntime interface and its NativeInputRuntime and TestInputRuntime implementations?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ritamerkl: I don't think so, since the InputManager MouseEvents were only sent from the InputModuleRegistration player loop callback when ::IsWorldPlayingThisFrame() was true, which basically is only true if in playmode, not editmode. Unless I miss the way [ExecuteInEditorMode] is working now.

untitled folder/Il2CppOutputProject/Source/il2cppOutput/il2cpp-compile.traceevents
ritamerkl: Yes, accidentaly pushed some remains of the local iPhone build. Removed here: [remove build](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/e4e88bbaec751e74c7f0af71be13283c39a1fc11)

Assets/Samples/MouseEvents/OnMouseEventsTest.cs
ritamerkl: Agree to both- applied suggestions here: [6429e90](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/6429e900890f36c85afd54e26163222c22c1dc9a)

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: Done: [937d7b6](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/937d7b6e2dffb077719b57523f4d8cc0d26df086)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ritamerkl: NativeInputRuntime is the name of the Bindings class in the module, I fail to understand what exactly you are referring to. 
I placed it here to have all updated data after the inputsystem Update, this happens for test as well.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: OK, then I believe we might be safe, would to good to make sure we haven't overlooked something like that though.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: I referred to the fact that this is calling NativeInputSystem directly. If you look at NativeInputRuntime.cs you will see it implements INativeInputRuntime as an abstraction of the module to allow stubbing it away in e.g. tests. Hence InputTestRuntime also exists stubbing these calls. It's basically a mocking strategy that has historically been used. If this is possible to auto-test without mocking, it should right? Then I see no reason why this would be just fine, but just wanted to let you know in case you had not noticed since it seems SetMouseEventsData is part of Input module based on where its placed/mapped?

ekcoh: And NativeInputRuntime is not the bindings class, it's just an abstraction within the package. The bindings class (P/Invoke) is called NativeInputSystem which is the one you are calling here

Assets/Samples/MouseEvents/OnMouseEventSample.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.EnhancedTouch;
+using Touch = UnityEngine.InputSystem.EnhancedTouch.Touch;
+
+public class OnMouseEventSample : MonoBehaviour
```
ekcoh: If the intent here is to serve as a sample (and maybe also our own way to manually test or debug this), I would suggest that we have all the supported implicit script mouse event callback functions present in this file.

ekcoh: E.g. missing OnMouseOver, OnMouseEnter, OnMouseExit, OnMouseUp, OnMouseUpAsButton

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,199 @@
+#if UNITY_6000_4_OR_NEWER
```
ekcoh: Great you added automated tests for all the events! This way we protect ourselves from regressions in this area. Similarly I think there is a lot to gain from having them covered in the sample as well.

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,199 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.TestTools;
+using Is = UnityEngine.TestTools.Constraints.Is;
+
```
ekcoh, line 10: Nitpick: Redundant empty row

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,199 @@
...
+        gameObject.transform.position = Vector3.zero;
+        var camera = new GameObject("MainCamera").AddComponent<Camera>();
+        camera.transform.position = new Vector3(0, 0, -2f);
+        camera.tag = "MainCamera";
+        return gameObject;
+    }
+
+    private static HashSet<Type> _testDevices = new HashSet<Type> { typeof(Mouse), typeof(Pen), typeof(Touchscreen) };
+    // OnMouseOver/Exit and Hover events are not supported for touch
+    private static HashSet<Type> _testDevicesNoTouch = new HashSet<Type> { typeof(Mouse), typeof(Pen) };
```
ekcoh: Isn't a HashSet overkill here (dynamic allocation etc) when function with a switch statement with 2 cases would be enough?

ekcoh: I can't see (or am I missing something) that we use lookup in the tests? If not, then I would recommend a plain static readonly array for these. (Less allocation, leaner, simpler) and also likely much faster. Also, when using HashSet, the iteration order is "random"

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,199 @@
...
+
+    [UnityTest]
+    [Category("MouseEvents")]
+    public IEnumerator MouseEvents_CanReceiveOnMouseDown([ValueSource(nameof(_testDevices))] Type pointerType)
+    {
+        var pointer = (Pointer)InputSystem.s_Manager.AddDevice(pointerType);
```
ekcoh: Nitpick: It looks like this 
```cs
var pointer = (Pointer)InputSystem.s_Manager.AddDevice(pointerType);
InputSystem.AddDevice(pointer);
```
Is repeated for all the tests. Can't we just embed it into SetUpScene() and have it as an out parameter from that function to reduce redundant code?

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,199 @@
...
+        InputSystem.AddDevice(pointer);
+
+        var gameObject = SetUpScene();
+        gameObject.AddComponent<OnMouseEventsTest>();
+        var vec = Camera.main.WorldToScreenPoint(gameObject.transform.position);
+
+        Set(pointer, new Vector2(vec.x, vec.y), true, false);
```
ekcoh, line 36: Throughout all of these tests I see we have the same pattern, e.g. Set(...); yield return null; (repeat)

I would be reassuring we also verified this works correctly without the yields, e.g. when events are not in separate frames. Maybe it's possible to parameterise the test with another variable e.g. `bool yieldBetweenSet` so that we could test both scenario by having the yields not happening when that is false. Maybe in that case we only need to yield at the end to trigger script callbacks?

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -0,0 +1,199 @@
...
+        gameObject.AddComponent<OnMousEventTestTwo>();
+        var vec = Camera.main.WorldToScreenPoint(gameObject.transform.position);
+        Set(pointer, new Vector2(vec.x, vec.y), false, false);
+        yield return null;
+        Assert.That(gameObject.GetComponent<Renderer>().material.color, Is.EqualTo(Color.blue), "No MouseOver event received.");
+    }
+
+    void Set(Pointer pointer, Vector2 point, bool pressed, bool released)
+    {
+        if (pointer is Touchscreen touchscreen)
```
ekcoh, line 132: Is Pen/Stylus falling under the same behaviour as mouse? Or is it excluded from this functionality somehow?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ however, it has to be formatted properly to pass verification tests.
 
 ### Added
 - Exposed MediaPlayPause, MediaRewind, MediaForward keys on Keyboard.
+- Added OnMouse events for the InputSystem from editor version 6.3, including samples and tests.
```
ekcoh: I would suggest saying "Added support for (`MonoBehavior` OnMouse-events) [https://docs.unity3d.com/ScriptReference/MonoBehaviour.html] when running the Input System on Unity 6.3 or newer."

Then I would make a separate entry for the sample linking to it. It makes more sense to say Unity 6.3 IMO since editor might indicate it would only be supported in editor but this is for players as well right?

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ekcoh: Based on how this have been done in InputManager.cs I wonder if my comment above doesn't make sense. It seems we only call the script once per frame which I wonder if its valid. E.g. that way there wouldn't be any events for e.g. two events in the same frame. Is this really how it works with InputManager? This is the main thing to sort out before I approve. Maybe its useful to test InputManager vs Input System with forced 1 FPS frame rate to sort this out?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3703,6 +3704,11 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             ////       mess in the event buffer
             ////       same goes for events that someone may queue from a change monitor callback
             InvokeAfterUpdateCallback(updateType);
+            //send pointer data to backend for OnMouseEvents
+#if UNITY_6000_4_OR_NEWER
+            if (Pointer.current != null && gameIsPlaying)
+                NativeInputSystem.SetMouseEventsData(Pointer.current.press.isPressed, Pointer.current.press.wasPressedThisFrame, Pointer.current.position.x.value, Pointer.current.position.y.value);
```
ekcoh: Thanks for putting this together @ritamerkl, I added a few suggestions and my only main concern is still whether this will behave similar to Input Manager when we only call into the SetMouseEventData(...) function once every frame - this looks incorrect to me. I would have expected that this would not be based on a polling-based evaluation like this here but rather sit inside the event processing loop above or from within the device state updates after event compression. Does this work for e.g. 1 FPS?

Maybe I am overlooking something.... but was Input Manager only calling this once every frame?

ekcoh:
I see CI is failing with "Packages\com.unity.inputsystem\InputSystem\InputManager.cs(3710,35): error CS0117: 'NativeInputSystem' does not contain a definition for 'SetMouseEventsData'" but this would be expected until the feature is available in Unity.

ritamerkl:
> I see CI is failing with "Packages\com.unity.inputsystem\InputSystem\InputManager.cs(3710,35): error CS0117: 'NativeInputSystem' does not contain a definition for 'SetMouseEventsData'" but this would be expected until the feature is available in Unity.

It is weird since this is wrapped in defines for unity 600.4 or newer, but still fails on 600.3 🤔

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: I was adding this same comment. We should call this as `m_Runtime.SetMouseEventsData(...);`. This avoids making other classes (like this one) dependent on the bindings as well. This way we only have `NativeInputRuntime` class as a single dependent of the Module bindings (`NativeInputSystem`) and make it consistent across the codebase as is.

And by making it part of `IInputRuntime` interface we leverage the mocking strategy @ekcoh mentioned as well. So I think this should be improved.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3703,6 +3704,11 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             ////       mess in the event buffer
             ////       same goes for events that someone may queue from a change monitor callback
             InvokeAfterUpdateCallback(updateType);
+            //send pointer data to backend for OnMouseEvents
+#if UNITY_6000_4_OR_NEWER
+            if (Pointer.current != null && gameIsPlaying)
```
jfreire-unity: This means we will be calling this every frame, regardless of having received an event from a Pointer device or not, right?

ritamerkl: right

ekcoh: We need to look at how its implemented under the hood, there might even be timeout logic in there which would require it to be once a frame.

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -36,377 +21,134 @@ internal GameObject SetUpScene()
...
     }
 
     [UnityTest]
     [Category("MouseEvents")]
-    public IEnumerator PenEvents_CanReceiveOnMouseUp()
+    public IEnumerator MouseEvents_CanReceiveOnMouseUp([ValueSource(nameof(_testDevices))] Type pointerType)
     {
-        var pen = InputSystem.AddDevice<Pen>();
+        var pointer = (Pointer)InputSystem.s_Manager.AddDevice(pointerType);
+        InputSystem.AddDevice(pointer);
```
jfreire-unity: I think it's great we have tests with the Pointer abstraction. But could we have also some real device tests, for Touch, Pen and Mouse devices to make sure the events are generated for these devices?

You can take advantage of `TestCase` attribute pattern already used in other tests in our codebase. Something like:

```csharp
    [TestCase("Pointer")]
    [TestCase("Pen")]
    [TestCase("Touchscreen")]
    [TestCase("Mouse")]
    MouseEvents_CanReceiveOnMouseUp(string deviceLayout)
    {
            var device = InputSystem.AddDevice(layoutName);
            ....
     }
   ```

This removes the need for private static devices such as `HashSet _testDevices`

File: Packages/com.unity.inputsystem/InputSystem/InputManager.cs [hunk 20]
```diff
@@ -3703,6 +3704,11 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             ////       mess in the event buffer
             ////       same goes for events that someone may queue from a change monitor callback
             InvokeAfterUpdateCallback(updateType);
+            //send pointer data to backend for OnMouseEvents
+#if UNITY_6000_4_OR_NEWER
```
ekcoh: I would advise against this. Instead, create an asmdef symbol that defines the feature/capability, e.g. UNITY_INPUTSYSTEM_HAS_MOUSE_SCRIPT_EVENTS_SUPPORTED, then let that symbol be conditionally compiled based on the specific version that contains the native changes, e.g. 6000.0.4.a7, that is more specific and also removes versions from the source code which only cares about a certain API being a certain way instead. And it centralises the version number mess. At least this is what we agreed upon in https://github.com/Unity-Technologies/InputSystem/pull/2176, and the same technique was used in https://github.com/Unity-Technologies/InputSystem/pull/2200 to handle the API difference in module API and behaviour .

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: > there might even be timeout logic in there which would require it to be once a frame.

I'm not sure I understand what you mean with this. Could you specify it a bit?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ritamerkl: Alright, so I understand what you are after. I confused the names :D 
I was considering a change until I noticed: 
1. including SetMouseEventsData into the IInputRuntime would mean we would need a implementation for InputTestRuntime as well, since there is no way to mock setting the data, but we really need to send the data to the module to achieve reliable test results, this would mean we have to include NativeInputSystem there (so there are no savings on bindings).
2. there is no way to mock sending the data for mouse events, so it does not really make sense to bloat the code with two extra methods in NativeInputRuntime/InputTestRuntime.

I would leave this as is, I guess that is fine for this kind. feel free to object if I missed something.

Assets/Samples/MouseEvents/OnMouseEventSample.cs
ritamerkl: Right, added here: [0bef5d2](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/0bef5d265bcfbcf5f0da3c7b179f43176a39bf6f)

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: Done: [0bef5d2](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/0bef5d265bcfbcf5f0da3c7b179f43176a39bf6f)

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: done: [0a65443](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/0a654430b83345cbe3f6cd7b79f96646a4a9a742)

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: done: [0a65443](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/0a654430b83345cbe3f6cd7b79f96646a4a9a742)

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: done: [0a65443](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/0a654430b83345cbe3f6cd7b79f96646a4a9a742)

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: No, only touchscreen needs an extra handling, mouse and pen are handled below. I will add a comment to clarify.

ritamerkl: [e0f10fb](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/e0f10fbed7e7a8cba8f1baf64b5b689aaf6a4522)

Packages/com.unity.inputsystem/CHANGELOG.md
ritamerkl: good idea, especially the link. Changed here: [a9cff76](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/a9cff7642b9fdba9ed4706b6bb844b0890b6772a)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ritamerkl: I think we talked about this before, but to clarify: before this was happening on a PlayerLoopCallback, which was set on module registration. The player loop happens once a frame, the timing of this is "PreUpdate". 
Unless I am terribly mistaking something here, I think that is the same logic, except the Input update is set to FixedUpdate or custom (and changed to any other time).

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: I think we did sort that out in our call - 
This logic is like the old behaviour of the InputManager OnMouse events, where the update happened once per frame (PlayerLoopCallback), too. See comment below.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ritamerkl: There is no timeout logic for the MouseEvents.cs if you mean that

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ritamerkl: actually really easier, did that: [c870b7e](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/c870b7e2a62b444dfdff0704bb405d9579e59428)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ritamerkl: Right, very good point. I did that here: [7b13b3e](https://github.com/Unity-Technologies/InputSystem/pull/2207/commits/7b13b3e9ea61c8dd88a24425c5cbe7d4aaee3e9a)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3705,7 +3705,7 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             ////       same goes for events that someone may queue from a change monitor callback
             InvokeAfterUpdateCallback(updateType);
             //send pointer data to backend for OnMouseEvents
-#if UNITY_6000_4_OR_NEWER
+#if UNITY_INPUTSYSTEM_HAS_MOUSE_SCRIPT_EVENTS_SUPPORTED
```
jfreire-unity: 🎉

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
jfreire-unity: Nice!

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: Hm, good point. Yeah, not sure how we could mock this in InputTestRuntime then. And from what you mentioned, it doesn't seem to make sense to have it in `InputTestRuntime` as well, if we really need to call the module.

I don't have a better answer right now so maybe this is fine as is. 

It just feels odd to be breaking the class contract that in `NativeInputRuntime` should be the only file referencing `NativeInputSystem`. But maybe that's ok in this case?

Anyway, this is not a blocker at all. I'll leave it approved.

jfreire-unity (approved):

Pauliusd01:
Without your trunk PR I get:

Packages\com.unity.inputsystem\InputSystem\InputManager.cs(3711,35): error CS0117: 'NativeInputSystem' does not contain a definition for 'DoSendMouseEvents'

This needs a version check? Also, should this really be a shipped sample? It seems more like a simple test scene (not to knock down its value, it is just maybe too niche?)

Pauliusd01 (changes requested):
Updating status

ekcoh (approved):
Changes looks good now. I marked a few comments I was involved in resolved.

ritamerkl:
> Without your trunk PR I get:
> 
> Packages\com.unity.inputsystem\InputSystem\InputManager.cs(3711,35): error CS0117: 'NativeInputSystem' does not contain a definition for 'DoSendMouseEvents'
> This needs a version check? 

Yes, it needs the exact version where the trunk [PR](https://github.cds.internal.unity3d.com/unity/unity/pull/73198) lands in. But that can only happen after it is merged (in order to have the guaranteed version).

>Also, should this really be a shipped sample? It seems more like a simple test scene (not to knock down its value, it is just maybe too niche?)

Good point, we can also use it as a test scene. It is probably widely known how to use OnMouse events already anyway from the InputManager and does not need introduction at this point.

Pauliusd01:
> > Without your trunk PR I get:
> > Packages\com.unity.inputsystem\InputSystem\InputManager.cs(3711,35): error CS0117: 'NativeInputSystem' does not contain a definition for 'DoSendMouseEvents'
> > This needs a version check?
> 
> Yes, it needs the exact version where the trunk [PR](https://github.cds.internal.unity3d.com/unity/unity/pull/73198) lands in. But that can only happen after it is merged (in order to have the guaranteed version).

I don't see why it needs that on older editors though? If you merge this as is it will keep erroring out everywhere that isn't trunk?

jfreire-unity:
> Without your trunk PR I get:
> 
> Packages\com.unity.inputsystem\InputSystem\InputManager.cs(3711,35): error CS0117: 'NativeInputSystem' does not contain a definition for 'DoSendMouseEvents'
> 
> This needs a version check? Also, should this really be a shipped sample? It seems more like a simple test scene (not to knock down its value, it is just maybe too niche?)

@Pauliusd01 I just tested with 2021.3 LTS and got no error. Maybe this issue is fixed?

Also CI only fails for 6.4 due to this. For them to pass, the PR in trunk needs to land first as Rita said.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -116,7 +116,12 @@
             "name": "Unity",
             "expression": "6000.3.0a6",
             "define": "UNITY_INPUT_SYSTEM_PLATFORM_POLLING_FREQUENCY"
-        }
+        },
+	{
+	    "name": "Unity",
+            "expression": "6000.4.0a1",
+            "define": "UNITY_INPUTSYSTEM_HAS_MOUSE_SCRIPT_EVENTS_SUPPORTED"
```
jfreire-unity: As suggested by @K-Tone 
```suggestion
            "define": "UNITY_INPUTSYSTEM_SUPPORTS_MOUSE_SCRIPT_EVENTS"
```

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3703,6 +3704,12 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
             ////       mess in the event buffer
             ////       same goes for events that someone may queue from a change monitor callback
             InvokeAfterUpdateCallback(updateType);
+            //send pointer data to backend for OnMouseEvents
+#if UNITY_INPUTSYSTEM_HAS_MOUSE_SCRIPT_EVENTS_SUPPORTED
```
jfreire-unity: As suggested by @K-Tone 
```suggestion
#if UNITY_INPUTSYSTEM_SUPPORTS_MOUSE_SCRIPT_EVENTS
```

jfreire-unity (commented):
I just added a renaming suggestion for the define based on Slack comments.

Pauliusd01 (approved):
Sample itself is fine, still looking into the trunk work. Just make sure to edit the version define once trunk merges and possibly move it out of the samples folder so it doesn't get pushed as a public sample. But that last one is just my personal opinion

ericksson:
Cancelling all jobs running on this branch due to Android jobs being broken and stuck - see https://unity.slack.com/archives/C94RMJJ5T/p1760644095487229

ekcoh:

PR #2208, opened 2025-07-29
FIX: Correct north, east, south, west bindings for Switch Pro Controller to match the physical layout of the device

vallerieknight-unity:
### Description

- Adds a new device for iOS, such that we can correctly parse the state of the A button as buttonEast (and so on)
- Corrects Desktop HID device's `aButton`, `bButton`, `zButton`, `yButton` binding to be correct for the Switch Pro Controller

### Testing status & QA

- Tested locally on iOS 18.5 (for iOS, connected via Bluetooth) and macOS 15.5 (for Desktop, connected Bluetooth + USB)
- Added new tests to check for Nintendo state being parsed correctly

### Overall Product Risks

**iOS** : This could break som projects if the project already has mitigations for this behaviour, but, the behaviour _was_ incorrect previously, and customers in this position can disable the new device.

**Standalone**: Less likely to break customer projects, but, could break UI / tutorialisation if a user is assuming `.aButton` is always the same as button south. Happy to revert the standalone change, but if you are a Unity employee, please reach out on Slack as I can explain this decision making process further.

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers

See above notes on some potential upgrade pain - would be good to have some feedback / thoughts from other folks on both of these risks.

### Checklist

Before review:

- [ X ] Changelog entry added.
- [ X  ] Tests added/changed, if applicable.
    - Performance tests.
    - Integration tests.
- [ X ] Docs for new/changed API's.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


vallerieknight-unity:
iOS CI issues appear to be instabilities also occurring on head of `develop`; will trigger reruns periodically. Linux issues appear to be `pvp-sign`.

Assets/Tests/InputSystem/Plugins/iOSTests.cs
```diff
@@ -56,13 +56,64 @@ public void Devices_SupportsiOSGamePad(string product, Type deviceType, Type par
...
         AssertButtonPress(gamepad, new iOSGameControllerState().WithButton(iOSButton.B), gamepad.buttonEast);
+        AssertButtonPress(gamepad, new iOSGameControllerState().WithButton(iOSButton.B), gamepad.bButton);
         AssertButtonPress(gamepad, new iOSGameControllerState().WithButton(iOSButton.LeftShoulder), gamepad.leftShoulder);
         AssertButtonPress(gamepad, new iOSGameControllerState().WithButton(iOSButton.RightShoulder), gamepad.rightShoulder);
     }
 
+    [Test]
+    [Category("Devices")]
+    // this is a new test, as we need to assert the Nintendo layout (e.g. buttonSouth == B button)
+    public void Devices_SupportsProControlleriOS()
```
jfreire-unity:
```suggestion
    public void Devices_SupportsSwitchProControlleriOS()
```

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
```diff
@@ -155,6 +155,34 @@ public class SwitchProControllerHID : Gamepad, IInputStateCallbackReceiver, IEve
...
+        /// </summary>
+        public new ButtonControl yButton => buttonWest;
+
+        /// <summary>
+        /// A Button for a Nintendo Switch Pro Controller.
+        /// If querying via script, ensure you cast the device to SwitchProControllerHID, rather than using the Gamepad class.
+        /// The gamepad class will return the state of buttonWest, whereas this class returns the state of buttonNorth
+        /// </summary>
+        public new ButtonControl xButton => buttonNorth;
+
```
jfreire-unity: This is a change impacting all platforms, right? The Changelog gives the idea we only changed this for iOS,.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/IOSGameController.cs
```diff
@@ -134,5 +195,41 @@ public class DualShock4GampadiOS : DualShockGamepad
     public class DualSenseGampadiOS : DualShockGamepad
     {
     }
+
+    /// <summary>
+    /// A Switch Pro Controller connected to an iOS device.
+    /// If you use InputSystem.GetDevice, you must query for this class rather than Gamepad in order for aButton, bButton, yButton and xButton to be correct
+    /// </summary>
+    [InputControlLayout(stateType = typeof(iOSGameControllerStateSwappedFaceButtons), displayName = "iOS Switch Pro Controller Gamepad")]
+    public class SwitchProControlleriOS : Gamepad
```
jfreire-unity: Do we need a new state? I wonder if we could actually create a similar pattern of what we do [here](https://github.com/Unity-Technologies/InputSystem/blob/6617fd65d82664c2dea1d2767129d4ea697b58ea/Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/IOSGameController.cs#L100)

Because the remapping seems to be done below, as we already did in `SwitchProControllerHID.cs`.

 Maybe we would need to create a `SwitchProController` class. Which would be inherited by `SwitchProControllerHID` and `SwitchProControlleriOS`.

Assets/Tests/InputSystem/Plugins/iOSTests.cs
vallerieknight-unity: Added locally in editor :) (sorry, I was checking up on the PR in my CLI and I didn't realise this was a suggestion rather than just a comment 😄).

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/IOSGameController.cs
vallerieknight-unity: Currently, if you query buttonSouth, you'll get the A button's state, despite the fact buttonSouth is on the east on a Pro Controller. The new state maps this all correctly, such that buttonSouth will actually return the south face button.

The changes below are intended for situations when you want a button by label, regardless of where it is on the controller (e.g. some users will query this, and their UI in a tutorial will show "Press (A)").

If it makes this easier to land currently, I can split the changes to update aButton => buttonEast and friends into a new PR. This PR would then only focus on getting in the new state that rotates around the buttons for the Pro Controller.

Also agreed on the SwitchProController class as a root class; I'm just about to log off for the weekend, so I can make that change on Monday.

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
vallerieknight-unity: Ah, that's a good point; thanks! If it sounds OK to you, I can add a new changelog item to reference the aliases being corrected and state that this is updated for both iOS and standalone.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/IOSGameController.cs
vallerieknight-unity: Just pushed change that adds new `SwitchProController` device :)

Packages/com.unity.inputsystem/InputSystem/Plugins/Switch/SwitchProControllerHID.cs
vallerieknight-unity: Just added new changelog item to help clear up why this was fixed here. If you think this should be a second PR, let me know - I combined it together, just as the actual issue there impacts both iOS & Standalone :)

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/IOSGameController.cs
Pillmatic: @vallerieknight-unity asking a lot here but a refactor like the `SwitchProController` for `DualSense`, `DualShock`, `Xbox` etc. is also much needed. We have so much boiler plate code as we otherwise get errors that specific controller features are not available on platform XY. We would expect that this is handled by the input system with an overall class per controller. Then if feature XY is not available on target XY it just does not perform the action and does not throw compile errors (e.g., PlayStation light bar). The same goes for creating actions and having to manually add targets to it due to different controller classes per platform. See the attached action we have for the select button on controllers to open a journal. You will notice we need several of the same controllers in there to make it work on all platforms the controller is used. Would bring a huge QoL improvement to using the input system.

action_example

vallerieknight-unity:
(Sorry for the noise! Rebased on latest develop to resolve conflicts with changelog after the release of 1.14.2 and to get PVP changes)

vallerieknight-unity:
Converting back to draft to investigate status of Pro Controllers on iOS 26

ekcoh:

vallerieknight-unity:
> Merged develop into this branch to get critical CI fix from #2260. Hope this doesn't caused any inconvenience for you.

NP, thanks for doing the merge & giving a heads up :)

vallerieknight-unity:
> Converting back to draft to investigate status of Pro Controllers on iOS 26

Took a hot second to find time for this 😅 . On latest iOS 26, this still works. I did correct the usages for iOS.

Packages/com.unity.inputsystem/InputSystem/Plugins/iOS/iOSSupport.cs
```diff
@@ -37,6 +37,12 @@ public static void Initialize()
                     .WithDeviceClass("iOSGameController")
                     .WithProduct("DualSense Wireless Controller"));
 
+            InputSystem.RegisterLayout<SwitchProControlleriOS>("SwitchProGamepadiOS",
+                matches: new InputDeviceMatcher()
+                    .WithInterface("iOS")
+                    .WithDeviceClass("iOSGameController")
+                    .WithProduct("Pro Controller"));
```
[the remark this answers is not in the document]
vallerieknight-unity: I agree with the bot in principle, but this is for consistency with everything else ("Controller" being the iOS game controller, "Gamepad" for specific devices).

jfreire-unity:
@vallerieknight-unity I moved the switch controller class into it's own file for consistency, even though its an abstract class. I think there are further changes  that could be made not to have yet a new state for iOS, but since I don't have a controller with me I won't be able to validate them.
Regardles, these changes are an improvement so I'll approve it. Thanks for putting your time into this.

jfreire-unity (approved):

Pauliusd01 (approved):
Checked the original switch pro controller functionality (I only own the switch 1 version, not sure if switch 2 refresh is different enough to warrant concerns) and didn't see any issues in-game, in debugger or with binding setting in editor. 

Note that there's a lot of CI failures, some more valid than others for example this one: 
`File contents must not match NDA regex patterns` from the code analyzer

PR #2221, opened 2025-08-06
FIX: InputControl is not updated from the first time [ISXB-1221]

K-Tone:
### Description

This aims fixing the problem where the InputControl picker won't apply the selected item every second time basically (only when auto-save was on). To be technically precise, it wasn't applying the selected item every time when the input actions window was dirty, which normally happens in two cases:
- you add a new input action, the window is marked dirty, and then if you try to select a new binding it won't be set
- you set a new binding, that marks the window dirty, and then you try selecting the binding another time, that won't work

The root cause for these issues is that our input picker drop-down menu is technically a separate EditorWindow that steals focus from the input actions window as it's open. This was unfortunate  because the input actions window subscribes to focus loss in order to save changes and reload all the state as it does so. Practically, if the input actions window was dirty, you click on the picker, that changes focus and triggers auto-save right there. It re-creates all the backing for the window, but the picker uses SerializedProperties, it doesn't talk to data directly. Because of that, when the new assets are loaded as part of auto-save, it won't pick the right data up.

Hereby we propose to address that by avoiding auto-save at that sensitive moment in time while we're opening the drop-down thing for the input control picker. 

### Testing status & QA

Local testing, QA pending

### Overall Product Risks

- Complexity:  Low
- Halo Effect: Low

### Comments to reviewers

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
@@ -217,7 +217,7 @@ namespace UnityEngine.InputSystem.Editor
         /// <summary>
         /// This property is only set from this class in order to communicate that we're showing the dropdown at the moment
         /// It's employed to skip auto-saving, because that complicates updating the internal SerializedProperties.
-        /// Unfortunately, we can't use IMGUIDropdownVisible from the setings provider because of the early-out logic in there. 
+        /// Unfortunately, we can't use IMGUIDropdownVisible from the setings provider because of the early-out logic in there.
         /// </summary>
         public static bool IsShowingDropdown { get; private set; }
     }
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
@@ -161,9 +161,11 @@ namespace UnityEngine.InputSystem.Editor
 
         private void ShowDropdown(Rect rect, SerializedProperty serializedProperty, Action modifiedCallback)
         {
-            #if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
+#if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
             InputActionsEditorSettingsProvider.SetIMGUIDropdownVisible(true, false);
-            #endif
+#endif
+            IsShowingDropdown = true;
+
             if (m_PickerDropdown == null)
             {
                 m_PickerDropdown = new InputControlPickerDropdown(
@@ -187,6 +189,8 @@ namespace UnityEngine.InputSystem.Editor
             m_PickerDropdown.SetExpectedControlLayout(m_ExpectedControlLayout);
 
             m_PickerDropdown.Show(rect);
+
+            IsShowingDropdown = false;
         }
 
         private void SetExpectedControlLayoutFromAttribute(SerializedProperty property)
@@ -209,6 +213,13 @@ namespace UnityEngine.InputSystem.Editor
 
         private InputControlPickerDropdown m_PickerDropdown;
         private readonly InputControlPickerState m_PickerState;
+
+        /// <summary>
+        /// This property is only set from this class in order to communicate that we're showing the dropdown at the moment
+        /// It's employed to skip auto-saving, because that complicates updating the internal SerializedProperties.
+        /// Unfortunately, we can't use IMGUIDropdownVisible from the setings provider because of the early-out logic in there. 
+        /// </summary>
+        public static bool IsShowingDropdown { get; private set; }
     }
 }
  #endif // UNITY_EDITOR
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
@@ -219,7 +219,7 @@ namespace UnityEngine.InputSystem.Editor
         /// It's employed to skip auto-saving, because that complicates updating the internal SerializedProperties.
         /// Unfortunately, we can't use IMGUIDropdownVisible from the setings provider because of the early-out logic in there.
         /// </summary>
-        public static bool IsShowingDropdown { get; private set; }
+        internal static bool IsShowingDropdown { get; private set; }
     }
 }
  #endif // UNITY_EDITOR
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
@@ -206,12 +206,9 @@ namespace UnityEngine.InputSystem.Editor
         private GUIContent m_PathLabel;
         private string m_ExpectedControlLayout;
         private string[] m_ControlPathsToMatch;
-        private InputControlScheme[] m_ControlSchemes;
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
```diff
@@ -161,9 +161,11 @@ public void OnGUI(Rect rect, GUIContent label = null, SerializedProperty propert
 
         private void ShowDropdown(Rect rect, SerializedProperty serializedProperty, Action modifiedCallback)
         {
-            #if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
+#if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
```
K-Tone, line 164: This was automatic by VSCode

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
```diff
@@ -206,12 +210,16 @@ private void SetExpectedControlLayoutFromAttribute(SerializedProperty property)
...
         private InputControlPickerDropdown m_PickerDropdown;
         private readonly InputControlPickerState m_PickerState;
-        private InputActionRebindingExtensions.RebindingOperation m_RebindingOperation;
+
+        /// <summary>
+        /// This property is only set from this class in order to communicate that we're showing the dropdown at the moment
+        /// It's employed to skip auto-saving, because that complicates updating the internal SerializedProperties.
+        /// Unfortunately, we can't use IMGUIDropdownVisible from the setings provider because of the early-out logic in there.
+        /// </summary>
+        public static bool IsShowingDropdown { get; private set; }
```
K-Tone: Yes, that is a static property. However, it's only set from one place (private set), and it won't leak memory. Also, if we somehow happen to have two drop-down editor windows at the same time (which QA will correct me is barely possible). Even if somehow somehow opens two drop-down thingies at the same time, I believe we're fine here, as the more drop-down windows we have, the less sense it makes to auto-save during those sensitive moments of time. Hope that makes sense.

K-Tone:
I don't think I can make an automated test for this myself, it's rather tricky UI/Editor case involving ImGUI and UIElements among other things.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
```diff
@@ -17,7 +17,6 @@ internal class InputActionsEditorSettingsProvider : SettingsProvider
         [SerializeField] InputActionsEditorState m_State;
         VisualElement m_RootVisualElement;
         private bool m_HasEditFocus;
-        private bool m_IgnoreActionChangedCallback;
```
K-Tone, line 20: This was never used, dead code, I'm removing as a harmless side effect.

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
```diff
@@ -206,12 +210,16 @@ private void SetExpectedControlLayoutFromAttribute(SerializedProperty property)
         private GUIContent m_PathLabel;
         private string m_ExpectedControlLayout;
         private string[] m_ControlPathsToMatch;
-        private InputControlScheme[] m_ControlSchemes;
-        private bool m_NeedToClearProgressBar;
 
         private InputControlPickerDropdown m_PickerDropdown;
         private readonly InputControlPickerState m_PickerState;
-        private InputActionRebindingExtensions.RebindingOperation m_RebindingOperation;
```
K-Tone, line 214: These three private properties were never really used, so I'm just cleaning it up as a harmless side effect on my main change.

Pauliusd01:
I only have access to a mac at the moment but this seemed fine to me. Only thing I noticed so far is that the "*" unsaved changes indicator is still present after reopening the editor if you had an unsaved input actions window open. (Steps are: have auto save on and make some changes -> do not close the window but close the editor itself -> your changes are actually saved but the indicator stays there on your next editor run. Probably not worth changing in case it could cause more trouble but mentioning regardless) Will recheck this again on Windows next week

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
ekcoh: Do this property need to have a public getter? (Noticed the editor for some reason isn't internal)

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
ekcoh: Thanks

ekcoh (approved):
Looks good as a fix for this problem from my perspective. I only have a small question and that is whether the new property really needs to be public (getter) or could be internal?

Packages/com.unity.inputsystem/InputSystem/Editor/ControlPicker/InputControlPathEditor.cs
K-Tone: Valid concern. Would you like me to make the property internal instead of public?

internal static bool IsShowingDropdown { get; private set; }

K-Tone:
> I only have access to a mac at the moment but this seemed fine to me. Only thing I noticed so far is that the "*" unsaved changes indicator is still present after reopening the editor if you had an unsaved input actions window open. (Steps are: have auto save on and make some changes -> do not close the window but close the editor itself -> your changes are actually saved but the indicator stays there on your next editor run. Probably not worth changing in case it could cause more trouble but mentioning regardless) Will recheck this again on Windows next week

This is not a regression from my changes is it? If not, I'd probably keep it a separate case...

K-Tone:
> Looks good as a fix for this problem from my perspective. I only have a small question and that is whether the new property really needs to be public (getter) or could be internal?

Just pushed the visibility change from public to internal, thanks for spotting that out.

Pauliusd01:
> > I only have access to a mac at the moment but this seemed fine to me. Only thing I noticed so far is that the "*" unsaved changes indicator is still present after reopening the editor if you had an unsaved input actions window open. (Steps are: have auto save on and make some changes -> do not close the window but close the editor itself -> your changes are actually saved but the indicator stays there on your next editor run. Probably not worth changing in case it could cause more trouble but mentioning regardless) Will recheck this again on Windows next week
> 
> This is not a regression from my changes is it? If not, I'd probably keep it a separate case...

You are 100% correct, it reproduces on develop as well. Ignore it

Pauliusd01 (approved):
LGTM, seems to be working just as well on Windows. Checked these with auto and manual save modes:
- Adding bindings, using the listen option and picking them directly from the list
- Creating actions, maps, duplicating, copying, cutting, renaming, etc, looking for inconsistencies when auto saving
- Making sure saving indicator acts as expected and expansions are maintained after saving 
- Saving behaviours in play mode, or when closing the editor

Note: It is still somewhat possible to reproduce the deselection issue in the project settings, project wide actions view. But I would consider that a non blocker. 
- Steps are: Reset actions to default -> without deselecting project settings immediately try to edit a binding -> your binding view deselects. This doesn't reproduce every single time for me, seems to relate to how many actions I have so I think we can ignore, this is still a huge improvement.

PR #2226, opened 2025-08-26
CHANGE: Migrate ci to recipe engine

windxu88:
### Description

This PR migrates CI to use Recipe Engine.

In Input CI, tests fall into two categories:` functional tests` and `performance tests`.
For each category, tests run on the following Runtime platforms:
- Editor on all three desktop OS platforms which are `Windows`, `Ubuntu` and `MacOS`
- Standalone on all three desktop OS platforms
- Standalone with il2cpp backend on all three desktop OS platforms
- Mobile platforms including `TvOS`, `iOS` and `Android`. For Android, we cover both `mono` and `il2cpp` backends.

In the migration, I created **four recipes for both `Functional` and `Performance` tests**.
For functional tests, they are:
- `EditorFunctionalTests`
- `StandaloneFunctionalTests`
- `StandaloneIl2CppFunctionalTests`
- `MobileFunctionalTests`

For performance tests, there are four equivalents:
- `EditorPerformanceTests`
- `StandalonePerformanceTests`
- `StandaloneIl2CppPerformanceTests`
- `MobilePerformanceTests`

For editor and standalone tests recipes, the tests will run on `WrenchPackage.DefaultEditorPlatforms` which are Windows, Mac and Ubuntu. That's what we want.
For mobile tests recipe, I added the following platforms in `InputSystemSettings.cs`:
- `MobileBuildPlatforms` 
- `MobileTestPlatforms`
Mobile platform settings are read from `mobile_config.json` file in `.yamato` folder.

Another difference between editor/standalone and mobile tests recipes is:
- Editor/Standalone tests recipes inherit from `BaseRecipe`
- Mobile tests recipes inherit from `MobileBaseRecipe` which inherits `BaseRecipe`. That's because I need to add support for both `mono` and `il2cpp` backends for Android. I did some customization for Android about how to create jobs for it. The details can be seen in `MobileBaseRecipe.cs`.

The major differences between functional tests recipes and performance tests recipes are in UtrCommand.
* Functional tests recipes use `WithCategory("!Performance")` and `--enable-code-coverage`.
* Performance tests recipes use `.WithCategory("Performance")` and `WithPerformanceDataReporting`.
Other than that, functional tests recipes and performance tests recipes are very similar.

**Please note that:** In the old Build & Run jobs on Desktop OS platforms, [upm-ci package test](https://github.com/Unity-Technologies/InputSystem/blob/0eb0e77ce7ffdb04929ce6e2fc3e5a8585c8b144/.yamato/upm-ci.yml#L28) can be replaced by Wrench validation jobs. That's why I didn't add it in the recipes. For example, the old PVS tests have been replaced by `upm-pvp test` in Wrench validation jobs. Wrench validation jobs also run tests inside the package, so there is no need to run package tests in other jobs.

### Testing status & QA

### Overall Product Risks

- Complexity: 
- Halo Effect: 

### Comments to reviewers

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Tools/CI/Recipes/EditorPerformanceTests.cs
```diff
@@ -0,0 +1,52 @@
...
+                .Add(UtrCommand.Run(platform.System, b => b
+                    .WithTestProject($"{ProjectPath}")
+                    .WithEditor(".Editor")
+                    .WithExtraArgs("--suite=Editor --suite=Playmode")
+                    .WithCategory("Performance")
+                    .WithExtraArgs("--clean-library", "--api-profile=NET_4_6")
+                    .WithRerun(1, true)
+                    .WithPerformanceDataReporting(true)
+                    .WithPerformanceProject("InputSystem")
+                    .WithExtraArgs("--report-performance-data --performance-project-id=InputSystem")
```
stefanunity: This line produces "--report-performance-data --performance-project-id=InputSystem" in the yml cmd line a 2nd time, since the two .Withs above it already say the same thing. 
Looks fine for the other performance test recipes.

Tools/CI/Recipes/Triggers.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    private ISet<IJobBuilder> GetTriggers()
+    {
+        HashSet<IJobBuilder> builders =
+        [
+            JobBuilder.Create("All Mobile Tests")
+                .WithDependencies(allMobileFunctionalTests)
+                .WithDependencies(allTvOSFunctionalBuildJobs),
+            
+            JobBuilder.Create("All Functional Tests")
```
stefanunity, line 42: Maybe a trigger could be added for "all functional editor jobs for trunk" to run during development to save some clicks and have a quick CI turnaround before the PR stage? But I'll leave that up to @jfreire-unity to decide. Just a thought.

stefanunity (approved):
Thanks for doing this! While this is now a bunch more files flying around than before, it is going to be easier to keep alive without everyone bolting every new thing onto that one yml file :D
Only two small comments from my side.

Tools/CI/Recipes/EditorPerformanceTests.cs
windxu88: Good catch!

Tools/CI/Recipes/InputBaseRecipe.cs
```diff
@@ -0,0 +1,86 @@
...
+using RecipeEngine.Api.Extensions;
+using RecipeEngine.Api.Platforms;
+using RecipeEngine.Api.Recipes;
+using RecipeEngine.Modules.Wrench.Models;
+using RecipeEngine.Platforms;
+using RecipeEngine.Unity.Abstractions.Packages;
+
+namespace InputSystem.Cookbook.Recipes;
+
+public abstract class InputBaseRecipe: RecipeBase
```
jfreire-unity: Nit pick: We're in the `InputSystem.Cookbook.Recipes` so I would think we don't need `Input` prefix for `inputBaseRecipe`, `InputMobileBaseRecipe`. But this is just my preference and I don't know if without those prefixes, those classes exist elsewhere. You will likely be more into this area so I trust your judgement.

Tools/CI/Recipes/MobileFunctionalTests.cs
```diff
@@ -0,0 +1,146 @@
...
+                .WithEditor(".Editor")
+                .WithSuite(UtrTestSuiteType.Playmode)
+                .WithPlatform(platform.System)
+                .WithCategory("!Performance")
+                .WithExtraArgs("--clean-library")
+                .WithRerun(1, true)
+                .WithBuildOnly()
+                .WithPlayerSavePath("build/players")
+                .WithArtifacts("build/logs")));
+        }
```
jfreire-unity: I see some repeated method calls such as `WithArtifacts` , `WithPlayersSavePath`, etc. I wonder if we can make all of the repetition go away so that it's clearer of what's different bewteeen the jobs.

Tools/CI/Recipes/MobileFunctionalTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            job.WithCommands(UtrCommand.Run(platform.System, b => b
+                .WithTestProject($"{ProjectPath}")
+                .WithEditor(".Editor")
+                .WithSuite(UtrTestSuiteType.Playmode)
+                .WithCategory("!Performance")
+                .WithExtraArgs("--platform=tvOS --clean-library")
+                .WithRerun(1, true)
+                .WithBuildOnly()
+                .WithPlayerSavePath("build/players")
+                .WithArtifacts("build/logs")));
```
jfreire-unity: I don't see a `WithPlatform` here. Is it expected?

Tools/CI/Recipes/MobileFunctionalTests.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        job.WithArtifact(new Artifact("players", "build/players/**/*"), new Artifact("logs", "build/logs/**/*"))
+            .WithInfrastructureInstabilityDetection<WrenchExtensions.CustomScriptInfo>();
+
+        if (platform.System == SystemType.IOS && float.Parse(unityVersion) > 6000.2f)
+            job.WithEnvironmentVariable("UNITY_HANDLEUIINTERRUPTIONS", 1);
```
jfreire-unity, line 51: Why is this only done for 6.2 and forward? Maybe some comments explaining this would help understand its need.

Tools/CI/Recipes/MobileFunctionalTests.cs
jfreire-unity: Also, you can take this comment to all the other places we are creating jobs. I don't know it the builder pattern needs to follow a specific order as well. Still, I think at least adding the methods that are different depending on the condition we check for would be beneficial to all maintainers.

Tools/CI/regenerate.bat
jfreire-unity, line 1: Is this file expected to be empty?

Tools/CI/regenerate.sh
```diff
@@ -1,2 +1,2 @@
 cd $(dirname "$0")/../../
-dotnet run --project Tools\CI\InputSystem.Cookbook.csproj
\ No newline at end of file
+dotnet run --project Tools/CI/InputSystem.Cookbook.csproj
```
jfreire-unity, line 2: When should this be run? Could we actually add a README.md to `Tools/CI` for us to know how it is setup and what to run when we make changes?

Tools/CI/Recipes/InputMobileBaseRecipe.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    public override IEnumerable<IJobBuilder> GetJobs()
+    {
+        List<IJobBuilder> builders = new();
+
+        var package = Settings.InputSystemPackage;
+        var platforms = GetJobPlatforms(package);
+        foreach (var platform in platforms)
+        {
+            var supportedVersions = package.SupportedEditorVersions;
```
jfreire-unity: So how does this work? How do we know for which versions the CI will run? Will it run against trunk? When a Editor version becomes EOL, does it automatically stop running for those versions?
We currently have the jons which run against `trunk` as not a not required to be green so that the PR can land (Setup through Github). Will this still be possible?

jfreire-unity (commented):
Looks like a good improvement, so thanks for this!

I added some comments to avoid code repetition and questions regarding matching the current functionality. 

Also, I think this area would benefit from having a README in case someone in the team needs to do a CI fix. I think a lot of us are not aware of how these CI Recipes so I'll try to book a meeting next week.

I'm not sure how we will "deploy" this. Are we able to test this on demand before landing the PR to make sure we have 1:1 matching with the current jobs that run?

My last comment/nitpick is about the size of the PR. We could have probably done this in smaller chunks so that we wouldn't have so many changes landing. Also would have been easier to review.

Nevertheless, this is great work, and thanks for improving our CI workflow! 🚀 🙏

Tools/CI/Recipes/InputMobileBaseRecipe.cs
windxu88: > So how does this work? How do we know for which versions the CI will run? Will it run against trunk? When a Editor version becomes EOL, does it automatically stop running for those versions?

Recipe Engine, specifically its Wrench module provides this `SupportedEditorVersions` for each package. The versions are from the minimum Editor version supported by the package to Trunk which currently is `6000.4`, **excluding the versions which have reached EOL**.
For Input, `SupportedEditorVersions` are `2021.3, 2022.3, 6000.0, 6000.2, 6000.3 and Trunk`.

When another version reaches EOL such as 6000.2 or Trunk becomes 6000.5, we need to regenerate the CI by running `Tools/CI/regenerate.sh[bat]` or running the `InputSystem.Cookbook` project in IDE, then `SupportedEditorVersions` will retrieve new values and apply them to CI.

windxu88: > We currently have the jons which run against `trunk` as not a not required to be green so that the PR can land (Setup through Github). Will this still be possible?

Yes, it is still possible. Recipe Engine is just a C# framework which generates Yamato jobs and serializes them to yml files. After generating the CI jobs, the rest remains the same as before.

Tools/CI/regenerate.sh
windxu88: We need to run regenerate.sh[bat] when any of following things happen:
- Changes made to InputSystem.Cookbook project
- An unity version reaches EOL
- Trunk version has changed which also means a new version has branched off from Trunk
- Minimum Editor version of Input package has changed
- Wrench version is updated

Tools/CI/regenerate.bat
windxu88: It is not empty if you open that file. It should be github display issue.

Tools/CI/Recipes/MobileFunctionalTests.cs
windxu88: Have added some comments

Tools/CI/Recipes/MobileFunctionalTests.cs
windxu88: "WithPlatform" doesn't work with tvOS. Instead I used `--platform=tvOS` directly.

Tools/CI/Recipes/MobileFunctionalTests.cs
windxu88: I have done some refactoring to reduce the duplication. Please re-review.

Tools/CI/Recipes/InputMobileBaseRecipe.cs
jfreire-unity: Thanks for clarifying!

jfreire-unity (approved):
Thanks for doing this! Approving to not leave this lingering longer. No blockers for me.

PR #2228, opened 2025-08-28
FIX: UISelectable's OnSubmit and OnCancel are not triggered when using InputTestFixture with mock keyboard [ISXB-1561]

K-Tone:
### Description

Here we attempt to address an ancient regression that happened when we moved from using the cached button states obtained via `performed` callbacks to using InputAction.wasReleasedThisFrame. The problem with wasReleasedThisFrame is that it gets updated before we can get control in tests, so there's really no direct way to trigger it from tests. That said, we have a simple caching approach here that  hopefully works well. 

### Testing status & QA

Local testing, 4 new regression tests to cover for the cases where we trigger events from a mock keyboard, both with a frame delay as well as without a frame delay.  QA pass pending.

### Overall Product Risks

- Complexity:  Low
- Halo Effect: Medium - as this is actually affecting both mock tests as well as regular user experience (InputSystemUIInputModule is needed for uGUI stuff to work with this package well, so potentially a lot of users)

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -2287,11 +2291,6 @@ private void OnMoveCallback(InputAction.CallbackContext context)
             m_NavigationState.device = context.control.device;
         }
 
-        private void OnSubmitCancelCallback(InputAction.CallbackContext context)
```
K-Tone, line 2290: Worry not, the device setting code is still present in the two callbacks that I added.

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -2318,6 +2317,26 @@ private void OnTrackedDevicePositionCallback(InputAction.CallbackContext context
 #endif
         }
 
+        private void OnSubmit(InputAction.CallbackContext context)
+        {
+            m_SubmitCancelState.device = context.control.device;
```
K-Tone, line 2322: This line is from OnSubmitCancelCallback that got removed

Packages/com.unity.inputsystem/InputSystem/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -2318,6 +2317,26 @@ private void OnTrackedDevicePositionCallback(InputAction.CallbackContext context
...
+
+            if (!context.ReadValueAsButton())
```
K-Tone, line 2324: So the way this is expected to work is that if we got into the callback, then this button has been interacted with. If it's been interacted with, and it is not pressed, then it's been released. Does this sound correct? Any corner cases here?

Assets/Tests/InputSystem/Plugins/MockKeyboardWithUITests.cs
```diff
@@ -0,0 +1,105 @@
...
+        m_ButtonGo = new GameObject("button", typeof(Button), typeof(CancellationHandler));
+        m_ButtonGo.transform.SetParent(m_CanvasGo.transform, false);
+
+        // Now creating the Input layer
+        m_EventSystemGo = new GameObject("eventSystem", typeof(EventSystem));
+        EventSystem.current.SetSelectedGameObject(m_ButtonGo);
+
+        var inputModule = m_EventSystemGo.AddComponent<InputSystemUIInputModule>();
+
+        m_InputActionAsset = ScriptableObject.CreateInstance<InputActionAsset>();
```
K-Tone, line 45: Hmm now that i'm reading it, it feels like this asset will leak, I should Destroy in in TearDown

Assets/Tests/InputSystem/Plugins/MockKeyboardWithUITests.cs
```diff
@@ -0,0 +1,105 @@
...
+    private Keyboard m_Keyboard;
+    private GameObject m_CanvasGo;
+    private GameObject m_EventSystemGo;
+    private GameObject m_ButtonGo;
+    private InputActionAsset m_InputActionAsset;
+
+    public class CancellationHandler : MonoBehaviour, ICancelHandler
+    {
+        public Action trigger;
+        public void OnCancel(BaseEventData eventData) => trigger();
```
K-Tone, line 24: This stuff is only needed so that I could use a lambda closure with assertion below.

ekcoh: I would suggest that a comment like this motivating its use would be so much better as a onliner e.g.
`// Needed so that we can use a lambda closure within assertion below`

Directly in the code. Then you quickly see why this exists directly in the code, and there is no need to clarify this in this review where the information will get lost.

Assets/Tests/InputSystem/Plugins/MockKeyboardWithUITests.cs
ekcoh: I believe you are right. Or use `ScopedDisposable` I added for this use case a year or so ago. Combine it with a `using` statement and you can apply the regular disposable pattern to Unity objects for RAII-like cleanup.

ekcoh: Then it cleans up even if there is an exception in the test code

ekcoh: In this case I realise you should cleanup in TearDown() though since you create in SetUp and you still get that guarantee from the test framework

Assets/Tests/InputSystem/Plugins/MockKeyboardWithUITests.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        GameObject.Destroy(m_EventSystemGo);
+        GameObject.Destroy(m_CanvasGo);
+
+        InputSystem.RemoveDevice(m_Keyboard);
+
+        base.TearDown();
+    }
+
+    private static bool[] FrameWaitingOptions = { true, false };
```
ekcoh, line 68: Great you added this to get coverage on both scenarios/flavours. Maybe even motivate this with an inline comment, e.g.
`// Test scenarios for separating Press and Release into two separate frames or processing them in the same frame.`
IMO it makes it easier to decipher the tests after that

ekcoh: Another option would be to make this an integer, framesToWait and let the above represent { 1, 0 }, then we could also easily extend it to case 2 as well where there is a "null frame" in between.

ekcoh (commented):
Thanks for looking into this, generally changes look good but I am still confused around why the code before the diff wasn't working as expected to I need to run this to give my final comments/feedback on the proposed fix. I'll get back to you.

ekcoh:
@K-Tone looks like formatter needs to be run on the code

K-Tone:
> Thanks for looking into this, generally changes look good but I am still confused around why the code before the diff wasn't working as expected to I need to run this to give my final comments/feedback on the proposed fix. I'll get back to you.

Yes. The problem as I perceive it is that action.WasReleasedThisFrame actually relies on when an EventSystem.Update happened (which is just a regular MonoBehaviour, so it's really when we update all MonoBehaviours). Because of that it's really challenging to trigger a Release in code such that we do it before EventSystem.Update is called this frame. In the user example:

```
// There was some EventSystem.Update at some point in time...

Release(someButton); // Releasing a button via InputTestFIxture.Release

// Right now, if we say action.WasReleasedThisFrame, it will go true of course, but the InputSystemUIInputModule doesn't know about that, it will only get updated next time EventSystem is updated, which will be next frame actually...

yield return null; // waiting for a frame

// EventSystem.Update will happen before we continue from the yield above, but that's already next frame so action.WasReleasedThisFrame is false. 

```
I believe it's not a problem for normal input that reads actual devices because it's assumably interacted with before scripts update. In these kinds of automations with mock keyboards we're just running off an unfortunate frequency/phase.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,7 +9,6 @@ Due to package verification, the latest version below is the unpublished version
 however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
-- Fixed InputControl picker not updating correctly when the Input Actions Window was dirty. [ISXB-1221](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1221)
```
K-Tone, line 12: I think my brain left the building when I filled that up, it should have totally been in another section. Moving now.

Assets/Tests/InputSystem/Plugins/MockKeyboardWithUITests.cs
K-Tone: Done, pushed

Assets/Tests/InputSystem/Plugins/MockKeyboardWithUITests.cs
K-Tone: Done pushed

Assets/Tests/InputSystem/Plugins/MockKeyboardWithUITests.cs
K-Tone: Right, fair. I have this comment right below, where it gets used: `We'd like to test both options to be safe with rapid actions that happened within a single frame`. Can totally do it here instead if you prefer. As to the integer wait frames - I don't know, can of course do that if you fancy but to me it looked there was no difference whether to wait for one frame or any other >0 amount of frames.

ekcoh: I gess it's fine as is, I cannot say if doing it in 2 non-consecutive frame impacts the logic or not.

ekcoh (approved):
I think this looks good. Great work @K-Tone

Pauliusd01:
Seems like the bug still reproes in the user project, waiting for changes

Pauliusd01 (changes requested):
Just updating status to remove it from my to review list and so I know I'm waiting for changes

K-Tone:
I'm not going to deliver this PR, as explained in the resolution note for the case it's aiming to address.

Issue #35, opened 2018-02-23
When a TouchControl is finished, it's phase becomes "Ended" and stays that way.  Should it eventually revert back to "None"?


jackpr-unity:
I was testing Touchscreen by querying Touchscreen.current.touches.  Per Rene's direction, valid touches have phase marked as something other than "None."  After a touch is first recorded, it will never return to the "None" phase.  Instead, when a touch is done it's phase will change to "Ended" and remain in that state.  Is it designed that way, or should a touch's state eventually return to "None"?

Rene-Damm:
Together with the recent API changes, I think this is fine. `.touches` now returns active touches only. Going to `.allTouchControls` will include all ten controls and these do have the `Ended` behavior. Seems fine to me. `None` means there wasn't ever a touch reported on the control. `Ended` means that was one but it has ended.

PR #2229, opened 2025-08-28
FIX: InputActionMap warnings when the UI map is not setup correctly (ISXB-1560)

Darren-Kelly-Unity:
### Description

Fix the warning being shown when no InputActionMap is setup properly. There was a missing file name in the warning.
I also now only show the warning if there is an action map as we default when there isn't one at all, therefore we don't need the warning.

Ticket
https://jira.unity3d.com/browse/ISXB-1560

### Testing status & QA

I have tested manually I recommend QA does the same. Following the repro steps in the ticket.

### Overall Product Risks

- Complexity: 1
- Halo Effect: 1

### Comments to reviewers

Nothing major to comment.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/InputSystem?pullRequest=2229) <br/>All committers have signed the CLA.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
```diff
@@ -70,7 +70,11 @@ public Context(InputActionAsset asset,
             private string GetAssetReference()
             {
                 var path = AssetDatabase.GetAssetPath(asset);
-                return path ?? asset.name;
+                if (path == string.Empty)
```
ekcoh: `return path ?? asset.name` is shorthand for:

```
if (path == null)
    path = asset.name;
return path;
```

The new code doesn't check for null, but https://docs.unity3d.com/ScriptReference/AssetDatabase.GetAssetPath.html, says it may return null, empty string or a path string. It seems I overlooked this when I implemented this, but the new change is also incorrect (may crash). Hence I suggest we change it to:

`return (path == null || path == string.Empty) ? asset.name : path;`

or the shorthand

`return path.IsNullOrEmpty ? asset.name : path` (I believe there is a static extension for this, do not remeber its exact name)

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
```diff
@@ -70,7 +70,11 @@ public Context(InputActionAsset asset,
             private string GetAssetReference()
             {
                 var path = AssetDatabase.GetAssetPath(asset);
-                return path ?? asset.name;
+                if (string.IsNullOrEmpty(path))
```
ekcoh: Great you did the change I suggested :) the comment got lost due to your update. Nitpick but maybe just:
`return string.IsNullOrEmpty(path) ? asset.name : path;`
As-is is also fine. Seems like I overlooked the empty string return value when I implemented this....

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
```diff
@@ -92,6 +96,11 @@ public void Verify(string actionNameOrId, InputActionType actionType, string exp
 
                     // Check if the map (if any) exists
                     var noMapOrMapExists = true;
+                    if (asset.actionMaps.Count == 0)
```
ekcoh: I do not understand why you are checking this? This code is trying to verify a single action (it's the intended purpose of the method). This method should not verify the asset, only the action within the asset. Additionally, based on the comments in the JIRA case, it seems like the desired outcome is to not generate all warnings when there is no "UI" map (relying on defaults), and not in the general case. If this is the goal with the fix is to disable all action verification when there is no UI map I suggest we remove this and instead add an early return to line #35 in this file. That early return I believe should also check for the presence of an "UI" action map and not check action map count in general. E.g. 

`if (asset.FindActionMap("UI", false) == null) return;`

See https://docs.unity3d.com/Packages/com.unity.inputsystem@1.14/api/UnityEngine.InputSystem.InputActionAsset.html#UnityEngine_InputSystem_InputActionAsset_FindActionMap_System_String_System_Boolean_

ekcoh (changes requested):
Thanks for looking into this. Some comments and suggestions....

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
ekcoh: Also consider adding a unit test (do not remember if tests for this class exist or not that may be extended) to avoid future regression.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 - Fixed an issue where using Pen devices on Android tablets would result in double clicks for UI interactions. [ISXB-1456](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1456)
 - Fixed an issue preventing an embedded platform from being released. It adds back some `#defines` to `XRSupport` and `InputDeviceCharacteristics`.
+- Fixed a warning not showing the asset name correctly when the input action map parameters for runtime GUI were not setup.[ISXB-1560]
```
ekcoh: Great you added this. Seems like you also realized we avoid linking internal issues but link public ones. 
Nitpick: Space after full-stop / period before issue reference.

ekcoh:
I guess the tests checkbox and docs checkbox should not be checked since the PR isn't currently touching any of those areas?

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
Darren-Kelly-Unity: Done! :)

Pauliusd01:
Might be something on my part but develop and pr seemed identical to me. (Green border is the PR and red one is Develop)

image

Packages/com.unity.inputsystem/CHANGELOG.md
Darren-Kelly-Unity: Fixed! :)

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
Darren-Kelly-Unity: Ok got it, I think misunderstood this one, I went by what Paulius said but I didn't understand the first time round from reading the comment. I'll try to check with him to confirm.

Darren-Kelly-Unity:
> I guess the tests checkbox and docs checkbox should not be checked since the PR isn't currently touching any of those areas?

I actually though initially I shouldn't need to add tests so checked it as I thought about it.

But I saw you left a comment to add a test so I can look into that!

Darren-Kelly-Unity:
@Pauliusd01, it wouldn't let me quote reply for some reason.

Edit: Nevermind, it was due to my misunderstanding Hakan pointed out. Now everything should work. I pushed a fix.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
Darren-Kelly-Unity: Yes, you are right! I understood from the ticket when there is nothing mapped in the input file, when it should have been no UI map there.

I made the change you suggested and now it seems to behave as expected fixing Paulius' issue below also.

Pauliusd01 (approved):
LGTM. Checked that the warning is no longer shown when UI map is missing and that it is still shown when the map is partially incomplete. Also, double checked that game UI (UGUI and UIToolkit) still works properly without having a UI map in the project wide actions asset and it does, the warning really was unnecessary in this scenario.

Packages/com.unity.inputsystem/InputSystem/Plugins/InputForUI/InputActionAssetVerifier.cs
Darren-Kelly-Unity: I have now added two unit tests also to test both that warnings are thrown when there are actions missing from the UI action map.

And another so that we do not have any warnings when no UI map exists.

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -559,8 +575,6 @@ public void ActionsWithoutUIMap_ShouldGenerateWarnings()
             LogAssert.Expect(LogType.Warning, new Regex($"^InputAction with path 'UI/RightClick' in asset '{link}' could not be found."));
             LogAssert.Expect(LogType.Warning, new Regex($"^InputAction with path 'UI/ScrollWheel' in asset '{link}' could not be found."));
         }
-        // else: expect suppression of child errors
-        LogAssert.NoUnexpectedReceived();
```
Darren-Kelly-Unity, line 563: Had an issue here with Teardown logging the warnings again causing the logs to be there twice and fail.

If you think I should do this another way let me know.

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -535,19 +536,34 @@ public void DefaultActions_ShouldNotGenerateAnyVerificationWarnings(bool useProj
...
+        LogAssert.NoUnexpectedReceived();
+    }
+    
+    [Test(Description = "Verifies that user-supplied project-wide input actions generates warnings if the UI map is present but actions are missing.")]
+    [Category(kTestCategory)]
+    public void ActionsWithUIMap_MissingActions_ShouldGenerateWarnings()
+    {
+        var asset = ProjectWideActionsAsset.CreateDefaultAssetAtPath(kAssetPath);
+        var uiActionMap = asset.FindActionMap("UI", true);
+        for (int i = uiActionMap.m_Actions.Length - 1; i >= 0; i--)
```
Darren-Kelly-Unity: I tried using Clear here instead but it didn't work / caused exceptions.

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
```diff
@@ -535,19 +536,34 @@ public void DefaultActions_ShouldNotGenerateAnyVerificationWarnings(bool useProj
         LogAssert.NoUnexpectedReceived();
     }
 
-    [Ignore("We currently allow a PWA asset without an UI action map and rely on defaults instead. This allows users that do not want it or use something else to avoid using it.")]
-    [Test(Description = "Verifies that user-supplied project-wide input actions generates warnings if action map is missing.")]
+    [Test(Description = "Verifies that user-supplied project-wide actions do not generate warnings if action map is missing. We use default actions in this case.")]
     [Category(kTestCategory)]
-    public void ActionsWithoutUIMap_ShouldGenerateWarnings()
+    public void ActionsWithoutUIMap_ShouldNotGenerateWarnings()
```
ekcoh, line 540: This test makes sense with your changes given that the new desired behaviour is no warnings.

Assets/Tests/InputSystem/Plugins/InputForUITests.cs
ekcoh: Not sure what you try to test here? Based on the name it sounds like it's a duplicate of my previous test `ActionMapWithNonExistentRequiredAction_ShouldGenerateWarning`? Maybe you are good with any additional tests? (Check test coverage)

ekcoh: Maybe this should just go?

Darren-Kelly-Unity: Ah, in this case. It's similar to your old test but the difference is we leave the UI map in and delete the contents.

This should show warnings as me & Paulius talked about this, the default is not used in a case where the UI map is present but not containing necessary actions.

Which is what I test for here.

ekcoh (approved):
LGTM now. Good work!

ekcoh:
One small thing, you need to rerun formatter on this PR to get CI green.

ritamerkl (approved):
LGTM

jfreire-unity:
@Darren-Kelly-Unity can you make the PR name more descriptive, please? Just so that the commit makes more sense. Thanks

ekcoh:
Suggested way forward:
- Need to agree to CLA
- If this branch is updated with latest develop 6000.1 build targets should go away.

PR #2075, opened 2024-12-04
NEW: Added support for F13 to F24 keys

bmalrat:
### Description

https://jira.unity3d.com/browse/UUM-44328

- Added support of F13-F24 keys for enabled editors(6000.0.38f1, 6000.1.0a10)
- Changed enum value `Key.IMESelected` to obsolete which was not a real key. Please use the ButtonControl `imeSelected`.
- Changed the version to 1.4 to accept api changes.

### Testing status & QA

tested on mac/windows/linux

### Overall Product Risks

_Please rate the potential complexity and halo effect from low to high for the reviewers. Note down potential risks to specific Editor branches if any._

- Complexity: 
- Halo Effect: 

### Comments to reviewers

_Please describe any additional information such as what to focus on, or historical info for the reviewers._

### Checklist

Before review:

- [ X] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - JIRA ticket linked, example ([case %<ID>%](https://issuetracker.unity3d.com/product/unity/issues/guid/<ID>)). If it is a private issue, just add the case ID without a link.
    - Jira port for the next release set as "Resolved".
- [X ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [X ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

- [ ] Create forward/backward port if needed. If you are blocked from creating a forward port now please add a task to ISX-1444.

PR #2230, opened 2025-09-03
FIX: [ISXB-1541] Keyboard.allKeys contain null KeyControl for deprecated key Key.IMESelected which results in exception if using sub-script operator for lookup during enumeration of allKeys.

ekcoh:
### Description

Provides a suggested bug fix for https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1541 which is reported as a regression for 1.14.0 due to https://github.com/Unity-Technologies/InputSystem/pull/2075.

In 1.13.1 subscript operator throws `ArgumentOutOfRangeException` for `Key.IMESelected` and `Key` enumeration type contain `IMESelected` defined as 111.

Identified the following potential solutions:
1) Since `IMESelected` was accidentally added to Key enumeration type when `Input System` was first released and later marked obsolete in https://github.com/Unity-Technologies/InputSystem/pull/2075, make sure we still provide a KeyControl for `IMESelected` but advise to not rely on it due to obsoletion status. Let `IMESelected` be part of `Keyboard.allKeys`.
2) Exclude `IMESelected` from `allKeys` (since it is not a key). This is cleaner since it is inline with 1.13.1 where `IMESelected` also wasn't part of `allKeys`.

This PR contains a fix according to 1). Investigating possibility for 2), hence draft status.

**Summary of changes:**
- Addressed issue by defining an aliased `KeyControl` for `IMESelected` bit and returning this `KeyControl` for any sub-script operator look-up using the deprecated key `Key.IMESelected`. This should be removed when `IMESelected` can be removed from API. This fix should be ok given the following API contracts:
   - `IMESelected` key control part of `Keyboard.keys` which makes sense since it has incorrectly been exposed as a key in `Key` enumeration type from version 1.0.
   - Sub-script operator states "Look up a key control by its key code." It isn't specific for whether KeyControl may be null or not so in theory a fix here could be to return null for Key.IMESelected since it is not a key. However, it also states "This is equivalent to allKeys[(int)key - 1]" which implies that result of this operation should be equivalent to subtracting one from key and looking it up via allKeys. This constraint is now also tested via `Device_Keyboard_SubscriptOperatorShouldBeEquivalentToShiftedAllKeysLookup` to capture this requirement.
- Added a new test class `CoreTests_Device_Keyboard.cs` and added some missing new test cases tied to `Keyboard` to this class. I removed some obsolete test cases from `CoreTests_Devices.cs`. I did not port existing keyboard tests from `CoreTests_Devices.cs` to the new file as part of this PR since this could be done separately and would make the PR more difficult to review. Hopefully the extended functional tests increase test coverage of keyboard functionality since it turned out many key properties were never accessed in tests. I introduced this file since a single enormous test file just makes it difficult to assess what is being tested. Generally all individual devices need their dedicated tests (which is the case for plugin devices but not for built-in it seems).
- Updated `FastKeyboard.cs` (precompiled layout) to match changes to `Keyboard.cs`.

### Testing status & QA

Fix should be covered by new (and existing) automated tests, but it is suggested to test keyboard functionality and IME as part of quality assuring this PR. Fix developed and tested on Unity 6000.0.56f1.

**Open issue:** Check if this really is a regression - investigate behaviour on version before 1.14.0 for using subscript operator with `IMESelected` before transitioning this out of draft mode.

### Overall Product Risks

Medium due to central functionality.

- Complexity: Small
- Halo Effect: Medium

### Comments to reviewers

Nothing special to call out apart from that it should be mentioned there is still no property API for IMESelected which is intentional to avoid accessing it.

### Checklist

Before review:

- [x] Changelog entry added.
    - Functional tests added `Devices_Keyboard_CanGetKeyCodeFromKeyboardKey`, `Devices_Keyboard_AllKeysEnumeratesAllKeyControls`, `Devices_Keyboard_AllKeysShouldContainKeyControlsCorrespondingToAllKeys`, `Devices_Keyboard_SubscriptOperatorCanLookupKeyControlOfCorrespondingKey`, `Devices_Keyboard_SubscriptOperatorThrowsForInvalidOrOutOfRangeKey`.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Tests/InputSystem/CorePerformanceTests.cs
+++ b/Assets/Tests/InputSystem/CorePerformanceTests.cs
@@ -126,7 +126,7 @@ internal class CorePerformanceTests : CoreTestsFixture
             int keyIndex = 0;
             foreach (var key in keyboard.allKeys)
             {
-                if (++keyIndex == (int)KeyEx.IMESelected)  // Skip IMESelected as it's not a real key.
+                if (++keyIndex == (int)KeyEx.IMESelected)  // Skip IMESelected as it's not a real key. TODO This is incorrect, it should not even be there
                     continue;
                 key.ReadValue();
             }
--- a/Assets/Tests/InputSystem/CorePerformanceTests.cs
+++ b/Assets/Tests/InputSystem/CorePerformanceTests.cs
@@ -126,8 +126,6 @@ internal class CorePerformanceTests : CoreTestsFixture
             int keyIndex = 0;
             foreach (var key in keyboard.allKeys)
             {
-                if (++keyIndex == (int)KeyEx.IMESelected)  // Skip IMESelected as it's not a real key. TODO This is incorrect, it should not even be there
-                    continue;
                 key.ReadValue();
             }
         })
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -2851,15 +2851,6 @@ partial class CoreTests
 
     [Test]
     [Category("Devices")]
-    public void Devices_CanGetKeyCodeFromKeyboardKey()
-    {
-        var keyboard = InputSystem.AddDevice<Keyboard>();
-
-        Assert.That(keyboard.aKey.keyCode, Is.EqualTo(Key.A));
-    }
-
-    [Test]
-    [Category("Devices")]
     public void Devices_CanLookUpKeyFromKeyboardUsingKeyCode()
     {
         var keyboard = InputSystem.AddDevice<Keyboard>();
@@ -5893,19 +5884,5 @@ partial class CoreTests
             for (var i = 0; i < TouchscreenState.MaxTouches + 5; ++i)
                 BeginTouch(i, new Vector2(i * 1.0f, i * 2.0f), time: 0);
         }, Throws.Nothing);
-    }
-
-    // Note: Tested inside this class for now since there are no dedicated device test classes for built-in devices,
-    //       only for plugin devices.
-    [Test, Description("https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1541")]
-    [Category("Devices")]
-    public void Devices_KeyboardAllKeys_EnumeratesAllKeyControls()
-    {
-        var keyboard = InputSystem.AddDevice<Keyboard>();
-        int index = 0;
-        foreach (var key in keyboard.allKeys)
-        {
-            Assert.NotNull(key, $"Key at index {index++} was null");
-        }
     }
 }
--- a/Assets/Tests/InputSystem/CoreTests_Devices.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices.cs
@@ -5894,4 +5894,18 @@ partial class CoreTests
                 BeginTouch(i, new Vector2(i * 1.0f, i * 2.0f), time: 0);
         }, Throws.Nothing);
     }
+
+    // Note: Tested inside this class for now since there are no dedicated device test classes for built-in devices,
+    //       only for plugin devices.
+    [Test, Description("https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1541")]
+    [Category("Devices")]
+    public void Devices_KeyboardAllKeys_EnumeratesAllKeyControls()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        int index = 0;
+        foreach (var key in keyboard.allKeys)
+        {
+            Assert.NotNull(key, $"Key at index {index++} was null");
+        }
+    }
 }
--- a/Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
@@ -160,8 +160,6 @@ partial class CoreTests
         var index = 0;
         foreach (var key in keyboard.allKeys)
         {
-            if (index == 109)
-                Debug.Log("TEMP");
             Assert.NotNull(key, $"Key at index {index++} was null");
         }
     }
--- a/Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
@@ -1,0 +1,86 @@
+using System;
+using System.Linq;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.InputSystem;
+
+// Functional testing of built-in device UnityEngine.InputSystem.Keyboard.
+
+partial class CoreTests
+{
+    [Test]
+    [Category("Devices")]
+    public void Devices_Keyboard_CanGetKeyCodeFromKeyboardKey()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+
+        Assert.That(keyboard.aKey.keyCode, Is.EqualTo(Key.A));
+        Assert.That(keyboard.bKey.keyCode, Is.EqualTo(Key.B));
+        Assert.That(keyboard.cKey.keyCode, Is.EqualTo(Key.C));
+        Assert.That(keyboard.dKey.keyCode, Is.EqualTo(Key.D));
+        Assert.That(keyboard.eKey.keyCode, Is.EqualTo(Key.E));
+        Assert.That(keyboard.fKey.keyCode, Is.EqualTo(Key.F));
+        Assert.That(keyboard.gKey.keyCode, Is.EqualTo(Key.G));
+        // TODO Make this complete
+    }
+    
+    [Test, Description("https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1541")]
+    [Category("Devices")]
+    public void Devices_Keyboard_AllKeysEnumeratesAllKeyControls()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        var index = 0;
+        foreach (var key in keyboard.allKeys)
+        {
+            if (index == 109)
+                Debug.Log("TEMP");
+            Assert.NotNull(key, $"Key at index {index++} was null");
+        }
+    }
+    
+    [Test]
+    [Category("Devices")]
+    public void Devices_Keyboard_AllKeysShouldContainKeyControlsCorrespondingToAllKeys()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        var allKeys = keyboard.allKeys;
+        foreach (var key in Enum.GetValues(typeof(Key)).Cast<Key>())
+        {
+            Assert.That(allKeys.Contains(keyboard.spaceKey), Is.True);
+        }
+    }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_Keyboard_SubscriptOperatorCanLookupKeyControlOfCorrespondingKey()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        foreach (var key in Enum.GetValues(typeof(Key)).Cast<Key>())
+        {
+            // Key.None is documented as an invalid key so skip it in this test and instead we verify in test below
+            // that exception is thrown if attempting to lookup key-control using it.
+            if (key == Key.None)
+                continue;
+            if (key == Key.IMESelected)
+                continue;
+            
+            var keyControl = keyboard[key];
+            Assert.That(keyControl, Is.Not.Null);
+            Assert.That(keyControl.keyCode, Is.EqualTo(key));
+        }
+    }
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_Keyboard_SubscriptOperatorThrowsForInvalidOrOutOfRangeKey()
+    {
+        var keyboard = InputSystem.AddDevice<Keyboard>();
+        var minKey = Enum.GetValues(typeof(Key)).Cast<Key>().Min();
+        var invalidMin = (Key)((int)minKey)-1;
+        var maxKey = Enum.GetValues(typeof(Key)).Cast<Key>().Max();
+        var invalidMax = (Key)((int)maxKey)+1;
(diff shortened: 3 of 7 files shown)


Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -2849,15 +2849,6 @@ public void Devices_CanGetNameOfCurrentKeyboardLayout()
         Assert.That(keyboard.keyboardLayout, Is.EqualTo("new"));
     }
 
-    [Test]
-    [Category("Devices")]
-    public void Devices_CanGetKeyCodeFromKeyboardKey()
```
ekcoh, line 2854: This became obsolete since it was replaced bye a more elaborate test `Devices_Keyboard_CanGetKeyCodeFromKeyboardKey` that verifies this for all keys and not only `A`.

ekcoh:
Opening this up for review for now since code generator seem to need to be fixed/changed to cope with nonconsecutive enum. So option (2) I haven't been able to solve yet. Another option would be to optionally just return null from subscript-operator if key is IMESelected.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -2370,7 +2375,9 @@ public KeyControl this[Key key]
             {
                 var index = (int)key - 1;
                 if (index < 0 || index >= m_Keys.Length)
-                    throw new ArgumentOutOfRangeException(nameof(key));
+                {
+                    throw new ArgumentOutOfRangeException($"{nameof(key)}: {key}");
+                }
                 return m_Keys[index];
```
ekcoh, line 2377: I am leaving towards it would be better to return null for index == (int)Key.IMESelected - 1?

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -2370,7 +2375,9 @@ public KeyControl this[Key key]
             {
                 var index = (int)key - 1;
                 if (index < 0 || index >= m_Keys.Length)
-                    throw new ArgumentOutOfRangeException(nameof(key));
+                {
+                    throw new ArgumentOutOfRangeException($"{nameof(key)}: {key}");
```
ekcoh, line 2375: It's very odd Codecov is mentioning these lines since there is now a test explicitly testing key/index outside of bounds. Need to investigate if the test is really running in CI (it did locally I believe)

Assets/Tests/InputSystem/CorePerformanceTests.cs
```diff
@@ -126,7 +126,7 @@ public void Performance_ReadEveryKey()
             int keyIndex = 0;
             foreach (var key in keyboard.allKeys)
             {
-                if (++keyIndex == (int)KeyEx.IMESelected)  // Skip IMESelected as it's not a real key.
+                if (++keyIndex == (int)KeyEx.IMESelected)  // Skip IMESelected as it's not a real key. TODO This is incorrect, it should not even be there
```
ekcoh: I will remove this comment, in this particular solution it unfortunately still needs to be there. Unless we can pull of (2) which currently wrack havoc with precompiled layouts.... :/

Pauliusd01:
Do different platforms have relevance here? Do I need to check Mac/Linux?

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -179,6 +180,10 @@ public unsafe struct KeyboardState : IInputStateTypeInfo
         [InputControl(name = "mediaRewind", displayName = "MediaRewind", layout = "Key", bit = (int)Key.MediaRewind)]
         [InputControl(name = "mediaForward", displayName = "MediaForward", layout = "Key", bit = (int)Key.MediaForward)]
         [InputControl(name = "IMESelected", layout = "Button", bit = (int)KeyEx.RemappedIMESelected, synthetic = true)] // Use the last bit to hold IME selected state.
+        // Disable deprecation warning to not generate warnings in user project about internal use of deprecated key
+#pragma warning disable 0618
```
bmalrat: with RemappedIMESelected should not have a deprecated warning

Assets/Tests/InputSystem/CorePerformanceTests.cs
bmalrat: The if (++keyIndex == (int)KeyEx.IMESelected) might be removed now since it's not null anymore

Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
```diff
@@ -0,0 +1,233 @@
...
+        Assert.That(keyboard.f18Key.keyCode, Is.EqualTo(Key.F18));
+        Assert.That(keyboard.f19Key.keyCode, Is.EqualTo(Key.F19));
+        Assert.That(keyboard.f20Key.keyCode, Is.EqualTo(Key.F20));
+        Assert.That(keyboard.f21Key.keyCode, Is.EqualTo(Key.F21));
+        Assert.That(keyboard.f22Key.keyCode, Is.EqualTo(Key.F22));
+        Assert.That(keyboard.f23Key.keyCode, Is.EqualTo(Key.F23));
+        Assert.That(keyboard.f24Key.keyCode, Is.EqualTo(Key.F24));
+        Assert.That(keyboard.mediaPlayPause.keyCode, Is.EqualTo(Key.MediaPlayPause));
+        Assert.That(keyboard.mediaRewind.keyCode, Is.EqualTo(Key.MediaRewind));
+        Assert.That(keyboard.mediaForward.keyCode, Is.EqualTo(Key.MediaForward));
```
bmalrat, line 152: We may need to also check the IME key until we remove it

Assets/Tests/InputSystem/CorePerformanceTests.cs
jfreire-unity: I would remove this TODO anyway or just write a real action to be done in the future. like "TODO Remove this when...."
To me it just reads like a comment that shouldn't be in the codebase..

Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
```diff
@@ -0,0 +1,233 @@
...
+
+// Functional testing of built-in device UnityEngine.InputSystem.Keyboard.
+
+partial class CoreTests
+{
+    private static readonly Key[] sKeys = Enum.GetValues(typeof(Key)).Cast<Key>().ToArray();
+
+    [Test]
+    [Category("Devices")]
+    public void Devices_Keyboard_CanGetKeyCodeFromKeyboardKey()
```
jfreire-unity, line 15: Nice to have this, particularly in a different file! 
I suggest we create a follow-up task/add a TODO/make a new PR with changes to relocate tests for keyboard inside `CoreTests_Devices` into this file.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -179,6 +180,10 @@ public unsafe struct KeyboardState : IInputStateTypeInfo
...
+        [InputControl(name = "IMESelectedObsoleteKey", layout = "Key", bit = (int)KeyEx.RemappedIMESelected, synthetic = true)]
```
jfreire-unity, line 182: I'm confused by this. Why are we adding yet another control for the same value at `KeyEx.RemappedIMESelected` ? Isn't the one with name "IMESelected" suficient? Or because it's a "Button" it's not considered a "Key", and thus not part of `this.keys`?

Assets/Tests/InputSystem/CoreTests_Devices.cs
jfreire-unity: great!

jfreire-unity (approved):
Added some questions but they shouldn't block the PR since it seems that it fixes the bug. Tests look good!

I definetly wished we could just obsolete and remove this :)

Assets/Tests/InputSystem/CorePerformanceTests.cs
ekcoh: Seems like there was confusion here.... for this solution (1) as @bmalrat is correct the whole if-statement can go so that needs updating.
My previous comment was to remove the TODO which is something I threw in there on first pass over the code to see what needed to be done as "note to self". So just as you say @jfreire-unity it should not be in the code base which is what I suggested 2 days ago.

Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
ekcoh: Cool, yeah these huge source files make no sense to me. Sure thing, that was the idea, I get it filed as part of landing this PR. We should likely also split out all the other areas. And maybe skip the partial class / CoreTests_ prefix which adds zero clarity/value IMO

Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
ekcoh: You are absolutely right, will add. Good catch.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
```diff
@@ -156,6 +156,7 @@ public unsafe struct KeyboardState : IInputStateTypeInfo
         [InputControl(name = "f8", displayName = "F8", layout = "Key", bit = (int)Key.F8)]
         [InputControl(name = "f9", displayName = "F9", layout = "Key", bit = (int)Key.F9)]
         [InputControl(name = "f10", displayName = "F10", layout = "Key", bit = (int)Key.F10)]
+
```
ekcoh: Will eliminate this diff, its a mistake it seems.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
ekcoh: Maybe it's a remainder from one of my variants on this, will check/fix and follow up. Fail to recall now why I added it :)

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
ekcoh: This is undesirable and why I rather have solution (2) mentioned in PR description. However, if this doesn't exist the layout precompiled layout thing will break down, need a KeyControl to reference if it shouldn't be null. However, it is at least not directly accessible in public API. Ideally I rather see we would return null for IMESelected but not have it in allKeys which is solution (2) I haven't been able to pull off yet due to problems (maybe bugs) in the layout precompiled code..... to be continued.

ekcoh:
> Do different platforms have relevance here? Do I need to check Mac/Linux?

Platform shouldn't matter as far as I am aware. There is no platform specific code at play here for the scope of this PR.

Pauliusd01 (approved):
LGTM, checked user project as well as IME testing scene in player/playmode using Japanese, Chinese, Russian, Hebrew, Arabic and Turkish

Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
ekcoh: How would that work? All the above are KeyControl properties on the Keyboard class. There is no such property for IME?

ekcoh: However, it is covered by other tests below when retrieved via allKeys for this potential fix

Assets/Tests/InputSystem/CorePerformanceTests.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2230/commits/b803b1d2e2e5d8df333c953fe839e9258361c804

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2230/commits/2eff10291de9699abbebfeafe9be896628bfea08

ekcoh:
> Added some questions but they shouldn't block the PR since it seems that it fixes the bug. Tests look good!
> 
> I definetly wished we could just obsolete and remove this :)

Addressed in https://github.com/Unity-Technologies/InputSystem/commit/b803b1d2e2e5d8df333c953fe839e9258361c804

Assets/Tests/InputSystem/CorePerformanceTests.cs
ekcoh: Marking this resolved, feel free to reopen if necessary

Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
ekcoh: Marking this revealed, @bmalrat feel free to reopen in case I overlooked something

Assets/Tests/InputSystem/CoreTests_Devices_Keyboard.cs
ekcoh: Filed ticket internal ticket ISX-2325. Marking this resolved.

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
ekcoh: This was fixed in https://github.com/Unity-Technologies/InputSystem/pull/2230/commits/2eff10291de9699abbebfeafe9be896628bfea08

bmalrat (approved):

Packages/com.unity.inputsystem/InputSystem/Devices/Keyboard.cs
jfreire-unity: I see, got it.

PR #2234, opened 2025-09-11
FIX: MouseDownEvent is triggered when changing from Scene view to Game view (ISXB-1671)

jfreire-unity:
### Description

Proposed solution to fix [ISXB-1671](https://issuetracker.unity3d.com/issues/mousedownevent-is-triggered-when-changing-from-scene-view-to-game-view).

The bug reports a UITK event that shouldn't be triggered. This ocurrs due to a `InputSystemProvider` UI Click action being performed once the Editor is in play-mode and we select Game view tab when previously in Scene View.

The action that is triggered provides the wrong mouse position. Essentially, an action is being performed for the mouse click/position that happened while out of focus in the Editor and triggered the focus regain. However, when `InputManager.OnUpdate()` is called, it assumes the game already is in focus, so there was no way to know if an event was triggered out of focus or not.

The proposed solution is to filter and discard the events that happened immediately before focus was regained, by checking if they were received before the focus regain. In the diagram below, essentially we are discarding the "ofending" event - in red - that triggered an Input Action, which would then dispatch a UITK event through `InputSystemProvider`.

image

### Unrelated changes to bug-fix

I also did a small refactoring to put some conditions in their own functions so that can be easier to read through `InputManager.OnUpdate`. This is just a personal preference as it has been historically complicated to reason about all the conditions we have. This is still not perfect, but I hope it's a small improvement for others as well.

If you want, you can have a look of the commit which contains the fix changes, 61574984776fdb6824c4a6d0fa50a816e7349336. This doesn't have any refactoring yet and is easier to reason about.

### PR template update

Removes a section that we no longer need to take into account.

### Testing status & QA

### Overall Product Risks

- Complexity: 1
- Halo Effect: 1

### Comments to reviewers

Verify that the original bug doesn't happen anymore, both on Win and macOS.

Do some other tests with changing Background settings, to make sure nothing has regressed. We can likely use on of the Mouse Visualisers for it.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -642,6 +642,55 @@ public void Actions_DoNotGetTriggeredByEditorUpdates()
         }
     }
 
+    [Test]
+    [Category("Actions")]
```
jfreire-unity, line 645: I still need to add a InputForUI specific test for this.

ritamerkl (commented):
I am hesitant introducing such a big change in the InputManager, also looking at it I wonder:
Isn't the wrong event the mousePosition sent from DeviceSync? Since the event before should not cause the MouseDown, but the second event (caused by the DeviceSync). Or did I get that wrong?

jfreire-unity:
> I am hesitant introducing such a big change in the InputManager, also looking at it I wonder: Isn't the wrong event the mousePosition sent from DeviceSync? 

The wrong event is the one _before_ the DeviceSync. That's why we should discard it and not process it.
With this PR we make sure that we only process events received AFTER focus was regained. 

> I am hesitant introducing such a big change in the InputManager

I don't see it as a big change but maybe the small refactoring is "polluting" the PR in that sense? If you look at https://github.com/Unity-Technologies/InputSystem/pull/2234/commits/61574984776fdb6824c4a6d0fa50a816e7349336 there are a bit smaller changes and I just factored out some methods.

ekcoh (commented):
It's a bit difficult at the moment to judge what is refactoring and what is bug fixing / change in behavior on this PR. I will come back with additional comments once I have realised this.

jfreire-unity:
> It's a bit difficult at the moment to judge what is refactoring and what is bug fixing / change in behavior on this PR. I will come back with additional comments once I have realised this.


@ekcoh could you try looking at https://github.com/Unity-Technologies/InputSystem/commit/61574984776fdb6824c4a6d0fa50a816e7349336 first, since it has the least amount of changes? Let me know if it's better.
The refactor is just encapsulating the previously existing conditions into single functions. 
But if it's still confusing, I can redo this without any of these changes.

.github/pull_request_template.md
```diff
@@ -11,8 +11,8 @@ _Please describe the testing already done by you and what testing you request/re
 
 _Please rate the potential complexity and halo effect from low to high for the reviewers. Note down potential risks to specific Editor branches if any._
 
-- Complexity: 
-- Halo Effect: 
+- Complexity:
```
ekcoh, line 14: Why is this PR modifying the template?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -642,6 +642,55 @@ public void Actions_DoNotGetTriggeredByEditorUpdates()
...
+    [TestCase(InputSettings.BackgroundBehavior.ResetAndDisableAllDevices)]
+    public void Actions_DoNotGetTriggeredByOutOfFocusEventInEditor(InputSettings.BackgroundBehavior backgroundBehavior)
+    {
+        InputSystem.settings.backgroundBehavior = backgroundBehavior;
+
+        var mouse = InputSystem.AddDevice<Mouse>();
+        var mousePointAction = new InputAction(binding: "<Mouse>/position", type: InputActionType.PassThrough);
+        mousePointAction.Enable();
+
+        using (var trace = new InputActionTrace(mousePointAction))
```
ekcoh, line 659: I would recommend letting the time step parameter also be an input parameter to the test case. It should still be valid for timeStep = 0.0f, since our logic need to handle event order as the determining factor when time isn't sufficient to tell them apart.

ekcoh: So I would recommend adding that test as well.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ however, it has to be formatted properly to pass verification tests.
 - Fixed the compilation warnings when used with Unity 6.4 (ISX-2349).
 - Fixed an issue where `InputSystemUIInputModule.localMultiPlayerRoot` could not be set to `null` when using `MultiplayerEventSystem`. [ISXB-1610](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1610)
 - Fixed an issue in `Keyboard` where the sub-script operator would return a `null` key control for the deprecated key `Key.IMESelected`. Now, an aliased `KeyControl`mapping to the IMESelected bit is returned for compability reasons. It is still strongly advised to not rely on this key since `IMESelected` bit isn't strictly a key and will be removed from the `Key` enumeration type in a future major revision. [ISXB-1541](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1541).
+- An issue where a UITK MouseEvent was triggered when changing from Scene View to Game View in the Editor has been fixed.  [ISXB-1671](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1671)
```
ekcoh: Nitpick. Double space before ISXB number.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3706,6 +3664,160 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+        }
+
+        /// <summary>
+        /// Determines if we should exit early from event processing without handling events.
+        /// </summary>
+        /// <param name="eventBuffer">The current event buffer</param>
+        /// <param name="canFlushBuffer">Whether the buffer can be flushed</param>
+        /// <param name="updateType">The current update type</param>
+        /// <returns>True if we should exit early, false otherwise.</returns>
+        private bool ShouldExitEarlyFromEventProcessing(InputEventBuffer eventBuffer, bool canFlushBuffer, InputUpdateType updateType)
```
ekcoh: I would prefer that we do not pass canFlushBuffer to this function, nor eventBuffer. Those two is the obvious things to check in the outer scope. E.g. 

var shouldExitEarly = eventBuffer.eventCount == 0 || canFlushBuffer || ShouldExitEarlyFromEventProcessing(updateType);

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2244,6 +2244,8 @@ internal struct AvailableDevice
         private bool m_NativeBeforeUpdateHooked;
         private bool m_HaveDevicesWithStateCallbackReceivers;
         private bool m_HasFocus;
+        private bool m_DiscardOutOfFocusEvents;
```
ekcoh, line 2253: A bit of a corner case, but what happens with these if there is a domain reload while in playmode? Will they be reinitialised to the same values they had before the domain reload was initiated?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3706,6 +3664,160 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+            // like to drop status events, and do not early out.
+            if (!gameIsPlaying && gameShouldGetInputRegardlessOfFocus && (eventBuffer.sizeInBytes > (100 * 1024)))
+            {
+                canEarlyOut = false;
+                return true;
+            }
+            return false;
+        }
+
+        /// <summary>
```
ekcoh, line 3774: Thanks for documenting complex internal functions. I think its helpful.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3706,6 +3664,160 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+        /// Determines if an event should be discarded based on timing or focus state.
+        /// </summary>
+        /// <param name="eventType">The type of the current event</param>
```
ekcoh: Nitpick: I would advise against using the term "current" here. It is just an event being evaluated right?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3706,6 +3664,160 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+                return true;
+
+            // Check if this is an out-of-focus event that should be discarded
+            if (ShouldDiscardOutOfFocusEvent(eventTime))
+                return true;
+
+            return false;
+        }
+
+        /// <summary>
```
ekcoh, line 3794: Generally this makes sense, but have you tested that the following scenario works.
Press a gamepad button in edit mode and keep it held. Enter play mode. Is the button still considered down?

ekcoh (commented):
Commenting for now since the PR states it will add more tests and I added some suggestions. Generally I believe this is an improvement and happy to see some refactoring happening on InputManager. Regarding the refactor I would just like to make sure we have test coverage on the conditions before and after to avoid regressing something. It looked quite similar though and I couldn't spot any mistake there.
My other concern is regarding whether the focus change callback is direct or indirect. If direct it may lead to out-of-order processing of focus change and events which is why I have mentioned previously that we should have our focus changes in the event queue to make it trivial to reason about. If indirect we have a problem since we do not have the focus events true timestamp. Getting a timestamp when the callback is invoked doesn't need to imply that its the true timestamp of the event on the event timeline.

ekcoh:
@jfreire-unity Please update this branch with changes from develop to get critical CI fix from https://github.com/Unity-Technologies/InputSystem/pull/2260. I do not dare updating it for you since the branch has conflicts.

.github/pull_request_template.md
jfreire-unity: I wanted to remove the After merge section as it's not relevant anymore.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: Yes. There's no change in this behavior. There's a device sync done when we are in focus.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: Hm, good question. How can I trigger a domain reload in play mode so I can test that?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: Makes sense, changed.

Assets/Tests/InputSystem/CoreTests_Actions.cs
jfreire-unity: In that case, this would not work. Time needs to _move_; otherwise, we can't know what events have occurred before and after the focus. This solution relies on event timestamps. Similar to what we do for transitions between Edit Mode and Entering Play Mode, see the conditions in `ShouldDiscardEditModeTransitionEvent()`. Let me know if I'm missing something.

In theory, if we had focus events in the queues, we wouldn't need to know about timestamps. However, as we discussed, there is a higher risk in introducing such a change. We should still conduct an exploration of the implementation, but I would prefer to land first a less risky solution for this particular bug. 

I'm pappy to prototype a solution together if you like, after this fix. We can sync on Slack.

Pauliusd01 (changes requested):
Issue still occurs for me if I switch the resolution from free aspect to one of the predefined ones. 
https://github.com/user-attachments/assets/9a530010-e9cb-432c-9c9a-b047a3a0b4e6

jfreire-unity:
> Issue still occurs for me if I switch the resolution from free aspect to one of the predefined ones. https://github.com/user-attachments/assets/9a530010-e9cb-432c-9c9a-b047a3a0b4e6

I can't access that image

benoitalain:
Just chiming in to bring back a point that was mentioned in our early discussions about this issue. In an ideal world, we'd have to be able to click inside the GameView directly and have the first click that happened before we changed focus be processed as though it had been a normal click after focus. Most software allow that, where you change your window focus and at the same time you click a button from the other window. In that case, if the coordinates of the click (as expressed in the new focused window) are valid, we should accept the click. I don't know how computing new coordinates after the fact is feasible in general though.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
LeoUnity: You can either:
 - Change a script manually, that will trigger a domain reload
 - Make a script hat calls into EditorUtility.RequestScriptReload

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3032,6 +3034,8 @@ internal void OnFocusChanged(bool focus)
             }
             else
             {
+                m_DiscardOutOfFocusEvents = true;
+                m_FocusRegainedTime = m_Runtime.currentTime;
```
LeoUnity, line 3044: I'm wondering if this is granular enough, as in the m_Runtime.currentTime might not be exactly the time in which the application got focus but a bit later than that, which means it would potentially discard less events than it should...
But I am not sure about this since I can't find where onPlayerFocusChanged is called from, I can only find it being invoked from a test, so I don't know how "fast" we get to this function from when a focus change happens.

jfreire-unity: I guess that's a valid concern for devices queuing events not from the main thread (likely HID). For that particular case, the best way would be to have these focus events in the input queue, which is more complex and as a higher hallo effect. 
I agree this solution is not the best but a small pragmatic improvement. The best would likely be having focus events in the queue. 
Do you see any other alternatives with better granularity?

`onPlayerFocusChanged` is called whenever `Application.focusChanged` is called.
From native is called by [Player.SetPlayerFocus() ](https://github.cds.internal.unity3d.com/unity/unity/blob/5f34ab5ab7805304900ef761e6ea34d9fadd195d/Runtime/Misc/Player.cpp#L281)

jfreire-unity:
> Issue still occurs for me if I switch the resolution from free aspect to one of the predefined ones. https://github.com/user-attachments/assets/9a530010-e9cb-432c-9c9a-b047a3a0b4e6

Could you test the same on develop but by adding a `EventSystem` game object which will also add `InputSystemUIInputModule`? Feel free to ping me on Slack.

I believe the issue you are reporting happens regardless of using InputForUI, so it wouldn't be a regression.

Pauliusd01:
> > Issue still occurs for me if I switch the resolution from free aspect to one of the predefined ones. https://github.com/user-attachments/assets/9a530010-e9cb-432c-9c9a-b047a3a0b4e6
> 
> Could you test the same on develop but by adding a `EventSystem` game object which will also add `InputSystemUIInputModule`? Feel free to ping me on Slack.
> 
> I believe the issue you are reporting happens regardless of using InputForUI, so it wouldn't be a regression.

Here's what it's looking to me:

With the event system object:
- Your PR: doesn't repro with free aspect option, reproes with custom ones
- Develop: doesn't repro with free aspect option, reproes with custom ones

Without the event system object (default user project)
- Your PR: doesn't repro with free aspect option, reproes with custom ones
- Develop: reproes with free aspect option, reproes with custom ones

So It's still an improvement and the custom resolution thing is not a new regression, if you want it to be a separate bug that sounds fine to me. Only thing though is I'm seeing a lot of CI failures

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: I think it's better to use timestamps than what we had before and in practise it's likely not a problem. Still safe to land as is though and guess we need it in the queue to have a strictly accurate implementation.

jfreire-unity:
> > > Issue still occurs for me if I switch the resolution from free aspect to one of the predefined ones. https://github.com/user-attachments/assets/9a530010-e9cb-432c-9c9a-b047a3a0b4e6
> > 
> > 
> > Could you test the same on develop but by adding a `EventSystem` game object which will also add `InputSystemUIInputModule`? Feel free to ping me on Slack.
> > I believe the issue you are reporting happens regardless of using InputForUI, so it wouldn't be a regression.
> 
> Here's what it's looking to me:
> 
> With the event system object:
> 
> * Your PR: doesn't repro with free aspect option, reproes with custom ones
> * Develop: doesn't repro with free aspect option, reproes with custom ones
> 
> Without the event system object (default user project)
> 
> * Your PR: doesn't repro with free aspect option, reproes with custom ones
> * Develop: reproes with free aspect option, reproes with custom ones
> 
> So It's still an improvement and the custom resolution thing is not a new regression, if you want it to be a separate bug that sounds fine to me. Only thing though is I'm seeing a lot of CI failures

Thanks for testing!

Yes, I would raise it as different bug since it is something already happenin with Event System + UIInputModule and will likely require trunk changes. It is not a focus change issue. 
I prefer to land this small improvements to at least fix the regression reported. 

I'm looking into the CI issues and will ping reviewers once those are fixed

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: Maybe we can also add an inline comment to this test case that with a strictly defined event queue containing all relevant inputs without downsampling the test case should be expanded to be time-agnostic for a fixed timestamp?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Did you conclude this @jfreire-unity ?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: It's a valid concern and the same concern I raised in a comment above but by definition background devices are unrelated to focus. Hence their filtering will never be perfect. ALL uses cases will be greatly simplified if we get focus into queue and move background behaviour filtering down into module though, I have captured it in the "roadmap draft".

ekcoh (approved):
Thanks for doing this improvement/fix. I will approve now since I believe this makes it more correct than existing solution.

Assets/Tests/InputSystem/CoreTests_Actions.cs
jfreire-unity: This is done now.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
jfreire-unity: They are initialized to their default values (false for the boolean, 0.0 to the timestamp).
I'm not seeing a reason why they should have their same values as before the domain reload. The timestamp will always be "old". But maybe I'm missing some detail or a pattern.

ekcoh: OK, thanks for following up, good to have checked the domain reload scenario.

Pauliusd01 (approved):

jfreire-unity:
I'll merge anyway since the failure is unrelated to this PR and happens on `develop`

PR #2237, opened 2025-09-16
FIX: out of range exception when pressing undo after creating and editing a new control scheme (ISXB-1607) 

Darren-Kelly-Unity:
### Description

You could cause multiple null or out of range exceptions if you press undo while editing a control scheme.
This is caused by undo causing the control scheme to be removed while you still have the edit window open.

Now the edit window will close immediately if you press undo while it's open.

Ticket:
https://jira.unity3d.com/browse/ISXB-1607

Video of fix (Usually undoing when editing the new control scheme would throw an exception / it could also cause other exceptions if the window was left open, therefore I now close it when this happens to avoid future issues):

https://github.com/user-attachments/assets/8c7f78ed-dfd8-4c2d-b24a-6f01363ce6a0

### Testing status & QA
Tested manually all the cases I could find until all exceptions were gone. There was multiple paths to create exceptions due to the asset being removed via undo.

### Overall Product Risks

- Complexity: 2
- Halo Effect: 1

### Comments to reviewers

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
```diff
@@ -104,6 +107,27 @@ public InputActionsEditorView(VisualElement root, StateContainer stateContainer,
...
             s_OnPasteCutElements.Add(this);
+
+            Undo.undoRedoPerformed += CheckForInvalidControlSchemeInOneFrame;
+        }
+
+        private async void CheckForInvalidControlSchemeInOneFrame()
+        {
+            try
+            {
+                await Task.Delay(1);
```
ekcoh: This await Task.Delay(1) raises warning flags to me. What guarantees this is valid to execute 1 second from now? 
- Why does it have to be delayed?
- It looks like this runs async but code being accessed do not seem to be thread safe?
- Is there a logical condition not based on time when this is valid to execute?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
```diff
@@ -104,6 +107,27 @@ public InputActionsEditorView(VisualElement root, StateContainer stateContainer,
                 });
 
             s_OnPasteCutElements.Add(this);
+
+            Undo.undoRedoPerformed += CheckForInvalidControlSchemeInOneFrame;
```
ekcoh: Maybe I am misunderstanding this PR/issue so I want to check whether this problem is if you undo text editing while in text editing mode? Such editing should not have been applied yet to any serializedProperty right or am I misunderstanding this issue?

ekcoh (commented):
Mainly asking question at this point since I am not sure I understand this fix or problem yet....

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
Darren-Kelly-Unity: It's a bit different than that, basically the last undo when opening a control scheme to edit is the creation of the asset.

Therefore undoing while editing the name actually deletes the asset.
This of course causes exceptions when inside the edit window if you try to alter the asset or even try to close it.

My solution I have landed on for now is to close the control scheme view when this happens and revert to selecting the all controls control scheme now instead.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/InputActionsEditorView.cs
Darren-Kelly-Unity: You are right, it was something I added at the end to solve an exception but it's not the correct way to do this.

I have now resorted to just closing the window when an undo is triggered if it's open.

This has one side effect though that it will trigger the view to close in a case where it might not need to, definitely less bad than all the exceptions it could cause previously.

The best solution that I could not figure out a way to do would be to add the opening of the control scheme editor popup to the undo stack so that pressing undo would close it, I did find a way to possibly do this but it would add a good amount of complexity and would be quite hacky.

I chose not to go this route in the end as the undo api is not really designed to be used with things that are not Objects/GameObjects.

ekcoh (approved):
Thanks for looking into this Darren, not sure I fully understand the fix but it seems legit. Since the editor is something where we do not have the greatest coverage, do you think it would be feasible to auto-test this regression somehow? If not or unfeasible I think it can land as-is with manual testing.

ekcoh:

Pauliusd01 (approved):

jfreire-unity (approved):

Darren-Kelly-Unity:
@ekcoh I tried adding a test for this like discussed in standup but ran into issues with the advanced dropdown not using UIToolkit causing issues.

PR #2241, opened 2025-09-22
CHANGE: ISX-2377: Rebinding sample improvements (Mouse sensitivity + Swap sticks) (Validation week), FIX: ISXB-1721 ApplyParameterOverride doesn't work if binding name is empty string.

ekcoh:
### Description

This PR improves the "rebinding sample" in the following ways:
- Adds a "swap bindings" button to showcase how to swap two similar bindings. This is currently applied to the move and look actions of the gamepad control scheme.
- Improved `RebindSaveLoad` by documenting it and extracting Load/Save to API surface of the class and only calling them from OnEnable/OnDisable. Also made the actual `PlayerPrefs` key a string property of the component. This allows this file to be a reusable component as-is is nice there is nothing hard-coded and it may be used on OnEnable/OnDisable or just programatically.
- Removed delta action binding from _"Look_" action since it is intentionally ignored by rebinding framework. This control is now instead replaced by a UI slider allowing to control/override mouse sensitivity of the associated action and persist the setting across runs.
- Refactored rebinding sample custom editor code to reduce code bloat and enable code reuse via composition.
- Extended "Reset All" button to also reset look sensitivity for "Keyboard & Mouse" scheme.
- Added `RebindActionParameterUI` which is a slider UI integration example of configuring action processor parameters directly by adding the monobehavior, configuring it and run. Supports overriding action or binding processors.
- Added `RebindActionParameterUIEditor` which is a custom editor for the above `MonoBehaviour`.

It also fixes bug ISXB-1721 which was required to make mouse sensitivity work properly. Looking through the logic close to the the fix/diff in `InputActionParameters.cs` suggests there are more empty-string bugs for the other match conditions that might need to be looked into and be properly tested since all but ID check seems to suffer from similar problems.

### Testing status & QA

Functionality has been tested during development by running the associated sample.

Automated test have been added to prove failure of ISXB-1721 and its fix and prevent future regressions.

### Overall Product Risks

Small, no changes to source, only to sample scripts.

- Complexity: Medium
- Halo Effect: Small

### Comments to reviewers

I am not 100% sure the scene save doesn't contain any accidental changes but I believe it should be fine.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Assets/Samples/RebindingUI/BindingUI.cs
```diff
@@ -0,0 +1,146 @@
...
+using System;
+using System.Linq;
+using UnityEditor;
+
+namespace UnityEngine.InputSystem.Samples.RebindUI
+{
+    /// <summary>
+    /// Common binding UI helper to allow editor composition.
+    /// </summary>
+    internal class BindingUI
```
ekcoh, line 12: This isn't consistently named with file and should be renamed.

ritamerkl:
I just tried the sample and I think the improvements are nice, I only have one point to critisize:
The swap button is very nondescript, I only discovered it after reading there is one. 
I will look at the code now.

Pauliusd01:
I'm not sure if the sensitivity option is working, I can still rotate the character the same amount of times from the same amount of mouse distance regardless whether I'm on low or high sensitivity. (Maybe the differences between low and high are just very low? Since I didn't feel a difference at first what I did was go to the left edge of the screen and drag my mouse to the right counting full rotations of the character with both low and high sensitivity. Both totaling to around 2 full rotations exactly) Also, like Rita mentioned the swap button is tiny. Maybe spelling it out alongside the arrow icon would work. A crude example: 

image

Otherwise looking good to me

ekcoh:
> I'm not sure if the sensitivity option is working, I can still rotate the character the same amount of times from the same amount of mouse distance regardless whether I'm on low or high sensitivity. (Maybe the differences between low and high are just very low? Since I didn't feel a difference at first what I did was go to the left edge of the screen and drag my mouse to the right counting full rotations of the character with both low and high sensitivity. Both totaling to around 2 full rotations exactly) Also, like Rita mentioned the swap button is tiny. Maybe spelling it out alongside the arrow icon would work. A crude example:
> 
> image
> Otherwise looking good to me

I have a look if the slider is working, might be that I broke something while cleaning this up. 

Regarding the swap button I can change it but vote to keep its current design but make it larger and to the right of the Reset buttons spanning the same height as the top of left stick button and bottom of right stick button.

ekcoh:
Screenshot 2025-10-10 at 11 21 16

Made the swap button larger as requested.

ekcoh:
I can confirm sensitivity slider is broken, will fix and rerequest review when done.

ekcoh:
@Pauliusd01 Had to extend this PR with a bug fix for ISXB-1721 (and test case to prove it failed before and passes after fix) which was needed to support mouse sensitivity properly so should be good now.

Pauliusd01 (changes requested):
Two issues I see, one is that the UI map is active during gameplay:
image

And the other is that I can't navigate the rebinding menus via gamepad anymore when I previously could. At first I thought it was the slider's fault but I can't jump to the right side of the menus even without having selected it so something else is going on there.

ekcoh:
> Two issues I see, one is that the UI map is active during gameplay: image
> 
> And the other is that I can't navigate the rebinding menus via gamepad anymore when I previously could. At first I thought it was the slider's fault but I can't jump to the right side of the menus even without having selected it so something else is going on there.

Thanks, I guess I messed up something in the UI navigation configuration, will check. The UI map being enabled is a mistake from me since it had to be enabled to support on screen controls. Since I removed them from this branch (was only temporarily there), I guess I forgot to undo that change.

ekcoh:
> one is that the UI map is active during gameplay:
This has been addressed in https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/35a4927bf17154a260dcf195819d33daf6845e5e, also removed script RequireTouichscreen and the on-screen menu button which was associated with on screen controls and shouldn't be included in this PR.

ekcoh:
Will look into UI navigation now.

ekcoh:
> Will look into UI navigation now.

Seems UGUIs automatic navigation mode will not be sufficient to get it right given current UI which seems to be indicated when visualising the automatic navigation links. I will set it up explicitly to work around it unless I find some culprit soon.

Culprit seems to be the slider, as long as it is included the buttons no longer get decent navigation links, so I guess we need to set it up.

ekcoh:
> And the other is that I can't navigate the rebinding menus via gamepad anymore when I previously could. At first I thought it was the slider's fault but I can't jump to the right side of the menus even without having selected it so something else is going on there.

I have now added explicit navigation links to make sure UI navigates correctly when Automatic is not sufficient. Ready to re-evaluate @Pauliusd01

Pauliusd01 (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,8 +10,15 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
 
+### Changed
+- Extended `RebindingUISample` to include example of mouse sensitivity slider and "swap controls" example.
```
ritamerkl: I recommend removing the "swap controls" part, since it is mentioned in the added section.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,8 +10,15 @@ however, it has to be formatted properly to pass verification tests.
...
+
+### Added
+- Added an example of how to switch two similar controls to the `RebindingUISample`. This is accessible via a "swap sticks button" which allows swapping the bindings of the "move" and "look" gamepad bindings.
```
ritamerkl: Minor: I would change this to
This is accessible via a button which ...
I would expect a button with the name "swap sticks button" otherwise.

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
```diff
@@ -846,9 +846,9 @@ public bool Matches(InputBinding binding)
         internal bool Matches(ref InputBinding binding, MatchOptions options = default)
         {
             // Match name.
-            if (name != null)
+            if (!string.IsNullOrEmpty(name))
             {
-                if (binding.name == null
+                if (string.IsNullOrEmpty(binding.name)
```
ritamerkl, line 851: nice!
The only question is... comparing an empty string with an empty string would fail now, but wouldn't before?

ritamerkl: I am wondering, maybe there is a reason we would want to allow that?

ritamerkl: Okay, looking at the test, I understand now. Will resolve.

Assets/Samples/RebindingUI/Game/GameplayManager.cs
```diff
@@ -256,7 +259,7 @@ void SpawnEnemy()
             --m_EnemySpawnCount;
 
             // Rent an enemy from the enemy pool
-            var enemyComponent = m_EnemyPool.Get();
+            var enemyComponent = m_EnemyPool.Get(); // TODO Null reference here on domain reload
```
ritamerkl: Is this solved?

Assets/Samples/RebindingUI/InputActionIndicator.cs
```diff
@@ -34,7 +34,9 @@ public class InputActionIndicator : MonoBehaviour
         private void OnEnable()
         {
             if (action != null && action.action != null)
-                action.action.performed += OnPerformed;
+            {
+                action.action.performed += OnPerformed; // TODO Problem here after domain reload, InputAction.addperformed(), CallbackArray.AddCallback,. InputArrayExtensions.Contains
```
ritamerkl: same question here, is this solved?

Assets/Samples/RebindingUI/RebindSaveLoad.cs
```diff
@@ -3,21 +3,79 @@
...
+        {
+            if (saveOnDisable)
+                Save();
+        }
+
+        private bool IsValidConfiguration()
+        {
+            if (actions == null)
+            {
+                Debug.LogWarning("Unable to apply binding overrides from player preferences without an associated " +
```
ritamerkl: Why the split in the string, and can we add an actionable solution?

Assets/Samples/RebindingUI/RebindSaveLoad.cs
```diff
@@ -3,21 +3,79 @@
...
+                    "action asset.");
+                return false;
+            }
+
+            if (string.IsNullOrEmpty(playerPreferenceKey))
+            {
+                Debug.LogWarning("Unable to load binding overrides from player preferences without a key");
```
ritamerkl: Warning message confusing without the specific context (add more context - what key?), same as above, can we add an actionable solution?

ritamerkl (approved):
Mainly nitpicks, the code looks good to me. 
There are two comments, which needs to be addressed and one outstanding file rename.

Assets/Samples/RebindingUI/RebindSaveLoad.cs
ritamerkl: I think that would be just nice for a sample to guide towards a solution, nothing mandatory for sure.

ekcoh: A preferences key-value map key. I can add "without a preferences key"? Its a field to set in the Inspector of this component

Assets/Samples/RebindingUI/InputActionIndicator.cs
ekcoh: Good catch, it's not, it's a separate bug, will file a separate ticket on it. Its not related to PR.

Assets/Samples/RebindingUI/Game/GameplayManager.cs
ekcoh: Good catch, it's not, it's a separate bug, will file a separate ticket on it. Its not related to PR.

Assets/Samples/RebindingUI/RebindSaveLoad.cs
ritamerkl: Yes, that would be helpful, also you could add something like fix this add a Preference key in PlayerSettings->.... ?

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Thanks, I think the Changed part should also be modified to say "Replaced the Look rebinding button with a mouse sensitivity slider as an example for how to customise mouse sensitivity for the associated control scheme."

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: I reconsider how this is formulated.

Assets/Samples/RebindingUI/RebindSaveLoad.cs
ekcoh: Too long row, not sure I manually split it or if the formatting tool did.

Assets/Samples/RebindingUI/RebindSaveLoad.cs
ekcoh: You just need to type in text in the components field called "Player Preference Key" and it just works. This is just to catch the case of the field being null or empty in which case a setting key cannot be created on the fly. 
But I can add something about either assigning it programatically or via Inspector.

ritamerkl: I leave that up to you, it's super minor and rather an idea than any proper suggestion

Assets/Samples/RebindingUI/RebindSaveLoad.cs
ritamerkl: oh ok, I did not know the tool would be doing something like that, I was just wondering why.

Assets/Samples/RebindingUI/BindingUI.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/2ed59380b440db182cd5ca139f760d63ce7a4333

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: This have been addressed in https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/a2ccf0cea46efbbdd311704b68b786321db6be58

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: This have been addressed in https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/a2ccf0cea46efbbdd311704b68b786321db6be58

Assets/Samples/RebindingUI/Game/GameplayManager.cs
ekcoh: Will also remove TODO

Assets/Samples/RebindingUI/InputActionIndicator.cs
ekcoh: Will also remove TODO

Assets/Samples/RebindingUI/Game/GameplayManager.cs
ekcoh: Filed as internal issue ISXB-1726 and linked it to ISX-2377.

Assets/Samples/RebindingUI/InputActionIndicator.cs
ekcoh: Filed as internal issue ISXB-1726 and linked it to ISX-2377.

ekcoh: TODO Removed in https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/d8cb402b223f18c40a00f91458b5447905c6804e

Assets/Samples/RebindingUI/Game/GameplayManager.cs
ekcoh: TODO Removed in https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/d8cb402b223f18c40a00f91458b5447905c6804e

Assets/Samples/RebindingUI/RebindSaveLoad.cs
ekcoh: Reverted this in https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/6948e6f61def98c0aa0c57a7a7c7d998af4a8ed7

Assets/Samples/RebindingUI/RebindSaveLoad.cs
ekcoh: I did a slight improvement. It also has tooltip description so hope it's good enough. See https://github.com/Unity-Technologies/InputSystem/pull/2241/commits/2984cf4b50e63df7fb0dbeab0a6d893509541a47

Packages/com.unity.inputsystem/InputSystem/Actions/InputBinding.cs
ekcoh: I noticed you marked this one resolved, but let's discuss to make sure there is no "devil in the details" here. The method attempts to figure out whether a "binding" and a "binding mask" matches. In reality they are both binding objects. I struggle to see why the previous behavior would be correct here.

The conditions are equivalent to before but have been extended to handle the empty string since CharacterSeparatedListsHaveAtLeastOneCommonElement is false for empty string. Now, if both are null they are considered to "match", if both are empty string they are considered to "match" (this was not the case before). And if they contain a string they match if strings are equal. IMO even if the string is empty it should be a string compare. The odd thing here is that "matching" is loosely defined and weaker than equality.

I am however afraid the similar checks below this one are also incorrect, but could be followed up outside this PR. I filed ISXB-1727 to follow-up with test cases, investigation and potential additional bug fixes for Matches.

ritamerkl: > if both are empty string they are considered to "match" (this was not the case before).

Yes, that's what I was worried about. Thinking about it twice- I realized, it makes sense it is that way. 
My concern was that if there are usages of this method that rely on the old way of comparing empty strings which led to a mismatch. That could potentially lead to bugs somewhere-

ekcoh: Yes, that is a valid concern, but the empty string being considered an, in that case, undocumented special case seems less likely to be something we want.

ekcoh: Will mark this resolved, let me know if you think differently.

PR #146, opened 2018-06-15
TEMP PR: Onscreen controller

Rene-Damm:
Temporary PR not meant for merging. Just meant as an intermediate review.

PR #2244, opened 2025-09-25
FIX: ISXB-1674 - Input actions asset not converted correctly when upgrading from 1.14.1 (Regression)

AswinRajGopal:
### Description

Bug: https://jira.unity3d.com/browse/ISXB-1674
Port: 1.14.X

Preserve original tokens and only touch the enum values. That means processor names, order, legacy processors, and formatting survive unchanged so the editor no longer collapses or marks them “Obsolete”.
Avoid whole string rewrites, which previously caused mismatches between what the editor expected and what was written back and only assign a new processor string if something actually changed.

### Testing status & QA

Verified manually with the attached repro project.

### Overall Product Risks

Complexity: 0
Halo Effect: 0

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


ekcoh:
I noticed this PR title was incorrect so I changed it to remove that failure.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
...
+            }
+
+            for (var mi = 0; mi < parsedJson.maps.Length; ++mi)
             {
-                for (var mi = 0; mi < parsedJson.maps.Length; ++mi)
+                var mapJson = parsedJson.maps[mi];
+                if (mapJson.actions == null || mapJson.actions.Length == 0)
+                    continue;
+
+                for (var ai = 0; ai < mapJson.actions.Length; ++ai)
```
ekcoh: Would it be possible to add some inline comments to the different parts of this large for loop or split it up so that is clear what the intent is? It is quite difficult to review given that there is no motivation given for what we do here.

ekcoh: It looks like 'ai' isn't really used so why not e.g foreach instead?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
...
+                    var actionJson = mapJson.actions[ai];
+                    var raw = actionJson.processors;
+                    if (string.IsNullOrEmpty(raw))
+                        continue;
+
+                    var parts = System.Text.RegularExpressions.Regex.Split(raw, @"\s*([,;])\s*");
+                    if (parts.Length == 0)
+                        continue;
+
+                    var tokens = new List<string>();
```
ekcoh: Since this is inside a loop I would recommend moving `tokens` outside the loop and just do `clear` here to avoid GC pressure.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
...
+                    for (int i = 0; i < parts.Length; i += 2)
+                        if (!string.IsNullOrEmpty(parts[i]))
+                            tokens.Add(parts[i]);
+
+                    if (tokens.Count == 0)
+                        continue;
+
+                    var parsed = new List<NameAndParameters>(tokens.Count);
```
ekcoh: Also this list may be moved to outer scope to avoid allocating on every iteration.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
...
+                    foreach (var t in tokens)
+                        parsed.Add(NameAndParameters.Parse(t));
+
+                    var rebuiltTokens = new List<string>(tokens.Count);
```
ekcoh: same with this list

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
...
+                        var nap = parsed[pi];
+
+                        var procType = InputSystem.TryGetProcessor(nap.name);
+                        if (procType == null || nap.parameters.Count == 0)
+                        {
+                            rebuiltTokens.Add(tokens[pi]);
                             continue;
+                        }
+
+                        var dict = new Dictionary<string, string>(nap.parameters.Count, System.StringComparer.OrdinalIgnoreCase);
```
ekcoh: This dict could also be outer scope?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
...
+                                continue;
+
+                            if (!dict.TryGetValue(field.Name, out var rawVal))
                                 continue;
-                            }
 
-                            var dict = nap.parameters.ToDictionary(p => p.name, p => p.value.ToString());
-                            var anyChanged = false;
-                            foreach (var field in procType.GetFields(BindingFlags.Public | BindingFlags.Instance).Where(f => f.FieldType.IsEnum))
+                            if (int.TryParse(rawVal, NumberStyles.Integer, CultureInfo.InvariantCulture, out var n))
```
ekcoh: I do not understand what the goal is here, could you add inline comments explaining?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
...
-                                        dict[field.Name] = Convert.ToInt32(values[ord]).ToString();
-                                        anyChanged = true;
+                                        dict[field.Name] = underlying.ToString(CultureInfo.InvariantCulture);
+                                        changedThisProcessor = true;
                                     }
                                 }
                             }
+                        }
+
+                        if (!changedThisProcessor)
```
ekcoh: Also code from here to end is fuzzy to me, it would be helpful with inline comments explaining what we are trying to achieve in order to review.

ekcoh (changes requested):
Thanks for looking into this, my main comment at this review round is to add inline comments explaining why we need to do the different parts since it is not clear to me from the code. I think this is important to make sure its maintainable.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: Sure I'll add comments in detail.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,31 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
         yield return null;
     }
+    
+    [UnityTest]
+    public IEnumerator Migration_FromLegacyJson_ShouldConvertOrdinal_KeepInvertVector2_AndSeparators()
+    {
+        var legacyJson = m_Asset.ToJson().Replace("\"version\": 1", "\"version\": 0").Replace("Custom(SomeEnum=10)", "Custom(SomeEnum=1)");
+
+        // Add a trailing processor to verify the semicolon separator is preserved.
+        if (!legacyJson.Contains(";InvertVector2(invertX=true)"))
```
LeoUnity: The fact we have an conditional treatment on our test data before we act on it is not a good sign, ideally the data we use for testing is known ahead of time and we can just apply the function under test on it.

Whenever we need to bump the current version to 2 the replace above will fail, which is not what the test is trying to test.

I am wondering if a better approach here would be to just have a static json string here instead.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1014,60 +1014,125 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
         {
             if (parsedJson.version >= JsonVersion.Version1)
                 return;
-            if ((parsedJson.maps?.Length ?? 0) > 0 && (parsedJson.version) < JsonVersion.Version1)
+
```
LeoUnity: Why do we need to do this parsing manually ourselves here? Is there anything special that we need in this parsing that is no needed elsewhere?
I'm assuming this data is used in other parts of the system, and there would be logic we can reuse... now if there is an issue with the parsing of jsonmaps we have to fix in 2 places...

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,31 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+        Assume.That(migratedAsset, Is.Not.Null, "Failed to load legacy JSON into an InputActionAsset.");
+
+        var migratedJson = migratedAsset.ToJson();
+        Assume.That(migratedJson, Is.Not.Null.And.Not.Empty, "Migrated JSON was empty.");
+
+        Assert.Less(migratedJson.IndexOf("InvertVector2(invertX=true)", StringComparison.Ordinal), migratedJson.IndexOf(",Custom(SomeEnum=20)", StringComparison.Ordinal),
+            "Expected a comma between the first and second processors, with InvertVector2 first."
+        );
+
+        Assert.Greater(migratedJson.IndexOf(";InvertVector2(invertX=true)", StringComparison.Ordinal), migratedJson.IndexOf("Custom(SomeEnum=20)", StringComparison.Ordinal),
```
LeoUnity: I'm also wondering if asserting on a substring is the most robust way here... The substring might be there but other things could make the json parsing fail... What we want to make sure is that after the migration the json can be parsed without issues and that the enum values are correct.

We even have the migratedAsset object, so let's just assert that the processors are set correctly.

LeoUnity (commented):
There are a few things you mention in the PR description that I don't understand why they are needed:

> Preserve original tokens and only touch the enum values

> Formatting survive unchanged
I don't understand why this would be a requirement, I believe that if the user uses our editor to edit the asset that was written manually and has custom formatting that will overwrite the users formatting. Trying to parse json with string replaces, regex is a losing battle, the only way this won't be brittle is by parsing it, converting the data and serializing it back to json.

> Avoid whole string rewrites, which previously caused mismatches between what the editor expected 
As above, I don't understand why do we want to avoid string rewrites, my current thinking is that we didnt catch this for 2 reasons, we are not reusing the parsing and serializing funcitons that is used elsewhere, and we didnt test this case.


I believe the proper way of moving forward here is that we use the exact same code to parse and serialize json for this type that is used elsewhere in the editor, if we try to do anything different we are creating an opportunity for a bug to be hidden here.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,58 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+
+        var action = map.FindAction("Move");
+        Assert.That(action, Is.Not.Null, "Expected Move action to exist.");
+
+        var processors = action.processors;
+
+        // Verify processor order and that enum was converted properly
+        Assert.That(processors, Does.Contain("StickDeadzone"), "StickDeadzone processor missing.");
+        Assert.That(processors, Does.Contain("InvertVector2(invertX=false)"), "InvertVector2 missing.");
+        Assert.That(processors, Does.Contain("Custom(SomeEnum=20)"), "Custom(SomeEnum=1) should migrate to SomeEnum=20 (OptionB).");
```
LeoUnity: This test is not verifying that the processor string can be parsed properly, which is what we want... we could still have a ";" separating them and those asserts would most likely pass.

@ekcoh do you know how we could get access to the parsed list of processors so we could assert on that instead? I could only find InputControl.processors or InputActionState.processors, but I don't know how to go from an InputActionAsset to InputControl or InputActionState

It seems InputAction.ResolveBindings end up calling InputBindingResolver.AddActionMap which then calls InstantiateWithParameters and that leads to the parsing and creation of processors, but I am not sure how we could call that from a test and then retrieve the list of processors to assert on that.

ekcoh: Good catch, since the API design of Input Actions is based on string you can put whatever into Processors. So this test needs to complemented with code resolving the actions and the only way to actually prove the processors is correctly parsed and applied I believe would be to enable the actions and do ReadValue on the action after being resolved. To drive that action you also likely need a fake device to drive it with e.g. 1.0 and then see that e.g. scale processor multiples by 10 or similar.

In CoreTests_Actions.cs there are various examples on how to achieve this, e.g. search for tests containing ReadValue.

ekcoh: This would have been trivial if the string based API design was not a thing and instead actions had a list of actual processors on them.

ekcoh (changes requested):
Same comment as @LeoUnity, let's complement the test by making sure it parses correctly (logically) and then we can conclude this I believe.

Packages/com.unity.inputsystem/InputSystem/Utilities/NameAndParameters.cs
```diff
@@ -27,6 +27,24 @@ public override string ToString()
             return $"{name}({parameterString})";
         }
 
+        internal static string SerializeMultiple(IEnumerable<NameAndParameters> list)
+        {
+            if(list == null)
```
ekcoh: Is this null check needed? It looks like this is never called with null?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1027,43 +1027,37 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
                             continue;
 
                         var list = NameAndParameters.ParseMultiple(raw).ToList();
-                        var rebuilt = new List<string>(list.Count);
+                        var converted = new List<NameAndParameters>(list.Count);
```
ekcoh: I still do not see why this isn't moved to outer scope and reused over iterations instead of being allocated on every iteration? Am I missing something?

Packages/com.unity.inputsystem/InputSystem/Utilities/NameAndParameters.cs
LeoUnity: I'm not sure, it was copied from the file above 
> private static string ToSerializableString(IEnumerable<NameAndParameters> parametersForEachListItem)

It was moved here just to make sure we can reuse the logic on those 2 places. But I have no idea if it can ever be null on the original place or not.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
AswinRajGopal: Hi @LeoUnity @ekcoh, our NameAndParameters.ParseMultiple() explicitly checks for both seperators `,`  and `;` between processors entries. Either way we get the same result however I agree with semantic issues from string only assertions. 
Should we go for 
`var parsed = NameAndParameters.ParseMultiple(action.processors).ToList();
 Assert.That(parsed.Count, Is.EqualTo(expectedCount)); ?
`

Packages/com.unity.inputsystem/InputSystem/Utilities/NameAndParameters.cs
ekcoh: OK, let it be then

Packages/com.unity.inputsystem/InputSystem/Utilities/NameAndParameters.cs
```diff
@@ -27,6 +27,24 @@ public override string ToString()
...
+                return string.Empty;
+
+            return string.Join(NamedValue.Separator, list.Select(x => x.ToString()).ToArray());
+        }
+
+        internal static NameAndParameters Create(string name, IList<NamedValue> parameters)
+        {
+            var result = new NameAndParameters
```
ekcoh: Nitpick: Why create a result variable here instead of just 

return new NameAndParameters { .... };

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/NameAndParametersListView.cs
```diff
@@ -83,11 +83,7 @@ private void OnParametersChanged(ParameterListView listView, int index)
 
         private static string ToSerializableString(IEnumerable<NameAndParameters> parametersForEachListItem)
```
ekcoh: Why is this function existing if it only forwards to NameAndParameters.SerializeMultiple? Seems redundant

ekcoh: Maybe keep the old name instead?

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
ekcoh: Wouldn't it just work to do this? Its simple and very little code

var legacyJson = ....
var modernJson = ....

var ported = InputActionAsset.FromJson(legacyJson).ToJson();
Assert.That(modernJson, Is.EqualTo(ported));

ekcoh: Ideally you would have compared InputActionAsset objects but unfortunately it's not supported :/ Hence full cycle allows you to avoid the details. If porting is successful I would expect them both to be equal. Or even

var legacyJson = ....
var modernJson = ....

var ported = InputActionAsset.FromJson(legacyJson).ToJson();
Assert.That(InputActionAsset.FromJson(modernJson).ToJson(), Is.EqualTo(ported));

Since that wouldn't break either if formatting is updated

AswinRajGopal: Yes it should work, earlier the cause of regression was not considering `;` separator in the migration logic it only splits `,` and keeps the `;` unchanged caused the serialization to break. Since now as @LeoUnity suggested the robust approach to use the NameAndParameters the above FromJson and ported json assertion should do good I believe.

LeoUnity: > our NameAndParameters.ParseMultiple() explicitly checks for both seperators , and ;

This is the source of the regression we have today and we are fixing on this PR. The comma and semicolon are both used, but in different contexts.

Take a look at this example, this should allow you to assert on the parsed processors:
```
        var inputActionAsset = AssetDatabaseUtils.CreateAsset<InputActionAsset>();
        var actionMap = inputActionAsset.AddActionMap("Action Map");
        actionMap.AddAction("Action", InputActionType.Value, processors: "Custom(SomeEnum=10)").AddBinding("<Gamepad>/leftStick");
        
        actionMap.ResolveBindings();
        
        Assert.That(actionMap.m_State.processors[0].GetType(), Is.EqualTo(typeof(CustomProcessor)));
        Assert.That((actionMap.m_State.processors[0] as CustomProcessor).SomeEnum, Is.EqualTo(SomeEnum.OptionA));
```

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
AswinRajGopal: the iterator `actionJson` is immutable as the ReadActionJson is a struct so we use for loop instead of foreach.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
AswinRajGopal: One binding was needed for the action map to resolve the bindings.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +112,14 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
 
         yield return null;
     }
+
+    [Test]
```
LeoUnity, line 113: We are missing a test case here that tests that an action map containing 2 processors are working. It's the whole reason we are doing this fix. We know that with a single processor everything was working.

LeoUnity (changes requested):
Missing an automated test case that reproduces the issue we are trying to fix here.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +112,24 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+    public void Migration_ShouldProduceValidActionAsset_WithAllProcessors_AndEnumConverted()
+    {
+        m_ActionMap.ResolveBindings();
```
LeoUnity: This test passes without the fix, this means this is not testing what we want. The asset in the editor is created with version 1 already, the migration code is not being run at all as part of this test.

There are 2 problems with the way we create the asset:
 - The asset is created once and shared across the different tests, which means the order in which the tests run could impact the results of the test, this is bad.
 - The migration code requires the version to be 0, creating the asset via the editor will create an asset with version 1.

Also the start data already has the CustomProcessor with the value being 10, the initial value should be 1, and then the conversion should run and upgrade it to 10.

This test is not a UI test, but its part of the UIToolkitBaseTestWindow class, I would recommend splitting this test into its own class that has no dependency on the UI.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionAsset.cs
```diff
@@ -1026,44 +1031,49 @@ internal void MigrateJson(ref ReadFileJson parsedJson)
                         if (string.IsNullOrEmpty(raw))
                             continue;
 
-                        var list = NameAndParameters.ParseMultiple(raw).ToList();
-                        var rebuilt = new List<string>(list.Count);
-                        foreach (var nap in list)
+                        if (!NameAndParameters.ParseMultiple(raw, ref parsedList) || parsedList.Count == 0)
```
ekcoh: Do we need `parsedList.Count == 0` here? It looks like this just needs to for loop below not being executed since Count == 0? If we can simplify logic I believe we should.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
 
         yield return null;
     }
+    
+    [Test]
+    public void Migration_ShouldProduceValidActionAsset_WithEnumProcessorConverted()
+    {
+        var legacyJson = @"
```
ekcoh: Why is this a var? It can be a `const string`.

ekcoh: Also, could you add an inline comment here that explains "what part" of json is part of interest or considered legacy. It is not clear from this test case which makes it difficult to maintain.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+        }";
+    
+        var asset = InputActionAsset.FromJson(legacyJson);
+        Assert.That(asset, Is.Not.Null, "Migration failed to produce a valid InputActionAsset.");
+    
+        var map = asset.FindActionMap("Player");
+        Assert.That(map, Is.Not.Null, "Expected Player map to exist.");
+    
+        var action = map.FindAction("Move");
+        action.AddBinding("<Gamepad>/leftTrigger");
```
ekcoh: I do not understand why this binding is added programatically instead of being part of the json string? Is it relevant for the test case? Otherwise consider making it part of the string (data).

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+        Assert.That(action, Is.Not.Null, "Expected Move action to exist.");
```
ekcoh: This check is not needed and can be removed. The test case would result in exception on line #146 if it was null. I suggest removing this line, the test will throw if action is not there. Otherwise move it before AddBinding.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+    
+        var parsed = UnityEngine.InputSystem.Utilities.NameAndParameters.ParseMultiple(action.processors).ToList();
+        Assert.That(parsed.Count, Is.EqualTo(3), "Expected three processors.");
```
ekcoh: You check that json contains 3 processors but only two of them are checked below, why?

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+        Assert.That(parsed[0].name, Is.EqualTo("StickDeadzone"));
+        Assert.That(parsed[1].name, Is.EqualTo("InvertVector2"));
+        
+        map.ResolveBindings();
```
ekcoh: Is this relevant for this test?

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
...
+        
+        var invertX = parsed[1].parameters.FirstOrDefault(p => p.name == "invertX");
+        Assert.That(invertX.name, Is.EqualTo("invertX"));
+        Assert.That(invertX.value.ToBoolean(), Is.False);
+
+        Assert.That(map.m_State.processors[2].GetType(), Is.EqualTo(typeof(CustomProcessor)));
```
ekcoh: Why check this via action state when you check the others only by name? If there is a motivation for it please document inline

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
 
         yield return null;
     }
+    
+    [Test]
+    public void Migration_ShouldProduceValidActionAsset_WithEnumProcessorConverted()
```
ekcoh, line 115: Can you document what the intent of this test case is inline in code? 

My interpretation is that you want to verify that processors defined in a certain way in .inputaction json are parsed/preserved correctly when imported (with migration)? If this is the case I suggest you convert this into a functional test instead of doing a lot of string comparisons. Here is a rough example (coded directly into web interface so it may contain errors but hope you get the idea):

```cs
const string legacyJson = "<as before>"; // Consider including binding in JSON to avoid writing code for it
var asset = InputActionAsset.FromJson(legacyJson);
var gamepad = InputSystem.AddDevice<Gamepad>();
var action = asset.FindAction("Player/Move");
action.Enable();

// Make sure deadzone processor is applied
Set(gamepad.leftStick, new Vector2(InputSystem.settings.defaultDeadzoneMin * 0.9f, 0.0f));
Assert.That(action.ReadValue<Vector2>(), Is.Equal(Vecto2.zero)); // Filtered by deadzone

// Make sure invert processor is applied
Set(gamepad.leftStick, new Vector2(1.0f, 0.0f));
Assert.That(action.ReadValue<Vector2>(), Is.Equal(-Vector2.right)); // Inverted

// Make sure custom processor is applied
Set(gamepad.leftStick, new Vector2(0.5f, 0.0f));
Assert.That(action.ReadValue<Vector2>(), Is.Equal(something)); // Not sure how custom should change value, so change this to what is appropriate
```

Ideally order dependency can be proven through the same test case.

ekcoh (changes requested):
Thanks for iterating on this and good to see it gets simpler and simpler. The test case still confuses me though. Providing a pseudo example of what I consider to be a stronger test.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
```diff
@@ -108,5 +108,57 @@ public IEnumerator ProcessorEnum_ShouldSerializeByValue_WhenSerializedToAsset()
 
         yield return null;
     }
+    
+    [Test]
```
ekcoh, line 113: Please add Description attribute here with a link to the public issue tracker to illustrate its a regression test case.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
AswinRajGopal: Sure I'll change it into a functional test.

ekcoh: Do you see any pros/cons with changing approach? Would such a functional test miss some aspect of this?

AswinRajGopal: My view about authoring a regression test is to cover only the migration case where we actually verify that a JSON with a custom processor and some neighboring processors like stickdeadzone or invertVector are parsed properly after migration. 

The functional tests already exists which covers most of the functionals in the input system generic tests. However it will be robust to club both of these for this case.

ekcoh: OK, that is also fine so keep this if you want. Not sure @LeoUnity has additional perspective?
Not sure we already test this specific processor order somewhere else?

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
AswinRajGopal: I've replaced this with enabling the associated action from the asset. So that it will resolve all of its binding so that we will be able to access the processors.

Assets/Tests/InputSystem.Editor/CustomProcessorEnumTest.cs
AswinRajGopal: Have now checked all of the processors via action state.

LeoUnity (approved):

LeoUnity (approved):
PR looks good, there are a few failing CI jobs.

ekcoh (approved):

ericksson:
The Android jobs are broken and stuck - see https://unity.slack.com/archives/C94RMJJ5T/p1760644095487229

ekcoh:

ekcoh:
Added @Pauliusd01 for QA approval

Pauliusd01:
Did not forget this one, looking at it today/tomorrow

Pauliusd01 (changes requested):
This doesn't seem to fix assets that were already broken by the regression. Is there a way you can fix them as well? If not is there a workaround we can offer? I tried downgrading and then upgrading back to your version but that doesn't fix it. 

@ekcoh @LeoUnity do you have any opinions about this?

AswinRajGopal:
We can bump the version so that the migration will be invoked again. Will wait for Leo and Hakan’s suggestion.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

LeoUnity:
I didn't consider that, I'd say let's land this as it is and do a release ASAP to reduce the window in which users can end up with a corrupted asset.

There is something we could do, but if nobody complains I'm not sure it's worth it. Worst case scenario users can edit the file manually and correct the ";" to ",".

Pauliusd01:
> I didn't consider that, I'd say let's land this as it is and do a release ASAP to reduce the window in which users can end up with a corrupted asset.
> 
> There is something we could do, but if nobody complains I'm not sure it's worth it. Worst case scenario users can edit the file manually and correct the ";" to ",".

Okay that does work, I'll mention this as a known issue on the next release

Pauliusd01 (approved):
LGTM, tried upgrading from an older version to this one and checked basic input action editor and player functionality.

PR #2245, opened 2025-09-25
FIX: Use correct data format for composite stick axes in HID layout builder

myss:
The generic HID layout builder made a faulty assumption that the X and Y axes of a composite "Stick" control always used a simple bitfield format (`SBIT`/`BIT`).

This assumption is incorrect for many modern devices, most notably 6DoF controllers like the 3Dconnexion SpaceMouse, which report their axes as 16-bit shorts (`SHRT`). This bug caused the builder to apply the wrong data format, resulting in incorrect parsing and unusable stick input for these devices.

This commit resolves the issue by replacing the brittle, hardcoded format logic with a call to the `element.DetermineFormat()` helper method. This is the same robust method used by the general-purpose element processor later in the same function.

This change:
- Fixes axis detection for 6DoF controllers and other HID devices that use `SHRT` or `BYTE` formats.
- Makes the special-case "Stick" logic consistent with the rest of the HID parser.

### Description

Fix HID handling for 3dconnexion SpaceMouse. Without the fix, the "/stick", "/stick/x" and "/stick/y" values contained garbage.

### Testing status & QA

Only tested on a 3dconnexion SpaceMouse Wireless.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - JIRA ticket linked, example ([case %<ID>%](https://issuetracker.unity3d.com/product/unity/issues/guid/<ID>)). If it is a private issue, just add the case ID without a link.
    - Jira port for the next release set as "Resolved".
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [ ] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

- [ ] Create forward/backward port if needed. If you are blocked from creating a forward port now please add a task to ISX-1444.

PR #2246, opened 2025-09-26
FIX: Incorrect HID Stick values for LogicalMinimum with negative values (ISXB-1735)

jfreire-unity:
### Description

This PR addresses the issues encountered when parsing data for devices that have LogicalMinimum values with negative values. Seems we weren't taking 2's complement values into account. Ticket: https://jira.unity3d.com/browse/ISXB-1735

When parsing, we are now casting the bytes read to signed values. It also fixes an error when shifting bits to construct a 4-byte value.

Also includes fixes from user PR #2245 

### Testing status & QA

Tested with a Gamepad stick which has both signed short and byte for LogicalMin/Max of [-127,127] and [-32.768,32.767]

### Overall Product Risks

- Complexity: 0
- Halo Effect: 1

### Comments to reviewers

This is not easy to test without real devices that we don't have a layout for so I at least hope existing devices would still work.

One strategy could be to test this on macOS and comment this line: https://github.com/Unity-Technologies/InputSystem/blob/d1d72d50aa4c9c75e1c4055965e2ba9d9455605e/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs#L3798
Then look if the sticks were well interpreted.

However, this only allows us to verify for devices that send Stick values as a signed short. For signed bytes I don't know of a device besides the ones users have reported in Discussions https://discussions.unity.com/t/input-system-reading-invalid-values-from-hall-effect-keyboards/1684840/3

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Tests/InputSystem/Plugins/HIDTests.cs
+++ b/Assets/Tests/InputSystem/Plugins/HIDTests.cs
@@ -317,12 +317,26 @@ internal class HIDTests : CoreTestsFixture
 
     [Test]
     [Category("HID Devices")]
-    public void Devices_CanCreateGenericHID_WithSignedLogicalMinAndMaxSticks()
+
+    // These descriptor values were generated with the Microsoft HID Authoring descriptor tool in
+    // https://github.com/microsoft/hidtools for the expexted values.
+    // Logical min 0, logical max 65535
+    [TestCase(16, new byte[] {0x16, 0x00, 0x00}, new byte[] { 0x27, 0xFF, 0xFF, 0x00, 0x00 }, 0, 65535, 0.01f)]
+    // Logical min -32768, logical max 32767
+    [TestCase(16, new byte[] {0x16, 0x00, 0x80}, new byte[] {0x26, 0xFF, 0x7F}, -32768, 32767, 0.01f)]
+    // Logical min 0, logical max 255
+    [TestCase(8, new byte[] {0x15, 00}, new byte[] {0x26, 0xFF, 0x00}, 0, 255, 0.01f)]
+    // Logical min -128, logical max 127
+    [TestCase(8, new byte[] {0x15, 0x80}, new byte[] {0x25, 0x7F}, -128, 127, 0.01f)]
+    // Logical min -16, logical max 15 (below 8 bit boundary)
+    [TestCase(5, new byte[] {0x15, 0xF0}, new byte[] {0x25, 0x0F}, -16, 15, 0)]
+    // Logical min 0, logical max 31 (below 8 bit boundary)
+    [TestCase(5, new byte[] {0x15, 0x00}, new byte[] {0x25, 0x1F}, 0, 31, 0)]
+    public void Devices_CanParseHIDDescritpor_WithSignedLogicalMinAndMaxSticks(byte reportSizeBits, byte[] logicalMinBytes, byte[] logicalMaxBytes, int logicalMinExpected, int logicalMaxExpected, float errorMargin)
     {
-        // This is a HID report descriptor for a simple device with two analog sticks;
-        // Similar to a user that reported an issue in Discussions:
-        // https://discussions.unity.com/t/input-system-reading-invalid-values-from-hall-effect-keyboards/1684840/3
-        var reportDescriptor = new byte[]
+        // Dynamically create HID report descriptor for two analog sticks with parameterized logical min/max
+
+        var reportDescriptorStart = new byte[]
         {
             0x05, 0x01,        // Usage Page (Generic Desktop Ctrls)
             0x09, 0x05,        // Usage (Game Pad)
@@ -330,14 +344,22 @@ internal class HIDTests : CoreTestsFixture
             0x85, 0x01,        //   Report ID (1)
             0x05, 0x01,        //   Usage Page (Generic Desktop Ctrls)
             0x09, 0x30,        //   Usage (X)
-            0x09, 0x31,        //   Usage (Y)
-            0x15, 0x81,        //   Logical Minimum (-127)
-            0x25, 0x7F,        //   Logical Maximum (127)
-            0x75, 0x08,        //   Report Size (8)
-            0x95, 0x02,        //   Report Count (2)
-            0x81, 0x02,        //   Input (Data,Var,Abs)
-            0xC0,              // End Collection
         };
+
+        var reportDescriptorEnd = new byte[]
+        {
+            0x75, reportSizeBits,   //   Report Size (8)
+            0x95, 0x01,             //   Report Count (1)
+            0x81, 0x02,             //   Input (Data,Var,Abs)
+            0xC0,                   //   End Collection
+        };
+
+        // Concatenate to form final descriptor based on test parameters where logical min/max bytes
+        // are inserted in the middle.
+        var reportDescriptor = reportDescriptorStart.Concat(logicalMinBytes).
+            Concat(logicalMaxBytes).
+            Concat(reportDescriptorEnd).
+            ToArray();
 
         // The HID report descriptor is fetched from the device via an IOCTL.
         var deviceId = runtime.AllocateDeviceId();
@@ -350,7 +372,7 @@ internal class HIDTests : CoreTestsFixture
             new InputDeviceDescription
             {
                 interfaceName = HID.kHIDInterface,
-                manufacturer = "TestVendor",
+                manufacturer = "TestLogicalMinMaxParsing",
                 product = "TestHID",
                 capabilities = new HID.HIDDeviceDescriptor
                 {
@@ -365,35 +387,22 @@ internal class HIDTests : CoreTestsFixture
         Assert.That(device, Is.Not.Null);
         Assert.That(device, Is.TypeOf<Joystick>());
 
-        // Stick vector 2 should be centered at (0,0).
+        var parsedDescriptor = JsonUtility.FromJson<HID.HIDDeviceDescriptor>(device.description.capabilities);
+
+        // Check we parsed the values as expected
+        foreach (var element in parsedDescriptor.elements)
+        {
+            if (element.usage == (int)HID.GenericDesktop.X)
+            {
+                Assert.That(element.logicalMin, Is.EqualTo(logicalMinExpected));
+                Assert.That(element.logicalMax, Is.EqualTo(logicalMaxExpected));
+            }
+            else
+                Assert.Fail("Could not find X and Y elements in descriptor");
+        }
+
+        // Stick vector 2 should be centered at (0,0) when initialized
         Assert.That(device.stick.ReadValue(), Is.EqualTo(new Vector2(0f, 0f)).Using(Vector2EqualityComparer.Instance));
-
-        // Queue event with stick pushed to bottom. We assume Y axis is inverted by default in HID devices.
-        // See HID.HIDElementDescriptor.DetermineParameters()
-        InputSystem.QueueStateEvent(device, new SimpleJoystickLayoutWithStickByte { reportId = 1, x = 0, y = 127 });
-        InputSystem.Update();
-
-        Assert.That(device.stick.ReadValue() , Is.EqualTo(new Vector2(0f, -1f)).Using(Vector2EqualityComparer.Instance));
-
-        InputSystem.QueueStateEvent(device, new SimpleJoystickLayoutWithStickByte { reportId = 1, x = 0, y = -127 });
-        InputSystem.Update();
-
-        Assert.That(device.stick.ReadValue(), Is.EqualTo(new Vector2(0f, 1f)).Using(Vector2EqualityComparer.Instance));
-
-        InputSystem.QueueStateEvent(device, new SimpleJoystickLayoutWithStickByte { reportId = 1, x = 127, y = 0 });
-        InputSystem.Update();
-
-        Assert.That(device.stick.ReadValue() , Is.EqualTo(new Vector2(1f, 0f)).Using(Vector2EqualityComparer.Instance));
-
-        InputSystem.QueueStateEvent(device, new SimpleJoystickLayoutWithStickByte { reportId = 1, x = -127, y = 0 });
-        InputSystem.Update();
-
-        Assert.That(device.stick.ReadValue(), Is.EqualTo(new Vector2(-1f, 0f)).Using(Vector2EqualityComparer.Instance));
-
-        InputSystem.QueueStateEvent(device, new SimpleJoystickLayoutWithStickByte { reportId = 1, x = 127, y = 127 });
-        InputSystem.Update();
-
-        Assert.That(device.stick.ReadValue(), Is.EqualTo(new Vector2(0.7071f, -0.7071f)).Using(Vector2EqualityComparer.Instance));
     }
 
     [Test]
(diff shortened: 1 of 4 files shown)


Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -1026,7 +1113,7 @@ public void Devices_GenericHIDConvertsXAndYUsagesToStickControl()
     }
 
     [StructLayout(LayoutKind.Explicit)]
-    struct SimpleJoystickLayout : IInputStateTypeInfo
+    struct SimpleJoystickLayoutWithStickUshort : IInputStateTypeInfo
```
jfreire-unity: Need to create a test with a report descriptor that fits this struct as well.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
```diff
@@ -257,6 +257,7 @@ internal static unsafe HIDDeviceDescriptor ReadHIDDeviceDescriptor(ref InputDevi
 
                     // Update the descriptor on the device with the information we got.
                     deviceDescription.capabilities = hidDeviceDescriptor.ToJson();
+                    Debug.Log($"Parsing HID descriptor from JSON for device '{deviceDescription.capabilities}'");
```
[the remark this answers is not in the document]
jfreire-unity: Got it, will remove.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
```diff
@@ -394,7 +394,7 @@ public InputControlLayout Build()
                         .WithProcessors(xElement.DetermineProcessors());
 
                     builder.AddControl(stickName + "/y")
-                        .WithFormat(yElement.isSigned ? InputStateBlock.FormatSBit : InputStateBlock.FormatBit)
+                        .WithFormat(yElement.DetermineFormat())
```
jfreire-unity, line 397: This change was included from #2245

ekcoh: Isn't it required to always determine format for any control? Also buttons support this  (but I cannot recall ever seeing it in practise), but definitely e.g. triggers go into same territory

Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -309,6 +291,111 @@ public void Devices_CanCreateGenericHID_FromDeviceWithBinaryReportDescriptor()
...
+        var reportDescriptor = new byte[]
+        {
+            0x05, 0x01,        // Usage Page (Generic Desktop Ctrls)
+            0x09, 0x05,        // Usage (Game Pad)
+            0xA1, 0x01,        // Collection (Application)
+            0x85, 0x01,        //   Report ID (1)
+            0x05, 0x01,        //   Usage Page (Generic Desktop Ctrls)
+            0x09, 0x30,        //   Usage (X)
+            0x09, 0x31,        //   Usage (Y)
+            0x15, 0x81,        //   Logical Minimum (-127)
```
ekcoh: Happy to see a test with an actual report! Great.
I would recommend we parameterise this this test so that we have logical minimum ranges as below. Range can be fixed e.g 127+127+1 different values = 255 in this example. It might also be good to test with range 2^8 =256 256 since it is quite common to utilise all the possible values and create a symmetry challenge instead. Hence, recommend:

// Range 255
[-127, 127] (currently tested)
[-254, 0]
[0, 254]
[0, 254]
[-381, -127]
[127, 382]

// range 256
[-127, 128]
[-255, 0]
[0, 255]
[0, 255]
[-382, -127]
[128, 384]

I think you get the idea, both sides of zero, spanning zero or shifted to all negative or all positive. It should smoke out any problems with the function handling the sign extension which could be tested in isolation and doesn't really need a full report test.

Still great you added tests for full parsing.

ekcoh (commented):
Thanks for looking into this. I am confused about our ReadData method, it is not yet clear to me how we handle sign-extended types of arbitrary bit sizes? E.g.
It's completely valid to have e.g. X with logical min -15 logical max 16, and hence Report size 5

Assets/Tests/InputSystem/Plugins/HIDTests.cs
ekcoh: What I am after is that shifting logical range should change nothing

jfreire-unity: Ah, good suggestion. I'll improve that.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
jfreire-unity: Yes, and we also did it before. Not sure why assigned it the SBit/Bit format before but we do read multiple "bits" if we had them (e.g. in **[stateBlock.ReadFloat()](https://github.com/Unity-Technologies/InputSystem/blob/fbb9034628473955f1d546693d6d0773b2d492a0/Packages/com.unity.inputsystem/InputSystem/State/InputStateBlock.cs#L384)).** So it might have been to deal with cases where we could use like "10 bits" to define an axis. I'll have a look and test this as well to see if there are any other changes needed. Thanks.

ekcoh:
I see CHANGELOG.md have not been updated yet, but if this PR picks up contribution work from [@myss](https://github.com/myss) I suggest closing that PR: https://github.com/Unity-Technologies/InputSystem/pull/2245, mentioned we re-reroute it via this PR and make sure to credit [@myss](https://github.com/myss) in the CHANGELOG.md similar to previous user-contributions.

Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -309,6 +291,120 @@ public void Devices_CanCreateGenericHID_FromDeviceWithBinaryReportDescriptor()
         ////TODO: check hat switch
     }
 
+    // This is used to mock out the IOCTL the HID device driver would use to return
```
ekcoh, line 294: Not sure I understand the "mock out" part mentioned in comment here? Isn't this just a refactor of the previous code with a dedicated unsafe method?

Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -309,6 +291,120 @@ public void Devices_CanCreateGenericHID_FromDeviceWithBinaryReportDescriptor()
...
+    // the report descriptor and its size.
+    unsafe void SetDeviceCommandCallbackToReturnReportDescriptor(int deviceId, byte[] reportDescriptor)
+    {
+        runtime.SetDeviceCommandCallback(deviceId,
+            (id, commandPtr) =>
+            {
+                if (commandPtr->type == HID.QueryHIDReportDescriptorSizeDeviceCommandType)
```
ekcoh, line 304: Note sure commandPtr is guaranteed to not be null here?

Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -309,6 +291,120 @@ public void Devices_CanCreateGenericHID_FromDeviceWithBinaryReportDescriptor()
...
+    [TestCase(16, new byte[] {0x16, 0x00, 0x80}, new byte[] {0x26, 0xFF, 0x7F}, -32768, 32767, 0.01f)]
+    // Logical min 0, logical max 255
+    [TestCase(8, new byte[] {0x15, 00}, new byte[] {0x26, 0xFF, 0x00}, 0, 255, 0.01f)]
+    // Logical min -128, logical max 127
+    [TestCase(8, new byte[] {0x15, 0x80}, new byte[] {0x25, 0x7F}, -128, 127, 0.01f)]
+    // Logical min -16, logical max 15 (below 8 bit boundary)
+    [TestCase(5, new byte[] {0x15, 0xF0}, new byte[] {0x25, 0x0F}, -16, 15, 0)]
+    // Logical min 0, logical max 31 (below 8 bit boundary)
+    [TestCase(5, new byte[] {0x15, 0x00}, new byte[] {0x25, 0x1F}, 0, 31, 0)]
+    public void Devices_CanParseHIDDescritpor_WithSignedLogicalMinAndMaxSticks(byte reportSizeBits, byte[] logicalMinBytes, byte[] logicalMaxBytes, int logicalMinExpected, int logicalMaxExpected, float errorMargin)
```
ekcoh: Do we need to pass errorMargin into test case? Isn't this given by 1 / (numerical logical range) + float.epsilon?

ekcoh: (If reinterpreted as floating point)

ekcoh: Would it make sense to have an odd test case above 8 bit boundary, e.g. 13 bits?

ekcoh (changes requested):
Generally looks very good, missing CHANGELOG.md entry it sounds like on your comment you intend to add more tests? Some minor comments.

Assets/Tests/InputSystem/Plugins/HIDTests.cs
jfreire-unity: Yes, it's just a refactor into a unsafe method since we're now using it multiple times. I can change the language if "mock" is not ideal but at least is the way I see it, since we're not actually getting anything from the platform layer in these tests.

Assets/Tests/InputSystem/Plugins/HIDTests.cs
jfreire-unity: Sure, I'll add a case for it.

Assets/Tests/InputSystem/Plugins/HIDTests.cs
jfreire-unity: No, it was used in my previous testing of tests and was a leftover. Removed.

jfreire-unity: Done!

jfreire-unity:
> Generally looks very good, missing CHANGELOG.md entry it sounds like on your comment you intend to add more tests? Some minor comments.

Thanks! Added changelog now and referenced the User PR in it. Feel free to do suggestions if it's not good enough

ericksson:
Cancelling all jobs running on this branch due to Android jobs being broken and stuck - see https://unity.slack.com/archives/C94RMJJ5T/p1760644095487229

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HID.cs
jfreire-unity: Our SBit reading is broken any way, as we are not doing bit sign extension. I'll fix that in a follow-up PR.

jfreire-unity:
@Pauliusd01 @ekcoh this is ready for final approval.

@ekcoh your Space Mouse test might probably be enough as I think it also has Logical Minimum and Maximum with signed values.

Best way to validate is to run the tests I added in `develop` to see that they fail.
I validated this with a real custom HID gamepad device on macOS so I'm not sure the Xbox One approach works.

Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -309,6 +291,130 @@ public void Devices_CanCreateGenericHID_FromDeviceWithBinaryReportDescriptor()
...
+                        UnsafeUtility.MemCpy(commandPtr->payloadPtr, ptr, reportDescriptor.Length);
+                        return reportDescriptor.Length;
+                    }
+                }
+
+                return InputDeviceCommand.GenericFailure;
+            });
+    }
+
+    [Test]
```
ekcoh, line 321: Still think it would be great to have a test case here for a value that only lies on negative side since that is missing from the test set [-72, -37]. Similarly the test cases to do not have any case spanning only the positive side, e.g. [+21, +38]. All values here are either symmetric around zero or start at zero.

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HIDParser.cs
```diff
@@ -277,7 +277,7 @@ private unsafe static int ReadData(int itemSize, byte* currentPtr, byte* endPtr)
                     return 0;
                 var data1 = *currentPtr;
                 var data2 = *(currentPtr + 1);
-                return (data2 << 8) | data1;
+                return (short)((data2 << 8) | data1);
```
ekcoh, line 280: I guess all of these yields interesting values when all bits are value bits, e.g. source data is unsigned? E.g. If source values has logical min 0, logical max 255 and has 8 value bits, this will return -127 when fully actuated? Since ReadData is signed this seems unavoidable but maybe something we should have a one-liner doc about? 
Let me know if this is incorrect. E.g. I have an old gamepad where usage id X is exactly the values above.

ekcoh: To be clear: I am not looking for implementing unsigned ReadValue here, rather that we create awareness about this behaviour if I am not mistaken here.

ekcoh (approved):
Approving this since PR looks great and fixes a critical issue. Would however suggest we fix this before landing: 
- Include negative range TestCase since it's currently not tested. (Both min and max negative)
- Positive range descriptor TestCase on added test where max value use all bits as an unsigned value to at least capture what is happening, e.g. overflow to negative? See comment for details.
- If overflowing when no bits are sign bits, make sure we document this behaviour (ignore if I am wrong, but I cannot see how it would work)

Pauliusd01:
Will take a look today

Packages/com.unity.inputsystem/InputSystem/Plugins/HID/HIDParser.cs
jfreire-unity: > this will return -127 when fully actuated?

Thats a good point. 
But that will not happen. This part deals with interpreting report descriptors data. What I've read and seen is that descriptors that provide logical min 0, logical max 255, have the logical max set to 2 bytes, so that the sign bit is in the MSB of the 2 byte value.
So, 255 for a logical max comes represented in the descriptor as 0xFF 0x00. I used an [HID descriptor generator from the HID spec to validate this.](https://github.com/microsoft/hidtools). I probably should make a small doc on how to use it for posterity as well.

This is of course separate from what we read in input reports, which is a 8 bit value. We then use the logical min and max to convert to the 8 bit value in the range of 0 and 255. But this code here is not dealing with that. And as you mentioned, this is done at some point inside ReadValue of an InputControl.

ekcoh:

Assets/Tests/InputSystem/Plugins/HIDTests.cs
jfreire-unity: Added. For parsing is not a problem but reading values for all positive logical min max is. I created a ticket to track it    https://jira.unity3d.com/browse/ISXB-1737

Pauliusd01 (approved):
LGTM, what's checked: ran the tests locally and checked sony and xbox gamepad functionality on playmode and player (Mac & Windows)

Assets/Tests/InputSystem/Plugins/HIDTests.cs
```diff
@@ -408,11 +411,8 @@ public void Devices_CanParseHIDDescritpor_WithSignedLogicalMinAndMaxSticks(byte
                 Assert.That(element.logicalMax, Is.EqualTo(logicalMaxExpected));
             }
             else
-                Assert.Fail("Could not find X and Y elements in descriptor");
+                Assert.Fail("Could not find X element in descriptor");
         }
-
-        // Stick vector 2 should be centered at (0,0) when initialized
-        Assert.That(device.stick.ReadValue(), Is.EqualTo(new Vector2(0f, 0f)).Using(Vector2EqualityComparer.Instance));
```
jfreire-unity: I removed this to limit this test to parsing the report descriptor logical values.  We have other problems reported in https://jira.unity3d.com/browse/ISXB-1737 when reading values and this would expand the scope of the test.

This is basically because we only seem to support values centered around 0 in the report size bit value range. Those are the only ones being read properly.

PR #1949, opened 2024-06-18
FIX: InputActionReference when using FEPM (ISX-1968)

timkeo:
### Description

When Domain Reloads are disabled, InputActionReference instances continue to reference the "old" InputAction object from the previous PlayMode session. This fix clears the InputAction reference when exiting PlayMode allowing it to be reloaded in the next session.

Although Configurable PlayMode options, aka Fast Enter PlayMode, isn't currently supported, a number of tickets have been filed against this issue including:

- [ISXB-601](https://jira.unity3d.com/browse/ISXB-601)
- [ISXB-718](https://jira.unity3d.com/browse/ISXB-718)
- [ISXB-900](https://jira.unity3d.com/browse/ISXB-900)
- [ISXB-469](https://jira.unity3d.com/browse/ISXB-469)
- [ISXB-598](https://jira.unity3d.com/browse/ISXB-598)

indicating FEPM is quite a popular feature and fixing this issue is beneficial to our customers.

Furthermore, the underlying problem is a bug in its own right regardless of FEPM: the underlying Action instance is invalid for EditMode and holding the reference is dangerous as it keeps the object "alive" along with the pointers to native memory buffers that are de-allocated when exiting PlayMode. In short, attempting to access the ActionReference after leaving PlayMode can cause an AV and crash the Editor.

### Changes made

When exiting PlayMode, all `InputActionReference` instances are queried and their `m_Action` field is reset. This forces a "reload" of the Action from the Asset the next time its accessed just as if a Domain Reload had occured.

### Notes

This issue is completely separate from my CoreCLR refactor and can be fixed independently of it.

### Checklist

Before review:

- [x] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - FogBugz ticket attached, example `([case %number%](https://issuetracker.unity3d.com/issues/...))`.
    - FogBugz is marked as "Resolved" with *next* release version correctly set.
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [x] Commit message for squash-merge is prefixed with one of the list:
[...]

PR #2248, opened 2025-09-29
FIX: ISXB-1584, ISXB-1688, ISXB-1692, ISXB-1693, ISXB-1694, ISXB-1701 InputActionReference bug fixes and improved test coverage.

ekcoh:
### Description

**Fixes the following issues:** 
- https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1584, where `InputActionReference.action` would return an incorrect (cached) action if set programmatically after the cached instance had been resolved.
- ISXB-1688 - `InputActionSetupExtensions.RemoveAction(action)` throws exception if attempting to remove an action that doesn't have any bindings.
- ISXB-1692 - Assigning an InputActionReference field stored in an .inputactions asset via `InputActionReference.Set` in playmode corrupts the referenced asset. 
- ISXB-1693 - `InputActionReference.Create(null)` returns null instead "of a reference object to the given action". If its desirable to assign a reference to null when action is null, one should not call Create to begin with.
- ISXB-1694 - `InputActionReference.Set(null)` does not update the cached input action reference nor the `ScriptableObject.name` leading to incorrect action being returned and reference showing up as the path to the previously assigned action in Inspector.
- ~~ISXB-1699 - `InputActionAsset.RemoveActionMap` doesn't remove actions within the map, leaving them with stale references back to the map.~~ This was a misinterpretation and have been reverted. Ticket marked "As Designed".
- ISXB-1701 - Inspector field of type `InputActionReference` isn't reflecting broken reference in case action is deleted from asset, action map is deleted from asset or the asset itself is deleted. 

**Revisiting/altering fixes of the previous tickets:**
- ISX-1968 "Investigate InputActionReferences problem with FEPM/CoreCLR"
- https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-900 "Action Map Must Be Contained in State" and other errors are thrown when re-entering Play mode without a domain reload, while using temporarily created InputBindings. #1949 

The above revisited tickets reference the problem of InputActionReference run-time objects not being reset when exiting play-mode with domain reloads disabled. To get this scenario covered, this scenario is now included in the added editor tests. Tested both with and without domain reloads enabled.

**Additional changes made**:
- Simplified code in InputSystemProvider.cs (UITK integration) since there is no reason to use `InputActionReference` in that class as long as there is no UI to edit it implemented. If such a UI is added later it can easily be converted back to `InputActionReference`.
- Renamed function arguments in non-public functions of `InputActionReference` to eliminate name-shadowing warnings. Didn't touch public ones since a parameter rename would be unexpected after API publishing.
- Simplified loading code of `InputActionReference` instances from assets within `InputActionImporter`.
- Adds `EditorPrefsTestUtils.cs` which is a test utility to simplify editor tests.
- Adds `CoreTests_Actions_References.cs` which extracts previous and new `InputActionReference` play-mode tests into their own (partial) file.
- Adds `DumpInputActionReferences.cs` which extends "QA Tools" with ability to dump a list of all input action references (loaded + assets) into the console via QA Tools Unity menu.

**Open issues/questions to reviewers:**
- Property drawer for input action references is currently conditionally compiled depending on project-wide input actions. This makes little sense now, but if the check is removed we get a dependency on search context that isn't available in that version, so we might want to land this post 2021.3 EoL or add a similar fix for 2021.3? (Currently missing on PR)

### Testing status & QA

Added a bunch of functional tests that replicates the problematic scenarios mentioned above. Both play-mode functional tests as well as edit-mode as well as transitional editor-mode to play-mode and play-mode to edit-mode editor tests were added.

Removed some code related to project-wide-actions in property drawer I believe @jfreire-unity added. I do not think it is needed as a special case anymore and I also think it was incorrect to change the users objects.

Altered a previous play-mode fix by @timkeo but it is essentially intact to previous behavior. I think there is opportunity to change this later to only use Object.FindObjectsByType(true) instead of Resources.FindObjectsByType() as an optimisation after this has landed?

Haven''t tested on 2021.3 which might be behaving differently but I do not know if we should bother since it is soon EoL.

### Overall Product Risks

- Complexity: Medium
- Halo Effect: Large

InputActionReferences are used all over the place so this is a serious bug that can have impact, especially on more complex projects.

### Comments to reviewers

Recommend double checking the repro project with this fix.

### Checklist

Before review:

- [X] Changelog entry added.
    - Functional tests added: `Actions_CanResolveActionReferenceAndThenSetItToAnotherAction`
    - Performance tests. - None
    - Integration tests. - None
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


ekcoh:
This solves the incorrect return value, but it seems it is not enough to solve the reported problem so moving this back to draft mode for more work.

timkeo (approved):
Reviewed the change to my bug fix and skimmed though some of the other changes.

Assets/Editor/DumpInputActionReferences.cs
```diff
@@ -0,0 +1,33 @@
...
+        DumpReferences(sb, "All objects:", Resources.FindObjectsOfTypeAll<InputActionReference>());
+        Debug.Log(sb.ToString());
+    }
+
+    [UnityEditor.MenuItem("QA Tools/Dump Input Action References to Console", false, 100)]
+    private static void Dump()
+    {
+        DumpReferences();
+    }
+}
```
K-Tone, line 33: Should this be part of our Input Debugger interface instead?

ekcoh: Maybe, right now I just put it into our QA asset scripts since I did it to investigate bugs and felt it had more value to keep than removing it, but maybe someone would be interested in it. I would say there is no obvious user-value in it, but maybe for ourselves. However, you likely want snapshots and this was simple and sufficient to solve such problems. I do not intend to move it there as part of this PR at least.

ekcoh: However its handy to be able to see which InputActionReferences exist in memory and on disc as serialised SOs

Pauliusd01:
Seeing some new warnings that don't show up in develop:

image

ekcoh:
> Seeing some new warnings that don't show up in develop:
> 
> image

Good catch, I implemented it with 2023.3 so should check newer which I assume this is.

jfreire-unity:
> Seeing some new warnings that don't show up in develop:
> 
> image

Yes, also saw this, so I will wait until it's fixed before approving.

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
```diff
@@ -12,37 +13,53 @@ namespace UnityEngine.InputSystem.Editor
...
+                property.objectReferenceValue = reference ?
+                    InputActionReference.Create(reference.action) : null;
             }
+
+            EditorGUI.EndProperty();
+        }
+
+        public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+        {
+            OnGUI(position, property, label);
```
jfreire-unity, line 75: Any reason to factor out the code to call a single function in this override?

Assets/Tests/InputSystem.Editor/EditorPrefsTestUtils.cs
```diff
@@ -0,0 +1,76 @@
+using UnityEditor;
+
+namespace Tests.InputSystem.Editor
+{
+    /// <summary>
+    /// Utility to simplify editor tests with respect to editor preferences.
+    /// </summary>
+    internal static class EditorPrefsTestUtils
```
jfreire-unity, line 8: 👍

ekcoh:
@jfreire-unity @Pauliusd01 I think I have eliminated the warnings on this branch now. The two key warnings must be present on develop and should be fixed separately. While fixing I realised the associated code could be simplified.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
```diff
@@ -227,10 +241,65 @@ internal static void ResetCachedAction()
...
+            {
+                // "Immutable" input action references are always sub-assets of InputActionAsset.
+                var isSubAsset = UnityEditor.AssetDatabase.IsSubAsset(reference);
+                if (!isSubAsset)
+                    return true;
+
+                // If we cannot get the path of our reference, we cannot be a persisted asset within an InputActionAsset.
+                var path = UnityEditor.AssetDatabase.GetAssetPath(reference);
+                if (path == null)
+                    return true;
```
jfreire-unity, line 279: Seems reasonable to have code coverage of this case? See all the 3 codecov warnings in this file (not sure if they are false positives as well).

ekcoh:
I guess we should investigate if we can have an automated test that could fail if any warning is generated that maps to our assemblies.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
```diff
@@ -118,26 +127,24 @@ public void Set(InputActionAsset asset, string mapName, string actionName)
...
         {
-            var actionMap = action.actionMap;
-            if (!asset.actionMaps.Contains(actionMap))
-                throw new ArgumentException(
-                    $"Action '{action}' is not contained in asset '{asset}'", nameof(action));
-
-            m_Asset = asset;
-            m_ActionId = action.id.ToString();
-            name = GetDisplayName(action);
+            CheckImmutableReference();
```
jfreire-unity, line 140: Are we now considering InputActionReference immutables during runtime (PlayMode/Standalone) and so Set can't be used?

jfreire-unity (commented):
Sorry to take so long to review. 

1. I'm not able to fix the main issue (ISXB-1584) as it now raises an error. I also don't understand what should I do based on the exception logged message. 
2. Does this mean that `InputActionReference.Set()` cannot be run at in Play-mode now?
3. There's no guidance in the PR description to make the main bug work be fixed with this change. That could have been helpful.
 
4. I'm much longer to review because there are a lot of things we're fixing in this PR and I'm not familiar with the impact of all the changes. 
 
 The only think I have to point out is that to me is that the changes raising an error for ISXB-1584 are a relatively impactfull breaking change, unless I'm seeing this wrong. 
 If I'm indeed correct, then let's just make this a bit more explicit and add maybe one example in a Discussions post or something. Happy to chat about it.

jfreire-unity:
This is the message I get when opening the ISXB-1584 project in 6.2.0f1 and using the package in this branch:
```
InvalidOperationException: Attempting to mutate an immutable InputActionReference instance. This is not allowed since it would modify the source asset.Instead use InputActionReference.Create(action) to create a new in-memory instance or serialize it as a separate asset if it survive domain reloads.
```

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
ekcoh: Set can be used if the reference is not an asset.

ekcoh: If we allow Set on asset we mutate our sub-asset references inside .inputactions which is a huge problem since the operation will be persistent

ekcoh: Hence

reference = InputActionReference.Create(something);
reference.Set(somethingElse);

is completely fine

ekcoh:
> This is the message I get when opening the ISXB-1584 project in 6.2.0f1 and using the package in this branch:
> 
> ```
> InvalidOperationException: Attempting to mutate an immutable InputActionReference instance. This is not allowed since it would modify the source asset.Instead use InputActionReference.Create(action) to create a new in-memory instance or serialize it as a separate asset if it survive domain reloads.
> ```

Could you elaborate a bit more on the reproduction steps? Just opening the repo root without doing anything?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionReference.cs
ekcoh: Thanks for calling this out, will check to make sure its tested with a reference not in an asset

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
ekcoh: This is just to allow some kind of testability of the property drawer, its not ideal but I couldn't come up with anything better unless you have ideas?

jfreire-unity:
> > This is the message I get when opening the ISXB-1584 project in 6.2.0f1 and using the package in this branch:
> > ```
> > InvalidOperationException: Attempting to mutate an immutable InputActionReference instance. This is not allowed since it would modify the source asset.Instead use InputActionReference.Create(action) to create a new in-memory instance or serialize it as a separate asset if it survive domain reloads.
> > ```
> 
> Could you elaborate a bit more on the reproduction steps? Just opening the repo root without doing anything?

I downloaded the "bug project", used this PR branch, and got that error when entering playmode. I'll test a 3rd time now from scratch

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
ekcoh: I see no way to mock the ObjectField otherwise

ekcoh:
> > > This is the message I get when opening the ISXB-1584 project in 6.2.0f1 and using the package in this branch:
> > > ```
> > > InvalidOperationException: Attempting to mutate an immutable InputActionReference instance. This is not allowed since it would modify the source asset.Instead use InputActionReference.Create(action) to create a new in-memory instance or serialize it as a separate asset if it survive domain reloads.
> > > ```
> > 
> > 
> > Could you elaborate a bit more on the reproduction steps? Just opening the repo root without doing anything?
> 
> I downloaded the "bug project", used this PR branch, and got that error when entering playmode. I'll test a 3rd time now from scratch

I now I understand, sorry, Friday-slowness, yes, the repro scenario cannot be allowed. It corrupts the assets. Try the original one if you are curious and see that after exiting playmode, asset A is containing a reference to a sub-asset in asset B......

ekcoh:
> > > > This is the message I get when opening the ISXB-1584 project in 6.2.0f1 and using the package in this branch:
> > > > ```
> > > > InvalidOperationException: Attempting to mutate an immutable InputActionReference instance. This is not allowed since it would modify the source asset.Instead use InputActionReference.Create(action) to create a new in-memory instance or serialize it as a separate asset if it survive domain reloads.
> > > > ```
> > > 
> > > 
> > > Could you elaborate a bit more on the reproduction steps? Just opening the repo root without doing anything?
> > 
> > 
> > I downloaded the "bug project", used this PR branch, and got that error when entering playmode. I'll test a 3rd time now from scratch
> 
> I now I understand, sorry, Friday-slowness, yes, the repro scenario cannot be allowed. It corrupts the assets. Try the original one if you are curious and see that after exiting playmode, asset A is containing a reference to a sub-asset in asset B......

I haven't replied to user yet, but the recommendation will be to instead.

reference = InputActionReference.Create(desiredAction);

jfreire-unity:
> InputActionReference.Create(desiredAction);

Got it. It works, thanks. For some reasons it didn't work before (maybe I had the project with asset already changed by using .Set() without these changes)

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
jfreire-unity: Ah good point 👍

jfreire-unity (approved):
I'm approving now that I managed to solve the customer issue. Good work.

This is a small change in behavior so I think we should communicate this somehow in a Discussions post. I can already predict more bugs related to this change.

ekcoh:
> > InputActionReference.Create(desiredAction);
> 
> Got it. It works, thanks. For some reasons it didn't work before (maybe I had the project with asset already changed by using .Set() without these changes)

Awesome, yeah I found that if you enter playmode once with the repro project you have corrupted the asset beyond what can be resolved by reimport or action editor changes.

ekcoh:
> I'm approving now that I managed to solve the customer issue. Good work.
> 
> This is a small change in behavior so I think we should communicate this somehow in a Discussions post. I can already predict more bugs related to this since.

Would probably make sense to provide an example with the release post. Regardless, the existing behaviour is clearly faulty so at least IMO this improves things and leads to less suprises with corrupt assets or cross-asset references.

ekcoh:
Hmmm..... maybe this still causes errors on 2021.3 since that version wouldn't have PWA define

ekcoh:
> Hmmm..... maybe this still causes errors on 2021.3 since that version wouldn't have PWA define

Confirmed, property drawer is mutating reference via serialised property when used on Unity versions where PWA isn't supported (rather SearchProvider isn't supported), a custom property drawer is also needed before that to not corrupt assets. Will add an update to this PR.

ekcoh:
> > Hmmm..... maybe this still causes errors on 2021.3 since that version wouldn't have PWA define
> 
> Confirmed, property drawer is mutating reference via serialised property when used on Unity versions where PWA isn't supported (rather SearchProvider isn't supported), a custom property drawer is also needed before that to not corrupt assets. Will add an update to this PR.

Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2248/commits/9c7cb4442a4ccc0307d999c373ce654c74e54943

jfreire-unity:
@ekcoh as a sanity check, I would check with packages like XR and Cinemachine if they have any immediate problems with this change.

ekcoh:
> @ekcoh as a sanity check, I would check with packages like XR and Cinemachine if they have any immediate problems with this change.

Good idea. I am have reached out in appropriate internal channels to see if additional feedback can be collected.

ekcoh:
According to @windxu88 the Ubuntu instability is expected in CI at the moment

Pauliusd01:
Been looking at this on and off for a while and it seems fine so far. Still looking into it but got minor notes:

- Some minor improvements for the error you get when trying to use InputActionReference.Set()  "I`nvalidOperationException: Attempting to mutate an immutable InputActionReference instance. This is not allowed since it would modify the source asset.Instead use InputActionReference.Create(action) to create a new in-memory instance or serialize it as a separate asset if **it survive domain reloads**.`" I assume the ending was supposed to be something like "...if it to survive domain reloads" or even expanding the sentence further.
- For [ISXB-1688](https://jira.unity3d.com/browse/ISXB-1688) is it about creating and modifying asset files in editor? As I didn't have any problems removing actions in runtime before or after your changes.
- For [ISXB-1692](https://jira.unity3d.com/browse/ISXB-1692) how is it different from [ISXB-1584](https://jira.unity3d.com/browse/ISXB-1584)? Just a bit of extra info would help as to me it just seemed the same.

ekcoh:
> @ekcoh as a sanity check, I would check with packages like XR and Cinemachine if they have any immediate problems with this change.

Cinemachine team confirmed it doesn't seem to cause any issues with Cinemachine.

ekcoh:
> Been looking at this on and off for a while and it seems fine so far. Still looking into it but got minor notes:
> 
> * Some minor improvements for the error you get when trying to use InputActionReference.Set()  "I`nvalidOperationException: Attempting to mutate an immutable InputActionReference instance. This is not allowed since it would modify the source asset.Instead use InputActionReference.Create(action) to create a new in-memory instance or serialize it as a separate asset if **it survive domain reloads**.`" I assume the ending was supposed to be something like "...if it to survive domain reloads" or even expanding the sentence further.

Thanks for the feedback. I look into making that exception message better.

> * For [ISXB-1688](https://jira.unity3d.com/browse/ISXB-1688) is it about creating and modifying asset files in editor? As I didn't have any problems removing actions in runtime before or after your changes.

No, it's about doing it programmatically and I have covered it with an automated test that failed when implemented. Then fixed the issue.

> * For [ISXB-1692](https://jira.unity3d.com/browse/ISXB-1692) how is it different from [ISXB-1584](https://jira.unity3d.com/browse/ISXB-1584)? Just a bit of extra info would help as to me it just seemed the same.

It's not fundamentally different, but I decided to see ISXB-1584 as the reported problem, that the action wasn't properly updated. This particular one is about corrupting it. There was many things happening at once in 1584 and I just filed specific tickets per problem. This one refers to the part that if you run the repro project without these changes after exiting playmode the first asset (originally containing A), will contain a reference pointing to an action within the asset containing B.

Pauliusd01 (approved):
LGTM, things checked: 
- each ticket's repro steps
- adding it to a big internal project (kinwarton) and testing its functionality
- exploratory testing around using/deleting/setting action references in various scenarios. (Such as: using project wide actions assets for references and making changes to it, using multiple references pointing to the same thing and then making changes,  domain reload enabled/disabled testing)

ekcoh:
@Pauliusd01 Attempted to improve exception message in https://github.com/Unity-Technologies/InputSystem/pull/2248/commits/3b9cceb87e42944374f2353ee7aee773c6985111

chris-massie (approved):
These changes did not seem to negatively impact XRI during my testing.

PR #2249, opened 2025-10-01
FIX: Sort out controls which are not contained in event for AnyButtonPressed event (ISXB-1005)

ritamerkl:
### Description

This PR fixes an [issue](https://jira.unity3d.com/browse/ISXB-1005) where InputSystem.anyButtonPress is triggered multiple times if another control of the same device is actuated while holding a button down. This is also reproducible on gamepads that receive events constantly, holding a button down then produces one anyButtonPress event per frame.

This was due to the fact that all controls of the device contained in the event were checked for actuated buttons, not only the control triggering the event. Then the magnitude check of the button causes the trigger of the callback over and over again.
This is prevented by the newly added line of code which checks if the control is the same as from the InputEvent.

This [issue](https://jira.unity3d.com/browse/ISXB-1628) is fixed with the PR too.

### Testing status & QA

Manual testing with Test project of customers.

### Overall Product Risks

Low

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers

The documentation states that the callback is only triggered once per button press.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Assets/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -179,6 +179,45 @@ public void Events_CanListenForButtonPresses()
         InputSystem.Update();
     }
 
+    [Test]
+    [Category("Events")]
+    public void Events_OnAnyButtonPressed_FiltersOutOtherControls()
```
jfreire-unity, line 184: Great to have a test. I assume this test would fail before your change, right?

jfreire-unity (approved):
Looks like a simple one liner and good to have yet another test for this!

ekcoh (approved):
Looks good, thanks for looking into this.

Pauliusd01:
Did not forget this one, taking a look today

Assets/Tests/InputSystem/CoreTests_Events.cs
ritamerkl: yes!

Pauliusd01 (approved):

PR #2254, opened 2025-10-07
FIX: ISXB-1637 Crash on memcpy when calling Press() in an InputTestFixture

ekcoh:
### Description

This PR addresses user reported issue: https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1637 in the following ways:
- Adds a direct null pointer check to `DeltaStateEvent.cs` where pointer arithmetic would be incorrect and lead to exception or crash when offsetting a nullptr.
- Adds an indirect null pointer check to `InputTestFixture.cs` (via new internal property `InputControl.hasState`) where accessing a device from "outside" test context would lead to the same symptoms as above. This is now instead checked and throws exception if this is attempted.
- Adds detection and throws exception if detecting `InputTestFixture` Set operations being used in an editor-assembly with [UnityTest] tests. Documented this and previously undocumented exceptions in xmldoc.
- Adds new test class `InputTestFixtureTests` that verifies normal behavior in `InputTestFixture` as well as inappropriate use similar to the reported repro project and verifies that expected exceptions are thrown for these scenarios.

The problem in the provided repro project is two-fold:
- Using `[OneTimeSetup]` and `[OneTimeTearDown]` to manage test devices. This will not work as intended since the device will be added to the real Input System context, then removed as part of `InputTestFixture.Setup()`. To reduce chance of similar accidental situations, I have added information recommending against this to `InputTestFixture` xmldoc. In addition, if this is attempted it should result in exception which is verified in the added tests.
- Using `InputTestFixture` in edit-mode tests `[UnityTest]` which isn't fully supported it seems.

Notice that I decided to block use of `InputTestFixture` and its Set-family of operations in editor-assembly tests. The reason for this is that `InputManager` will not process such updates, it will switch to editor buffers and then early out. I do not see any scenario where this may cause problems but if you do let me know and we update accordingly. The changes on this PR is intended to detect and inform about unintended or unsupported use of a test fixture.

### Testing status & QA

Added multiple automated edit-mode tests into `InputTestFixtureTests`

### Overall Product Risks

- Complexity: Small
- Halo Effect: Small

### Comments to reviewers

Good to check that you are aligned on this clarification and blocking certain behavior.

### Checklist

Before review:

- [x] Changelog entry added.
    - Functional tests: Multiple tests added in `InputTestFixtureTests`
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Events/DeltaStateEvent.cs
```diff
@@ -76,6 +76,8 @@ public static NativeArray<byte> From(InputControl control, out InputEventPtr eve
             if (!device.added)
                 throw new ArgumentException($"Device for control '{control}' has not been added to system",
                     nameof(control));
+            if (control.currentStatePtr == null) // Protects statePtr assignment below
+                throw new ArgumentNullException($"Control '{control}' do not have an associated state");
```
K-Tone:
```suggestion
                throw new ArgumentNullException($"Control '{control}' does not have an associated state");
```

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
```diff
@@ -533,6 +544,12 @@ public void Set<TValue>(InputDevice device, string path, TValue state, double ti
         /// Note that this parameter will be ignored if the test is a <c>[UnityTest]</c>. Multi-frame
         /// playmode tests will automatically process input as part of the Unity player loop.</param>
         /// <typeparam name="TValue">Value type of the given control.</typeparam>
+        /// <exception cref="ArgumentNullException">If control is null.</exception>
+        /// <exception cref="ArgumentException">If the device associated with <paramref name="control"/> has not
+        /// been added to the system or if the control do not have any associated state. The latter may only
```
K-Tone:
```suggestion
        /// been added to the system or if the control does not have any associated state. The latter may only
```

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
```diff
@@ -545,11 +562,22 @@ public void Set<TValue>(InputControl<TValue> control, TValue state, double time
...
+            if (!control.hasState)
+            {
+                throw new ArgumentException($"Control '{control}' does not have any associated state. " +
+                    "Make sure the control or device was added after executing Setup().",  nameof(control));
+            }
 
             if (IsUnityTest())
+            {
+                if (IsEditMode())
+                    throw new NotSupportedException("InputTestFixture.Set do not support edit mode (editor assembly) [UnityTest].");
```
K-Tone:
```suggestion
                    throw new NotSupportedException("InputTestFixture.Set does not support edit mode (editor assembly) [UnityTest].");
```

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,7 +10,9 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
 
-
+### Fixed
+- Fixed an issue in `DeltaStateEvent.From` where unsafe code would throw exception or crash if internal pointer `currentStatePtr` would be `null`. [ISXB-1637](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1637).
```
K-Tone:
```suggestion
- Fixed an issue in `DeltaStateEvent.From` where unsafe code would throw exception or crash if internal pointer `currentStatePtr` was `null`. [ISXB-1637](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1637).
```

Assets/Tests/InputSystem.Editor/InputTestFixtureTests.cs
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    [Test]
+    public void Click_ShouldThrow_WithinPlayModeTestFixtureContextIfInvalidDevice()
+    {
+        Click(correctlyUsedDevice.spaceKey);
+        Assert.Throws<ArgumentException>(() => Click(incorrectlyUsedDevice.spaceKey));
+    }
+
+    // TODO Add remaining
```
K-Tone: Maybe local Copilot can help adding the rest of the combinations if needed? Otherwise, we should remove this TODO, as it comes in a newly added file. This produces an impression of an unfinished test, whereas the currently available set is already pretty good in my opinion.

K-Tone (changes requested):
I'd ask to remove TODO and adjust spelling in a few spots, otherwise looks pretty good

Assets/Tests/InputSystem.Editor/InputTestFixtureTests.cs
ekcoh: Mistake on my side, I see if I can get them added. Its the right thing to do even if they happen to call the same function its an implementation detail.

ekcoh:
@K-Tone Thanks for the feedback. It was very valuable since I realised I had missed the same problem with touch APIs. It's not addressed and have been committed with additional tests to cover touch APIs as well as Vector2 API.

Assets/Tests/InputSystem.Editor/InputTestFixtureTests.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2254/commits/57ef549e82cc15641cbf386bf30f4e7acce03ce8

K-Tone (approved):
Looks good to me, thanks for addressing the comments!

Pauliusd01 (approved):

ekcoh:
Interesting that CI behaves differently on Windows than on other platforms, some new editor tests failing. Will need to investigate on Windows I guess.

ekcoh:
I am not understanding these failures. They are all green when I execute them locally on Windows 10.

ekcoh:
I updated the approach taken here to a more direct approach after I realised push/pop in Input manager isn't consistently pushing/popping devices and also to not manage dangling references in such devices. This is likely another area that needs attention (test scope only).

ekcoh:
Re-requesting review on this one since the previous approach had flaws. Se commits from https://github.com/Unity-Technologies/InputSystem/pull/2254/commits/ef7b23d9886dce7b87d5484cde42510a8a805e96 and forward.

Pauliusd01 (approved):

K-Tone (approved):

PR #2262, opened 2025-10-21
NEW: Test case for monotonic timestamps.

ekcoh:
### Description

As part of a pending Unity PR we ended up in a discussion whether Input System requires strictly increasing timestamps or not. This PR proves with an automated tests that timestamps do not have to be strictly incrementing, but should have timestamps greater or equal to the last received timestamp for the same source (device).

Skipping CHANGELOG update since test only.

### Testing status & QA

Adds a single test.

### Overall Product Risks

- Complexity: Minimal
- Halo Effect: Minimal

### Comments to reviewers

Nothing special.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


ekcoh:
This currently fails for Android due to package containing Android specific path which ignores timestamps.

jfreire-unity (commented):
Thanks for adding this, makes a lot of sense to have this.

How should we proceed with regards to Android CI errors? I'm fine approving once CI is green

Assets/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -2256,6 +2256,31 @@ public void Events_IfOldStateEventIsSentToDevice_IsIgnored()
...
+    {
+        var device = InputSystem.AddDevice<Keyboard>();
+
+        InputSystem.QueueStateEvent(device, new KeyboardState(), 1.0);
+        InputSystem.Update();
+
+        // Accepted: equal
+        InputSystem.QueueStateEvent(device, new KeyboardState(Key.Space), 1.0);
+        InputSystem.Update();
+        Assert.That(device.spaceKey.ReadValue(), NUnit.Framework.Is.EqualTo(1.0f).Within(1e-6));
```
jfreire-unity, line 2271: I'm wondering why we don't use the already defined `using Is = UnityEngine.TestTools.Constraints.Is` that we use a lot across all codebase tests?
The same applies to other `NUnit.Framework.Is.Equal` calls below.

ekcoh: Will probably switch to `UnityEngine.TestTools.Constraints.Is` but it's just terribly annoying to have hundreds of code warnings inside the file leading to the quick navigation in scrollbar being unusable due to all warnings. The above `Is` extends the class I used and that generates a warning, e.g. "you should not access static functions via derived class", which makes sense. But maybe we should just add the required pragmas to the test classes to see actual warnings and fix them if this noise is removed.

ekcoh:
> Thanks for adding this, makes a lot of sense to have this.
> 
> How should we proceed with regards to Android CI errors? I'm fine approving once CI is green

Good question, I want to sync up with Android team how they decide to land and backport the pending PR first. Then we either need to add an exception to the test (until it can be removed from InputManager), or let this just sit here until it can be removed from Input Manager. It might be that the exception becomes version specific, but ideally not.

Other ideas @jfreire-unity ?

Assets/Tests/InputSystem/CoreTests_Events.cs
ekcoh: I create a separate PR to fix these messy using/aliasing.

PR #2265, opened 2025-10-27
FIX: Change type aliasing to use the type implementing static interface instead of using derived type.

ekcoh:
### Description

The https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/api/UnityEngine.TestTools.Constraints.AllocatingGCMemoryConstraint.html example is misleading since it guides to alias the extended Is from this package (adding a single extension). 

This leads to warnings in certain compilers/IDEs, e.g. https://www.jetbrains.com/help/resharper/AccessToStaticMemberViaDerivedType.html

However, the extension is extending NUnit.Framework.Is and hence the aliasing of Is can be made to this type instead to still support the extensions while also not generating warnings in aforementioned IDE.

This PR has no functional impact but allows noticing warnings in test code in Rider instead of getting hundreds of warnings about the above problem.

Skipping CHANGELOG entry since this is mainly internal.

### Testing status & QA

Just checked it compiles and no warnings in Rider 2025.2.3 with Unity 6000.2.7f2. 

### Overall Product Risks

Minimal.

- Complexity: Minmal
- Halo Effect: Minimal

### Comments to reviewers

No functional impact at all is expected.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Actions.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Actions.cs
@@ -21,11 +21,10 @@ using UnityEngine.Profiling;
 using UnityEngine.TestTools;
 using UnityEngine.TestTools.Utils;
 using UnityEngine.TestTools.Constraints;
-using Is = UnityEngine.TestTools.Constraints.Is;
+
+using Is = NUnit.Framework.Is;
 
 #pragma warning disable CS0649
-[SuppressMessage("ReSharper", "AccessToStaticMemberViaDerivedType")]
-
 // As should be obvious from the number of tests in here, the action system rivals the entire combined rest of the system
 // in terms of complexity.
 partial class CoreTests
--- a/Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
+++ b/Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
@@ -6,7 +6,6 @@ using NUnit.Framework;
 using UnityEngine;
 using UnityEngine.InputSystem;
 using UnityEngine.TestTools;
-using Is = UnityEngine.TestTools.Constraints.Is;
 
 partial class CoreTests
 {
--- a/Assets/Tests/InputSystem/CoreTests_Controls.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Controls.cs
@@ -16,7 +16,7 @@ using UnityEngine.InputSystem.Processors;
 using UnityEngine.InputSystem.Utilities;
 using UnityEngine.Profiling;
 using UnityEngine.TestTools.Constraints;
-using Is = UnityEngine.TestTools.Constraints.Is;
+using Is = NUnit.Framework.Is;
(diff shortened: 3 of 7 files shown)


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -21,11 +21,10 @@
 using UnityEngine.TestTools;
 using UnityEngine.TestTools.Utils;
 using UnityEngine.TestTools.Constraints;
-using Is = UnityEngine.TestTools.Constraints.Is;
 
-#pragma warning disable CS0649
-[SuppressMessage("ReSharper", "AccessToStaticMemberViaDerivedType")]
+using Is = NUnit.Framework.Is;
```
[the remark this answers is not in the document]
ekcoh: Yes, but that is a very specific use case that is still supported since the Unity extension is based on the same class being aliased here. Not using the base is considered bad practise by e.g. JetBrains Rider.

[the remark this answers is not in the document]
ekcoh: UnityEngine.TestTools.Constraints just need a using statement to make it accessible.

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
```diff
@@ -6,7 +6,6 @@
 using UnityEngine;
 using UnityEngine.InputSystem;
 using UnityEngine.TestTools;
-using Is = UnityEngine.TestTools.Constraints.Is;
```
ritamerkl, line 9: was this not being used?

ritamerkl (approved):
LGTM, thanks for adding.

Assets/Tests/InputSystem/CoreTests_MouseEvents.cs
ekcoh: I think when UnityEngine.TestTools.Constraints isn't included (with using) there is no ambiguous behavior that needs explicit aliasing to be resolved, but I probably need to double check with a version that enables this code.

ekcoh: There was no GC memory allocation check so UnityEngine.TestTools.Constraints isn't needed at least

ekcoh: using NUnit.Framework; should be enough

ritamerkl: thanks for checking!

jfreire-unity (approved):

PR #2268, opened 2025-10-28
FIX: Autosave closes add control scheme ui element on save. (ISXB-1713)

Darren-Kelly-Unity:
### Description

Fix an issue where the control scheme UI would close when pressing the + for adding a device type when there was a pending change to be auto saved.

Ticket:
https://jira.unity3d.com/browse/ISXB-1713

### Testing status & QA

I have tested this manually in the editor on Unity version 6.000.1.5f1

### Overall Product Risks

- Complexity: 2
- Halo Effect: 1

### Comments to reviewers

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


K-Tone (approved):
Looks okay to me! Thanks for working on it. 

CC @ekcoh for viz, this fix seems to build up on my other idea of not saving during the sensitive moments in time. Not cool to be extending it, but what else could be done before our auto-save code is redesigned to be less crazy than it is now? :)

Pauliusd01:
It still reproduces with project wide actions view but if it is complicated to fix I'd be fine with merging as is. 

Steps to repro: Open project settings -> input system view -> make sure you have a PWA asset if not make one -> make some changes (this window always uses auto save) -> start editing the control scheme as bug repro describes -> focus is lost

https://github.com/user-attachments/assets/f3dfcffa-075c-46d2-a11a-75da6946f51a

Pauliusd01 (changes requested):
Updating status

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
```diff
@@ -175,7 +175,12 @@ private void OnFocusOut(FocusOutEvent @event = null)
 
         private void ValidateAndSaveAsset(InputActionAsset asset)
         {
-            ProjectWideActionsAsset.Verify(asset); // Ignore verification result for save
+            if (!InputEditorUserSettings.autoSaveInputActionAssets || ControlSchemesView.IsShowingControlSchemeView)
```
ritamerkl: maybe add a comment to make clear this is related to auto-save/focus problems, so we can find it easily by the time we work on a root cause fix.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
```diff
@@ -14,9 +14,16 @@ internal class ControlSchemesView : ViewBase<InputControlScheme>
         private string m_NewName;
         public event Action<ViewBase<InputControlScheme>> OnClosing;
 
+        /// <summary>
+        /// This property is only set from this class in order to communicate that we're showing the control schemes view at the moment
+        /// It's employed to skip auto-saving, because that complicates the save causing this window to close when adding a new device type.
+        /// </summary>
+        internal static bool IsShowingControlSchemeView { get; private set; }
```
ritamerkl: It's static, but it's editor window stuff (not runtime), so fine for me :D And I see there is a similar implementation already in InputControlPathEditor.

ritamerkl (approved):
Nice workaround

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
Darren-Kelly-Unity: Yeah I copied the implementation from InputControlPathEditor, I think it makes sense in this case also as we may not have an instance of the class yet in certain circumstances.

Later we can hopefully remove both of these static bools.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorSettingsProvider.cs
Darren-Kelly-Unity: Added! :)

ekcoh (approved):
Thanks for doing this fix @Darren-Kelly-Unity. The staticness of these states look a bit dangerous to me, but if it solves the problem lets land this fix since it quite condensed.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
ritamerkl: Could it be possible to have the situation where we have several instances of the window, where this can be problematic?

Darren-Kelly-Unity: That's a very good point! I just tested and this would cause the bug to happen again if you manage to have both control scheme view windows open at the same time while doing things, it's quite an edge case though. I'll try to see if I can solve this.

Darren-Kelly-Unity: It's quite hard to do though as the control scheme view closes when focus is lost, not sure how I did it now.

ritamerkl: It is very edge case, and I'd be okay in landing this, since editing from two places is probably not the way users want to do that anyway. And it's still better than this happening everytime. The only thing that could potentially happen would be that the auto save is not happening even though we don't necessary need to prevent it, right?

ritamerkl: I'm just thinking of worst case that could happen if there are side effects with two instances, if the worst case is that the bug is still there in a very edge case scenario, that's fine I think.

Darren-Kelly-Unity:
> Thanks for doing this fix @Darren-Kelly-Unity. The staticness of these states look a bit dangerous to me, but if it solves the problem lets land this fix since it quite condensed.

I actually changed it now due to a comment from Rita asking about with two windows open, now I just check the view to see if the ControlScheme is currently active for each instance which is better.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Views/ControlSchemesView.cs
Darren-Kelly-Unity: Solved this now & we talked in slack! :)

Pauliusd01 (approved):

PR #2270, opened 2025-10-29
FIX: [ISXB-1716] Add support for all existing InputControl types in InputTestFixture.Trigger 

K-Tone:
### Description

There was a problem reported that if you invoke InputTestFixture.Trigger with anything other than a Button or an Axis, there would be a NotImplementedException thrown.

Here, we address it by logically extending the approach originally taken by the Axis implementation (change the value slightly as an effect from Trigger) to all other possible InputControl variations that we have. We only have two outliers here -- the integer control just gets its value incremented by one instead of the regular small change. Also the TouchState thing gets the next state in the list as well. 

### Testing status & QA

Local testing 

### Overall Product Risks

- Complexity:  Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
+++ b/Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
@@ -13,7 +13,6 @@ using UnityEngine.TestTools;
 using UnityEngine.TestTools.Utils;
 using UnityEngine.InputSystem.XR;
 
-
 #if UNITY_EDITOR
 using UnityEditor;
 using UnityEngine.InputSystem.Editor;
--- a/Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
+++ b/Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
@@ -11,6 +11,9 @@ using UnityEngine.InputSystem.LowLevel;
 using UnityEngine.InputSystem.Utilities;
 using UnityEngine.TestTools;
 using UnityEngine.TestTools.Utils;
+using UnityEngine.InputSystem.XR;
+
+
 #if UNITY_EDITOR
 using UnityEditor;
 using UnityEngine.InputSystem.Editor;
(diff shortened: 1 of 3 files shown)


ProjectSettings/EditorBuildSettings.asset
```diff
@@ -5,4 +5,7 @@ EditorBuildSettings:
   m_ObjectHideFlags: 0
   serializedVersion: 2
   m_Scenes: []
-  m_configObjects: {}
+  m_configObjects:
+    com.unity.input.settings.actions: {fileID: -944628639613478452, guid: 0cfafdaf330c8481b9ae4210786b578d,
+      type: 3}
+  m_UseUCBPForAssetBundles: 0
```
jfreire-unity: Just chiming in as this is probably due to a Project Wide Actions asset in the project. We should probably remove it as it's likely part of our automated tests

K-Tone: Whoops, you're right

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
```diff
@@ -728,32 +730,110 @@ public void Trigger(InputAction action)
...
+
+                    return;
+                }
+
+                if (control is IntegerControl integerControl)
+                {
+                    Set(integerControl, integerControl.ReadValue() + 1);
+
+                    return;
+                }
```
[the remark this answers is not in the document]
K-Tone: Not sure we need to do that really, there's barely any contract that Trigger should produce a minimal change

ekcoh (approved):
Thanks for doing this @K-Tone and I think this fills the gap called out in the ticket.
However I also believe this makes it even more clear why this is incorrectly implemented to begin with which I think is outside of this PR but wanted to mention anyway for alignment. Since Input System claims to be an extendible system, and decided to expose this test fixture, it means that the number of InputControl or InputDevice types is not fixed. Hence, IMO, implementing something like `Trigger` here is not scalable. It would have been much better to not have such a function in the fixture at all, and instead add such support via extension methods. Then each plugin implementing the control could provide a test assembly with an extension method providing the needed Trigger implementation for that particular type only. The same goes for Press, Release and all "test extensions" defined in this file. Something to think about for when we do v.2.0 API of this. Maybe we should even file a ticket on it or what do you think? It would be a straightforward refactor. If `InputTestFixture` would not be public, this would be less important, but still better design IMO, or what do you think?

K-Tone:
> Thanks for doing this @K-Tone and I think this fills the gap called out in the ticket. However I also believe this makes it even more clear why this is incorrectly implemented to begin with which I think is outside of this PR but wanted to mention anyway for alignment. Since Input System claims to be an extendible system, and decided to expose this test fixture, it means that the number of InputControl or InputDevice types is not fixed. Hence, IMO, implementing something like `Trigger` here is not scalable. It would have been much better to not have such a function in the fixture at all, and instead add such support via extension methods. Then each plugin implementing the control could provide a test assembly with an extension method providing the needed Trigger implementation for that particular type only. The same goes for Press, Release and all "test extensions" defined in this file. Something to think about for when we do v.2.0 API of this. Maybe we should even file a ticket on it or what do you think? It would be a straightforward refactor. If `InputTestFixture` would not be public, this would be less important, but still better design IMO, or what do you think?

Totally agreed, that's what I've been thinking about while working on the ticket.

K-Tone:
Probably not much to review here (no fix), but still CC for viz, so that you're aware this happened @Pauliusd01

PR #2272, opened 2025-10-30
NEW: Insights analytics support for the Input System

K-Tone:
### Description

This PR brings support for the Insights analytics module to Input System.

Requires native side PR to land first: https://github.cds.internal.unity3d.com/unity/unity/pull/83832

Currently, we send these kinds of data to Insights:

1) When a new input device is connected, we send this information:
* product (" DUALSHOCK 4 Wireless Controller")
* serial ("A0:5A:5E:92:0F:3E")
* version ("1.0")
* interface ("HID")

2) When an input device is disconnected, we send this information:
* serial (really intended to be matched with the event above at the time of analysis)

3) When a new Input Action Map is being loaded or activated, we scan for all the Input Actions that are being brought to life and for each one, we send the following data:
* name (e.g. "Jump")
* type (e.g. "Button")

### Testing status & QA

Local testing

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -10,6 +10,7 @@ using Unity.Profiling;
 using UnityEngine.InputSystem.Utilities;
 
 using ProfilerMarker = Unity.Profiling.ProfilerMarker;
+using UnityEngineInternal.Input;
 
 ////TODO: now that we can bind to controls by display name, we need to re-resolve controls when those change (e.g. when the keyboard layout changes)
 
@@ -957,6 +958,14 @@ namespace UnityEngine.InputSystem
                 NotifyListenersOfActionChange(InputActionChange.ActionEnabled, map.m_SingletonAction);
             else
                 NotifyListenersOfActionChange(InputActionChange.ActionMapEnabled, map);
+
+#if UNITY_INPUT_SYSTEM_SUPPORTS_INSIGHTS
+            // also report all actions to insights
+            foreach (var inputAction in map.actions)
+            {
+                NativeInputSystem.LogInputActionInsight(inputAction.name, inputAction.type.ToString());
+            }
+#endif
         }
 
         private void EnableControls(InputActionMap map)
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -6,11 +6,9 @@ using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
 using UnityEngine.InputSystem.Controls;
 using UnityEngine.InputSystem.LowLevel;
-using Unity.Profiling;
 using UnityEngine.InputSystem.Utilities;
 
 using ProfilerMarker = Unity.Profiling.ProfilerMarker;
-using UnityEngineInternal.Input;
 
 ////TODO: now that we can bind to controls by display name, we need to re-resolve controls when those change (e.g. when the keyboard layout changes)
 
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -963,7 +963,7 @@ namespace UnityEngine.InputSystem
             // also report all actions to insights
             foreach (var inputAction in map.actions)
             {
-                NativeInputSystem.LogInputActionInsight(inputAction.name, inputAction.type.ToString());
+                InputSystem.s_Manager.m_Runtime.LogInputActionInsight( inputAction.name, inputAction.type.ToString() );
             }
 #endif
         }
--- a/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
@@ -963,7 +963,7 @@ namespace UnityEngine.InputSystem
             // also report all actions to insights
             foreach (var inputAction in map.actions)
             {
-                InputSystem.s_Manager.m_Runtime.LogInputActionInsight( inputAction.name, inputAction.type.ToString() );
+                InputSystem.s_Manager.m_Runtime.LogInputActionInsight( inputAction );
             }
 #endif
         }
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -2531,9 +2531,7 @@ namespace UnityEngine.InputSystem
             // Parse description, if need be.
             var description = device?.description ?? InputDeviceDescription.FromJson(deviceDescriptor);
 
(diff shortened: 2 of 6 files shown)


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -957,6 +958,14 @@ public void EnableAllActions(InputActionMap map)
                 NotifyListenersOfActionChange(InputActionChange.ActionEnabled, map.m_SingletonAction);
             else
                 NotifyListenersOfActionChange(InputActionChange.ActionMapEnabled, map);
+
+#if UNITY_INPUT_SYSTEM_SUPPORTS_INSIGHTS
+            // also report all actions to insights
+            foreach (var inputAction in map.actions)
+            {
+                NativeInputSystem.LogInputActionInsight(inputAction.name, inputAction.type.ToString());
```
ekcoh: I would suggest to add these functions to an interface, e.g. `IInputSystemInsights`, which NativeInputRuntime can implement (similar to IInputRuntime` and call via that interface instead. This way, the test harness can also mock it (similar to IInputRuntime).

ekcoh: If not this would log to insights also when running tests

ekcoh: And not be testable?

ekcoh (commented):
Looks good, but would suggest to use interface similar to other module facing calls given the current test fixture design.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
K-Tone: I"ll take a look at that, thanks

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -92,11 +92,16 @@
...
-	    "name": "Unity",
+        {
+            "name": "Unity",
             "expression": "6000.4.0a4",
             "define": "UNITY_INPUTSYSTEM_SUPPORTS_MOUSE_SCRIPT_EVENTS"
-	}
+        },
+        {
+            "name": "Unity",
+            "expression": "6000.5.0a5",
```
K-Tone, line 102: This will have to be changed to whatever is the actual Unity version we happen to release Input insights in. We will know that once the native PR lands. Then we will change this for every backport

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
K-Tone: I added proxy functions to IInputRuntime for now, for simplicity sake. Thanks for bringing this up, totally escaped my attention.

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestRuntime.cs
```diff
@@ -481,6 +481,19 @@ public void SendAnalytic(InputAnalytics.IInputAnalytic analytic)
             #endif // UNITY_2023_2_OR_NEWER
         }
 
-        #endif // UNITY_ANALYTICS || UNITY_EDITOR
+        // We don't want to populate Insights from within tests, even when running them in players
+        public void LogDeviceConnectedInsight(InputDeviceDescription description)
```
ekcoh, line 485: Might want to store these in harness in case you want to test it in the package

ekcoh (approved):

ekcoh:
Looks like you need to (re)run formatter on this one

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2531,6 +2531,8 @@ private void OnNativeDeviceDiscovered(int deviceId, string deviceDescriptor)
             // Parse description, if need be.
             var description = device?.description ?? InputDeviceDescription.FromJson(deviceDescriptor);
 
+            m_Runtime.LogDeviceConnectedInsight(description);
+
```
rohit-moni, line 2535: I see that logging the input action is bracketed by the new define `UNITY_INPUT_SYSTEM_SUPPORTS_INSIGHTS`. Should this also be conditional on it?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3597,6 +3599,8 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
                         {
                             RemoveDevice(device, keepOnListOfAvailableDevices: false);
 
+                            m_Runtime.LogDeviceDisconnectedInsight(device.description);
+
```
rohit-moni, line 3603: Similar: should this be #if UNITY_INPUT_SYSTEM_SUPPORTS_INSIGHTS?

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -423,6 +425,27 @@ public void SendAnalytic(InputAnalytics.IInputAnalytic analytic)
...
+        {
+#if UNITY_INPUT_SYSTEM_SUPPORTS_INSIGHTS && !UNITY_EDITOR
+            NativeInputSystem.LogDeviceDisconnectedInsight(description.serial);
+#endif
+        }
+
+        public void LogInputActionInsight(InputAction action)
+        {
+#if UNITY_INPUT_SYSTEM_SUPPORTS_INSIGHTS && !UNITY_EDITOR
+            NativeInputSystem.LogInputActionInsight(action.name, action.type.ToString());
```
rohit-moni, line 445: Not too familiar with input actions, I'm worried that enabling this will add a string allocation to every input action, resulting in two potential issues: performance degradation of all inputs (allocation perf) and frequent / large garbage collection (after many disparate allocations). Can this be cached? Do we have the action type string already stored somewhere? Is there another reason that invalidates the concern?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -957,6 +956,14 @@ public void EnableAllActions(InputActionMap map)
...
             else
                 NotifyListenersOfActionChange(InputActionChange.ActionMapEnabled, map);
+
+#if UNITY_INPUT_SYSTEM_SUPPORTS_INSIGHTS
+            // also report all actions to insights
+            foreach (var inputAction in map.actions)
+            {
+                InputSystem.s_Manager.m_Runtime.LogInputActionInsight( inputAction );
+            }
+#endif
```
rohit-moni, line 966: Was looking at this again and am a little confused. Is `LogInputActionInsight` for when the InputAction happens, or for when it's enabled?

I think as it is right now, we're sending insights for all the actions, but only once (when they're all enabled). We're not tracking when the actions happen. Is that the intent?

PR #2274, opened 2025-11-05
CHANGE: Remove code that supported Unity versions older than 2022 LTS

K-Tone:
### Description

Removing mentions of Unity older than 2022 LTS from code and manifests.

### Testing status & QA

Local complilation

### Overall Product Risks

- Complexity:  Low
- Halo Effect:  Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Explains the change in `Changed`, `Fixed`, `Added` sections.
    - For API change contains an example snippet and/or migration example.
    - JIRA ticket linked, example ([case %<ID>%](https://issuetracker.unity3d.com/product/unity/issues/guid/<ID>)). If it is a private issue, just add the case ID without a link.
    - Jira port for the next release set as "Resolved".
- [ ] Tests added/changed, if applicable.
    - Functional tests `Area_CanDoX`, `Area_CanDoX_EvenIfYIsTheCase`, `Area_WhenIDoX_AndYHappens_ThisIsTheResult`.
    - Performance tests.
    - Integration tests.
- [ ] Docs for new/changed API's.
    - Xmldoc cross references are set correctly.
    - Added explanation how the API works.
    - Usage code examples added.
    - The manual is updated, if needed.

During merge:

- [x] Commit message for squash-merge is prefixed with one of the list:
    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:

- [ ] Create forward/backward port if needed. If you are blocked from creating a forward port now please add a task to ISX-1444.

PR #2277, opened 2025-11-11
CHANGE: Remove the UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS compile-time define

K-Tone:
### Description

Hereby we propose to remove support for the case where we run without the project-wide input actions feature via the corresponding compile-time define (UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS). 

It was necessary back in the day for two main reasons:

1. Feature development approach where we introduced new features under defines, allowing easy opt-in while hopefully reducing potential breakages in the old paths. Project-wide input actions has been there for a while so far, so we're confident we don't need to disable it for the stability reasons. 
2. We had a number of limitations in Unity 2021.x relating to UITK, so we literally couldn't have a fully-functional Input Actions Editor Window there. Because of that, we had a rule in manifest to unset UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS in case pre Unity 2022 was detected. Now that the minimum editor version has been bumped to 2022.3 LTS, the define is always on in fact. 

In this PR, alongside a myriad of trivial changes where we remove the define set/check, we also removed the old imgui-based Input Actions Editor Window that we don't show any more via normal means. It's been possible to enable it using features, but we no longer want to have that variability. With it, we're cutting `InputActionEditorToolbar`. We would have loved to cut more imgui-related stuff, in particular `InputActionPropertiesView` but that proved to be impossible -- since it's still used in custom property drawers. 

### Testing status & QA

Local testing / compilation. 

### Overall Product Risks

- Complexity:  Low
- Halo Effect:  Low

### Comments to reviewers

Needs to land after https://github.com/Unity-Technologies/InputSystem/pull/2274

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

After merge:


Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
```diff
@@ -1,26 +1,12 @@
...
 namespace UnityEngine.InputSystem.Editor
 {
-    // TODO: Remove when UIToolkit editor is complete and set as the default editor
-    [InitializeOnLoad]
-    internal static class EnableUITKEditor
-    {
-        static EnableUITKEditor()
-        {
-        }
-    }
```
K-Tone, line 22: I couldn't notice anyone using this class actually

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporterEditor.cs
```diff
@@ -124,36 +121,24 @@ private static bool IsProjectWideActionsAsset(InputActionAsset asset)
...
-            // Redirect to UI-Toolkit window editor if not configured to use IMGUI explicitly
-            if (!InputSystem.settings.useIMGUIEditorForAssets)
-                InputActionsEditorWindow.OpenEditor(asset);
-            else
-                InputActionEditorWindow.OpenEditor(asset);
-#else
-            // Redirect to IMGUI editor
-            InputActionEditorWindow.OpenEditor(asset);
-#endif
+            InputActionsEditorWindow.OpenEditor(asset);
```
K-Tone, line 141: I would have loved to remove InputSystem.settings.useIMGUIEditorForAssets as well, but it looks like we need it still because we still need (?) to support imgui-based CustomEditors. Once we make sure all of our own editors are ported to UITK, maybe time to deprecate that feature as well? For now I propose to cut the InputActionEditorWindow that is in fact no longer maintained at all, and is no longer accessible through normal means. That's the intent behind this change.

josepmariapujol-unity (approved):
LGTM! Approved

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ however, it has to be formatted properly to pass verification tests.
 
 ## [Unreleased] - yyyy-mm-dd
 
-
+### Changed
+- Project-Wide Input Actions can no longer be disabled (removed the UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS define)
```
ekcoh: I would change this statement to say it can no longer be excluded from compilation. It is very possible to disable it by not marking any .inputactions asset as project-wide.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/AxisComposite.cs
```diff
@@ -219,13 +219,12 @@ internal class AxisCompositeEditor : InputParameterEditor<AxisComposite>
 
         public override void OnGUI()
         {
-#if UNITY_INPUT_SYSTEM_PROJECT_WIDE_ACTIONS
-            if (!InputSystem.settings.useIMGUIEditorForAssets) return;
-#endif
+            if (!InputSystem.settings.useIMGUIEditorForAssets)
```
ekcoh, line 222: As part of this change, or maybe as a separate follow-up this setting should probably be deprecated? At least IMO we should remove the old InputActionEditor

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporterEditor.cs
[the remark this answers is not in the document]
ekcoh: I agree that the useIMGUIEditorForAssets setting should not be used anymore and be marked deprecated?

ekcoh: I am not sure I follow the reasoning here, you are explicitly deleting the editor on this PR? Do not confuse the action asset editor with the Inspector editors or do I misunderstand the reasoning here, mind elaborating?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorWindow.cs
ekcoh: I believe it was forcing the class to be JITed on load which otherwise could be deferred. I have been confused around this previously as well. Not sure we need it or not.

ekcoh (commented):
Great to see this clean-up happening. Could approve already now but want to sort out why we cannot at least deprecate and stop using the imgui setting?

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporterEditor.cs
K-Tone: Sorry for being unclear. In this PR, in addition to removing the project-wide input actions define, I'm also removing the old imgui-based InputActionEditorWindow class. In addition to this class, I wanted to remove its siblings  that represented the panels of the dialog, there are two of them: one for the toolbar, and the other one for the bindings/actions view. I could delete the toolbar thing indeed, but the bindings/actions view is still needed, because it's used in the custom editors thing, when you double-click on an ActionMap or bindings reference in the inspector. I would have loved to remove useIMGUIEditorForAssets too, but it felt like we still need it for at least until we have all of our inspectors ("custom editors") ported to uitk.

K-Tone: The line we're having this discussion under is intended to make it so that no matter whether we want to use imgui-based inspectors or not, we won't open the old Input Actions Editor Window.

Packages/com.unity.inputsystem/InputSystem/Actions/Composites/AxisComposite.cs
K-Tone: The old InputActionEditorWindow is gone in this PR. Agreed about deprecating useIMGUIEditorForAssets

jfreire-unity (approved):
Thanks for doing this. Great to do some house cleaning! I share the sentiment of removing old IMGUI inspector code. Maybe we could also figure out how many inspectors there are and plan work to remove them in 6.7?

My only nitpick is to the PR title, because I got confused at first.
 We're actually only removing the conditional compilation for the project-wide actions feature AND removing unused IMGUI code. 
But changing it to to "Remove project-wide input action feature conditional compilation" or "Project-wide input actions feature is always compiled" would be slightly better, but it's just an opinion.

K-Tone:
CC @Pauliusd01, here is a PR where we remove some more defines. Do you want to be added as a reviewer, or is it fine to just have you mentioned for viz?

K-Tone:
Going back to removing InputSystem.settings.useIMGUIEditorForAssets, all the spots I checked actually have a UITK path that is currently enabled by default. 

image

Check out this image, it's typical. There are 14 files like that. I'd propose this:
* After landing this PR, open a new one specifically for useIMGUIEditorForAssets removal
* Mark useIMGUIEditorForAssets deprecated, make it do nothing; I guess being public, we can't remove it now
* For all spots like on the screenshot above, we remove the OnGUI function from all of those classes (if possible) - because it doesn't do anything except checking for useIMGUIEditorForAssets
* Where we have GUIContent, split it into strings - text & tooltip so that we don't depend on imgui concepts any more

How does that sound @ekcoh @jfreire-unity ?

Pauliusd01 (approved):
LGTM, tested: 
- Input binding controls in the inspector
- Input Actions window and ProjectWideActions window functionality, saving, controls, focus, etc
- Upgrading from 1.7 Input system to this PR

jfreire-unity:
> * After landing this PR, open a new one specifically for useIMGUIEditorForAssets removal
> * Mark useIMGUIEditorForAssets deprecated, make it do nothing; I guess being public, we can't remove it now
> * For all spots like on the screenshot above, we remove the OnGUI function from all of those classes (if possible) - because it doesn't do anything except checking for useIMGUIEditorForAssets
> * Where we have GUIContent, split it into strings - text & tooltip so that we don't depend on imgui concepts any more
> 
> How does that sound @ekcoh @jfreire-unity ?

Sounds like a good plan to me 👍

Packages/com.unity.inputsystem/InputSystem/Editor/AssetImporter/InputActionImporterEditor.cs
ekcoh: Thanks for the explanation, that makes full sense since they were code-refused (thankfully) between editor and inspector. For reference, we also sorted out my confusion around this and the other related PR being in flight. Will mark this resolved and approve now.

ekcoh (approved):

K-Tone:
Failures overview:

**InputSystem-MobileFunctionalTests - 6000.5 - IOS**

This is a known problem that is unrelated to this branch, Wind is working on it at the moment.

**Test Sample Projects trunk**

It's caused by a surprising instability that is a compilation issue caused by some Timeline script upgrader shenanigans. Do we even depend on timeline though? 

Library/PackageCache/com.unity.timeline@6b9e48457ddb/Editor/TimelineHelpers.cs(31,20): error CS0234: The type or namespace name 'GUID' does not exist in the namespace 'UnityEditor' (are you missing an assembly reference?)

Nothing of that is related to my changes so I'm going to force-merge now.

PR #2285, opened 2025-11-18
FIX: Mitigate memory allocation driven by reflective introspection of binding types ISXB-1766

ekcoh:
### Description

Mitigates effects reported in ISXB-1766 (internal). This is not a fix but a mitigation (reducing number of scenarios when Reflection is used). A parallel track is being investigated in https://github.com/Unity-Technologies/InputSystem/pull/2288 relying on Roslyn source generators and some changes to type registration and management to avoid Reflection instead. This parallel track is less of a quick fix.

The goal is to reduce memory allocations from reflective type loading to only happen when an unresolved type is encountered in an input action asset. The drawback of this is that GC pressure may be postponed from outside intialization. The benefit is that no reflective type loading occurs if no custom type extensions are referenced from `.inputaction` assets, which would previously unconditionally load an introspect all assemblies unconditionally.

Some scenarios:
- In case PWIA or a regular input actions asset do not reference custom processors, interactions or composites, reflection based type registration never runs.
- In case PWIA references custom processors, interactions or composites, reflection based type registration is lazily trigger as part of the initialisation sequence of the Input System, and hence there is not much difference to before.
- In case regular input actions asset reference custom processors, interactions or composites, the reflection based type registration is triggered when the asset is resolved instead of during initialisation which could be harmful (GC pressure and slow execution). This can in turn be mitigated by doing explicit type registrations before enabling the actions which cause the reflection based type registration to never run.

A proper fix is also investigated in https://github.com/Unity-Technologies/InputSystem/pull/2288 (pending), with higher risk and halo effect.

### Testing status & QA

Only quickly verified with a custom processor type in a local project so far.

### Overall Product Risks

May have halo effects, e.g. on XRI.

- Complexity: small
- Halo Effect: large

### Comments to reviewers

This doesn't solve the reported behavior, only reduces to number of occasions when it happens.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2053,8 +2053,14 @@ void RegisterCustomTypes(Type[] types)
             }
         }
 
-        void RegisterCustomTypes()
+        internal bool hasCustomTypesBeenRegistered { get; private set; }
```
bmalrat: should it be also reset after UninstallGlobals or in the init to be sure ?

Packages/com.unity.inputsystem/InputSystem/Utilities/TypeTable.cs
```diff
@@ -83,7 +83,15 @@ public Type LookupTypeRegistration(string name)
             var internedName = new InternedString(name);
             if (table.TryGetValue(internedName, out var type))
                 return type;
-            return null;
+
+            // Failed to look-up type, either type do not exist or it is a custom type.
+            // Check whether we have attempted to load custom types and otherwise lazily load
+            // types only when referenced and reattempt looking up type by name. (ISXB-1766)
+            if (InputSystem.s_Manager == null || InputSystem.s_Manager.hasCustomTypesBeenRegistered)
```
bmalrat: that's doesn't seems clean to add a global reference to the supposed owner of it.
It breaks the isolation.
We may add to TypeTable a generic fallback call that the owner can configure.

Packages/com.unity.inputsystem/InputSystem/Utilities/TypeTable.cs
```diff
@@ -83,7 +83,15 @@ public Type LookupTypeRegistration(string name)
...
+                return null;
+
+            InputSystem.s_Manager.RegisterCustomTypes();
+            return LookupTypeRegistration(name);
```
bmalrat: maybe longer but internedName is already resolved so this should be enough
```
table.TryGetValue(internedName, out type);
return type;
```

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
ekcoh: Overlooked the hidden instance reference to InputManager hiding inside the static API actually using a static push/popped instance. Added a reset in https://github.com/Unity-Technologies/InputSystem/pull/2285/commits/858f1b7d1e0a86c7143d9f5727af27f82ae694eb

ekcoh: I agree it's safest to reset since type registration is bound to instate. The assembly type loading result is not.

Packages/com.unity.inputsystem/InputSystem/Utilities/TypeTable.cs
ekcoh: I agree, I added an InputManager reference instead passed via Initialize since it's always "owned" (at least managed) by InputManager. Its strong coupling, but there is currently no point of having an indirection (with a cost).

ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2285/commits/5cb4617686c194acc18850eca639f4460a594dcf

Packages/com.unity.inputsystem/InputSystem/Utilities/TypeTable.cs
ekcoh: I switched to `internedName` after also reducing the scope of the recursive call in https://github.com/Unity-Technologies/InputSystem/pull/2285/commits/5cb4617686c194acc18850eca639f4460a594dcf

MorganHoarau:
> This can in turn be mitigated by doing explicit type registrations before enabling the actions which cause the reflection based type registration to never run.

* Is that related to creating input custom processor? 
* Is this something user can do and if so, is that something documented for users? 
  * If so, is there documentation update needed to warm about the additional allocation cost if not done properly?

jfreire-unity (approved):
I'm approving this so it doesn't get blocked. I think it's the best approach to solve the issue in a short time. It's still unclear what the best solution will be going forward, so perhaps we should track that work after this lands.

My only question is about the steps to validate this. How can I validate (and QA as well) that this change doesn't allocate the +19MB of memory? Would there be a way to automate this and add a test for it?

ekcoh:
> > This can in turn be mitigated by doing explicit type registrations before enabling the actions which cause the reflection based type registration to never run.
> 
> * Is that related to creating input custom processor?
> * Is this something user can do and if so, is that something documented for users?
>   
>   * If so, is there documentation update needed to warm about the additional allocation cost if not done properly?

Yes, it's documented in the manual. It was easier to decipher before Processor page was edited in recent versions IMO, but it is at least in API docs linked from "Remarks" here: https://docs.unity3d.com/Packages/com.unity.inputsystem@1.16/api/UnityEngine.InputSystem.InputSystem.html
Before the "reflective solution" was added, this was the only way to do it which worked fine until Project-wide-input-actions was added since then the registration needs to happen before "SubsystemRegistration" which is not possible and hence the type registration changed to the "reflective solution" which drives the registration from within Input System initialisation instead of requiring the "user" to do it before enabling action assets.

bmalrat (approved):
sounds good to me to mitigate the issue

ekcoh:
> I'm approving this so it doesn't get blocked. I think it's the best approach to solve the issue in a short time. It's still unclear what the best solution will be going forward, so perhaps we should track that work after this lands.
> 
> My only question is about the steps to validate this. How can I validate (and QA as well) that this change doesn't allocate the +19MB of memory? Would there be a way to automate this and add a test for it?

IMO, the potential solution I can see which requires a bit of exploration would be to:
- Defer enabling PWIA until just before OnBeforeSceneLoad - might require custom hook since it needs to happen after SubsystemRegistration but before OnBeforeSceneLoad.
- Add a source generator that would pick up all related types via implemented interfaces and generate in-memory source that explicitly registers these types using a SubsystemRegistration hook. (Cannot be OnBeforeSceneLoad since it's too late). 
- Remove reflection type registration code since no longer needed.

See internal JIRA for more details.

bmalrat:
> I'm approving this so it doesn't get blocked. I think it's the best approach to solve the issue in a short time. It's still unclear what the best solution will be going forward, so perhaps we should track that work after this lands.
> 
> My only question is about the steps to validate this. How can I validate (and QA as well) that this change doesn't allocate the +19MB of memory? Would there be a way to automate this and add a test for it?

You may be able to check it within the profiler marker InputManager.RegisterCustomTypes which should not be called.

LeoUnity (approved):

Pauliusd01:
Not seeing any functionality issues yet but I'm having trouble profiling it. Hoping to finish by tomorrow morning

chris-massie:
This change seems to be fine for XR packages. We generally use `InputSystem.InputSystem.Register-` methods in static constructors/InitializeOnLoadMethod which should help with not needing to use reflection in the first place for custom input action assets.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -19,7 +19,9 @@ however, it has to be formatted properly to pass verification tests.
 ### Fixed
 - An issue where a UITK MouseEvent was triggered when changing from Scene View to Game View in the Editor has been fixed. [ISXB-1671](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1671)
 - Fix documentation error in file AndroidGameController.cs mentioning a wrong controller. [DOCATT-9806]
-
+- Deferred auto-registration of processors, interactions and composite binding types referenced by `InputActionAsset`
+  to only happen once when an unresolved type reference is found in an action definition. This avoids reflective
+  type loading from assemblies for all cases where the Input System is not extended. (ISXB-1766).
```
chris-massie, line 24:
```suggestion
- Deferred auto-registration of processors, interactions and composite binding types referenced by `InputActionAsset` to only happen once when an unresolved type reference is found in an action definition. This avoids reflective type loading from assemblies for all cases where the Input System is not extended. (ISXB-1766).
```
I don't think you should manually word wrap.

chris-massie (approved):

ekcoh:
> This change seems to be fine for XR packages. We generally use `InputSystem.InputSystem.Register-` methods in static constructors/InitializeOnLoadMethod which should help with not needing to use reflection in the first place for custom input action assets.

I also got this impression looking at the code but it doesn't help if you rely on PWIA, a potential solution for that is being looked at in parallel which might require some orchestration on both sides. But happy to hear you have kept your manual registration code. Lets keep that until there is clarity on how to properly handle this without reflective runtime operations long term.

Pauliusd01 (approved):
Honestly couldn't tell much of a difference when profiling different scenes, I didn't see any reflection code showing up with or without custom processors but the functionality seemed unaffected so it should be fine 🤞 Release testing will hopefully reveal anything missed

PR #2296, opened 2025-11-27
FIX: IndexOutOfRangeException when enabling InputActionMap. (ISXB-1767)

ekcoh:
### Description

This PR is a fix for a bug reported in ISXB-1767 (internal ticket). The reported problem is an `IndexOutOfRangeException` being thrown from `InputManagerStateMonitors.AddStateChangeMonitor` which leaves the Input System in an incomplete and inconsistent state preventing automatic recovery.

Recreating the exact scenario reported in the ticket was difficult since the reproduction case was a very large project, but I believe I have finally been able to recreate the exact scenario now in test case `Actions_CanHandleDeviceDisconnectWithControlSchemesAndReconnect` which is also added to this PR to prevent future regression. The scenario involves at least 2 devices being present and being associated with separate control schemes while triggering an action and disconnecting the associated device while disabling and reenabling an action map from the action cancellation callback.

The fix involves not allowing adding state monitors to a device with an invalid device index which prevents the problematic exception. In addition to throwing, the exception also has side-effects and leaves the system in an inconsistent state which leads to additional exceptions being thrown afterwards and prevents input bindings from going back into a resolved state. What is interesting is that the presence of control schemes is what creates this issue and it is not completely clear to me why this is essential.

### Testing status & QA

Added a more or less exact reproduction of the problematic scenario in `InputManagerStateMonitors.AddStateChangeMonitor`. I would also recommend verifying the final version of this PR with the repro project to make sure it's properly solved.

Found another issue while testing this bug that I filed separately into ISXB-1776.

### Overall Product Risks

- Complexity: Medium - Central logic for detecting input and enabling actions modified
- Halo Effect: Medium - Might have implication on any use of actions

### Comments to reviewers

Recommend looking for other aspects or side-effects I might have overlooked. Additional insights of the control scheme aspect with relation to this bug is valuable.

To manually trigger the problem do the following reproduction steps:
1. Setup an .inputactions asset with bindings for multiple devices, e.g. keyboard and gamepad including associated control schemes. Simplest is just setting Project-wide Input Actions up with the default created asset since it already contain these ingredients.
2. Create a script that disables and enables an action map associated with the actions being tested, e.g. "Player" map, from within the cancellation callback for an action, e.g. "Jump" action. See example below:
```cs
using UnityEngine;

public class Trigger : MonoBehaviour
{
    // Assumes "Jump" binding is used to trigger the problematic behavior.
    void Start()
    {
        var map = InputSystem.actions.FindActionMap("Player");
        var jump = map.FindAction("Jump");
        jump.performed += (ctx) => Debug.Log("Jump Performed");
        jump.canceled += ctx =>
        {
            map.Disable();
            map.Enable(); // <-- Triggers exception
        };
    }
}
```
If using `ProjectWideInputActionsSampleScene` the above script can just be attached to the "Cube" object or any other object.
3. Connect a gamepad, e.g. DualSense via USB cable (important since step 6 will be difficult otherwise). 
4. Enter play-mode.
5. Press Jump button (E.g. "X" on a DualSense) and keep it pressed to ensure that the action is still in performed state.
6. With your free hand, disconnect the USB cable from the device while still holding "X".
7. Observe failed assert and exceptions in console. It seems the Input System cannot recover from this even if device is reconnected. With this PR the system shall both handle the error and auto-resolve bindings when device is reconnected.

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1153,6 +1153,13 @@ private void EnableControls(int mapIndex, int controlStartIndex, int numControls
                 if (IsControlEnabled(controlIndex))
                     continue;
 
+                // We might en up here if an action map is enabled from e.g. an event processing callback such as
```
ekcoh:
```suggestion
                // We might end up here if an action map is enabled from e.g. an event processing callback such as
```

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -17,6 +17,7 @@
 using UnityEngine.InputSystem.Processors;
 using UnityEngine.InputSystem.Utilities;
 using UnityEngine.InputSystem.XInput;
+using UnityEngine.PlayerLoop;
```
jfreire-unity: Is this needed?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -12581,4 +12582,67 @@ public void Actions_ActionMapDisabledDuringOnAfterSerialization()
...
+
+        var action = map.AddAction(name: "Toggle", InputActionType.Button);
+        action.AddBinding("<Gamepad>/leftTrigger");
+        action.started += context => ++ started;
+        action.performed += context => ++ performed;
+        action.canceled += (context) =>
+        {
+            // In reported issue, map state is changed from cancellation callback.
+            map.Disable();
+            map.Enable();
```
jfreire-unity, line 12614: Interesting that this is done in the project. Do you know what's the goal with this?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1153,6 +1153,13 @@ private void EnableControls(int mapIndex, int controlStartIndex, int numControls
                 if (IsControlEnabled(controlIndex))
                     continue;
 
+                // We might end up here if an action map is enabled from e.g. an event processing callback such as
+                // InputAction.cancel event handler (ISXB-1766). In this case we must skip controls associated with
+                // a device that is not connected to the system (Have deviceIndex < 0). We check this here to not
+                // cause side effects if aborting later in the call-chain.
+                if (!controls[controlIndex].device.added)
```
jfreire-unity, line 1160: Could a similar thing happen if a device were disabled?

jfreire-unity (approved):
Nice one-line fix with test reproducing the issue 🎉 
Added some minor comments but no blockers.

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: I wouldn't think so must be a mistake.

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: I reply in DM on this one.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
ekcoh: I thought so which is why I extended the test to disable, enable, disable, enable but this seems to be fine. I have to admit why this is not clear and would also feel better by doing this symmetric. Should we explore making it symmetric or what do you think?

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh:
```suggestion
```

ekcoh: Removed it

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs
```diff
@@ -1153,6 +1153,13 @@ private void EnableControls(int mapIndex, int controlStartIndex, int numControls
                 if (IsControlEnabled(controlIndex))
                     continue;
 
+                // We might end up here if an action map is enabled from e.g. an event processing callback such as
+                // InputAction.cancel event handler (ISXB-1766). In this case we must skip controls associated with
```
ekcoh:
```suggestion
                // InputAction.cancel event handler (ISXB-1767). In this case we must skip controls associated with
```

Pauliusd01:
Just updating that I am still looking at this, potentially found issues but need to nail down the repro first

Pauliusd01 (changes requested):
Found an issue with the user repro project, steps are: 

Complete steps 2-11 from the ISXB-1767 bug -> instead of disconnecting the gamepad in isolation use WASD keyboard keys to move your character while you do it -> the gamepad disconnects and the console gets spammed with errors for each keyboard key pressed -> this persists even if you leave and reenter play mode

I wasn't able to reproduce this on a new project or without your changes so seems like a legit bug to me. I will paste the full errors to you over slack

ekcoh:
Thanks for describing the repro steps @Pauliusd01. I need to investigate and also reproduce that scenario with automated tests I believe to see if we can replicate in CI setting. Suspect it relates to control scheme switching in the middle of disconnect.

Pauliusd01 (approved):
LGTM, checked basic input samples, disconnecting keyboard/gamepad and checking the debugger for unexpected problems, disconnecting devices on the user project (user steps, while using a different device, disconnecting more than one device, etc)

PR #2303, opened 2025-12-15
FIX: InputActionReference not set on Prefab and ScriptableObject (ISBX-1787)

MorganHoarau:
### Description

This PR addresses a regression caused by [this change](https://github.com/Unity-Technologies/InputSystem/pull/2248/files?file-filters%5B%5D=.cs&show-viewed-files=true#diff-862d86d5ea01423b7c307f08579cb5538da6de4ab18aebb083223ed10c872bb6R67). 

The root cause of the issue was that the new implementation of the `InputActionReferencePropertyDrawer` was trying to assigned a virtual input asset ref (not yet registered in asset database) to the `SerializedObject.objectReferenceValue`. This is not supported as the field needs to be assigned to an asset written on the disk.

The fix is to assign directly the retrieved asset reference. 

* [x] Jira ticket: [ISXB-1787](https://jira.unity3d.com/browse/ISXB-1787)
* [x] CHANGELOG.md

### Testing status & QA

Used the following minimal reproduction steps:
1. Create a new project
2. Create a new MonoBehaviour and ScriptableObject that both implement a public `InputActionReference` field.
3. Try to assign the an input action to the field.

**Result:**
fixes-inputactiondrawer

**No automated tests added as testing property drawers directly is challenging because they require GUI interaction.**

### Overall Product Risks

- Complexity: Low
- Halo Effect: low

### Comments to reviewers

### Checklist

Before review:

- [X] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -14,6 +14,7 @@ however, it has to be formatted properly to pass verification tests.
 
 - Fixed warnings being generated on Unity 6.4 and 6.5. (ISX-2395).
 - Fixed extra empty lines being displayed in the control binding list when mouse buttons are pressed [ISXB-1677](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1677)
+- Fixed InputActionReference not being set when attempting to set it on a Prefab or ScriptableObject [ISXB-1787](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1787)
 
 ### Changed
 
--- a/Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/InputActionReferencePropertyDrawer.cs
@@ -48,9 +48,7 @@ namespace UnityEngine.InputSystem.Editor
             // Only assign the value was actually changed by the user.
             if (EditorGUI.EndChangeCheck() && !Equals(candidate, current))
             {
-                var reference = candidate as InputActionReference;
-                property.objectReferenceValue = reference ?
-                    InputActionReference.Create(reference.action) : null;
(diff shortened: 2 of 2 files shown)


ekcoh:
I noticed you ticked the "Test added/changed" on this PR but I cannot see you have added any automated tests to verify the prefab behaviour? Forgot to commit tests or accidentally checked checkbox?

ekcoh (approved):
Thanks for fixing this mistake I introduced (and quickly spotting the problem as well) @MorganHoarau. I crossed checked this against earlier changes and I think we should still be good. This change implies that we pick an actual `InputActionReference` without creating a copy from the editor which makes a lot of sense. It will however have impact on anyone trying to programatically resetting the reference without assigning a new InputActionReference SO instance (the anti-pattern I tried to address in previous / linked PR). In that case, the checks I added should still capture that anti-pattern and avoid corrupting the object at the source, but users needs to mindful to replace any reference with a newly created instance instead.

The only thing I am missing with this PR is an automated test to verify behaviour with prefabs to make sure we avoid future regressions in this area. I do not think it needs to block the fix from landing, but it would be great to add an automated test verifying the exact scenario. Do you think that would be feasible to add to e.g. https://github.com/Unity-Technologies/InputSystem/blob/develop/Assets/Tests/InputSystem.Editor/InputActionReferenceEditorTests.cs? I assume it might be possible to test by assigning via SerializedObject similar to what the editor does in case I overlooked it.

ekcoh:
@MorganHoarau Seems like you need to sign CLA for your user (see failed CI check)

MorganHoarau:
I ticked it because I added that "No tests were added" in the testing status, but didn't realized it was just for case were tests were added/changed. 

As the repro requires user input to cause the property drawer interaction, I don't think there is a way to automate it. But maybe there is? 

I just saw your second comment. I will have a look at the test file.

Pauliusd01 (approved):

PR #2311, opened 2026-01-07
FIX: Handle consecutive wildcards in StringMatches (1/x)

josepmariapujol-unity:
### Description

Normalize consecutive wildcards ('*' /  '**') semantics are handled correctly. Consecutive wildcard characters ('*') used in input control-paths are now collapsed into a single wildcard when multiple consecutive wildcard characters are present.

NOW, if we have multiple **** result the same behaviour than only one *
BEFORE, multiple **** result in a different behaviour than only one *

JIRA: [ISX-2467](https://jira.unity3d.com/browse/ISX-2467)

### Testing status & QA

Type multiple ****, should behave the same way as if you typed only one *.
First editor contains PR changes, second editor uses old input system version (hence, not working with multiple ***).

https://github.com/user-attachments/assets/f3a4907a-97bb-41b4-9105-d25c323644bb

### Overall Product Risks

- Complexity: 1
- Halo Effect: 1

### Comments to reviewers

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


ekcoh (changes requested):
Change looks good to me but I would suggest making sure this is covered in a unit test. If this logic currently lacks any unit test I would suggest adding at least one covering this specific scenario.

ekcoh (commented):
Another thing to address, there is no ticket trace nor any CHANGELOG entry for this fix, please add that.

Assets/Tests/InputSystem/CoreTests_Controls.cs
```diff
@@ -1607,4 +1607,30 @@ static void CheckChildControls(Type type, HashSet<Type> checkedTypes)
             Assert.That(setMethod.IsPublic, Is.True, inputControlMessage);
         }
     }
+
+    [Test]
```
ekcoh, line 1611: Thanks for adding this test. This makes the intent clearer and avoids regressions.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ however, it has to be formatted properly to pass verification tests.
 ## [Unreleased] - yyyy-mm-dd
 
 ### Changed
-
+- Updated handle consecutive wildcards in StringMatches in InputControlPath.cs.
```
ekcoh: Fine as-is, but maybe can be more crisp for actual effect? E.g. "Consecutive wildcard characters ('*') used in input control-paths are now collapsed into a single wildcard when multiple consecutive wildcard characters are present". Or what do you think?

ekcoh (approved):

Packages/com.unity.inputsystem/CHANGELOG.md
josepmariapujol-unity: Much better 👍

Pauliusd01:
Did not forget about this one, taking a look

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ however, it has to be formatted properly to pass verification tests.
 ## [1.18.0] - 2026-01-14
 
 ### Changed
-
+- Consecutive wildcard characters ('*') used in input control-paths are now collapsed into a single wildcard when multiple consecutive wildcard characters are present.
```
Pauliusd01: This is under the wrong version

Pauliusd01:
image

Just an FYI in case it is relevant but if you use `/*/* ` you get these two errors. It is the same in develop but thought I'd mention it anyway.

Pauliusd01 (approved):

PR #2321, opened 2026-01-12
CHANGE: Split Runtime and Editor code into separate folders to prepare for Module migration

jfreire-unity:
### Background & Goal

_NOTE 🤯: This PR is **huge** and hard to review, but a lot of the changes are moving files and folders around. I know it's not easy to navigate these huge changes so I tried my best to do a good PR description._

The Unity Input System package historically had `UnityEditor.*` references scattered throughout its runtime code, guarded by `#if UNITY_EDITOR` preprocessor directives, but still in the same files as runtime logic. 

This was making the codebase hard to port into a Unity module in trunk (`unity/unity`), and it was also harder to reason about the Editor vs Standalone player boundary.

This PR tries to break the dependency of Editor references in Runtime code to help us migrate to a Unity module and does the following changes:

- **Zero `UnityEditor.*` references remain in the runtime source files.**
- All editor-specific logic has been moved to dedicated editor-side files or wired up via internal callback hooks.
- The **single `Unity.InputSystem` assembly is kept** . The folder reorganisation (`Runtime/` and `Editor/` subfolders) is the externally visible result, making forward-ports and moving `git` history to trunk and eventual module migration easier. We decided not to introduce 2 assemblies as this would duplicate our upgrade path when projects upgrade to a Unity version using the new Input System module.

Ideally, we would have liked to also have a Runtime and Editor assembly. 
That work was initially done but then we reverted it due to the last point above.

This PR was done with the help of agentic coding as well.

---

### High-Level Changes

| Area | Change |
|------|--------|
| **Folder structure** | All runtime source moved to `InputSystem/Runtime/`, all editor source to `InputSystem/Editor/`. 1,201 files renamed (history preserved via `git mv`). |
| **Assembly** | `Unity.InputSystem.asmdef` kept as single monolith. A `Unity.InputSystem.Editor` asmdef was prototyped and then rolled back — the folder split remains as foundation for a future module migration. |
| **Runtime → Editor communication** | Replaced every direct `UnityEditor.*` call in runtime files with internal `static` callback fields (`Action`, `Func`). The Editor registers these at startup via `[InitializeOnLoad]`. In player builds these fields are `null` and all call sites are null-safe (`?.Invoke()`). |
| **New file: `InputSystemEditorInitializer.cs`** | Central `[InitializeOnLoad]` class in `Editor/` that owns all editor lifecycle concerns previously scattered across the runtime (play-mode changes, domain reload, asset tracking, analytics, Unity Remote, remoting, etc.). |
| **New file: `InputSystemState.cs`** | Extracted domain-reload state (previously coupled to `InputSystemObject` and editor-specific paths) into its own runtime class to help with the dependency boundary. |
| **Editor classes extracted** | 13 `*Editor` classes (Interaction editors, Composite editors, Processor editors, OnScreen editors) removed from runtime files and placed in dedicated files under `Editor/`. |
| **Editor-only files relocated** | 11 files that were already fully wrapped in `#if UNITY_EDITOR` moved to `Editor/` (e.g. `PlayerInputEditor`, `UnityRemoteSupport`, `iOSPostProcessBuild`). |
| **Test fixtures updated** | `InputTestRuntime`, `InputTestStateManager` updated to reflect the new interface contracts (e.g. `onPlayModeChanged: Action<int>` instead of `Action<PlayModeStateChange>`). |

---

### Pattern used for these changes

The approach used throughout this PR is maybe worth understanding before reviewing individual files:

```csharp
// Runtime side — declares an internal hook, never calls UnityEditor
internal static Action<InputActionAsset> s_OnActionsChanging;

// usage — null-safe, works in player (null) and editor (set)
s_OnActionsChanging?.Invoke(asset);

// Editor side — InputSystemEditorInitializer registers at load time
[InitializeOnLoad]
static class InputSystemEditorInitializer
{
    static InputSystemEditorInitializer()
    {
        InputSystem.s_OnActionsChanging = ValidateAndTrackActions;
        // ... all other hooks ...
    }
}
```

This pattern appears in: `InputSystem.cs`, `InputManager.cs`, `NativeInputRuntime.cs`, `InputActionSetupExtensions.cs`, `InputActionAsset.cs`, `InputActionReference.cs`, `EnhancedTouchSupport.cs`, `RemoteInputPlayerConnection.cs`, `OnScreenStick.cs`, `TrackedPoseDriver.cs`, `InputSystemUIInputModule.cs`.

⚠️ ❗❗❗ Feel free to let me know if you disagree with this approach and if you have a better one. I'm open to suggestions ❗❗❗ ⚠️

---

### Review Guide Suggestions

**1. `InputSystem/Runtime/InputSystem.cs`**
- Confirm `using UnityEditor;` is gone.
- Check the four new internal hooks (`s_OnSettingsChanged`, `s_OnActionsChanging`, `s_ShouldEnableActions`, `s_OnPlayModeChangeCallback`) are used consistently and null-safely.
- The `#if UNITY_EDITOR` block for `OnPlayModeChange` is intentional — it is a forwarding shim for tests that call it directly.

**2. `InputSystem/Runtime/NativeInputRuntime.cs`**
- Previously contained direct `EditorApplication.isPlaying`, `EditorApplication.isPaused`, `InternalEditorUtility.isApplicationActive`, and `EditorAnalytics` calls.
- Now uses plain `internal bool` fields (`m_IsInPlayMode`, `m_IsEditorPaused`, `m_IsEditorActive`) and callbacks (`m_SendEditorAnalytic`, `m_SetUnityRemoteMessageHandler`, etc.).
- Verify the `UpdateEditorState()` method in `InputSystemEditorInitializer` is pumping those fields on `EditorApplication.update`.

**3. `InputSystem/Runtime/InputManager.cs`**
- Replaced `EditorApplication.isPaused` → `m_Runtime.isEditorPaused` (from `IInputRuntime`).
- `InputSystem.s_SystemObject` references replaced with local `m_ExitEditModeTime` / `m_EnterPlayModeTime` fields.
- `ProjectWideActionsBuildProvider.actionsToIncludeInPlayerBuild` replaced with `s_GetProjectWideActions` callback.

**4. `InputSystem/Runtime/IInputRuntime.cs`**
- `onPlayModeChanged` changed from `Action<PlayModeStateChange>` to `Action<int>` — removes the `UnityEditor` enum from the interface.
- Added `bool isEditorPaused { get; }`.

**5. `InputSystem/Editor/InputSystemEditorInitializer.cs`** *(new file)*
- This is the single source of truth for all editor-runtime "wiring". Verify that the callback registered here have a corresponding field in the runtime.
- Check play-mode lifecycle: `OnEditorPlayModeStateChanged` → sets `m_IsInPlayMode`, `m_IsEditorPaused`, dispatches `DispatchPlayModeChange`, etc.
- Check domain-reload path: `InputSystemState` is loaded/restored here.

**6. Extracted `*Editor` classes** (`Editor/Actions/Interactions/`, `Editor/Actions/Composites/`, `Editor/Controls/Processors/`, `Editor/Plugins/OnScreen/`)
- These were verbatim moves from their parent runtime files — no logic changes.
- Maybe worth a quick sanity check that nothing is missing.

**7. `InputSystem/Runtime/Actions/InputActionReference.cs`, `InputActionAsset.cs`, `EnhancedTouchSupport.cs`**
- Light-touch changes: direct `AssetDatabase`/`AssemblyReloadEvents` calls replaced with callback fields.
- Confirm the callback fields are `#if UNITY_EDITOR`-guarded.

**8. Test fixtures — `Tests/TestFixture/`**
- `InputTestRuntime.cs`: `onPlayModeChanged` is now `Action<int>` — make sure test call sites cast correctly.
- `InputTestStateManager.cs`: there were some naming changes, so something to be aware.

**9. Assembly / `.asmdef` files**
- `Unity.InputSystem.asmdef` should still reference a single assembly (no Editor split).
- Downstream asmdefs (samples, test projects) should all still point to `Unity.InputSystem` only.
- `AssemblyInfo.cs`: `InternalsVisibleTo` entries cover test assemblies — verify no entry was accidentally dropped.
- A new .asmdef was created for `InputSystemForUI.Editor.asmdef`, since the previous folder containing `InputSystemForUI.asmdef` was split into Editor and Runtime. This is part of internal API so I don't see any breakages in this area when upgrading.

### Testing status & QA

Built and run Samples on both Windows and macOS Play-mode and Standalone builds.

I recommend validating all the Samples work and do kind of the sames tests as we would do for a Release since this PR touches a lot of things. 

### Overall Product Risks

- Complexity: Medium (the main changes are straight forward but the fact that are so many changes maybe overwhelming to rewview) 
- Halo Effect: High

### Comments to reviewers

Some sanity check guides:

- [ ] Zero `using UnityEditor;` in any file under `InputSystem/Runtime/`
- [ ] `InputSystemEditorInitializer` registers every callback that has a field on the runtime side
- [ ] `InputTestRuntime` compiles and tests pass with the updated `Action<int>` signature
- [ ] No asmdef downstream of `Unity.InputSystem` was accidentally wired to a now-deleted `Unity.InputSystem.Editor` assembly. Hopefully there are not left overs of this initial work.

One thing is not in place is a mechanism to guarantee we don't leak UnityEditor references in the Runtime folder. I'll follow up on that in another PR.

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/Editor/Plugins/HID/HIDSupportEditorInitializer.cs
```diff
@@ -0,0 +1,34 @@
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.InputSystem.Editor;
+using UnityEngine.InputSystem.HID.Editor;
+
+namespace UnityEngine.InputSystem.HID
```
jfreire-unity: This should probably be changed to `UnityEngine.InputSystem.HID.Editor`

Packages/com.unity.inputsystem/Runtime/NativeInputRuntime.cs
```diff
@@ -84,6 +80,8 @@ public unsafe InputUpdateDelegate onUpdate
                         {
                             eventBufferPtr->eventCount = 0;
                             eventBufferPtr->sizeInBytes = 0;
+                            eventBufferPtr->capacityInBytes = 0;
+                            eventBufferPtr->eventBuffer = (void*)0;
```
josepmariapujol-unity: Set to "0" eventBufferPtr->capacityInBytes = 0; and eventBufferPtr->capacityInBytes.

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
```diff
@@ -258,7 +267,7 @@ public bool runInBackground
         // In older version this is stored here and package override module/platform.
         private float m_PollingFrequency = 60.0f;
         #endif
-        private bool m_DidCallOnShutdown = false;
+        private bool m_DidCallOnShutdown;
```
josepmariapujol-unity, line 295: By default equals `false`

Packages/com.unity.inputsystem/Editor/InputSystemObject.cs
josepmariapujol-unity: Shouldn't it be `namespace UnityEngine.InputSystem.Editor`?

josepmariapujol-unity:
Added: Create ModuleTypeForwarders.cs for Input [#93627](https://github.cds.internal.unity3d.com/unity/unity/pull/93627)

Packages/com.unity.inputsystem/Runtime/InputManager.cs
```diff
@@ -4303,7 +4303,13 @@ internal void RestoreStateWithoutDevices(SerializedState state)
...
             #endif
 
-            ////REVIEW: instead of accessing globals here, we could move this to when we re-create devices
+            // Restore available devices list so that GetUnsupportedDevices() and similar queries
+            // return the correct result before RestoreDevicesAfterDomainReload() runs.
+            /*if (state.availableDevices != null && state.availableDevices.Length > 0)
+            {
+                m_AvailableDevices = state.availableDevices;
+                m_AvailableDeviceCount = state.availableDevices.Length;
+            }*/
```
jfreire-unity: Is this to be removed or?

Packages/com.unity.inputsystem/Runtime/NativeInputRuntime.cs
```diff
@@ -103,6 +99,8 @@ public unsafe InputUpdateDelegate onUpdate
...
```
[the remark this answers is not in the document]
jfreire-unity: Seems about right. This shouldn't be done just because the events count is zero.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
```diff
@@ -216,13 +216,13 @@ public static Command DeleteSelectedControlScheme()
                 if (indexOfArrayElement > serializedArray.arraySize - 1)
                     return state.With(
                         selectedControlSchemeIndex: serializedArray.arraySize - 1,
-                        selectedControlScheme: new InputControlScheme(serializedArray.GetArrayElementAtIndex(serializedArray.arraySize - 1)), selectedDeviceRequirementIndex: -1);
+                        selectedControlScheme: InputControlSchemeEditorExtensions.FromSerializedProperty(serializedArray.GetArrayElementAtIndex(serializedArray.arraySize - 1)), selectedDeviceRequirementIndex: -1);
```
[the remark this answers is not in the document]
jfreire-unity: @u-pr what do you suggest as a fix?

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystem.cs
```diff
@@ -57,18 +51,16 @@ namespace UnityEngine.InputSystem
...
-#if UNITY_EDITOR
-        static readonly ProfilerMarker k_InputInitializeInEditorMarker = new ProfilerMarker("InputSystem.InitializeInEditor");
-#endif
-
         static InputSystem()
         {
+            // [InitializeOnLoad] only works in the editor and is managed by the static constructor in
+            // InputSystemEditorInitializer class.
+            // Here we only need to make sure that the static constructor is called in the player.
+            #if !UNITY_EDITOR
```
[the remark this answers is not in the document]
jfreire-unity: @u-pr what's your suggestion to fix this?

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
```diff
@@ -187,7 +187,7 @@ public InputActionsEditorState(InputActionsEditorState other, SerializedObject a
             {
                 if (m_selectedControlSchemeIndex >= controlSchemesArrayProperty.arraySize)
                     m_selectedControlSchemeIndex = 0;
-                m_ControlScheme = new InputControlScheme(
+                m_ControlScheme = InputControlSchemeEditorExtensions.FromSerializedProperty(
                     controlSchemesArrayProperty.GetArrayElementAtIndex(other.m_selectedControlSchemeIndex));
```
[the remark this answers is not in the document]
jfreire-unity: I won't change this in this PR, but something to look into.

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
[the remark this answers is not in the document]
jfreire-unity: @u-pr was this problem not there before the change this PR introduces?

Assets/Tests/InputSystem/CoreTests_Remoting.cs
```diff
@@ -268,13 +268,8 @@ public void Remote_ResettingDevice_WillSendChangeToRemotes()
...
 #if UNITY_EDITOR
-        // In the editor, RemoteInputPlayerConnection is a scriptable singleton. Creating multiple instances of it
-        // will cause an error messages - but will work nevertheless, so we expect those errors to let us run the test.
-        // We call RemoteInputPlayerConnection.instance once to make sure that we an instance is created, and we get
-        // a deterministic number of two errors.
-        var instance = RemoteInputPlayerConnection.instance;
-        UnityEngine.TestTools.LogAssert.Expect(LogType.Error, "ScriptableSingleton already exists. Did you query the singleton in a constructor?");
-        UnityEngine.TestTools.LogAssert.Expect(LogType.Error, "ScriptableSingleton already exists. Did you query the singleton in a constructor?");
+        // Ensure the singleton is initialized deterministically in editor.
+        _ = RemoteInputPlayerConnection.instance;
```
K-Tone: what is the the underscore thing here in this expression?

Darren-Kelly-Unity (approved):

Packages/com.unity.inputsystem/InputSystem/Editor/PropertyDrawers/GamepadButtonPropertyDrawer.cs
```diff
@@ -106,4 +106,4 @@ private int GetEnumIndex(int enumValue)
         private string[] m_EnumDisplayNames;
     }
 }
- #endif // UNITY_EDITOR
+#endif
```
VeraMommersteeg: would still be good to keep the  // UNITY_EDITOR behind this, so its easier to see what this endif belongs too

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionReference.cs
VeraMommersteeg: wouldnt it make more sense to move this into some form of editor utility file and then only wrap the callsite in an ifdef for now. that way it will be easier to deal with in the future

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionReference.cs
```diff
@@ -266,21 +273,25 @@ private void CheckImmutableReference()
             // editor and not in final builds. The alternative would be to set a non-serialized field on the reference
             // when importing assets which would simplify this class, but it adds complexity to import stage and
             // is more difficult to assess from a asset version portability perspective.
-            static bool CanSetReference(InputActionReference reference)
+            bool CanSetReference(InputActionReference reference)
```
VeraMommersteeg: why do we still need a function inside a function? this is no longer static

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionSetupExtensions.cs
```diff
@@ -323,13 +323,35 @@ public static BindingSyntax AddBinding(this InputAction action, string path, str
...
+            ControlSchemeWithBindingGroup,
+            ControlSchemeWithDevice,
+            ControlSchemeWithRequiredDevice,
+            ControlSchemeWithOptionalDevice,
+            ControlSchemeOrWithRequiredDevice,
+            ControlSchemeOrWithOptionalDevice
         }
 
-        #endif
+        private static void RegisterApiUsage(int api)
```
VeraMommersteeg: This function doesnt properly describe what it does. RegisterAnalyticsApi would be clearer. before you could see that it was an analytics call being registered, but now RegisterApiUsage((int)Api.AddBinding) doesnt say much

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystem.cs
```diff
@@ -3485,227 +3482,7 @@ internal static void EnsureInitialized()
...
-                    }
-
-                    // reload input assets marked as dirty from disk
-                    DirtyAssetTracker.ReloadDirtyAssets();
-                    break;
-            }
-        }
-
-        // We have this function to hide away instanceId -> entityId migration that happened in Unity 6.3
-        public static bool HasNativeObject(Object obj)
```
jfreire-unity, line 3679: We have removed this API so we need to reintroduce-it somehow/deprecate it.

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/OnScreen/OnScreenButton.cs
```diff
@@ -41,41 +37,13 @@ public void OnPointerDown(PointerEventData eventData)
 
         [InputControl(layout = "Button")]
         [SerializeField]
-        private string m_ControlPath;
+        internal string m_ControlPath;
```
VeraMommersteeg, line 40: was there a reason to make this internal vs private?

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/UI/InputSystemUIInputModule.cs
```diff
@@ -1606,19 +1602,19 @@ public InputActionReference trackedDeviceSelect
 
 #if UNITY_EDITOR
         /// <inheritdoc/>
-        protected override void Reset()
+        protected override         void Reset()
```
VeraMommersteeg: formatting

Packages/com.unity.inputsystem/InputSystem/Runtime/IInputRuntime.cs
```diff
@@ -187,21 +186,26 @@ internal unsafe interface IInputRuntime
 
         // If analytics are enabled, the runtime receives analytics events from the input manager.
         // See InputAnalytics.
-#if UNITY_ANALYTICS || UNITY_EDITOR
+        #if UNITY_ANALYTICS || UNITY_EDITOR
```
VeraMommersteeg: nitpick: can we instead keep all the preprocessor defines without indent? I know there isnt a coding standard for preprocessor formatting in unity, but now there are places in the codebase where they are indented, and other places they are not. I'm personally on the side of no indent, as it allows you to quickly see what is inside a preprocessor block, while when indented, this might be easier to miss

Packages/com.unity.inputsystem/InputSystem/Runtime/InputAnalytics.cs
```diff
@@ -13,6 +10,9 @@ internal static class InputAnalytics
     {
         public const string kVendorKey = "unity.input";
 
+        public static Func<bool> s_IsNewSystemBackendsEnabled { get; set; }
```
VeraMommersteeg: all public properties in unity are pascal case

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
```diff
@@ -187,26 +183,37 @@ public Action<int, string> onDeviceDiscovered
             set => NativeInputSystem.onDeviceDiscovered = value;
         }
 
+        // Callbacks set by Editor to handle shutdown subscription
+        // In Editor, we use EditorApplication.wantsToQuit which expects Func<bool>
+        internal Action<Func<bool>> m_RegisterWantsToQuit;
```
VeraMommersteeg, line 188: why not make this private as this only needs to exist in the current scope

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestFixture.cs
```diff
@@ -668,7 +668,7 @@ public void EndTouch(int touchId, Vector2 position, float pressure, Vector2 delt
         public void CancelTouch(int touchId, Vector2 position, Vector2 delta = default, bool queueEventOnly = false,
             Touchscreen screen = null, double time = -1, double timeOffset = 0)
         {
-            SetTouch(touchId, TouchPhase.Canceled, position, delta, queueEventOnly: queueEventOnly, screen: screen, time: time, timeOffset: timeOffset);
+            SetTouch(touchId, TouchPhase.Canceled, position, 1, delta, queueEventOnly: queueEventOnly, screen: screen, time: time, timeOffset: timeOffset);
```
VeraMommersteeg: what does 1 mean here? can you change it instead to include the named argument like the other ones?

Packages/com.unity.inputsystem/InputSystem/Runtime/IInputRuntime.cs
```diff
@@ -187,21 +186,26 @@ internal unsafe interface IInputRuntime
...
+        #endif // UNITY_ANALYTICS || UNITY_EDITOR
 
-#if UNITY_EDITOR
-        Action<PlayModeStateChange> onPlayModeChanged { get; set; }
+        #if UNITY_EDITOR
+        /// <summary>
+        /// Callback for play mode state changes. The int parameter corresponds to PlayModeStateChange enum values:
+        /// 0 = EnteredEditMode, 1 = ExitingEditMode, 2 = EnteredPlayMode, 3 = ExitingPlayMode
+        /// </summary>
+        Action<int> onPlayModeChanged { get; set; }
```
VeraMommersteeg: why not keep the enum instead, and have one in runtime?

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestRuntime.cs
```diff
@@ -418,7 +419,7 @@ public void SetUnityRemoteGyroUpdateInterval(float interval)
             unityRemoteGyroUpdateInterval = interval;
         }
 
-        public Action<PlayModeStateChange> onPlayModeChanged { get; set; }
+        public Action<int> onPlayModeChanged { get; set; }
```
VeraMommersteeg: why was this changed to an int? why not have an enum in runtime?

VeraMommersteeg (changes requested):
Overall looks good, I have a few questions and some minor feedback. I noticed there were quite a few changes unrelated to the editor/runtime split. It might've been better to do those changes in a separate pr to create less noise.

ekcoh:
@u-pr "PR Reviewer Guide" seems outdated. Can you update it based on latest commit or can this only be done manually by author @jfreire-unity?

josepmariapujol-unity:
@copilot resolve the merge conflicts in this pull request

Packages/com.unity.inputsystem/InputSystem/Editor/InputSystemEditorInitializer.cs
```diff
@@ -0,0 +1,600 @@
...
+
+                InputEditorUserSettings.s_Settings = new InputEditorUserSettings.SerializedState();
+                InputSystem.onSettingsChange += OnSettingsChanged;
+
+                #if !UNITY_DISABLE_DEFAULT_INPUT_PLUGIN_INITIALIZATION
+                InputSystem.PerformDefaultPluginInitialization();
+                #endif
+            }
+
+            var existingStateManagers = Resources.FindObjectsOfTypeAll<InputSystemStateManager>();
```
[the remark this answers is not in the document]
jfreire-unity: This is part of the FEPM PR and outside of the scope of this PR.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputAnalytics.cs
```diff
@@ -13,6 +10,9 @@ internal static class InputAnalytics
     {
         public const string kVendorKey = "unity.input";
 
+        public static Func<bool> IsNewSystemBackendsEnabled { get; set; }
+        public static Func<bool> IsOldSystemBackendsEnabled { get; set; }
```
jfreire-unity: This should probably be internal, as we don't need to expose this to users.

Packages/com.unity.inputsystem/Runtime/NativeInputRuntime.cs
jfreire-unity: This is not good, will be reverted.

jfreire-unity:
@u-pr can you check which review problems you mentioned have been addressed?

Packages/com.unity.inputsystem/InputSystem/Editor/InputSystemEditorInitializer.cs
```diff
@@ -0,0 +1,602 @@
...
+        }
+
+        private static System.Reflection.MethodInfo GetUnityRemoteAPIMethod(string methodName)
+        {
+            var editorAssembly = typeof(EditorApplication).Assembly;
+            var genericRemoteClass = editorAssembly.GetType("UnityEditor.Remote.GenericRemote");
+            if (genericRemoteClass == null)
+                return null;
+
+            return genericRemoteClass.GetMethod(methodName);
```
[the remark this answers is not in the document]
jfreire-unity: This won't be addressed in this PR as it's out of scope.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionReference.cs
jfreire-unity: Good point. We probably could, but I'm fine with keeping it here since I think we don't have more than 1 of utility for InputActionReference. 
Also, this logic was already here, so I would propose we don't change this (there's already plenty of scope 🫠 ). At least, I tried my best not to expand the scope of the changes necessary to achieve this "code split".

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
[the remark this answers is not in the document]
jfreire-unity: Out of scope for this PR, I'll add this to be fixed in https://jira.unity3d.com/browse/ISX-2555

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/InputActionsEditorState.cs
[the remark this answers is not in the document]
jfreire-unity: Out of scope for this PR, I'll add this to be fixed in https://jira.unity3d.com/browse/ISX-2555

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
```diff
@@ -4335,11 +4335,13 @@ internal struct GlobalState
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.BeforeSceneLoad)]
         private static void InitializeGlobalActionState()
         {
+            #if UNITY_EDITOR
             // Appears we shouldn't really reset globals in case the domain reload is enabled.
             // This is because in that case, we've just had the whole system init'ed via static ctors
             // Moreover, later in GlobalInialize we skip initialization specifically in this case.
-            if (!InputSystem.IsDomainReloadDisabledForPlayMode())
+            if (!InputSystem.s_IsDomainReloadDisabled())
```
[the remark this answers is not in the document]
jfreire-unity: @u-pr can you suggest a commit change for this?

Packages/com.unity.inputsystem/InputSystem/Editor/Actions/Composites/AxisCompositeEditor.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            var modeField = new EnumField(m_WhichAxisWinsLabel.text, target.whichSideWins)
+            {
+                tooltip = m_WhichAxisWinsLabel.tooltip
+            };
+
+            modeField.RegisterValueChangedCallback(evt =>
+            {
+                target.whichSideWins = (AxisComposite.WhichSideWins)evt.newValue;
+                onChangedCallback();
```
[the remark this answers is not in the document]
jfreire-unity: @u-pr can you suggest a commit change for this?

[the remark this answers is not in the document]
jfreire-unity: I now noticed this out of scope for this PR (code was already there from `develop`). I'll add this to be fixed in https://jira.unity3d.com/browse/ISX-2555

jfreire-unity:
@u-pr can you check which review problems you mentioned have been addressed?

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestStateManager.cs
```diff
@@ -139,7 +142,7 @@ public void Restore()
 
 #if UNITY_EDITOR
             InputEditorUserSettings.s_Settings = state.userSettings;
-            JsonUtility.FromJsonOverwrite(state.systemObject, InputSystem.domainStateManager);
+            JsonUtility.FromJsonOverwrite(state.systemObject, InputSystemEditorInitializer.domainStateManager);
```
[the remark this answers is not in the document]
jfreire-unity: @u-pr is this guaranteed in the develop branch of this repository or is the problem still there as well?

Packages/com.unity.inputsystem/Tests/TestFixture/InputTestRuntime.cs
jfreire-unity: Good point, addressed. We discussed that this small duplication will make it better than just having an int. I'll also comment the duplicated enum so that it's clear why it exists.

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
jfreire-unity: Since InputSystemEditorInitializer is calling it, it can't be only visible to the current class only. Same to other changes from private to internal.

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/OnScreen/OnScreenButton.cs
jfreire-unity: I reverted this bad change from Copilot.
OnScreenButtonEditor.cs now checks for this. Private only allows it to be used within the class.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionSetupExtensions.cs
jfreire-unity: Let me know if what our friend did is enough. I think this is out of scope of this PR (since the name was already this in develop branch)

Assets/Tests/InputSystem/APIVerificationTests.cs
```diff
@@ -212,12 +212,12 @@ internal static bool IgnoreTypeForDocsByNamespace(string @namespace)
 
     [Test]
     [Category("API")]
-    [TestCase("Keyboard", "Devices/Precompiled/FastKeyboard.cs")]
-    [TestCase("Mouse", "Devices/Precompiled/FastMouse.cs")]
-    [TestCase("Touchscreen", "Devices/Precompiled/FastTouchscreen.cs")]
+    [TestCase("Keyboard", "InputSystem/Runtime/Devices/Precompiled/FastKeyboard.cs")]
```
ekcoh: Why pass the InputSystem/Runtime/  when it seems to always be part of the arg? Either pass the full path or only pass the relevant part IMO since concatenated on line 220. Also would prefer `Path.Combine` but maybe its irrelevant.

Assets/Tests/InputSystem/CoreTests_Editor.cs
```diff
@@ -334,12 +334,12 @@ public void Editor_DomainReload_CanRemoveDevicesDuringDomainReload()
         InputSystem.AddDevice<Keyboard>(); // just to make sure keyboard stays as-is
 
         currentTime = 1;
-        InputSystem.OnPlayModeChange(PlayModeStateChange.ExitingEditMode);
+        InputSystemEditorInitializer.OnPlayModeChange(PlayModeStateChange.ExitingEditMode);
```
ekcoh, line 337: Doesn't have to be addressed by this PR, but I believe we should make sure we have a ticket filed on replacing these "faked play-mode state transitions" with real ones when this already sits in a proper editor test?

Assets/Tests/InputSystem/CoreTests_Remoting.cs
ekcoh: Exactly, most commonly seen in in inline lambdas though. This looks very odd. Why do we need for this "singleton" to instantiate on this row? It seems like it's created below. Is there some brittle design that might need addressing here?

Packages/com.unity.inputsystem/InputSystem/Editor/Controls/Processors/AxisDeadzoneProcessorEditor.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using UnityEngine.InputSystem.Editor;
+using UnityEngine.UIElements;
+
+namespace UnityEngine.InputSystem.Processors
+{
+    #if UNITY_EDITOR
```
ekcoh: Shouldn't this #if and matching #endif be enclosing all the content of this file?

Packages/com.unity.inputsystem/InputSystem/Editor/Controls/Processors/StickDeadzoneProcessorEditor.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+using UnityEngine.InputSystem.Editor;
+using UnityEngine.UIElements;
+
+namespace UnityEngine.InputSystem.Processors
+{
+    #if UNITY_EDITOR
```
ekcoh: I would recommend enclosing all the content of this file with this #if, and move corresponding #endif to the end as well.

Packages/com.unity.inputsystem/InputSystem/Editor/Controls/Processors/AxisDeadzoneProcessorEditor.cs
```diff
@@ -0,0 +1,45 @@
...
+            m_MinSetting.OnDrawVisualElements(root, onChangedCallback);
+            m_MaxSetting.OnDrawVisualElements(root, onChangedCallback);
+            CustomOrDefaultSetting.AddSharedDefaultSettingsFooter(root,
+                new[] { m_MinSetting, m_MaxSetting });
+        }
+
+        private CustomOrDefaultSetting m_MinSetting;
+        private CustomOrDefaultSetting m_MaxSetting;
+    }
+    #endif
```
ekcoh: IMO, it would be best to enclose all content, other files do this.

Packages/com.unity.inputsystem/InputSystem/Editor/Controls/Processors/StickDeadzoneProcessorEditor.cs
```diff
@@ -0,0 +1,45 @@
...
+            m_MinSetting.OnDrawVisualElements(root, onChangedCallback);
+            m_MaxSetting.OnDrawVisualElements(root, onChangedCallback);
+            CustomOrDefaultSetting.AddSharedDefaultSettingsFooter(root,
+                new[] { m_MinSetting, m_MaxSetting });
+        }
+
+        private CustomOrDefaultSetting m_MinSetting;
+        private CustomOrDefaultSetting m_MaxSetting;
+    }
+    #endif
```
ekcoh: This #endif can terminate the file

Packages/com.unity.inputsystem/InputSystem/Editor/UITKAssetEditor/Commands/ControlSchemeCommands.cs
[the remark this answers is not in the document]
ekcoh: I would suggest making a separate bug ticket on this specific issue and related fix, then fix it directly after landing this.

Packages/com.unity.inputsystem/InputSystem/Editor/InputSystemStateManager.cs
```diff
@@ -8,31 +8,6 @@
 
 namespace UnityEngine.InputSystem
 {
```
ekcoh, line 10: I notice this structure was previously included in EITHER editor OR development builds. To me it's not obvious why that would be, is to be able to access remoting or something else? Just want to make sure this won't make a regression and there might be reasoning from your side around this?

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionAsset.cs
```diff
@@ -895,10 +894,14 @@ IEnumerator IEnumerable.GetEnumerator()
         internal void MarkAsDirty()
```
ekcoh: Hmmm... who calls this? Given that the function is a no-op outside editor builds I wonder if this should only be tracked in the editor context, e.g. list of InputActionAsset references which are dirty instead of doing it this way.

ekcoh (commented):
Adding a first batch of comments and will resume from InputActionReference in the diff when I can review more.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionReference.cs
```diff
@@ -262,36 +269,10 @@ public InputAction ToInputAction()
...
-
-                // We can only allow setting the reference if it is not part of an persisted InputActionAsset.
-                return (mainAsset is not InputActionAsset);
-            }
-
             // Prevent accidental mutation of the source asset if this InputActionReference is a persisted object
             // residing as a sub-asset within a .inputactions asset.
             // This is not needed for players since scriptable objects aren't serialized back from within a player.
-            if (!CanSetReference(this))
+            if (!CanSetReference())
```
ekcoh: I would suggest that we do this differently. It seems whole `CheckImmutableReference` is really only relevant for editor use cases so it could move to editor code. That eliminates the need for those function objects. Instead, pass a single function object instead that is only called (if its set) at line where call to `CheckImmutableReference()` happens, then this is fully separated between editor and player right?

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionSetupExtensions.cs
```diff
@@ -375,7 +397,7 @@ public static BindingSyntax AddBinding(this InputAction action, InputControl con
         public static BindingSyntax AddBinding(this InputAction action, InputBinding binding = default)
         {
             #if UNITY_EDITOR
-            RegisterApiUsage(UnityEngine.InputSystem.Editor.InputExitPlayModeAnalytic.Api.AddBinding);
+            RegisterAnalyticsApi((int)Api.AddBinding);
```
ekcoh: Why are we casting this (and following places to `int`?)

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystemState.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using UnityEngine;
+using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.Utilities;
+
+namespace UnityEngine.InputSystem
+{
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
```
ekcoh, line 8: I wonder if this is all needed, shouldn't this structure be stripped if not referenced in a release build?

ekcoh (changes requested):
I think `InputActionReference` need some work. Will post again when I have reviewed this whole PR.

Tools/CI/Settings/InputSystemSettings.cs
```diff
@@ -120,35 +120,35 @@ public InputSystemSettings()
...
-            unityEditor.EditorPlatforms.Items[EditorPlatformType.Win10] 
-                = new EditorPlatform(EditorPlatformType.Win10, 
-                    new Agent("package-ci/win10:v4", FlavorType.BuildLarge, ResourceType.Vm));
-            
-            unityEditor.EditorPlatforms.Items[EditorPlatformType.MacOs13] 
-                = new EditorPlatform(EditorPlatformType.MacOs13, 
-                    new Agent("package-ci/macos-13:v4", FlavorType.BuildExtraLarge, ResourceType.VmOsx));
+            unityEditor.EditorPlatforms.Items[EditorPlatformType.Win10]
+                = new EditorPlatform(EditorPlatformType.Win10,
+                new Agent("package-ci/win10:v4", FlavorType.BuildLarge, ResourceType.Vm));
```
ekcoh, line 144: I guess the formatter decided to move this `new` on 144 to less indentation which seems odd. If this is the result of formatter just leave it, but find it odd.

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
```diff
@@ -346,110 +368,80 @@ public Func<IntPtr, bool> onUnityRemoteMessage
...
-                #else
-            // The preprocessor filtering is a workaround for the fact that the AnalyticsResult enum is not available before 2023.1.0a14 when not using the built-in Unity Analytics module.
-                    #if UNITY_INPUT_SYSTEM_ENABLE_ANALYTICS || UNITY_2023_1_OR_NEWER
-            var info = analytic.info;
-            EditorAnalytics.RegisterEventWithLimit(info.Name, info.MaxEventsPerHour, info.MaxNumberOfElements, InputAnalytics.kVendorKey);
-            EditorAnalytics.SendEventWithLimit(info.Name, analytic);
-                    #endif // UNITY_INPUT_SYSTEM_ENABLE_ANALYTICS || UNITY_2023_1_OR_NEWER
-                #endif // UNITY_2023_2_OR_NEWER
-            #elif (UNITY_ANALYTICS) // Implicitly: !UNITY_EDITOR
+            // In Editor, use the callback set by InputSystemEditorInitializer
```
ekcoh, line 399: Happy to see this messy function getting some cleanup. However, the previous approach of having a single SendAnalytic function is probably a bad idea given a clean division between runtime and editor analytics and it would make more sense to fully separate them? E.g. `SendRuntimeAnalytic`, `SendEditorAnalytic`

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
```diff
@@ -346,110 +368,80 @@ public Func<IntPtr, bool> onUnityRemoteMessage
...
-                    var addMethod = GetUnityRemoteAPIMethod("AddMessageHandler");
-                    addMethod?.Invoke(null, new[] { value });
-                    m_UnityRemoteMessageHandler = value;
-                }
+                m_UnityRemoteMessageHandler = value;
+                m_SetUnityRemoteMessageHandler?.Invoke(value);
             }
         }
 
         public void SetUnityRemoteGyroEnabled(bool value)
```
ekcoh: I get the impression the Unity Remote functions made available here and associated internal state above is only usable in edit mode? Wouldn't it make sense to only surface it in editor then or why do we surface in runtime and indirectly call like this? Is this additional tech debt that needs to be addressed and couldn't be addressed in this PR?

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystemTestHooks.cs
```diff
@@ -0,0 +1,64 @@
...
+
+#if UNITY_EDITOR || UNITY_INCLUDE_TESTS
+
+namespace UnityEngine.InputSystem
+{
+    /// <summary>
+    /// Extension of class to provide test-specific functionality.
+    /// Editor-specific operations are handled via callbacks set by the Editor assembly.
+    /// </summary>
+    public static partial class InputSystem
```
ekcoh: Any reason why this wasn't split in 2 parts? I fail to see why this should even be part of InputSystem partial class since it only seems to be about accessing internals which is made available anyway by `InternalsVisibileTo`. Would it be possible to have only `InputSystemRuntimeTestHooks` and `InputSystemEditorTestHooks` splitting this and separating into separate source trees to prepare for future assembly split? May this even reside in a "test assembly" / "test extension" only maybe?

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystemState.cs
ekcoh: (The conditionals)

Packages/com.unity.inputsystem/InputSystem/Runtime/IInputRuntime.cs
```diff
@@ -210,6 +213,20 @@ internal static class InputRuntime
...
     }
 
+#if UNITY_EDITOR
+    /// <summary>
+    /// Mirrors <c>UnityEditor.PlayModeStateChange</c> for use in runtime code without a UnityEditor dependency.
+    /// The underlying int values are identical so the two enums can be cast between each other freely.
+    /// </summary>
+    internal enum InputPlayModeChange
+    {
+        EnteredEditMode = 0,
```
ekcoh: Not sure whether the values were added now or not, but this is the default enum values, any reason why not only relying on compiler assigned values?

Assets/Tests/InputSystem/APIVerificationTests.cs
jfreire-unity: Agree with the suggestion. Will make sure we only pass the precompiled layout filename instead for now.

Assets/Tests/InputSystem/CoreTests_Editor.cs
jfreire-unity: Yes, I wouldn't want to address it in this PR but we can for sure improve it. [Ticket](https://jira.unity3d.com/browse/ISX-2556) created

Packages/com.unity.inputsystem/InputSystem/Editor/InputSystemStateManager.cs
jfreire-unity: The code was moved to Runtime/InputSystemState with `#if UNITY_EDITOR || DEVELOPMENT_BUILD`. The remaining code here was only guarded by `UNITY_EDITOR` so I believe this split is correct.
Seems InputRemoting does depend on `DEVELOPMENT_BUILD` so I guess that's the reason why we guarded this code with it as well.

jfreire-unity: Tracked in https://jira.unity3d.com/browse/ISX-2557

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionAsset.cs
jfreire-unity: To answer your question:
image

Seems mostly to undo changes that were done in playmode? The hash we are populating is inspected in `UnityEngine.InputSystem.Utilities.DirtyAssetTracker.ReloadDirtyAssets()`

I don't oppose doing what you mentioned to avoid this hook mechanism. We can probably improve things here so I'll create a ticket to do it. https://jira.unity3d.com/browse/ISX-2558

I could potentially renamed it to `EditorMarkAsDirty` or `MarkAsDirtyInEditor()` for now so it's more explicit for now?

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystemState.cs
jfreire-unity: Resolving as this seems a duplicated from a previous comment.

Tools/CI/Settings/InputSystemSettings.cs
jfreire-unity: it's just a formatter change and from a previous merge commit that has landed in `develop`: 89f8ecc5685f16f480748df2aeba5c06473fa1be

Packages/com.unity.inputsystem/InputSystem/Runtime/IInputRuntime.cs
jfreire-unity: No particular reason, just thought it was better to be explicit. But I saw now that the original Editor enum this mirrors has no values set, so I'll just remove the explicit setting of enum values. Thanks for pointing it out.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionSetupExtensions.cs
jfreire-unity: Fixed in 3161712b2b26cde1e9413d9d9af44539895336d6

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
jfreire-unity: I think your comment makes sense. I'm considering that out of scope because it's not immediately clear to me the "moments" where we want to do Editor and Runtime analytics.
So far, to me it seems it's mostly a "backend" difference? Like, if we are in PlayMode / Edit Mode we send data through the EditorAnalytics API? And if it's in player builds we use a different analytics function, like ` Analytics.Analytics.SendEvent` ?

jfreire-unity: https://jira.unity3d.com/browse/ISX-2559

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
jfreire-unity: Good point. I checked and seemed this code is mostly Editor only but then the IInputRuntime depends on it? Seems we created InputTestRuntime tests but I don't think we actually need this to be part of it? Not sure what was the motivation behind it.

With the help of LLM friends, I decoupled it from IInputRuntime so let me know if this is enough for now or I should just keep it like it was and spend a bit more time later on.

Changed in:  7653046151377ebc2856c020807202bcc02d264d

jfreire-unity:
❌ Need to fix 670f071108ba1f8bb7a74f574244fc5179d1c2a9, InputAnalytics seems to not be the best place to have the Api analytics enum as is,

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionReference.cs
jfreire-unity: Changed in 7e50a2a375f275ced0058e3b3f82cfeff056e736

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystemTestHooks.cs
jfreire-unity: This is from FEPM PR  #2343 so I didn't look much into this. But I can have a second look (didn't address it in the review from today, will check tomorrow if this is low risk - should be).

jfreire-unity:
> ❌ Need to fix [670f071](https://github.com/Unity-Technologies/InputSystem/commit/670f071108ba1f8bb7a74f574244fc5179d1c2a9), InputAnalytics seems to not be the best place to have the Api analytics enum as is,

✅ 🤞 Should be fixed by [2c323e4](https://github.com/Unity-Technologies/InputSystem/pull/2321/commits/2c323e4e6a76f8d3a446e6a1a8b28c61164ec6f2)

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionSetupExtensions.cs
```diff
@@ -323,13 +323,35 @@ public static BindingSyntax AddBinding(this InputAction action, string path, str
...
-        private static void RegisterApiUsage(UnityEngine.InputSystem.Editor.InputExitPlayModeAnalytic.Api api)
+
+        // Internal hook for Editor to register analytics tracking
+        internal static Action<int> s_ApiUsageCallback;
+
+        // Internal hook for Editor to suppress analytics
+        internal static Action<bool> s_SuppressAnalytics;
+
+        // API constants for analytics (matches Editor InputExitPlayModeAnalytic.Api enum)
+        internal enum Api
```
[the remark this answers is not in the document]
ekcoh: Status? - Would be great with clear follow up on the concern

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionAsset.cs
ekcoh: That's sounds good, then it sticks out even more. Since we are always in playmate for any player build it makes sense. Thanks for filing a follow up as well.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystemTestHooks.cs
ekcoh: OK, thanks. I do not want to halt this from landing, but any corrections that are not too complex might be beneficial to avoid piling up follow up work that might have to land in 2 places.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystem.cs
```diff
@@ -2985,23 +2963,25 @@ internal static bool ShouldDrawWarningIconForBinding(string bindingPath)
         #region Actions
 
 
-        // This is called from InitializeInEditor() and InitializeInPlayer() to make sure
+        // This is called to make sure
         // project-wide actions are all active in they are active in all of these MonoBehavior methods:
```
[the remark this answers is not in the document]
ekcoh: Looks simple to fix?

Packages/com.unity.inputsystem/InputSystem/Runtime/IInputRuntime.cs
```diff
@@ -220,10 +220,10 @@ internal static class InputRuntime
     /// </summary>
     internal enum InputPlayModeChange
     {
-        EnteredEditMode = 0,
-        ExitingEditMode = 1,
-        EnteredPlayMode = 2,
-        ExitingPlayMode = 3,
+        EnteredEditMode ,
```
ekcoh: Not sure why I cannot add a simple review suggestion at the moment, but there is excessive white space here that should be removed

Packages/com.unity.inputsystem/InputSystem/Runtime/InputAnalytics.cs
```diff
@@ -6,12 +6,34 @@
...
 
         internal static Func<bool> IsNewSystemBackendsEnabled { get; set; }
         internal static Func<bool> IsOldSystemBackendsEnabled { get; set; }
+        /// <summary>
+        /// Code authoring APIs on used for analytics.
+        /// </summary>
+        /// <remarks>
+        /// This enumeration may be added to, but NEVER changed, since it would break older data.
+        /// </remarks>
+        internal enum AuthoringApi
```
ekcoh, line 21: I am confused why this moved to Runtime? As far as I remember this is an editor-only analytic and irrelevant for runtime builds?

ekcoh: What was the problem with keeping it in editor? I believe the culprit is the shared submission interface via runtime interface. Editor analytics should probably not be piped through that interface, instead they should be taking a different path that is editor only (as the underlying SDK API)

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionReference.cs
```diff
@@ -224,6 +219,14 @@ internal static void InvalidateAll()
                 ((InputActionReference)obj).Invalidate();
         }
 
+#if UNITY_EDITOR
+        /// <summary>
+        /// Set by the editor to prevent mutating <see cref="InputActionReference"/> sub-assets inside
+        /// <see cref="InputActionAsset"/> files. Not used in player builds.
+        /// </summary>
+        internal static Action<InputActionReference> s_CheckImmutableReference;
```
ekcoh, line 227: Thanks for doing this change, this is at least cleaner even though #if UNITY_EDITOR is still present. IMO it wouldn't be the end of the world to have this unconditional and hope it's optimized away, but this is better than previous.

ekcoh (commented):
Follow up review done based on commits since last review. Thanks for doing changes to `InputActionReference`, but now I have shifted to be more concerned about the analytics. Most analytics are editor only (except initialise, shutdown analytics which are fuzzy what they are - the older analytics). I might need a third pass on this but not sure if you are changing additional things with analytics or if planned changes are in? Probably need to re-review that part in combined (full review mode then).

Assets/Tests/InputSystem/CoreTests_Remoting.cs
jfreire-unity: Very good point (I missed the subtlety of this). I looked into this now and made the change; we no longer need the previous check.

With these changes, `RemoteInputPlayerConnection` is always the same type, and `RemoteInputPlayerConnectionEditor` is the editor equivalent, which only hooks into the editor in https://github.com/Unity-Technologies/InputSystem/blob/c0a3892fa0fdf7a793d6c047acefac40ed9107f4/Packages/com.unity.inputsystem/InputSystem/Editor/InputSystemEditorInitializer.cs#L333 to the `RemoteInputPlayerConnection` instance.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputAnalytics.cs
jfreire-unity: The problem is that `InputActionSetupExtensions.cs` uses it, and it's a Runtime class. Even though it's usages it's guarded by `UNITY_EDITOR`, we can't pull in any Editor type to Runtime classes. This would be clearer if we also had assemblies split in this PR. (and will be clearer once we are a module in trunk).
The constraint is Runtime code can't have any `InputSystem.Editor` references (or any types inside the Editor folder) or any UnityEditor references as well.

Runtime folder => play mode (editor) and player builds
Editor folder => Editor only functionality.

jfreire-unity: I can guarde it with `UNITY_EDITOR` so that it doesn't step into player builds at all (I don't remember the constraint of `UNITY_ANALYTICS`...) Let me know what u think

Packages/com.unity.inputsystem/InputSystem/Runtime/IInputRuntime.cs
jfreire-unity: fixed!

Packages/com.unity.inputsystem/InputSystem/Runtime/InputAnalytics.cs
jfreire-unity: The alternative is to keep the enum type in an Editor class, and Runtime references of it are just an int. This then requires us to cast the enum.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSystemTestHooks.cs
jfreire-unity: I moved this out of the Input System partial class only. I didn't see any reason to keep it, and it will help in the future. I hope removing the `public static partial class InputSystem` is not a breaking change.

Changes in b11fcc6d037ec5421b7addeac373a4bc9b613807

I did try separating this into Editor and Runtime files. But honestly, it got a bit more confusing because the tests then contained both Editor and Runtime methods being called. It looked like the tests would also need more changes to be less confusing, and I sensed a rabbit hole was coming.
Also, the runtime side was still guarded by `UNITY_EDITOR` defines.

I think this is not critical since there is no Editor-specific dependency in this file. I'll leave them in the Runtime folder in this PR for now.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputAnalytics.cs
ekcoh: OK, a bit of a mess I assume. Forgot about the reference the enum part. OK, if you think it's best to leave it as is for now then maybe that is best. I will approve the PR since I see no need to block this.

ekcoh (approved):

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
```diff
@@ -103,6 +99,7 @@ public unsafe InputUpdateDelegate onUpdate
                         {
                             eventBufferPtr->eventCount = 0;
                             eventBufferPtr->sizeInBytes = 0;
+                            eventBufferPtr->capacityInBytes = 0;
```
jfreire-unity: I'll revert this, this was some Copilot change that I did not request.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionSetupExtensions.cs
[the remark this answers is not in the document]
jfreire-unity: This was moved to InputAnalytics.cs so we have a single enum for now. Since it needs to be used by Runtime classes, it will live there.

Pauliusd01 (approved):
LGTM, tested the current samples and qa scenes on windows and android and went through the tests in this generated test plan: https://github.com/Unity-Technologies/InputSystem/pull/2321#issuecomment-4233834351

VeraMommersteeg (approved):
Lgtm, Ty for doing the changes

PR #2335, opened 2026-01-29
FIX: handled events still triggering actions when switching PlayerInput device

AswinRajGopal:
### Description

  **Bug:** https://issuetracker.unity3d.com/issues/setting-inputeventptr-dot-handled-to-true-does-not-prevent-actions-from-triggering-when-changing-devices
  When `InputUser.onUnpairedDeviceUsed` marks `eventPtr.handled` = true, we expect no action to fire from that same
  button press.
  But on devices like DualSense, a single press can generate multiple input events. The handled flag only stops the
  one event, and another event in the same press still triggers the action.

  Root cause
  The action system is driven by state change events. We can see different eventIds for the handled event vs the
  action‑triggering event, so the action was coming from a separate input event.

  **Fix**

  - While dispatching `InputSystem.onEvent`, stop calling remaining listeners once handled becomes true.
  - When an event is handled (policy = SuppressStateUpdates), temporarily suppress action processing for that device
    for the rest of the update (and next update).
  - In `InputActionState`, skip processing control changes if the device is suppressed.

### Testing status & QA

Manually verified using repro project + dualsense controller. 

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers
  - Only applies when policy is `SuppressStateUpdates` and a listener explicitly sets handled = true.
  - Normal input flow remains unchanged.
  - This simply makes “handled” cover the entire physical press, even if it arrives as multiple events.
### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Events.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Events.cs
@@ -10,6 +10,7 @@ using Unity.Collections.LowLevel.Unsafe;
 using UnityEngine;
 using UnityEngine.Scripting;
 using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.Users;
 using UnityEngine.InputSystem.Controls;
 using UnityEngine.InputSystem.Layouts;
 using UnityEngine.InputSystem.LowLevel;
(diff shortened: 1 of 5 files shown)


AswinRajGopal:
Closing this fork. Created new PR from develop.

PR #2340, opened 2026-02-03
FIX: handled events still triggering actions when switching PlayerInput

AswinRajGopal:
### Description

  **Bug:** https://issuetracker.unity3d.com/issues/setting-inputeventptr-dot-handled-to-true-does-not-prevent-actions-from-triggering-when-changing-devices
  When `InputUser.onUnpairedDeviceUsed` marks `eventPtr.handled` = true, we expect no action to fire from that same
  button press.
  But on devices like DualSense, a single press can generate multiple input events. The handled flag only stops the
  one event, and another event in the same press still triggers the action.

  Root cause
  The action system is driven by state change events. We can see different eventIds for the handled event vs the
  action‑triggering event, so the action was coming from a separate input event.

  **Fix**

  - While dispatching `InputSystem.onEvent`, stop calling remaining listeners once handled becomes true.
  - When an event is handled (policy = SuppressStateUpdates), temporarily suppress action processing for that device
    for the rest of the update (and next update).
  - In `InputActionState`, skip processing control changes if the device is suppressed.

### Testing status & QA

Manually verified using repro project + dualsense controller. 

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers
  - Only applies when policy is `SuppressStateUpdates` and a listener explicitly sets handled = true.
  - Normal input flow remains unchanged, however documentation update seems necessary. 
  - This simply makes “handled” cover the entire physical press, even if it arrives as multiple events.
  -  handled + no state update leaves the button “unpressed” in system state, and gamepad polling makes
  the next event look like a new press which is now prevented. 
### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/CoreTests_Events.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Events.cs
@@ -1335,24 +1335,24 @@ partial class CoreTests
     // Step 3: Release button north and stick while no longer being suppressed.
     // Step 4: Press gamepad north and stick.
 
-    // Press event is detected in step 2 (false positive) with default interaction
+    // Press event is suppressed in step 1/2 with default interaction
     [TestCase(InputEventHandledPolicy.SuppressStateUpdates, // policy
         null, // interactions
-        new int[] { 0, 0, 1, 1, 2}, // started
-        new int[] { 0, 0, 1, 1, 2}, // performed
-        new int[] {0, 0, 0, 1, 1})] // cancelled
+        new int[] { 0, 0, 0, 0, 1}, // started
+        new int[] { 0, 0, 0, 0, 1}, // performed
+        new int[] {0, 0, 0, 0, 0})] // cancelled
     // Press event is not detected in step 1/2 with default interaction
     [TestCase(InputEventHandledPolicy.SuppressActionEventNotifications,
         null,
         new int[] { 0, 0, 0, 0, 1},
         new int[] { 0, 0, 0, 0, 1},
         new int[] {0, 0, 0, 1, 1})]
-    // Press event is detected in step 2 (false positive) with explicit press interaction
+    // Press event is suppressed in step 1/2 with explicit press interaction
     [TestCase(InputEventHandledPolicy.SuppressStateUpdates,
         "press",
-        new int[] { 0, 0, 1, 1, 2},
-        new int[] { 0, 0, 1, 1, 2},
-        new int[] {0, 0, 0, 1, 1})]
+        new int[] { 0, 0, 0, 0, 1},
+        new int[] { 0, 0, 0, 0, 1},
+        new int[] {0, 0, 0, 0, 0})]
     // Press event is not detected in step 1/2 (false positive) with explicit press interaction
     [TestCase(InputEventHandledPolicy.SuppressActionEventNotifications,
         "press",
@@ -1439,7 +1439,7 @@ partial class CoreTests
         Assert.That(action.WasPressedThisFrame, Is.EqualTo(performedThisFrame));
         releasedThisFrame = expectedCancelled[2] - expectedCancelled[1] > 0;
         Assert.That(action.WasReleasedThisFrame, Is.EqualTo(releasedThisFrame));
-        Assert.That(action.IsPressed, Is.True); // Note: This is not an event and hence not suppressed
+        Assert.That(action.IsPressed, Is.EqualTo(seesControlChangesUnderSuppression)); // Note: This is not an event and hence not suppressed
 
         Assert.That(Gamepad.current.buttonNorth.wasPressedThisFrame, Is.EqualTo(!seesControlChangesUnderSuppression));
         Assert.That(Gamepad.current.buttonNorth.wasReleasedThisFrame, Is.False);
--- a/Assets/Tests/InputSystem/CoreTests_Events.cs
+++ b/Assets/Tests/InputSystem/CoreTests_Events.cs
@@ -10,6 +10,7 @@ using Unity.Collections.LowLevel.Unsafe;
 using UnityEngine;
 using UnityEngine.Scripting;
 using UnityEngine.InputSystem;
+using UnityEngine.InputSystem.Users;
 using UnityEngine.InputSystem.Controls;
 using UnityEngine.InputSystem.Layouts;
 using UnityEngine.InputSystem.LowLevel;
@@ -1254,6 +1255,48 @@ partial class CoreTests
         InputSystem.Update();
 
         Assert.That(device.rightTrigger.ReadValue(), Is.EqualTo(0.0).Within(0.00001));
+    }
+
+    [Test]
+    [Category("Events")]
+    public void Events_HandledFirstEvent_DoesNotTriggerAction_OnSecondEventSameUpdate()
+    {
+        var gamepad = InputSystem.AddDevice<Gamepad>();
+        var action = new InputAction(type: InputActionType.Button, binding: "<Gamepad>/buttonSouth");
+        action.Enable();
+
+        var performedCount = 0;
+        action.performed += _ => ++performedCount;
+
+        var previousPolicy = InputSystem.s_Manager.inputEventHandledPolicy;
+        InputSystem.s_Manager.inputEventHandledPolicy = InputEventHandledPolicy.SuppressStateUpdates;
+
+        InputUser.listenForUnpairedDeviceActivity++;
+        var handledFirstEventId = 0;
+        Action<InputControl, InputEventPtr> onUnpaired = (control, eventPtr) =>
+        {
+            if (handledFirstEventId == 0)
+            {
+                handledFirstEventId = eventPtr.id;
+                eventPtr.handled = true;
+            }
+        };
+        InputUser.onUnpairedDeviceUsed += onUnpaired;
+
+        try
+        {
+            InputSystem.QueueStateEvent(gamepad, new GamepadState().WithButton(GamepadButton.South));
+            InputSystem.QueueStateEvent(gamepad, new GamepadState().WithButton(GamepadButton.South));
+            InputSystem.Update();
+
+            Assert.That(performedCount, Is.EqualTo(0));
+        }
+        finally
+        {
+            InputUser.onUnpairedDeviceUsed -= onUnpaired;
+            InputUser.listenForUnpairedDeviceActivity--;
+            InputSystem.s_Manager.inputEventHandledPolicy = previousPolicy;
+        }
     }
 
     [Test]
--- a/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/InputManager.cs
@@ -3463,14 +3463,7 @@ namespace UnityEngine.InputSystem
                     //       create simulated events from.
                     if (m_EventListeners.length > 0)
                     {
-                        if (DelegateHelpers.InvokeCallbacksSafeUntilHandled(ref m_EventListeners, currentEventPtr, device, k_InputOnEventMarker, "InputSystem.onEvent",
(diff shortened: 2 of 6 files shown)


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3454,24 +3455,31 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
                     if (m_EventListeners.length > 0)
                     {
-                        DelegateHelpers.InvokeCallbacksSafe(ref m_EventListeners,
-                            new InputEventPtr(currentEventReadPtr), device, k_InputOnEventMarker, "InputSystem.onEvent");
-
-                        // If a listener marks the event as handled, we don't process it further.
-                        if (m_InputEventHandledPolicy == InputEventHandledPolicy.SuppressStateUpdates &&
-                            currentEventReadPtr->handled)
+                        if (DelegateHelpers.InvokeCallbacksSafeUntilHandled(ref m_EventListeners, currentEventPtr, device, k_InputOnEventMarker, "InputSystem.onEvent",
+                            m_InputEventHandledPolicy == InputEventHandledPolicy.SuppressStateUpdates))
```
jfreire-unity: Are we saying here that if a single listener handles the event, we don't let other listeners process the event? Should this be part of the listener logic instead? Otherwise it's a unclear for the user why was their method hooked on `InputSystem.onEvent` not being triggered 🤔

Assets/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -1292,24 +1335,24 @@ class SuppressedActionEventData
...
-    // Press event is detected in step 2 (false positive) with explicit press interaction
+    // Press event is suppressed in step 1/2 with explicit press interaction
     [TestCase(InputEventHandledPolicy.SuppressStateUpdates,
         "press",
-        new int[] { 0, 0, 1, 1, 2},
-        new int[] { 0, 0, 1, 1, 2},
-        new int[] {0, 0, 0, 1, 1})]
+        new int[] { 0, 0, 0, 0, 1},
+        new int[] { 0, 0, 0, 0, 1},
+        new int[] {0, 0, 0, 0, 0})]
```
jfreire-unity, line 1355: This looks like a behavior change. What is the reason to change this?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3890,6 +3896,31 @@ private void InvokeAfterUpdateCallback(InputUpdateType updateType)
...
+                return false;
+
+            var deviceIndex = device.m_DeviceIndex;
+            if (m_SuppressActionsForDeviceUpdate == null || deviceIndex >= m_SuppressActionsForDeviceUpdate.Length)
+                return false;
+
+            var lastUpdate = (int)m_SuppressActionsForDeviceUpdate[deviceIndex];
+            var currentUpdate = (int)InputUpdate.s_UpdateStepCount;
+            return lastUpdate == currentUpdate || lastUpdate + 1 == currentUpdate;
+        }
```
jfreire-unity, line 3919: I'm sure why this code is needed.. Can you clarify what is this doing and why we need it?

jfreire-unity (commented):
Thanks, left some comments for clarification

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
AswinRajGopal: Thanks for pointing that out, we should not prevent notifying the other listeners. I've removed this call and the helper function.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
AswinRajGopal: Sorry for the confusion. I've added comments on top of the function - This function is needed to track the update step when an event is marked handled so that we can suppress action processing for `the device` in that update and the next update, while still allowing all onEvent listeners to run as usual. This keeps `SuppressStateUpdates behavior unchanged after removing the early exit from onEvent.

Without this new tracking, a listener can set `eventPtr.handled = true` and actions would still process in that update and potentially in the next in case the action system runs on a subsequent update, which breaks `SupressStateUpdates` intended behavior. This minimal mechanism is to preserve handled suppresses actions semantics while keeping onEvent broadcasting intact.

Assets/Tests/InputSystem/CoreTests_Events.cs
AswinRajGopal: We needed a explicit way to keep action suppression working across update steps. It will ensure that actions doesn't trigger on next update which shows up as a false positive press like reported in the bug. Please correct me if I'm wrong - my understanding is, 
 * With `SupressStateUpdates`, a handled press should not create started or performed in steps 1 or 2.
 * This change is removing the 2nd step false positive press and keeps the policy intact.

Pauliusd01:
Did not forget this one, still looking into it

Pauliusd01 (approved):
LGTM: Checked using the user project as well as our gamepad related sample scenes. Tried switching between mouse/keyboard/multiple gamepads in player and playmode, tried disconnecting/reconnecting as well as some of AI generated tests although those were a little bit "out there" and inconclusive. (Multiple listeners, Device switching, CrossUpdate supression, Rebinding supression) If there's anything specific you need me to recheck or you make changes after this then feel free to re-request the review

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3454,22 +3455,22 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
                         }
                     }
 
+                    var currentEventPtr = new InputEventPtr(currentEventReadPtr);
```
jfreire-unity, line 3458: I'm wondering if we need to cache this explicitly or if doing it the previous way is enough, by using `currentEventReadPtr` ? I don't see a reason but maybe I'm missing something.

Assets/Tests/InputSystem/CoreTests_Events.cs
jfreire-unity: I see. I'll keep looking at it to see if I find any problems. Meanwhile maybe @ekcoh has a more immediate answer since he added these tests.

ekcoh:
Thanks for looking into this @AswinRajGopal and creating this PR.
I am just starting to look into this PR so will first just comment on what I am reacting to on the description level:

> But on devices like DualSense, a single press can generate multiple input events. The handled flag only stops the
> one event, and another event in the same press still triggers the action.

A device like DualSense is often served by a polling (sampling) backend. It means it continuously reads its state and hence generate many events. It is true the handled flag only stops one event, but one should not stop any event from the perspective to update device state (this will break by design - which is why I added suppress mode originally since the original behaviour doesn't make any sense IMO). But if the transition from false, (suppressed), true yields a press interaction event we are already overboard. This is the exact same problem as @MorganHoarau is currently working with from the perspective of rotating a screen. Hence I believe this is a RCS/RCA issue again about false positives. I need to look into the diff here to say more but just wanted to mention this. I suspect the root cause for this issue lies in InputActionState if it allows interaction state to update when previous state was consumed/suppressed.

> Root cause
> The action system is driven by state change events. We can see different eventIds for the handled event vs the
> action‑triggering event, so the action was coming from a separate input event.

This goes against the previous statement a bit. If there was a transition from false to true and true was suppressed, the following true should not be seen as a transition, since the transition is really true to true which is not a change.

These two comments aside, I agree with you if you concluded that you wanted to change suppression mode. This is problematic since it would lead to a behavioural change. I initially struggle to see why test case asserts would change though. I need some more time to give better feedback on this PR and analyse it properly.

Assets/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -1292,24 +1335,24 @@ class SuppressedActionEventData
...
     // Step 4: Press gamepad north and stick.
 
-    // Press event is detected in step 2 (false positive) with default interaction
+    // Press event is suppressed in step 1/2 with default interaction
     [TestCase(InputEventHandledPolicy.SuppressStateUpdates, // policy
         null, // interactions
-        new int[] { 0, 0, 1, 1, 2}, // started
-        new int[] { 0, 0, 1, 1, 2}, // performed
-        new int[] {0, 0, 0, 1, 1})] // cancelled
+        new int[] { 0, 0, 0, 0, 1}, // started
```
ekcoh, line 1341: If any of this changes I suspect it will regress the rebinding sample.

Assets/Tests/InputSystem/CoreTests_Events.cs
ekcoh: Ideally, IMO, action suppression should be bound to event.handled and state propagation would always happen, but I believe some PoC and solution design is needed to validate such a change and evaluate any possible side-effects of such a change. The number of repetitions or number of frames should not be part of equation, only transitions. Hence why event.handled looks like of broken since it desynchronises from actual state and doesn't worked for polled sources that may provide an arbitrary number of repeats.

ekcoh: This test was added to verify suppression behaviour and it is not clear to me why it would have to change?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3454,22 +3455,22 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+                    }
 
-                        // If a listener marks the event as handled, we don't process it further.
-                        if (m_InputEventHandledPolicy == InputEventHandledPolicy.SuppressStateUpdates &&
-                            currentEventReadPtr->handled)
-                        {
-                            m_InputEventStream.Advance(false);
-                            continue;
-                        }
+                    if (m_InputEventHandledPolicy == InputEventHandledPolicy.SuppressStateUpdates && currentEventPtr.handled)
```
ekcoh, line 3469: Not sure I understand how this is supposed to be working? Would you mind elaborating, maybe with inline comments?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3890,6 +3889,35 @@ private void InvokeAfterUpdateCallback(InputUpdateType updateType)
...
+            if (m_InputEventHandledPolicy != InputEventHandledPolicy.SuppressStateUpdates || device == null)
+                return;
+
+            var deviceIndex = device.m_DeviceIndex;
+            if (m_SuppressActionsForDeviceUpdate == null || deviceIndex >= m_SuppressActionsForDeviceUpdate.Length)
+                Array.Resize(ref m_SuppressActionsForDeviceUpdate, deviceIndex + 1);
+            m_SuppressActionsForDeviceUpdate[deviceIndex] = InputUpdate.s_UpdateStepCount;
+        }
+
+        // Check whether actions for this device should be suppressed due to a handled event
```
ekcoh, line 3905: I do not think any kind of "previous step" logic could work, e.g. a DualSense may repeat its button state 2000 times or more in row? Or maybe I am reading this incorrectly?

ekcoh:
@AswinRajGopal I will perform a re-review of this PR, investigate the original issue and look into the possibility of providing a fix for this with less modifications and that can prevent false triggering. I get back to you when I have a clear picture of the problem and side-effects and whether my previous concerns have merit or not once I have more information. It seems @Pauliusd01 confirmed this PR solved the reported issue but I still have concerns around how we addressed it so I will invest some time into investigating this more in-depth so we can hopefully get this sorted.

PR #2349, opened 2026-02-18
FIX: inactive touch driving invalid synthetic click on state reset

MorganHoarau:
### Description

Addresses UUM-100125 (phantom click after device rotation).

Follow up PR about a UITKInputModule failing test locally: #2367 

**Issue:**
* After rotation, InputSystem queue a [device configuration event](https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/InputSystem.cs#L2730) which causes `InputActionState` to run initial-state catch-up in [`OnBeforeInitialUpdate`](https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Actions/InputActionState.cs#L1278). (See **Comments To Reviewers** section for overview)
* Touch position state can still be non-default after touch end (because [`position`](https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs#L86) is `dontReset = true`). So the stale state (touch position being non-default) was treated as fresh actuation and trigger unexpected callbacks.

**Changes**
* Added a guard in initial-state check to skip replay for [**inactive**](https://github.com/Unity-Technologies/InputSystem/blob/develop/Packages/com.unity.inputsystem/InputSystem/Devices/Touchscreen.cs#L201) touch state.
* Added regression test: `Actions_InitialStateCheckAfterConfigurationChange_DoesNotTriggerForInactiveTouch`
* Updated existing test [`Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works(true)`](https://github.com/Unity-Technologies/InputSystem/blob/develop/Assets/Tests/InputSystem/CoreTests_Actions.cs#L12436) to align the new expected behaviour. The test still preserves the original ISXB-98 guarantee: after enabling, real subsequent touches must still trigger the composite action correctly.

**TODO:**
* [x] Check existing documentation around the topic (Touch, device configuration event, dontReset) - need update?
> Also, unlike both [Button](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.8/api/UnityEngine.InputSystem.InputActionType.html#UnityEngine_InputSystem_InputActionType_Button) and [PassThrough](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.8/api/UnityEngine.InputSystem.InputActionType.html#UnityEngine_InputSystem_InputActionType_PassThrough) actions, [Value](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.8/api/UnityEngine.InputSystem.InputActionType.html#UnityEngine_InputSystem_InputActionType_Value) actions perform what's called "initial state check" on the first input update after the action was enabled. What this does is check controls bound to the action and if they are already actuated (that is, at non-default value), the action will immediately be started and performed. What this means in practice is that when a value action is bound to, say, the left stick on a gamepad and the stick is already moved out of its resting position, then the action will immediately trigger instead of first requiring the stick to be moved slightly.
> From [InputActionType](https://docs.unity3d.com/Packages/com.unity.inputsystem@1.8/api/UnityEngine.InputSystem.InputActionType.html?q=Initial%20state%20check) documentation.
* [x] Changelog

### Testing status & QA

* Reproduced issue in repro project.
* Added new regression test for this bug path.
* Ran existing tests and identified conflict with: `Actions_WithMultipleCompositeBindings_WithoutEvaluateMagnitude_Works(true)`

### Overall Product Risks

- Complexity: Low
- Halo Effect: Medium - affects TouchControl that supports initial state check.

### Comments to reviewers

Tracing the whole flow can help understand the broader context. `DeviceConfigurationEvent` flow diagram:
```mermaid
flowchart TD
    subgraph Native["Native / Backend"]
        A[Device rotates]
        B[Queue DeviceConfigurationEvent]
    end

    subgraph IM["InputManager"]
        C[Process DeviceConfigurationEvent]
        D[NotifyConfigurationChanged]
        E[InputActionState.OnDeviceChange ConfigurationChanged]
    end

    subgraph IASResolve["InputActionState.Resolve phase"]
        F[Full binding re-resolve]
        G[Set initialStateCheckPending = true]
    end

    subgraph IASInitial["InputActionState.OnBeforeInitialUpdate"]
        H[Loop bindings and controls]
        I{Control belongs to TouchControl?}
        J{touch.isInProgress?}
        K[Default CheckStateIsAtDefault path]
    end

    subgraph OldPath["Old behavior"]
        L[SignalStateChangeMonitor for inactive touch]
        M[Started/Performed from stale touch state]
        N[Phantom UI click]
    end

    subgraph NewPath["Current proposed fix"]
        O[Skip inactive touch control]
        P[No synthetic Started/Performed]
        Q[No phantom UI click]
    end

    A --> B
    B --> C
    C --> D
    D --> E
    E --> F
    F --> G
    G --> H
    H --> I
    I -- no --> K
    I -- yes --> J
    J -- yes --> K
    J -- no old --> L
    L --> M
    M --> N
    J -- no new --> O
    O --> P
    P --> Q
```

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Darren-Kelly-Unity (approved):

ekcoh (approved):
I think this fix is cleanly executed and really like that tests have been added for this specific scenario. I still wonder if a touch-specific solution is really needed, but see no reason to not land this now. We can always revisit this together with later interaction robustness improvements if needed since all is internal. Good job on sorting this out @MorganHoarau!

Pauliusd01:
I will probably only be able to take a look tomorrow

Pauliusd01 (approved):

PR #2361, opened 2026-02-26
CHANGE: Upgrade to Wench 2.X

stefanunity:
### Description

- upgrading Wrench to 2.x
- there are some breaking changes (including moving to System.Text.Json)
- cleaned up APV package exceptions

### Testing status & QA

-green CI
-jobs generation looking the "same"

### Overall Product Risks

- Complexity: low
- Halo Effect: low, internal


Tools/CI/.claude/settings.local.json
```diff
@@ -0,0 +1,7 @@
+{
```
jfreire-unity: I wouldn't leave this here, maybe we need to add this local settings file to the .gitignore.

jfreire-unity (changes requested):

Tools/CI/.claude/settings.local.json
stefanunity: ups, this was commited by accident, will delete

ritamerkl (approved):

.gitignore
```diff
@@ -45,6 +45,7 @@ Packages/com.unity.package-manager-doctools/
 
 Tools/CI/[Bb]in
 Tools/CI/[Oo]bj
+Tools/CI/.claude/**
```
jfreire-unity, line 48: 👌

jfreire-unity (approved):

PR #2365, opened 2026-03-03
NEW: Added Focus Events to Input Event Queue

VeraMommersteeg:
### Description

These are the managed changes for the package to introduce a focus event into the event buffer. It no longer listens to the Application.OnFocusChanged and instead changes focus based on the queued event from native. This will help us deal with desync focus state issues to make sure events are being processed in the right state. 

Jira tickets:
[ISX-2426](https://jira.unity3d.com/browse/ISX-2426)
[ISX-2427](https://jira.unity3d.com/browse/ISX-2427)

### Testing status & QA

Testing done:
- Ran editor and playmode tests locally
- Manually tested several Editor Playmode Input Behaviours
- Built samples as a player and tried several background behaviour settings compared to develop

Testing by QA:
- General pass against other samples

### Overall Product Risks

- Complexity: Medium - Focus event changes the way events are processed
- Halo Effect: Low - Most changes are encapsulated to InputManager.cs

### Comments to reviewers

I've spend some time cleaning up the OnUpdate a bit, didn't go too crazy and mostly just split up the code a bit in a couple of functions. But since the git diff makes it a bit difficult to see the changes, it might be easier to pull the branch and do a diff that way.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3129,151 +3063,142 @@ internal bool ShouldRunUpdate(InputUpdateType updateType)
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1809:AvoidExcessiveLocals", Justification = "TODO: Refactor later.")]
         private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer eventBuffer)
         {
-            // NOTE: This is *not* using try/finally as we've seen unreliability in the EndSample()
-            //       execution (and we're not sure where it's coming from).
-            k_InputUpdateProfilerMarker.Begin();
-
-            if (m_InputEventStream.isOpen)
+            using (k_InputUpdateProfilerMarker.Auto())
```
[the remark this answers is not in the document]
VeraMommersteeg: older versions didnt support auto. this is no longer the case

ekcoh:
@VeraMommersteeg Looks like the PR currently has conflicts that are in need of being resolved:
- InputManager.cs
- InputTestFixture.cs

Likely due to CEPM/FEPM branch being merged.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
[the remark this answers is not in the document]
ekcoh: Scoped lifetime management is very much preferred IMO due to proper exception safety. I think this comment can be resolved.

ekcoh:
@VeraMommersteeg Noticed there are multiple test failures, I suspect you have already noticed and is looking into it?

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
```diff
@@ -158,11 +158,10 @@ public void EnhancedTouch_SupportsEditorUpdates(InputSettings.UpdateMode updateM
         Assert.That(Touch.activeTouches, Has.Count.EqualTo(1));
 
         // And make sure we're not seeing the data in the editor.
-        runtime.PlayerFocusLost();
+        ScheduleFocusEvent(false);
```
ekcoh: Minor: It is only semantics, by I personally think a test fixture helper like is should be named what it does instead of reflecting implementation details (that changing focus leads to scheduling an event). E.g. `ChangeFocus(applicationHasFocus: false)` is a lot clearer. The fact that this leads to a deferred change event is a detail. But the actual change happens when that line is executed.

Nitpick: When using trivial arguments like this, I think it's helpful to pass as named argument since `false` could be mean many things. It makes it a lot easier to read the code, especially in web-interface where there are no IDE smartness that highlights the name of the argument.

ekcoh: I only write this comment once even though it applies to all tests here, e.g. row 174 `ScheduleFocusEvent(true)` etc.

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
```diff
@@ -171,8 +170,25 @@ public void EnhancedTouch_SupportsEditorUpdates(InputSettings.UpdateMode updateM
...
-        runtime.PlayerFocusGained();
-        InputSystem.Update();
+        ScheduleFocusEvent(true);
+
+        // We have to schedule the specific update type that the player is configured to process events in,
+        // otherwise we will end up using defaultUpdateType, which due to the fact we do not have focus in pre-update yet,
+        // will be Editor, which means that we will end up swapping buffers to the editor buffer, and retreiving the wrong active touch
+        // The only way to properly fix this, is to remove defaultUpdateType, and split the player/editor update loops into separate methods, which would be a breaking change.
+        // Until then, we have to make sure to schedule the correct update type here.
+        switch (updateMode)
```
ekcoh, line 180: I think it would be cleaner, and separate concerns, if the type transform from `InputSettings.UpdateMode` to `InputUpdateType` would happen in its own single-purpose (internal) extension method. This whole switch statement would then effectively be part of that static method and this would instead become:
`InputSystem.Update(updateMode.ToInputUpdateType());`

I suspect that conversion happens in multiple places so such a conversion is likely reusable to reduce code-bloat as well.

Assets/Tests/InputSystem/Plugins/UserTests.cs
```diff
@@ -1268,8 +1268,16 @@ public void Users_DoNotReactToEditorInput()
         ++InputUser.listenForUnpairedDeviceActivity;
         InputUser.onUnpairedDeviceUsed += (control, eventPtr) => Assert.Fail("Should not react!");
 
-        runtime.PlayerFocusLost();
-
+        // We now have to run a dynamic update before the press, to make sure the focus state is up to date.
```
ekcoh: Thanks for adding this comment, very helpful. A slight improvement could be to reference what you actually refer to by "pre-update". It is not currently clear exactly what this means. It might help developers circling back to this in the future.

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -4604,7 +4605,13 @@ void DeviceChangeCallback(InputDevice device, InputDeviceChange change)
         InputSystem.onDeviceChange += DeviceChangeCallback;
 
         var eventCount = 0;
-        InputSystem.onEvent += (eventPtr, _) => ++ eventCount;
+        InputSystem.onEvent += (eventPtr, _) =>
```
ekcoh, line 4608: Nitpick: I have no problem with this approach or filtering at all. The first thing that strikes me though is that it sounds like the test is based on observing an exact number of events. This means that the test is very brittle and would brake down when any new (or user plugin injected) event is added. To avoid that, maybe it would make sense to instead only count the events relevant to the test to make it future proof? The system should be resilient to new events being added IMO. What do you think?

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5275,13 +5284,10 @@ void DeviceChangeCallback(InputDevice device, InputDeviceChange change)
...
                             "Sync Mouse", "Sync Mouse2", "Sync Mouse3",
-                            "Sync Keyboard", "Reset Joystick"
-                        }));
-                        // Enabled devices that don't support syncs get reset.
-                        Assert.That(changes, Is.EquivalentTo(new[]
-                        {
-                            "SoftReset Mouse1", "SoftReset Mouse3", "HardReset Joystick", "SoftReset TrackedDevice2"
+                            "Sync Keyboard"
                         }));
+                        // Enabled devices that don't support syncs dont get reset for Ignore Focus as we do not want to cancel any actions.
```
ekcoh, line 5289: Makes a lot of sense IMO.

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5318,7 +5324,13 @@ public void Devices_CanSkipProcessingEventsWhileInBackground()
         Assert.That(performedCount, Is.EqualTo(1));
 
         // Lose focus
-        runtime.PlayerFocusLost();
+        ScheduleFocusEvent(false);
+#if UNITY_INPUTSYSTEM_SUPPORTS_FOCUS_EVENTS
+        // in the new system, we have to process the focus event to update the state of the devices.
```
ekcoh, line 5329: Nitpick: Avoid terms like "new" and "old", instead be specific, e.g. "Since Unity 6000.5.a8, we have..." (or whatever is the appropriate version). "New" is relative to something, and it is not clear to what, so this won't age well I am afraid.

ekcoh: Alternatively just skip what is before the "comma" since the define should also states the change version.

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5318,7 +5324,13 @@ public void Devices_CanSkipProcessingEventsWhileInBackground()
...
+        // In the old system, this wouldn't work and would make the test fal
```
ekcoh, line 5330: Instead of "In the old system" maybe just "Previously"?

Packages/com.unity.inputsystem/InputSystem/Utilities/DelegateHelpers.cs
```diff
@@ -12,25 +12,26 @@ public static void InvokeCallbacksSafe(ref CallbackArray<Action> callbacks, Prof
         {
             if (callbacks.length == 0)
                 return;
-            marker.Begin();
-            callbacks.LockForChanges();
-            for (var i = 0; i < callbacks.length; ++i)
+            using (marker.Auto())
```
ekcoh, line 15: I think it is great you did the conversion to scoped markers. I have wanted it for a long time :) Happy to see C# version now allows it.

VeraMommersteeg:
Yes Im aware I have changes to fix it locally

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -33,10 +33,19 @@ namespace UnityEngine.InputSystem
 {
     using DeviceChangeListener = Action<InputDevice, InputDeviceChange>;
     using DeviceStateChangeListener = Action<InputDevice, InputEventPtr>;
-    using LayoutChangeListener = Action<string, InputControlLayoutChange>;
     using EventListener = Action<InputEventPtr, InputDevice>;
+    using LayoutChangeListener = Action<string, InputControlLayoutChange>;
```
ekcoh, line 38: Looks like irrelevant ordering diff?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -152,10 +161,13 @@ public InputUpdateType defaultUpdateType
         {
             get
             {
-                if (m_CurrentUpdate != default)
+                if (m_CurrentUpdate != InputUpdateType.None)
                     return m_CurrentUpdate;
 
-                #if UNITY_EDITOR
+#if UNITY_EDITOR
+                // We can no longer rely on checking the current focus state, due to this check being used pre-update
```
ekcoh, line 247: Thanks for adding this comment. I think the general take on it is that previously the current focus state was used regardless if deferred events in the queue reflected input prior to the current focus state.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -181,15 +193,36 @@ public InputSettings.ScrollDeltaBehavior scrollDeltaBehavior
...
+#endif
+            }
+        }
+
         public float pollingFrequency
         {
             get
             {
-                #if UNITY_INPUT_SYSTEM_PLATFORM_POLLING_FREQUENCY
+#if UNITY_INPUT_SYSTEM_PLATFORM_POLLING_FREQUENCY
```
ekcoh, line 300: Thanks for correcting my old sins from the past :) Irrelevant to focus but adds formatting consistency to this file. I am surprised this is not altered/fixed by the formatting utility.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -398,12 +431,13 @@ public bool runPlayerUpdatesInEditMode
 #else
             true;
 #endif
+        private bool applicationHasFocus => (focusState & FocusFlags.ApplicationFocus) != 0;
```
ekcoh: Nitpick: This is fine, but the C# standard way of such a check is typically `(focusState & FocusFlags.ApplicationFocus) != FocusFlags.None`. Happy to see you do not use flags extension since I learned it applies boxing under the hood.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -1354,50 +1383,43 @@ public InputDevice AddDevice(InputDeviceDescription description)
...
+                        var command = DisableDeviceCommand.Create();
+                        m_Runtime.DeviceCommand(deviceId, ref command);
+                    }
+                    return null;
                 }
 
-                k_InputAddDeviceMarker.End();
-                return null;
+                var device = AddDevice(layout, deviceId, deviceName, description, deviceFlags);
+                device.m_Description = description;
```
ekcoh, line 1492: Not introduced by your PR but refactored, but isn't this `device.m_Description = description;` redundant? It seems to be passed in `AddDevice` on the line above? (I would assume it is the same)

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3129,151 +3063,142 @@ internal bool ShouldRunUpdate(InputUpdateType updateType)
...
-
-                            currentEventReadPtr = m_InputEventStream.Advance(leaveEventInBuffer: true);
-                        }
-                    }
+                        ProcessBeforeRenderStateEvents(out device, out currentEventReadPtr);
 
                     if (m_InputEventStream.remainingEventCount == 0)
                         break;
 
                     var currentEventTimeInternal = currentEventReadPtr->internalTime;
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -152,10 +161,13 @@ public InputUpdateType defaultUpdateType
         {
             get
             {
-                if (m_CurrentUpdate != default)
+                if (m_CurrentUpdate != InputUpdateType.None)
                     return m_CurrentUpdate;
 
-                #if UNITY_EDITOR
+#if UNITY_EDITOR
+                // We can no longer rely on checking the curent focus state, due to this check being used pre-update
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -2096,23 +2118,29 @@ internal void InstallRuntime(IInputRuntime runtime)
...
-            #if UNITY_EDITOR
+#if UNITY_EDITOR
             m_Runtime.onPlayerLoopInitialization = OnPlayerLoopInitialization;
-            #endif
+#endif
             m_Runtime.pollingFrequency = pollingFrequency;
-            m_HasFocus = m_Runtime.isPlayerFocused;
+
+            focusState = Application.isFocused ? focusState |= FocusFlags.ApplicationFocus
+                : focusState &= ~FocusFlags.ApplicationFocus;
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -1522,8 +1523,9 @@ public void Devices_CanReconnectDevice_WhenDisconnectedWhileAppIsOutOfFocus()
         Assert.That(device, Is.Not.Null);
 
         // Loose focus.
-        runtime.PlayerFocusLost();
-        InputSystem.Update();
+        ScheduleFocusEvent(false);
+        InputSystem.Update(InputUpdateType.Dynamic);
+        //InputSystem.Update();
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -1534,8 +1536,9 @@ public void Devices_CanReconnectDevice_WhenDisconnectedWhileAppIsOutOfFocus()
         Assert.That(InputSystem.devices, Is.Empty);
 
         // Regain focus.
-        runtime.PlayerFocusGained();
-        InputSystem.Update();
+        ScheduleFocusEvent(true);
+        InputSystem.Update(InputUpdateType.Dynamic);
+        //InputSystem.Update();
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5275,13 +5286,10 @@ void DeviceChangeCallback(InputDevice device, InputDeviceChange change)
...
                             "Sync Mouse", "Sync Mouse2", "Sync Mouse3",
-                            "Sync Keyboard", "Reset Joystick"
-                        }));
-                        // Enabled devices that don't support syncs get reset.
-                        Assert.That(changes, Is.EquivalentTo(new[]
-                        {
-                            "SoftReset Mouse1", "SoftReset Mouse3", "HardReset Joystick", "SoftReset TrackedDevice2"
+                            "Sync Keyboard"
                         }));
+                        // Enabled devices that don't support syncs dont get reset for Ignore Forcus as we do not want to cancel any actions.
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
```diff
@@ -171,8 +170,25 @@ public void EnhancedTouch_SupportsEditorUpdates(InputSettings.UpdateMode updateM
         Assert.That(Touch.activeTouches[0].touchId, Is.EqualTo(2));
 
         // Switch back to player.
-        runtime.PlayerFocusGained();
-        InputSystem.Update();
+        ScheduleFocusEvent(true);
+
+        // We have to schedule the specificic update type that the player is configured to process events in,
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
```diff
@@ -1171,7 +1187,8 @@ public void EnhancedTouch_ActiveTouchesGetCanceledOnFocusLoss_WithRunInBackgroun
         else
         {
             // When not running in the background, the same thing happens but only on focus gain.
-            runtime.PlayerFocusGained();
+            //runtime.PlayerFocusGained();
```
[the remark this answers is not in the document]
VeraMommersteeg: fixed

Packages/com.unity.inputsystem/InputSystem/LegacyInputManager.cs
```diff
@@ -0,0 +1,267 @@
...
+    /// </summary>
+    internal enum FocusFlags : ushort
+    {
+        /// <summary>
+        /// No focus state is active.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// The application has focus.
```
ekcoh: Maybe we should make this comment more crisp (potentially both here and in module), to clarify that ApplicationFocus means "GameView focus" when running in the editor, see https://docs.unity3d.com/6000.3/Documentation/ScriptReference/MonoBehaviour.OnApplicationFocus.html. This could avoid confusion and strengthen the definition.

Packages/com.unity.inputsystem/InputSystem/InputManager.LegacyFocusHandling.cs
```diff
@@ -0,0 +1,267 @@
+using System.Runtime.CompilerServices;
```
ekcoh, line 1: The name of this class can be very confusing since it is similar to what Input Manager in Unity engine is often called. Since this is a partial class containing a selected set of functionality I would suggest renaming it to `InputManager.LegacyFocusHandling.cs` or similar instead.

ekcoh: (But it make a lot of sense to extract this functionality into its own file)

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
```diff
@@ -206,8 +207,15 @@ public Action<bool> onPlayerFocusChanged
...
+#endif
 
-        public bool isPlayerFocused => Application.isFocused;
+        private FocusFlags m_FocusState = FocusFlags.None;
+        public FocusFlags focusState
+        {
+            get => m_FocusState;
+            set => m_FocusState = value;
+        }
+        public bool isPlayerFocused => (m_FocusState & FocusFlags.ApplicationFocus) != 0;
```
ekcoh: Nitpick: Compare `FocusFlags.None` instead of `0`?

ekcoh (commented):
Just commenting for now as a half-way review. Need to clone this PR in order to have a realistic chance to provide relevant feedback on InputManager.cs diff. Looking at the diff alone is very complex.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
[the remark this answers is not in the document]
VeraMommersteeg: fixed

ekcoh:
Tested the current branch locally on 2022.3 and noticed the following:
- 🟢 Playmate tests in editor succeeding
- 🔴 Player tests - hangs the player/editor (looks like recursive call to InputManager.Update)  
- 🔴 Editor tests - 3 failures (regressions of behaviour on versions older than 6000.5a8):
   - Move_ShouldMutateDeviceState_WithinPlayModeTestFixtureContext
   - Press_ShouldMutateDeviceState_WithinPlayModeTestFixtureContext 
   - Touch_ShouldMutateDeviceState_WithinPlayModeTestFixtureContext

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
VeraMommersteeg: yeah true, that is the standard, c++ brain kicking in

Assets/Tests/InputSystem/CoreTests_Devices.cs
VeraMommersteeg: That whole test is a mess IMO, it's the 800 line test and needs to be completely broken up, and yes having the test depend on event count makes it brittle. That is the reason I had to do the check there so it only counts the tests that it should for that test. I would say we leave it for now and instead just do a full refactor of this test as a separate pr

VeraMommersteeg: I'll make a jira card for this as something that we need to do at some point. Which is looking at all the tests there and see if they make sense

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
VeraMommersteeg: fixed

Packages/com.unity.inputsystem/InputSystem/LegacyInputManager.cs
VeraMommersteeg: fixed

Packages/com.unity.inputsystem/InputSystem/InputManager.LegacyFocusHandling.cs
VeraMommersteeg: fixed

Packages/com.unity.inputsystem/InputSystem/NativeInputRuntime.cs
VeraMommersteeg: fixed

Assets/Tests/InputSystem/Plugins/UserTests.cs
VeraMommersteeg: fixed

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
VeraMommersteeg: no idea, can look into it in a follow-up pr

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3478,213 +3315,515 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+            {
+                if (possibleFocusEvent->type == focusEventType && !gameShouldGetInputRegardlessOfFocus)
+                {
+                    // If the next event is a focus event and we're not supposed to get input of the current update type in the next one, drop current event.
+                    // This ensures that we don't end up with a half processed events due to swapping buffers between editor and player,
+                    // such as InputActionPhase.Started not being finished by a InputActionPhase.Performed and ending up in a pressed state in the previous update type
+                    m_InputEventStream.Advance(false);
+                    return true;
+                }
+            }
```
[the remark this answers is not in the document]
VeraMommersteeg: Solved it for now that it soft resets the device so it sends out a canceled

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
VeraMommersteeg: This only happens for this one test because this is the only one that switches update type in one test. Adding a function for a switch case for one test is a bit overkil imo 😅

Assets/Tests/InputSystem/Plugins/EnhancedTouchTests.cs
VeraMommersteeg: I've changed it to ScheduleFocusChangedEvent since we are scheduling a change focus event, as we do not want to mask the details so it helps people understand what the test does

ekcoh:
> Tested the current branch locally on 2022.3 and noticed the following:
> 
> * 🟢 Playmate tests in editor succeeding
> * 🔴 Player tests - hangs the player/editor (looks like recursive call to InputManager.Update)
> * 🔴 Editor tests - 3 failures (regressions of behaviour on versions older than 6000.5a8):
>   
>   * Move_ShouldMutateDeviceState_WithinPlayModeTestFixtureContext
>   * Press_ShouldMutateDeviceState_WithinPlayModeTestFixtureContext
>   * Touch_ShouldMutateDeviceState_WithinPlayModeTestFixtureContext

This seems fine now after last batch of commits.

VeraMommersteeg:
Those tests are passing on CI though, when I was fixing some tests it ended up in a bad state, might want to restart editor and delete any generated files

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3592,236 +3434,557 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
-                            }
+#if UNITY_EDITOR
+        // Handles editor-specific focus/background early-out behavior and advances the stream accordingly.
+        // Returns true if event should be skipped (stream advanced), false otherwise.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private unsafe bool SkipEventDueToEditorBehaviour(InputUpdateType updateType, FourCC currentEventType, bool dropStatusEvents, double currentEventTimeInternal)
+        {
+            var focusEventType = new FourCC(FocusConstants.kEventType);
+
+            // Check if the next event is a FocusEvent. If so, cancel any in-progress actions before
```
ekcoh, line 3502: I guess this complexity would disappear with split contexts since that would imply that event queue is processed regardless?

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3592,236 +3434,557 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+#if UNITY_EDITOR
+                m_Diagnostics?.OnEventTimestampOutdated(new InputEventPtr(currentEventReadPtr), device);
+#elif UNITY_ANDROID
+                // Android keyboards can send events out of order: Holding down a key will send multiple
+                // presses after a short time, like on most platforms. Unfortunately, on Android, the
+                // last of these "presses" can be timestamped to be after the event of the key release.
+                // If that happens, we'd skip the keyUp here, and the device state will have the key
+                // "stuck" pressed. So, special case here to not skip keyboard events on Android. ISXB-475
+                // N.B. Android seems to have similar issues with touch input (OnStateEvent, Touchscreen.cs)
+                if (!(device is Keyboard))
```
ekcoh, line 3746: This looks odd, is it so that return is conditioned on android but otherwise always happen? If that is the case I would recommend #else here and have two return statements, splitting an unbraced if makes it hard to read.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3592,236 +3434,557 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private unsafe void ProcessIMECompositionEvent(InputDevice device, InputEvent* currentEventReadPtr)
+        {
+            var imeEventPtr = (IMECompositionEvent*)currentEventReadPtr;
+            var textInputReceiver = device as ITextInputReceiver;
+            textInputReceiver?.OnIMECompositionChanged(imeEventPtr->compositionString);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void ProcessDeviceRemoveEvent(InputDevice device)
```
ekcoh, line 3830: Thanks for splitting out each event processing into its own function, makes it a lot easier to digest and aggressive inlining shouldn't lead to any perf complications anyway.

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
```diff
@@ -3592,236 +3434,557 @@ private unsafe void OnUpdate(InputUpdateType updateType, ref InputEventBuffer ev
...
+
+            bool runInBackground =
+#if UNITY_EDITOR
+                // In the editor, the player loop will always be run even if the Game View does not have focus. This
+                // amounts to runInBackground being always true in the editor, regardless of what the setting in
+                // the Player Settings window is.
+                //
+                // If, however, "Game View Focus" is set to "Exactly As In Player", we force code here down the same
+                // path as in the player.
+                //if we are in editor, and the editor input behaviour is that everything that is pressed goes to the game view, or we are allowed to run in the background
```
ekcoh, line 3877: Nitpick: Missing space before if.

ekcoh (approved):
Thanks for doing this comprehensive work @VeraMommersteeg. Changes looks great and I couldn't spot anything that looks incorrect, but added some additional nitpick-level comments. 
I like that legacy behaviour has been split out from InputManager since this was already partial that just simplifies removing legacy behaviour later.
I also like that you have added inline comments motivating complex behaviour in update loop (or indirect sub-parts of update), I also think its great you took some time to refactor / break up the update flow and its great you managed to keep many parts shared between legacy and new.

The only thing I would recommend checking either before or after this lands, would be to compare performance test results locally or in observer to check that the differences in event queue processing doesn't generate any performance regression. Functional correctness of course has higher priority, but we could negatively affect the update loop with these changes. IMO such performance issues should not prevent this from landing, but if we find anomalies we should likely try to address it. Great job!

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
VeraMommersteeg: Yes this was the best workaround I could do for now without dropping the event, as this will still send out a canceled and make sure we dont execute any hold interaction when we regain focus

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
VeraMommersteeg: I didnt change any of the code in here, so not sure what exactly that would imply. Might be good to look at this in a separate pr

Packages/com.unity.inputsystem/InputSystem/InputManager.cs
VeraMommersteeg: will add it to the changes to do in a followup pr

Pauliusd01 (changes requested):
Just leaving a status update about an issue we discussed with ps5 controllers remaining active in the background even when focus is lost

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5318,7 +5391,13 @@ public void Devices_CanSkipProcessingEventsWhileInBackground()
         Assert.That(performedCount, Is.EqualTo(1));
 
         // Lose focus
-        runtime.PlayerFocusLost();
+        ScheduleFocusChangedEvent(applicationHasFocus: false);
+#if UNITY_INPUTSYSTEM_SUPPORTS_FOCUS_EVENTS
+        // in the new system, we have to process the focus event to update the state of the devices.
+        // In the old system, this wouldn't work and would make the test fal
```
ekcoh, line 5330:
```suggestion
        // when focus changes were out-of-sync with event queue, this wouldn't work and would make the test fail
```

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5318,7 +5391,13 @@ public void Devices_CanSkipProcessingEventsWhileInBackground()
         Assert.That(performedCount, Is.EqualTo(1));
 
         // Lose focus
-        runtime.PlayerFocusLost();
+        ScheduleFocusChangedEvent(applicationHasFocus: false);
+#if UNITY_INPUTSYSTEM_SUPPORTS_FOCUS_EVENTS
+        // in the new system, we have to process the focus event to update the state of the devices.
```
ekcoh, line 5329:
```suggestion
        // when focus change is an event, we have to process the focus event to update the state of the devices.
```

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5296,6 +5296,73 @@ void DeviceChangeCallback(InputDevice device, InputDeviceChange change)
         #pragma warning restore CS0162
     }
 
+    [Test]
```
ekcoh: Thanks for adding this back for the background scenarios, it reduces risk of future regression.

ekcoh:
Looked through the latter commits, nothing major to call out so looks good to me. Added two suggestions for wording previously commented on.

Assets/Tests/InputSystem/CoreTests_Devices.cs
VeraMommersteeg: im removing the test again, as theres an odd timing with it, so will just do that in a separate pr. and would like to land this for now

Pauliusd01 (approved):

VeraMommersteeg:
failing test is the same as develop. will bypass and push

PR #2402, opened 2026-03-31
NEW: Input consumption - Priorities instead of complexity [ISX-2510]

Darren-Kelly-Unity:
## Purpose of this PR

Implements priority-based shortcut overlap resolution for [ISX-2510](https://jira.unity3d.com/browse/ISX-2510). Adds a new `InputAction.Priority` property and a `shortcutKeysUseActionPriority` setting so users can explicitly control which actions win when multiple bindings share the same control — replacing the opaque automatic complexity calculation that caused unpredictable results.

## Release Notes

- **[ISX-2510]** Added `InputAction.Priority` property (range 0–65535) to let users assign explicit per-action priority for shortcut overlap resolution.
- **[ISX-2510]** Added `InputSettings.shortcutKeysUseActionPriority` setting. When enabled, overlap resolution uses action priority instead of automatic composite complexity. Priority field is shown in the Input Actions editor when this mode is active.
- **[ISX-2510]** `shortcutKeysConsumeInput` (complexity-based resolution) and `shortcutKeysUseActionPriority` (priority-based resolution) are now independent; priority mode takes precedence when both are enabled.

## Functional Testing status

- New test file `CoreTests_ActionsPriority.cs` (971 lines) covers priority assignment, clamping, overlap suppression, same-priority co-performance, and interaction with `shortcutKeysConsumeInput`.
- Existing `CoreTests_Actions.cs` updated for renamed internal `controlGroupingAndComplexity` → `controlGroupingAndPriority` field.
- Manual testing: verified Priority field appears/disappears in the Input Actions editor when toggling `shortcutKeysUseActionPriority` in Project Settings.

## Performance Testing Status

No performance impact. The priority resolution path follows the same O(n²) grouping loop as the existing complexity path; no additional allocations.

## Overall Product Risks

**Technical Risk:** 2 — Replaces internal grouping logic in `InputActionState` that has many edge cases; complexity path is preserved as a fallback, but the two paths share the same memory layout (`controlGroupingAndPriority`). Regression risk is mitigated by the new test suite.

**Halo Effect:** 1 — Change is opt-in via a new setting; existing projects using `shortcutKeysConsumeInput` are unaffected unless they also enable `shortcutKeysUseActionPriority`.

## Class diagram

Changes across `origin/develop...HEAD` — new `InputActionStateMonitorIndex` struct, `InputAction.Priority`, `InputSettings.shortcutKeysUseActionPriority`, and the internal `ControlGroupingTable`.

<details>
<summary>Class diagram</summary>

```mermaid
classDiagram
  direction TB

  class InputAction {
    +int Priority
    +int MinPriority$
    +int MaxPriority$
    +ClampPriority(int) int$
    -int m_Priority
    ~OnActionPriorityChanged()
  }

  class InputSettings {
    +bool shortcutKeysConsumeInput
    +bool shortcutKeysUseActionPriority
    ~bool IsShortcutResolutionUsingActionPriority
    ~bool IsShortcutResolutionUsingComplexity
    ~bool IsShortcutComplexityModifierOrderActive
    -bool m_ShortcutKeysUseActionPriority
  }

  class InputActionState {
    -ushort* controlGroupingAndPriority
    +InitializeControlGrouping()
    ~OnActionPriorityChanged(InputAction)
    -GetControlMonitorGroupIndex(int) uint
    -GetControlBindingPriority(int) int
  }

  class ControlGroupingTable {
    <<static>>
    +int Stride$
    +GroupElementIndex(int) int$
    +PriorityElementIndex(int) int$
  }

  class InputActionStateMonitorIndex {
    <<struct>>
    +long Packed
    +int ControlIndex
    +int BindingIndex
    +int MapIndex
    +int Priority
    +Create(int,int,int,int) InputActionStateMonitorIndex$
    +FromPacked(long) InputActionStateMonitorIndex$
  }

  class InputActionMap {
    ~InputActionState m_State
  }

  InputActionState --> ControlGroupingTable : uses
  InputActionState --> InputSettings : reads
  InputActionState --> InputActionStateMonitorIndex : creates
  InputAction --> InputActionMap : belongs to
  InputActionMap --> InputActionState : owns
  InputAction ..> InputActionState : notifies priority
```

</details>

## Documentation Impact

No new pages required. Both `InputAction.Priority` and `InputSettings.shortcutKeysUseActionPriority` are already documented on this branch:

- **Settings.md** — shortcut resolution settings table (rows covering both modes)
- **Actions.md** — "Overlapping bindings and action priority" section with code example


Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
         Assert.That(!action3.WasPerformedThisFrame());
     }
 
+    [Test]
+    [Category("Actions Priority")]
```
ekcoh: If "Actions Priority" is its own category (it makes sense so I like it), could we also  move out those tests into e.g. "CoreTests_Actions_Priority.cs" (since partial), to simplify finding these and working with these. This file is way too large.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+    [TestCase("ctrl", "x", true)]
+    [TestCase("ctrl", "x", false)]
+    public void Actions_Priority_PressingModifierShortcutWithSameBinding_TriggersHighestPriorityAction(string sharedModifier, string binding1, bool legacyComposites)
```
ekcoh: We some inline doc be added here to explain the arguments? It's not clear to me what binding1 or legacyComposites mean unless I read the whole test. And reading it I am still confused what legacyComposites mean/imply?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+        action2.AddCompositeBinding((legacyComposites ? "ButtonWithOneModifier" : "OneModifier") + "(overrideModifiersNeedToBePressedFirst)")
+            .With("Modifier", "<Keyboard>/" + sharedModifier)
+            .With(legacyComposites ? "Button" : "Binding", "<Keyboard>/" + binding1);
+
+        action1.Priority = 9;
+        action2.Priority = 10;
+
+        map.Enable();
+
+        Assert.That(action1.WasPerformedThisFrame(), Is.False);
```
ekcoh: Remove these asserts, what do they test? Seems like a sanity check or another test? "Actions should not trigger if there is no input"

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+
+        var action1 = map.AddAction("action1");
+        action1.AddBinding("<Keyboard>/" + binding1);
+
+        var action2 = map.AddAction("action2");
+        action2.AddCompositeBinding((legacyComposites ? "ButtonWithOneModifier" : "OneModifier") + "(overrideModifiersNeedToBePressedFirst)")
+            .With("Modifier", "<Keyboard>/" + sharedModifier)
+            .With(legacyComposites ? "Button" : "Binding", "<Keyboard>/" + binding1);
+
+        action1.Priority = 9;
```
ekcoh: Is this also valid for 
action1.Priority = -2
action2.Priority = -1
?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
         Assert.That(!action3.WasPerformedThisFrame());
     }
 
+    [Test]
+    [Category("Actions Priority")]
+    [TestCase("ctrl", "x", true)]
```
ekcoh: I find this choice of parameterisation confusing, but maybe it's just me. However it looks like the test case structure here is generally:

public void Actions_PriorityFirstActionShouldFire_IfFirstActionHasHigherPriority(Action first, Action second)

and just keep the last part of the test. Actions could be constructed as a one-liner instead and then the same test could test multiple pairs that should fulfil the condition. Worthwhile?

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: Is there a similar test that verifies both are fired if same priority?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+        // Action1 is not performed because action2 has higher priority on the shared key.
+        Assert.That(action1.WasPerformedThisFrame(), Is.False);
+        Assert.That(action2.WasPerformedThisFrame(), Is.True);
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    [TestCase("ctrl", "x", true)]
+    [TestCase("ctrl", "x", false)]
+    public void Actions_Priority_SimpleBindingHigherPriority_PreemptsCompositeOnSharedKey(string sharedModifier, string binding1, bool legacyComposites)
```
ekcoh: I must admit I do not understand what this test is intended to verify? Mind elaborating since I fail to decipher the requirement? I believe the general naming pattern to be:

<FunctionOrFeature>_ShouldXXX_IfYYY, e.g.
Actions_Priority_ShouldOnlyAllowActionWithHighestPriorityToSignalPerformed_IfTwoActionsArePerformedSimultaneouslyButHaveDifferentPriorities

Or keep it shorter but add a TestDescription("...") telling the same story.

Maybe I would understand this test better by reviewing implementation but since requested to look at tests only I did not ATM.

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+
+        Assert.That(action1WasPerformed, Is.False);
+        Assert.That(action2.WasPerformedThisFrame(), Is.True);
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    [TestCase("ctrl", "shift", "x", false)]
+    [TestCase("ctrl", "shift", "x", true)]
+    public void Actions_Priority_EqualPriority_ConflictingShortcuts_FirstActionInMapWins(string sharedModifier, string sharedModifier2, string binding1, bool legacyComposites)
```
ekcoh: This is surprising IMO. Why wouldn't both fire with equal priority?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+        // Action1 is not performed because action2 has higher priority on the shared key.
+        Assert.That(action1WasPerformed, Is.False);
+        Assert.That(action2.WasPerformedThisFrame(), Is.True);
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    [TestCase("ctrl", "shift", "x", false)]
+    [TestCase("ctrl", "shift", "x", true)]
+    public void Actions_Priority_TwoShortcuts_HigherNumericPriorityWins(string sharedModifier, string sharedModifier2, string binding1, bool legacyComposites)
```
ekcoh: How is this different from the first test in this file? I think the tests would be simpler to understand if parameterised differently (maybe on Action) and reducing redundancy code-wise and keeping the exact same naming conventions between them to spot the differences/flavours.

Assets/Tests/InputSystem/CoreTests_Actions.cs
ekcoh: I see next test is both fire with equal priority but this one is not, I fail to see why?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+
+        Assert.That(action1.WasPerformedThisFrame(), Is.True);
+        Assert.That(action2WasPerformed, Is.False);
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    [TestCase("ctrl", "shift", "x", false)]
+    [TestCase("ctrl", "shift", "x", true)]
+    public void Actions_Priority_BothPrioritiesZero_ConflictingShortcuts_BothPerform(string sharedModifier, string sharedModifier2, string binding1, bool legacyComposites)
```
ekcoh: This is also true if both priorities are e.g. 3 right?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+
+        Assert.That(action1WasPerformed, Is.True);
+        Assert.That(action2WasPerformed, Is.True);
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    [TestCase("ctrl", "shift", "x", false)]
+    [TestCase("ctrl", "shift", "x", true)]
+    public void Actions_Priority_ShortcutConsumeDisabled_BothPerformDespiteDifferentPriorities(string sharedModifier, string sharedModifier2, string binding1, bool legacyComposites)
```
ekcoh: I was under the impression that shortcutKeysConsumeInput would be irrelevant after adding priority but it seems that might have changed along the way?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+
+        InputSystem.Update();
+
+        Assert.That(action1.WasPerformedThisFrame(), Is.False);
+        Assert.That(action2.WasPerformedThisFrame(), Is.False);
+    }
+
+    [UnityTest]
+    [Category("Actions Priority")]
+    public IEnumerator Actions_Priority_TwoNonConflictingShortcuts_ReversedPriorityOrder_BothStillPerform()
```
ekcoh: This could probably be combined with the other similar test and just pass parameterised actions in different order? Test logic can be fixed and just alter argument order to test?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+
+        InputSystem.Update();
+
+        Assert.That(action1.WasPerformedThisFrame(), Is.False);
+        Assert.That(action2.WasPerformedThisFrame(), Is.False);
+    }
+
+    [UnityTest]
+    [Category("Actions Priority")]
+    public IEnumerator Actions_Priority_TwoNonConflictingShortcuts_ReverseActionDeclarationOrder_BothPerform()
```
ekcoh: Again, just a small mutation but code bloat

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -299,6 +299,482 @@ public void Actions_WhenShortcutsEnabled_CanConsumeInput(bool legacyComposites)
...
+
+        InputSystem.Update();
+
+        Assert.That(action1.WasPerformedThisFrame(), Is.False);
+        Assert.That(action2.WasPerformedThisFrame(), Is.False);
+    }
+
+    [UnityTest]
+    [Category("Actions Priority")]
+    public IEnumerator Actions_Priority_TwoNonConflictingShortcuts_EqualHighPriority_BothPerform()
```
ekcoh: Just parameter difference so why another test?

ekcoh (commented):
Thanks for putting up this draft PR. I looked through the tests only so far and I think we should have a discussion around it which would be more helpful than doing asynchronous feedback on the comments. Test cases generally make sense but I fail to decipher some differences and rules in them I fail to understand. Also I am missing tests using priorities when interaction is not a shortcut modifier, I would assume the same rules apply regardless of binding? I think there is opportunity to reduce code diff here by doing some code reuse in the tests without having tests full of if statements.

I also think it would be good to have tests with repeated input which I mentioned before. I didn't spot any such tests unless I missed something. This would look like e.g. with a binding 'X', here value of X button provided as binary

```
X:                 0 1 1
Action 1 (Prio 1): 0 - 0
Action 2 (Prio 2): 0 1 0
                   a b c

   a: button up, nothing happens
   b: button down, triggers both, but only action 2 fires notification since higher prio
   c: button down again (repeat), triggers none since action 1 saw and reacted to state transition but did not fire
```

Assets/Tests/InputSystem/CoreTests_Actions.cs
K-Tone: We added it back for now while debugging. We need to have a task in the epic around how we migrate from the current input consumption to the new one.

Assets/Samples/ProjectWideActions/ProjectWideActionsExample.cs
```diff
@@ -29,6 +31,8 @@ void Start()
                 previous = InputSystem.actions.FindAction("Player/Previous");
                 sprint = InputSystem.actions.FindAction("Player/Sprint");
                 crouch = InputSystem.actions.FindAction("Player/Crouch");
+                b = InputSystem.actions.FindAction("Player/B");
+                shiftB = InputSystem.actions.FindAction("Player/Shift B");
```
K-Tone: I would probably not ship this code. If we want a dedicated sample - we should deliver something like my visual test project. If we don't - then this code is purely for debugging and we're better off removing it for now. What do you think?

ekcoh:
@Darren-Kelly-Unity @K-Tone Is this ready for review? It looks to have multiple u-pr comments on it and failing tests indicating it might not be but at the same time it is not a Draft nor having the "work-in-progress" label? I would suggest to either convert back to Draft or add that label for clarity and follow up on outstanding comments to make it clear what the current state is. Feel free to re-request review when ready.

**Current things to address:**
- Fix failing tests
- Fix formatting
- Follow up on outstanding review comments from previously
- Run AI review on new state of code

Assets/Tests/InputSystem/CoreTests_Actions.cs
Darren-Kelly-Unity: Yes, this has been done! :)

Assets/Tests/InputSystem/CoreTests_Actions.cs
Darren-Kelly-Unity: These tests have now been refactored completely and put in a new file since I last looked at these changes!

Assets/Tests/InputSystem/CoreTests_Actions.cs
Darren-Kelly-Unity: These tests have now been refactored completely and put in a new file since I last looked at these changes!

The new changed tests can be found in CoreTests_ActionsPriority.

Assets/Tests/InputSystem/CoreTests_Actions.cs
Darren-Kelly-Unity: These tests have now been refactored completely and put in a new file since I last looked at these changes!

The new changed tests can be found in CoreTests_ActionsPriority.

Assets/Tests/InputSystem/CoreTests_Actions.cs
Darren-Kelly-Unity: This may still be a valid thing to check my side so I took note of it.


But maybe re-reviewing the changed tests can be found in CoreTests_ActionsPriority first is best.

Darren-Kelly-Unity:
> ```cs
>             #if UNITY_ANALYTICS || UNITY_EDITOR
>             InputAnalytics.Initialize(this);
>             m_Runtime.onShutdown = () => InputAnalytics.OnShutdown(this);
> #endif
> 
>             m_StateMonitors = new InputManagerStateMonitors(() => m_DevicesCount, () => isProcessingEvents, m_Runtime);
>         }
> ```

For issue 1:


For issue 2:
From looking at the code I can't seem to see any issue here.

As a product risk for shipping players, it is low today because production code does not use reInstallRuntime after initial creation.

> ## PR Reviewer Guide 🔍
> #### (Review updated until commit [ced22e0](https://github.com/Unity-Technologies/InputSystem/commit/ced22e051ccae23391e1b1c3a9b02ca019fd439a))
> Here are some key observations to aid the review process:
> 
> **🎫 Ticket compliance analysis ✅**
> **[ISX-2510](https://jira.unity3d.com/browse/ISX-2510) - PR Code Verified**
> 
> Compliant requirements:
> 
> * Improve reliability of overlapping shortcut interactions instead of relying only on automatic complexity.
> * Document the behavior changes.
> * Complete programming for the feature.
> * Complete automated testing for the feature.
> 
> Requires further human verification:
> 
> * Align the shortcut-overlap solution with XR expectations/dependencies.
> * Complete planning and design sign-off for the feature.
> * Merge the work to `develop`.
> * Confirm manual/editor UX behavior for the new settings and Priority field visibility.
> 
> ⏱️ **Estimated effort to review**: 4 🔵🔵🔵🔵⚪
> This PR is large and cross-cutting, touching runtime action resolution, monitor infrastructure, editor UI, serialization, analytics, docs, and many tests.
> 
> 🏅 **Score**: 81
> The change is well-covered and thoughtfully implemented, but there are a couple of runtime edge cases that can break the advertised shortcut-resolution behavior.
> 
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Hold leakage**](https://github.com/Unity-Technologies/InputSystem/pull/2402/files#diff-b77be6054c214ebc44deeb7f6b4309194f2dab8ee301458f1d22869aeba538caR2637-R2648)
> Priority-based consumption only suppresses overlaps when there is a current input event to mark as handled. Interactions that reach `Performed` from a timeout instead of a fresh event can still complete on lower-priority overlapping actions after the higher-priority action has already won. A concrete case is two overlapping `hold` actions such as `B` and higher-priority `Shift+B`: keep both held long enough and both can perform.
> 
> ```cs
> // When we perform an action, we mark the event handled such that FireStateChangeNotifications()
> // can reset other monitors (priority: strictly lower priority; complexity: same group after sort).
> var settings = InputSystem.settings;
> var secondary = GetControlBindingPriority(trigger.controlIndex);
> var shouldConsumeHandled = settings.IsShortcutResolutionUsingActionPriority
>     ? secondary > 0
>     : secondary > 1;
> if (shouldConsumeHandled &&
>     // we can end up switching to performed state from an interaction with a timeout, at which point
>     // the original event will probably have been removed from memory, so make sure to check
>     // we still have one
>     m_CurrentlyProcessingThisEvent.valid)
> ```
> 
> [**Monitor reset**](https://github.com/Unity-Technologies/InputSystem/pull/2402/files#diff-6e654fc7619f66d16ae0e01244aaf63335c15a9539524c813e2c304b7e74d492R2291-R2297)
> `InstallRuntime` now recreates the entire state-monitor manager, which drops all registered control monitors and pending timeouts. If the runtime is reinstalled after actions are already enabled, those actions can stop receiving state changes or timer callbacks until they are re-resolved or re-enabled.
> 
> ```cs
>             #if UNITY_ANALYTICS || UNITY_EDITOR
>             InputAnalytics.Initialize(this);
>             m_Runtime.onShutdown = () => InputAnalytics.OnShutdown(this);
> #endif
> 
>             m_StateMonitors = new InputManagerStateMonitors(() => m_DevicesCount, () => isProcessingEvents, m_Runtime);
>         }
> ```
> 
> * [x]  Update review
> 
> 🤖 Helpful? Please react with 👍/👎 | Questions❓Please reach out in Slack [#ask-u-pr](https://unity.enterprise.slack.com/archives/C08MXQ4MDNW)

I looked into the Hold leakage mentioned and it doesn't seem to be much of an issue:
When Hold reaches Performed only because the timer expired, that transition often does not go through m_CurrentlyProcessingThisEvent.valid == true, and does not set handled.

From the current implementation: the code knowingly avoids setting handled when there is no valid event—likely to avoid writing handled on invalid/stale memory—so this is an intentional safety tradeoff with observable leakage for timeouts.

Pauliusd01:
https://github.com/user-attachments/assets/fb7f7ede-c123-48a0-833e-b4116d9b66c6

the priority field auto completes after entering a single digit so you can never enter anything about 9 without pasting it in (forgot to enable the key visualiser in the video but I am typing 10 in the field and only 1 is allowed to go through, then I paste it manually)

Pauliusd01:
image

The naming should be tweaked as the text is too long and does not fit (I assume we don't want to move the checkmark forward and make it out of line with the others) How about dropping the "shortcuts" part in the end, they're already under the shortcut support category. For example: Complexity-based & Action Priority. 

Also, some subjective tooltip suggestions below:

"When several enabled actions share the same control, the action with the higher Priority value is ordered first and consume input so lower-priority actions do not trigger on the same event. Set Priority on each in the Input Actions editor. Serialized priority values are kept when the option is disabled" 
-> 
"When several enabled actions are bound to the same control, the one with the highest Priority is evaluated first and consumes the input, so lower-priority actions don't also trigger from that event. Set each action's Priority in the Input Actions editor. Priority values stay saved on the asset even when Action Priority Shortcut Resolution is turned off."

and

"Composite bindings can consume overlapping input using binding-chain depth (complexity): actions with more composite parts are considered before simpler bindings on the same controls. This works well for shortcut keys; when two composites have the same depth, resolution can be non-deterministic. Conflicts can occur between different action maps (for example UI navigation versus gameplay on the same keys) but not between different action assets. Since consumption applies to enabled actions only, disable maps or actions you do not need in the current context to reduce surprises"
->
"Overlapping shortcuts are resolved by composite complexity: an action whose composite has more parts (e.g. Ctrl+Shift+S) wins over one with fewer parts — or a plain binding — on the same control, and consumes the input so the simpler action doesn't also fire. Two composites of equal complexity have no guaranteed order between them. Resolution applies across action maps within the same asset (for example UI navigation vs. gameplay on the same keys), but never across separate action assets. Only enabled actions consume input, so disable maps or actions you don't need in the current context to avoid unexpected behaviour." (or behaviours, don't know which is best. Maybe even keeping surprises is fine but sounds odd to me)

Pauliusd01:
I've tested the plan scenarios from here: https://github.com/Unity-Technologies/InputSystem/pull/2402#issuecomment-4561206339 and everything besides this part passes:

>Hold Interaction suppression: Test the "Hold leakage" scenario mentioned in PR comments. Verify that a higher-priority action suppresses a lower-priority action even when the transition is triggered by a timer (e.g., two overlapping Holds). 

Although I don't think I would agree that's a real bug, probably needs clarity. Otherwise seems good to me, does what it is supposed to do and much better than the old one

Tests if you want to double check: 
[InputPRTests.zip](https://github.com/user-attachments/files/28380715/InputPRTests.zip)

Darren-Kelly-Unity:
> I've tested the plan scenarios from here: [#2402 (comment)](https://github.com/Unity-Technologies/InputSystem/pull/2402#issuecomment-4561206339) and everything besides this part passes:
> 
> > Hold Interaction suppression: Test the "Hold leakage" scenario mentioned in PR comments. Verify that a higher-priority action suppresses a lower-priority action even when the transition is triggered by a timer (e.g., two overlapping Holds).
> 
> Although I don't think I would agree that's a real bug, probably needs clarity. Otherwise seems good to me, does what it is supposed to do and much better than the old one
> 
> Tests if you want to double check: [InputPRTests.zip](https://github.com/user-attachments/files/28380715/InputPRTests.zip)

Hi Paulius,

I also looked into this Test the "Hold leakage" scenario mentioned in PR comments. Verify that a higher-priority action suppresses a lower-priority action even when the transition is triggered by a timer (e.g., two overlapping Holds).

But I didn't see it as a bug either to be honest. The fix would be messy and change a lot so it wouldn't be worth it in my opinion to try to solve this.

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
```diff
@@ -0,0 +1,964 @@
...
+        // Cleanup key presses
+        ReleaseBindingsForActions(keyboard, action1, action2);
+
+        Assert.That(action1.WasPerformedThisFrame(), Is.False);
+        Assert.That(action2.WasPerformedThisFrame(), Is.False);
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    [TestCaseSource(nameof(k_TwoInputActionTestCases))] // TODO: Darren, Should both actions be performed this frame here??
```
ekcoh: TODO left in source code, needs action?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
```diff
@@ -0,0 +1,964 @@
...
+
+        Assert.That(roundTrip.MapIndex, Is.EqualTo(created.MapIndex));
+        Assert.That(roundTrip.ControlIndex, Is.EqualTo(created.ControlIndex));
+        Assert.That(roundTrip.BindingIndex, Is.EqualTo(created.BindingIndex));
+        Assert.That(roundTrip.Priority, Is.EqualTo(created.Priority));
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    public void Actions_Priority_InputActionStateMonitorIndex_PriorityRoundTripsFullSixteenBits()
```
ekcoh, line 450: To me it's not fully clear what the intent of this test really is? Wouldn't priority and complexity be indirectly covered by other tests? Seems like this test is non-functional and just validates the design?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
```diff
@@ -0,0 +1,964 @@
...
+        Assert.That(index300.Priority, Is.EqualTo(300));
+        Assert.That(InputActionState.GetComplexityFromMonitorIndex(index300.Packed), Is.EqualTo(300));
+
+        var index65535 = InputActionStateMonitorIndex.Create(0, 1, 0, priority: 65535);
+        Assert.That(index65535.Priority, Is.EqualTo(65535));
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    public void Actions_Priority_PrioritiesAbove255_ResolveInOrder()
```
ekcoh: What does this mean? Is there some magic limit of 255 priority where behavior changes?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
```diff
@@ -0,0 +1,964 @@
...
+        Assert.That(higher.WasPerformedThisFrame(), Is.True);
+        Assert.That(lower.WasPerformedThisFrame(), Is.False);
+
+        Release((ButtonControl)keyboard.xKey);
+        InputSystem.Update();
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    public void Actions_Priority_ChangingPriorityWhileEnabled_ReplacesStateMonitorInsteadOfDuplicating()
```
ekcoh, line 487: Do we need to support this? Allowing priority to change while being enabled sounds like it adds complexity that would generally not be needed IMO, isn't it sufficient to only allow priority to be set up front and treat it as immutable instead?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
```diff
@@ -0,0 +1,964 @@
...
+        Press((ButtonControl)keyboard.xKey, queueEventOnly: true);
+        InputSystem.Update();
+        Assert.That(performedCount, Is.EqualTo(1));
+
+        Release((ButtonControl)keyboard.xKey);
+        InputSystem.Update();
+    }
+
+    [Test]
+    [Category("Actions Priority")]
```
ekcoh, line 537: Could we add a test description to this test explaining what kind of requirement it verifies? I am not understanding it by looking at the name and the code

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
```diff
@@ -0,0 +1,964 @@
...
+        Assert.That(PackedPriorityForMonitor(state, keyboard.leftShiftKey), Is.EqualTo(3));
+
+        shiftB.Priority = 7;
+
+        Assert.That(PackedPriorityForMonitor(state, keyboard.bKey), Is.EqualTo(7));
+        Assert.That(PackedPriorityForMonitor(state, keyboard.leftShiftKey), Is.EqualTo(7));
+    }
+
+    [Test]
+    [Category("Actions Priority")]
```
ekcoh, line 580: What is this really testing? Could maybe be clearer if index wasn't using `var`?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
```diff
@@ -0,0 +1,964 @@
...
+                return;
+            }
+        }
+
+        Assert.Fail("Expected two control slots bound to the same physical control.");
+    }
+
+    [Test]
+    [Category("Actions Priority")]
+    public unsafe void Actions_Complexity_ControlGrouping_WritesPerControlSlotComplexity_NotActionPriority()
```
ekcoh, line 799: These seems to be new tests covering internals - that is probably great to get in but they are quite complex and do not map to any clear requirements so maybe add a test description for tests like these to make them possible to understand?

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+| Setting | Behavior |
+| ------- | -------- |
+| [Complexity-Based Shortcut Resolution](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysConsumeInput) | Orders overlapping bindings by composite **complexity** (binding-chain depth). This is the default develop behavior when action priority is off. |
+| [Action Priority Shortcut Resolution](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysUseActionPriority) | Orders overlapping bindings by each action's [`InputAction.Priority`](xref:UnityEngine.InputSystem.InputAction.Priority). When enabled, it **takes precedence** over complexity-based resolution even if both options are on. |
+
+Each action has a [`Priority`](xref:UnityEngine.InputSystem.InputAction.Priority) property (range **0**–**65535**, clamped when set). The value applies to all bindings on that action. Serialized priority is always stored on the asset; at runtime it is used only when **Action Priority Shortcut Resolution** is enabled. In that case the **Priority** field is also shown in the [Input Actions Editor](xref:input-system-configuring-input).
```
ekcoh: Might be good to point out the direction of the priority system, e.g. is priority 1 higher or lower than priority 2? That is not obvious.

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+When action priority resolution is active:
+
+- Higher priority actions are notified before lower-priority actions on the same control.
+- When an action reaches the **Performed** phase, priority **0** does **not** mark the input event as handled, so lower-priority actions in the same overlap group can still respond on that event.
```
ekcoh: My changes in https://github.com/Unity-Technologies/InputSystem/pull/2412 which landed while this has been in development should allow this even when marked handled. Did I miss something in my PR or is something changing on this PR related to this?

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+- Any priority **greater than zero** can mark the event handled and suppress strictly lower-priority actions in the same group for that event.
```
ekcoh, line 168: I get a bit nervous about this statement, are you relying on the handled flag to handle priority? What happens then if you have two interactions/bindings like these:

Shift+B - Prio 1
Shift+C - Prio 2

OR 

Shift+B - Prio 2
Shift+C - Prio 1

And the received event sequence is:

First keyboard event: Press shift
Second keyboard event: Both B and C gets pressed in the same atomic event?

ekcoh (commented):
(Intermediate review comment submission) - Only through half of the PR so far so will provide more feedback once through it.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionStateMonitorIndex.cs
```diff
@@ -0,0 +1,45 @@
+namespace UnityEngine.InputSystem
+{
+    /// <summary>
+    /// Bit-packed id registered with <see cref="InputManager"/> for <see cref="InputActionState"/> state change monitors.
+    /// Layout must match <c>TriggerState.kMaxNum*</c> limits on <see cref="InputActionState"/>.
+    /// </summary>
+    internal readonly struct InputActionStateMonitorIndex
+    {
+        readonly long m_Packed;
```
ekcoh, line 14: This seems to be a bit-packed field so I would recommend at least adding a one liner here with what it contains?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionStateMonitorIndex.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public InputActionStateMonitorIndex(long packed)
```
ekcoh: Why is this constructor only taking a pre-packed field? Wouldn't it make more sense to delegate the packing into this constructor to hide that complexity from the caller?

ekcoh: I see now that there is a `Create` function for that, but does that mean this constructor should be private instead?

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
```diff
@@ -1027,6 +1087,74 @@ private void EnableControls(InputAction action)
             }
         }
 
+        internal void OnActionPriorityChanged(InputAction action)
+        {
+            if (!InputSystem.settings.IsShortcutResolutionUsingActionPriority)
+                return;
+
+            Debug.Assert(action != null, "Action must not be null");
```
ekcoh: These Debug.Assert statements could be moved into the if statement below to avoid double checking in debug builds

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManagerStateMonitors.cs
```diff
@@ -5,126 +5,115 @@
 
 namespace UnityEngine.InputSystem
 {
-    internal partial class InputManager
+    /// <summary>
```
ekcoh, line 8: It seems this class have been refactored while making changes which made it very difficult to review the logical flow - so just want to flag I might have overlooked things in this class.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManagerStateMonitors.cs
```diff
@@ -401,75 +273,190 @@ internal unsafe void FireStateChangeNotifications(int deviceIndex, double intern
...
+
+                needToUpdateOrderingOfMonitors = true;
+            }
+
+            public void Remove(IInputStateChangeMonitor monitor, long monitorIndex, bool deferRemoval)
+            {
+                if (listeners == null)
+                    return;
+
+                for (var i = 0; i < signalled.length; ++i)
```
ekcoh, line 377: Is formatting off here or is this an unstopped for loop with a scoped if "child"? Would recommend using scoped and review formatting here

ekcoh (approved):
I found it very difficult to review this PR due to scattered changes but see no reason to block it - I would however recommend extensive testing of both the legacy and new mode for shortcuts to verify that it doesn't introduce regressions. Left some comments, suggestions and questions.

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
Darren-Kelly-Unity: Removed!

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
Darren-Kelly-Unity: There were multiple bugs in this area where things didn't work as expected that I fixed.

That's why I thought the test made sense.

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
Darren-Kelly-Unity: This was also in response to a bug found, I renamed & added a comment to help with understanding.

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
Darren-Kelly-Unity: Added!

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
Darren-Kelly-Unity: This was added with the help of AI to make sure things were packed correctly as they were not always. Also found by AI.

Changed the index to not use var also.

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
Darren-Kelly-Unity: Yeah I added quite a few tests for the subtle differences that show up between complexity and priority.

They are quite complex in the end / adding this backwards compatibility caused quite a lot of issues which then had to be tested.

I am still not sure it was worth adding all the complexity to our codebase but we had to do it so that's where a lot of these tests come from.

Added comments to quite a few more tests now to explain.

Packages/com.unity.inputsystem/Documentation~/Actions.md
Darren-Kelly-Unity: Done!

Packages/com.unity.inputsystem/Documentation~/Actions.md
Darren-Kelly-Unity: Tests should already be covering these cases!

Some info with the help of AI also:

Priority suppression in FireStateChangeNotifications (via signals.ClearBit) only clears monitors within the same overlap group. Groups are assigned per physical control — Shift is in one group, B is in another, C in another. So when B and C arrive in the same atomic event, they're in separate groups and neither suppresses the other. Both Shift+B and Shift+C fire regardless of priority ordering.

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionStateMonitorIndex.cs
Darren-Kelly-Unity: Made these changes! I thought it was ok due to the class being internal.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManagerStateMonitors.cs
Darren-Kelly-Unity: This is a remnant from separating classes / refactoring and copy pasting from the old code I believe.

But you are right, added braces!

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
ekcoh: OK, not sure I understand why the 8-byte boundary should be interesting due to that but maybe there were some 8-bit narrowing under the hood before driving the need for these tests?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
ekcoh: @Darren-Kelly-Unity Noticed you marked this resolved without a comment. If you think I am barking up the wrong tree - that is fine but I am curious why we have to support this?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
ekcoh: To me this looks like it should be debug asserted internally with the packing logic in that case?

Assets/Tests/InputSystem/CoreTests_ActionsPriority.cs
ekcoh: OK, thanks. I know there was a push to preserve the backwards compatibility path here - it still looks non-trivial to remove later though since its not contained or under a switch - I would minimum recommend filing a ticket containing exactly what needs to be done when that happens while this is fresh - if not the whole "exploration cost" of this PR needs to be taken again when its time to remove the complexity path?

Packages/com.unity.inputsystem/InputSystem/Actions/InputActionStateMonitorIndex.cs
ekcoh: Of course its OK but if the intent is not to be able to pass arbitrary longs IMO it is better to prevent misuse (even if internal)

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
 You can't change certain aspects of the configuration, such as action bindings, while an action is enabled. To stop actions or action maps from responding to input, call  [`Disable`](xref:UnityEngine.InputSystem.InputAction.Disable).
 
 While enabled, an action actively monitors the [controls](xref:input-system-controls) it's bound to. If a bound control changes state, the action processes the change. If the control's change represents an [interaction](xref:input-system-interactions) change, the action creates a response. All of this happens during the Input System update logic. Depending on the [update mode](xref:input-system-settings#update-mode) selected in the input settings, this happens once every frame, once every fixed update, or manually if updates are set to manual.
+
+## Overlapping bindings and action priority
+
+When several enabled actions share the same physical control (for example a plain **B** key action and a **Shift+B** composite), the Input System can resolve which action should respond first. This is controlled in **Project Settings** > **Input System Package** under [Improved Shortcut Support](xref:input-system-settings#improved-shortcut-support).
```
suearkinunity:
```suggestion
When several enabled actions share the same physical control (for example, a plain **B** key action and a **Shift**+**B** composite), the Input System can resolve which action should respond first. 

You can control this behavior in **Project Settings** > **Input System Package** > [**Improved Shortcut Support**](xref:input-system-settings#improved-shortcut-support).
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+| Setting | Behavior |
+| ------- | -------- |
+| [Complexity-Based Shortcut Resolution](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysConsumeInput) | Orders overlapping bindings by composite **complexity** (binding-chain depth). This is the default develop behavior when action priority is off. |
```
suearkinunity:
```suggestion
| [Complexity-Based Shortcut Resolution](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysConsumeInput) | Orders overlapping bindings by composite complexity (that is, the binding-chain depth). This is the default develop behavior when the Action Priority option is disabled. |
```

suearkinunity: What does "develop behavior" mean?

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+| [Action Priority Shortcut Resolution](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysUseActionPriority) | Orders overlapping bindings by each action's [`InputAction.Priority`](xref:UnityEngine.InputSystem.InputAction.Priority). When enabled, it **takes precedence** over complexity-based resolution even if both options are on. |
```
suearkinunity:
```suggestion
| [Action Priority Shortcut Resolution](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysUseActionPriority) | Orders overlapping bindings by each action's [`InputAction.Priority`](xref:UnityEngine.InputSystem.InputAction.Priority). When enabled, this takes precedence over complexity-based resolution, even if both options are enabled. |
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+Each action has a [`Priority`](xref:UnityEngine.InputSystem.InputAction.Priority) property (range **0**–**65535**, clamped when set; higher value = higher priority, notified first). The value applies to all bindings on that action. Serialized priority is always stored on the asset; at runtime it is used only when **Action Priority Shortcut Resolution** is enabled. In that case the **Priority** field is also shown in the [Input Actions Editor](xref:input-system-configuring-input).
```
suearkinunity:
```suggestion
Each action has a [`Priority`](xref:UnityEngine.InputSystem.InputAction.Priority) property. The range is from 0 to 65535, and is clamped when set. A higher value means a higher priority, notified first. 

The `Priority` value applies to all bindings on that action. Serialized priority is always stored on the asset; at runtime it's used only when **Action Priority Shortcut Resolution** is enabled. In that case, the **Priority** field is also shown in the [Input Actions Editor](xref:input-system-configuring-input).
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+When action priority resolution is active:
+
+- Higher priority actions are notified before lower-priority actions on the same control.
+- When an action reaches the **Performed** phase, priority **0** does **not** mark the input event as handled, so lower-priority actions in the same overlap group can still respond on that event.
```
suearkinunity:
```suggestion
- When an action reaches the Performed phase, priority `0 doesn't mark the input event as handled, so lower-priority actions in the same overlap group can still respond on that event.
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+- Any priority **greater than zero** can mark the event handled and suppress strictly lower-priority actions in the same group for that event.
```
suearkinunity, line 168:
```suggestion
- Any priority greater than zero can mark the event as handled and suppress strictly lower-priority actions in the same group for that event.
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+- Actions with the **same** priority are not suppressed relative to each other; both can perform in the same update if their bindings fire.
```
suearkinunity, line 169:
```suggestion
- Actions with the same priority are not suppressed relative to each other; both can perform in the same update if their bindings fire.
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+Set priority in code:
```
suearkinunity, line 171:
```suggestion
To set priority in code, use `.Priority`:
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+```CSharp
+fireAction.Priority = 10;
+reloadAction.Priority = 5;
+```
+
+Or edit the **Priority** field on an action in the Input Actions Editor when **Action Priority Shortcut Resolution** is enabled.
```
suearkinunity:
```suggestion
You can also edit the **Priority** field on an action in the Input Actions Editor when **Action Priority Shortcut Resolution** is enabled.
```

Packages/com.unity.inputsystem/Documentation~/Actions.md
```diff
@@ -152,3 +152,32 @@ When you enable an action, the Input System resolves its bindings, unless it has
...
+
+For composite shortcuts and complexity ordering, see [Multiple input sequences (such as keyboard shortcuts)](xref:input-system-action-bindings#multiple-input-sequences-such-as-keyboard-shortcuts).
```
suearkinunity:
```suggestion
For information about composite shortcuts and complexity ordering, refer to [Multiple input sequences (such as keyboard shortcuts)](xref:input-system-action-bindings#multiple-input-sequences-such-as-keyboard-shortcuts).
```

Packages/com.unity.inputsystem/Documentation~/Settings.md
```diff
@@ -120,6 +120,19 @@ To force the Editor to add all locally available Devices, even if they're not in
 > [!NOTE]
 > This setting is stored as a user setting, not a project setting. This means other users who open the project in their own Editor do not share the setting.
 
+## Improved Shortcut Support
+
+Under __Improved Shortcut Support__, two independent options control how the Input System resolves overlapping bindings on the same control (for example a plain __B__ key action versus __Shift+B__):
```
suearkinunity:
```suggestion
Under **Improved Shortcut Support**, two independent options control how the Input System resolves overlapping bindings on the same control (for example, a plain **B** key action as opposed to **Shift**+**B**):
```

Packages/com.unity.inputsystem/Documentation~/Settings.md
```diff
@@ -120,6 +120,19 @@ To force the Editor to add all locally available Devices, even if they're not in
...
+
+| UI / API | Description |
```
suearkinunity, line 127: We don't need to repeat information. I would suggest condensing it all to one page, and linking between them.

Packages/com.unity.inputsystem/Documentation~/ActionBindings.md
```diff
@@ -892,6 +892,9 @@ By using the [Pass-Through](xref:input-system-responding#pass-through) action ty
 > [!NOTE]
 > The mechanism described here only applies to actions that are part of the same [`InputActionMap`](xref:UnityEngine.InputSystem.InputActionMap) or [`InputActionAsset`](xref:UnityEngine.InputSystem.InputActionAsset).
 
+> [!NOTE]
+> In **Project Settings** > **Input System Package**, enable __Complexity-Based Shortcut Resolution__ ([`InputSettings.shortcutKeysConsumeInput`](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysConsumeInput)) and leave __Action Priority Shortcut Resolution__ ([`InputSettings.shortcutKeysUseActionPriority`](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysUseActionPriority)) disabled to use automatic composite complexity as described below. If __Action Priority Shortcut Resolution__ is enabled, overlaps are resolved using each action's [`InputAction.Priority`](xref:UnityEngine.InputSystem.InputAction.Priority) instead. See [Input settings](xref:input-system-settings#improved-shortcut-support) for details.
```
suearkinunity:
```suggestion
> To use automatic composite complexity as described below, in **Project Settings** > **Input System Package**, enable __Complexity-Based Shortcut Resolution__ ([`InputSettings.shortcutKeysConsumeInput`](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysConsumeInput)) and leave __Action Priority Shortcut Resolution__ ([`InputSettings.shortcutKeysUseActionPriority`](xref:UnityEngine.InputSystem.InputSettings.shortcutKeysUseActionPriority)) disabled. If __Action Priority Shortcut Resolution__ is enabled, overlaps are resolved using each action's [`InputAction.Priority`](xref:UnityEngine.InputSystem.InputAction.Priority) instead. Refer to [Input settings](xref:input-system-settings#improved-shortcut-support) for details.
```

suearkinunity (approved):
There's a lot of repetition between Actions.md and Settings.md; please condense the information to one page and link to it from the other.

Assets/Tests/InputSystem/CoreTests_Actions.cs
Darren-Kelly-Unity: Yes there is!

Packages/com.unity.inputsystem/Documentation~/Actions.md
Darren-Kelly-Unity: Discussed and resolved!

Packages/com.unity.inputsystem/Documentation~/Settings.md
Darren-Kelly-Unity: Makes sense! I just have it in one location now and link to it from the other.

Pauliusd01 (approved):
Rechecked the tests from: https://github.com/Unity-Technologies/InputSystem/pull/2402#issuecomment-4561206339 did not see any new issues

PR #2405, opened 2026-04-07
CHANGE: Rework Core Platforms Menu scene used for internal testing

Pauliusd01:
### Description

Reworks the menu scene visuals, adds a search bar, adds the feature to auto fill the list with relevant scenes and improves gamepad controls. (You need to hold Start to go back to the main menu using the gamepad)

image

### Testing status & QA

Run the menu scene and try loading the samples/qa scenes in playmode and build

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers

N/A

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Pauliusd01:
@u-pr can you check that all issues were resolved? Note that some had to be skipped such as:
```
Intrusive scene suspension -- Intentional design choice. Blocking raycasts alone doesn't prevent scenes from processing their own InputActions, so disabling root objects is necessary for the overlay to work reliably.

Editor namespace in non-Editor folder -- Invalid concern. Assets/Tools/ has an asmdef with "includePlatforms": ["Editor"], so it's already Editor-only.
```

ekcoh:
Think I found an issue:

Run the sample
1. QA Tools > Open Core Scene Menu.
2. Press "Play" in editor.

Clicking the top-left sample takes me to that sample scene but "Menu" button in top-left corner doesn't take me back. It looks like some dialog is shown but "behind the current UI" where the dialog and the sample UI is fighting so I cannot click the dialog.

Or is the design intended to be so that you need to hold mouse button down after pressing Menu, then move to dialog with button down and then click your choice? In that case it works as expected.

Assets/QA/Tests/Core Platform Menu/ReturnToMenuOverlay.cs
```diff
@@ -0,0 +1,410 @@
...
+            EventSystem.current.SetSelectedGameObject(target);
+    }
+
+    // ── Scene suspend / resume ──────────────────────────────────
+
+    void SuspendActiveScene()
+    {
+        m_SuspendedRoots.Clear();
+        var scene = SceneManager.GetActiveScene();
+        foreach (var root in scene.GetRootGameObjects())
```
ekcoh, line 202: Might want to skip disabling the camera or add another. camera while active scene is disabled to avoid the "No active camera" overlay clashing with the dialog.

ekcoh (approved):
Nice, added a suggestion since I didn't like the clashing overlays due to no camera being active. Apart from that, looks much better and setting things up programatically had a great impact on reducing size (rows removed is great).

Added a note on unexpected behaviour (having to hold mouse button while navigating overlay instead of toggle)

Assets/Tools/AddScenesToBuild.cs
```diff
@@ -1,104 +1,102 @@
...
-        for (int i = 0; i < sceneCount; i++)
-        {
-            string scenePath = EditorPrefs.GetString($"BuildSettingsScenePath_{i}", "");
-            bool sceneEnabled = EditorPrefs.GetBool($"BuildSettingsSceneEnabled_{i}", false);
-
-            buildScenes[i] = new EditorBuildSettingsScene(scenePath, sceneEnabled);
-        }
-        EditorBuildSettings.scenes = buildScenes;
+        for (int i = 0; i < kExcludedSegments.Length; i++)
+            if (path.IndexOf(kExcludedSegments[i], StringComparison.OrdinalIgnoreCase) >= 0)
```
MorganHoarau: nit. if `kExcludedSegments[i]` is empty, `path.IndexOf` returns 0 which would return true for all path. (ref: https://learn.microsoft.com/en-us/dotnet/api/system.string.indexof?view=net-10.0#system-string-indexof(system-string-system-stringcomparison))
This should never happens given the current excluded segments, but the code could be more correct.


```suggestion
            if (path.Contains(kExcludedSegments[i], StringComparison.OrdinalIgnoreCase) >= 0)
```

MorganHoarau (approved):
Minor change requested.

Not a fan of building entire uGUI from script, but good enough for internal testing.

Looks great, but haven't tested locally.

PR #2407, opened 2026-04-14
FIX: ISXB-1319 Input recorder game view focus

MorganHoarau:
### Description

PR addresses https://jira.unity3d.com/browse/ISXB-1319
* The first issue is fixed with a deemed acceptable trade-off.
* The second issue is documented as known limitation in Events.md

Tech Roundtable Slide-deck: https://docs.google.com/presentation/d/18zL10TRcnVfBiSSBe_YT5EYVkJLyt7gFvD3s7EL0JKY/edit?usp=drive_link

Note: _the fix adds a tight coupling between `InputManager` and `InputEventTrace.ReplayController`._

#### Issue 1: replay requires Game View focus

Root cause: three independent focus-gating layers that block replayed events when Game View is unfocused (event deferral, device-disabled discard, UI module skip). 
The fix implemented uses an InputManager-level `isReplayActive` flag that bypasses all three (**Editor-only**).

<details>

<summary>Issues 1 Details</summary>
  
When Game View loses focus -> `applicationHasFocus=false` -> `gameHasFocus=false`. Replay events are blocked at three independent layers:

```mermaid
sequenceDiagram
    participant IM as InputManager
    participant RC as ReplayController
    participant Dev as Mouse/Keyboard
    participant UIIM as InputSystemUIInputModule
    participant ES as EventSystem

    Note over IM,ES: Game View loses focus
    IM->>IM: gameHasFocus = false
    IM->>Dev: DisableDevice(DisabledWhileInBackground)
    ES->>ES: OnApplicationFocus(false) → m_HasFocus = false

    Note over IM,ES: Replay starts
    RC->>IM: QueueEvent(mouseStateEvent)
    IM->>IM: ❌ A — event deferred (ShouldDefer)
    IM->>IM: ❌ B — event discarded (device disabled)
    ES->>UIIM: Process()
    UIIM->>UIIM: ❌ C — !isFocused → skip ProcessPointer/ProcessNavigation
```

</details>

**Known trade-off**: The bypass doesn't distinguish replayed events from real hardware input, so real keyboard/mouse input also reaches the game during replay even with Game View unfocused. This is acceptable for typical usage (replay in isolation in Editor).

#### Issue 2: TextField text not replayed

**Architectural limitation**. UIToolkit TextField receives text through the native IMGUI event queue (`Event.PopEvent`), which is a separate pipeline from InputSystem. Replayed TextEvents fire correctly through `Keyboard.OnTextInput` but have zero subscribers. The IMGUI queue is only populated by native OS input. Keyboard state (key presses) does replay correctly; only character delivery to TextFields is broken. Fixing this would require engine-side changes (InputForUI / IMGUI bridge), outside scope of this ticket.

This is a consistent pattern across Unity's UI framework: IMGUI, uGUI, and UIToolkit all use `Event.PopEvent()` for text  processing.

JIRA user story: https://jira.unity3d.com/browse/ISX-2562

### Testing status & QA

* Tested on 6.0 using InputRecorderSample.unity scene in WinEditor.
* Need to test on 6.5 without the changes to see if latest changes from Vera affect the behavior.

https://github.com/user-attachments/assets/f4263385-7854-4943-9299-374754aabdb2

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Comments to reviewers

* Paulius: Could you please evaluate the known trade-off for Issue 1?
   * **To Test**: Record a long enough input sequence, then during playback, type on the keyboard or move the mouse while the Game View is unfocused and observe that the real inputs reaches the game.
   * **Context**: The fix works by bypassing the focus gate for the duration of replay — a side effect is that real hardware input (keyboard, mouse) can also pass through during that window, even though the Game View is not focused. This is the trade-off. Rita would like your assessment of how impactful this is in practice, and whether it is acceptable for shipping Issue 1's fix as-is.

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/Documentation~/Events.md
+++ b/Packages/com.unity.inputsystem/Documentation~/Events.md
@@ -224,7 +224,7 @@ trace.Dispose();
 Dispose event traces after use, so that they do not leak memory on the unmanaged (C++) memory heap.
 
 > [!NOTE]
-> **Keyboard text input is not replayed to UI text fields.** Keyboard state (key presses) is captured and replayed correctly and remains accessible via `Keyboard.current`. However there is a known limitation with character delivery to UI Framework components (uGUI `InputField` or UIToolkit `TextField`). These components receive text through a separate native pipeline that is not fed by event replay. As a result, text typed into UI text fields during recording will not appear during playback.
+> **Keyboard text input is not replayed to UI text fields.** Keyboard state (key presses) is captured and replayed correctly and remains accessible via `Keyboard.current`. However, there is a known limitation with character delivery to UI Framework components (uGUI `InputField` or UI Toolkit `TextField`). These components receive text through a separate native pipeline that is not fed by event replay. As a result, text typed into UI text fields during recording will not appear during playback.
 
 You can also write event traces out to files/streams, load them back in, and replay recorded streams.
 
--- a/Packages/com.unity.inputsystem/Documentation~/Events.md
+++ b/Packages/com.unity.inputsystem/Documentation~/Events.md
@@ -223,6 +223,9 @@ trace.Dispose();
 
 Dispose event traces after use, so that they do not leak memory on the unmanaged (C++) memory heap.
 
+> [!NOTE]
+> **Keyboard text input is not replayed to UI text fields.** Keyboard state (key presses) is captured and replayed correctly and remains accessible via `Keyboard.current`. However there is a known limitation with character delivery to UI Framework components (uGUI `InputField` or UIToolkit `TextField`). These components receive text through a separate native pipeline that is not fed by event replay. As a result, text typed into UI text fields during recording will not appear during playback.
+
 You can also write event traces out to files/streams, load them back in, and replay recorded streams.
 
 ```CSharp
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
@@ -1106,6 +1106,11 @@ namespace UnityEngine.InputSystem.LowLevel
             // editor updates where they reach the editor UI instead of the game.
             private void BeginReplayBypass()
             {
+                if (m_ClearReplayBypassCallback != null)
+                {
+                    InputSystem.onAfterUpdate -= m_ClearReplayBypassCallback;
+                }
+
                 if (!m_ReplayBypassActive)
                 {
                     m_ReplayBypassActive = true;
@@ -1121,8 +1126,12 @@ namespace UnityEngine.InputSystem.LowLevel
                 if (!m_ReplayBypassActive)
                     return;
 
-                if (m_ClearReplayBypassCallback == null)
-                    m_ClearReplayBypassCallback = EndReplayBypass;
+                if (m_ClearReplayBypassCallback != null)
+                {
+                    return;
+                }
+
+                m_ClearReplayBypassCallback = EndReplayBypass;
                 InputSystem.onAfterUpdate += m_ClearReplayBypassCallback;
             }
 
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
@@ -1148,6 +1148,7 @@ namespace UnityEngine.InputSystem.LowLevel
                     --InputSystem.s_Manager.m_ActiveReplayCount;
                 }
             }
+
 #endif
             /// <summary>
             /// Replay events recorded from <paramref name="recordedDevice"/> on device <paramref name="playbackDevice"/>.
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
@@ -1135,6 +1135,7 @@ namespace UnityEngine.InputSystem.LowLevel
                 InputSystem.onAfterUpdate -= DeferredStopEditorEventPassthrough;
                 StopEditorEventPassthrough();
             }
+
 #endif
             /// <summary>
             /// Replay events recorded from <paramref name="recordedDevice"/> on device <paramref name="playbackDevice"/>.
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
@@ -1072,8 +1072,10 @@ namespace UnityEngine.InputSystem.LowLevel
             private double m_StartTimeAsPerRuntime;
             private int m_AllEventsByTimeIndex = 0;
             private List<InputEventPtr> m_AllEventsByTime;
+#if UNITY_EDITOR
             private bool m_ReplayBypassActive;
             private Action m_ClearReplayBypassCallback;
+#endif
 
             internal ReplayController(InputEventTrace trace)
             {
@@ -1090,13 +1092,15 @@ namespace UnityEngine.InputSystem.LowLevel
             {
                 InputSystem.onBeforeUpdate -= OnBeginFrame;
                 finished = true;
+#if UNITY_EDITOR
                 EndReplayBypass();
-
+#endif
                 foreach (var device in m_CreatedDevices)
                     InputSystem.RemoveDevice(device);
                 m_CreatedDevices = default;
             }
 
+#if UNITY_EDITOR
             // Signals InputManager to treat events as if game view has focus, bypassing
             // editor focus routing that would otherwise defer pointer/keyboard events to
             // editor updates where they reach the editor UI instead of the game.
@@ -1135,7 +1139,7 @@ namespace UnityEngine.InputSystem.LowLevel
                     --InputSystem.s_Manager.m_ActiveReplayCount;
                 }
             }
-
+#endif
             /// <summary>
             /// Replay events recorded from <paramref name="recordedDevice"/> on device <paramref name="playbackDevice"/>.
             /// </summary>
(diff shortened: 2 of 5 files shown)


Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
```diff
@@ -1088,12 +1090,52 @@ public void Dispose()
...
+            // events (via onAfterUpdate). This ensures events already queued in the native buffer
+            // are still processed with the bypass active before it is removed.
+            private void ScheduleEndReplayBypass()
+            {
+                if (!m_ReplayBypassActive)
+                    return;
+
+                if (m_ClearReplayBypassCallback == null)
+                    m_ClearReplayBypassCallback = EndReplayBypass;
+                InputSystem.onAfterUpdate += m_ClearReplayBypassCallback;
```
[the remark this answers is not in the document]
MorganHoarau: no reason for it to happen but agree that it would raise the quality of the code so applied the feedback.

Packages/com.unity.inputsystem/Documentation~/Events.md
```diff
@@ -223,6 +223,9 @@ trace.Dispose();
 
 Dispose event traces after use, so that they do not leak memory on the unmanaged (C++) memory heap.
 
+> [!NOTE]
+> **Keyboard text input is not replayed to UI text fields.** Keyboard state (key presses) is captured and replayed correctly and remains accessible via `Keyboard.current`. However there is a known limitation with character delivery to UI Framework components (uGUI `InputField` or UIToolkit `TextField`). These components receive text through a separate native pipeline that is not fed by event replay. As a result, text typed into UI text fields during recording will not appear during playback.
```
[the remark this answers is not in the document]
MorganHoarau:
```suggestion
> **Keyboard text input is not replayed to UI text fields.** Keyboard state (key presses) is captured and replayed correctly and remains accessible via `Keyboard.current`. However, there is a known limitation with character delivery to UI Framework components (uGUI `InputField` or UI Toolkit `TextField`). These components receive text through a separate native pipeline that is not fed by event replay. As a result, text typed into UI text fields during recording will not appear during playback.
```

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
```diff
@@ -3371,15 +3381,18 @@ private unsafe void ProcessEventBuffer(InputUpdateType updateType, ref InputEven
 
                     // If device is disabled, we let the event through only in certain cases.
                     // Removal and configuration change events should always be processed.
+                    // During replay, allow events through for devices disabled due to background
+                    // focus loss — the replay intentionally re-injects events for those devices.
                     if (device != null && !device.enabled &&
+#if UNITY_EDITOR
+                        !isReplayActive &&
```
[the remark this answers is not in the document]
MorganHoarau: Known trade-off.

MorganHoarau:
waiting for CI to be green before adding final reviewers

ekcoh:
Thanks for a very good and descriptive PR description. For issue #2 I would recommend filing a backlog ticket referencing the findings from this investigation and possibly this PR as well so this can be looked into in the future.

Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
```diff
@@ -1249,6 +1306,9 @@ public ReplayController Rewind()
             public ReplayController PlayAllFramesOneByOne()
             {
                 finished = false;
+#if UNITY_EDITOR
```
ekcoh, line 1295: Generally, when this kind of behaviour is needed (with these editor checks all over the place), I wonder if we shouldn't add a proper optional callback for these replay events, e.g. `ReplayBegun`, there already seems to be an `m_OnFinished` that could be used to handle the call to editor code `ScheduleEndReplayBypass()`?

Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
```diff
@@ -1381,6 +1447,11 @@ private void Finished()
             {
                 finished = true;
                 InputSystem.onBeforeUpdate -= OnBeginFrame;
+#if UNITY_EDITOR
+                // Schedule bypass removal for after the next OnUpdate, so any events already
+                // queued into the native buffer this frame are still processed with the bypass active.
+                ScheduleEndReplayBypass();
```
ekcoh: Couldn't editor code hook to m_OnFinished in this case handle it that way?

Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
ekcoh: e.g. this could be e.g. (not sure about naming - conceptually) `m_OnReplayStart?();`

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
ekcoh: Minor, but I don't specifically see why this is "tightly replay", it looks like an internal feature "ignoreGameViewFocus". That might age/scale better and better indicates what this is all about. It's a specific mode where game-view focus is ignored in event filtering, or what do you think?

ekcoh (approved):
Generally very nice and clean PR with a great description. I have no reason to not approve but I would urge considering the questions and design aspects raised in comments since I believe this can be further improved by handling the ignoreFocusFiltering as a clear internal feature (that might become relevant for more things) and consider if editor/source splitting of the replay logic is potentially possible to make cleaner by adding (or relying on existing) events tied to the replay mechanism to avoid as much `#if UNITY_EDITOR` checks. The only thing I am missing here is a test case since this adds a "new internal feature" which doesn't seem to be covered for regressions which indicates it needs to be rested manually which is costly in the long run.

Darren-Kelly-Unity (approved):

Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
MorganHoarau: I took a bit of time to address this because when taking API surface and future editor migration, there are only least-worse solutions:

| Option | Implementation | Cons |
| --- | --- | --- |
| Partial methods |  3 partial void stubs, editor partial implements calling Start/Stop | hard to migrate to editor assembly |
| Direct calls | `InputManager.StartEditorEventPassthrough()` behind thin #if UNITY_EDITOR | need thin editor code in `InputEventTrace` to call it |
| public Callbacks | Caller (InputRecorder) chains `.OnReplayStart(Start).OnFinished(Stop)` | new API + caller must know/remember to call `InputManager` start/stop event passthrough |
 
1. A new `m_OnReplayStart` action would mean a new public API for it (if we want to let the caller handle the call to `InputManager` ref count - `InputRecorder` is a sample, not part of the input system assembly). The current API for Finished is replace-style (not add) and unless we change the contract, it means we would expect the call of the `ReplayController` to increment the `InputManager` count ref.
2. This passthrough functionality is internal requirement for all `InputEventTrace.ReplayController`, not just `InputRecorder`. It makes more sense to acquire/release the passthrough count privately. 

From a migration perspective, direct call to InputManager wrapped in `#if UNITY_EDITOR` is also the easiest to detect and migrate later on.

So I've trimmed down the code in InputEventTrace and migrated the ref-counting and null-safety in `InputManager.StopEditorEventPassthrough`.

Do you agree with the approach?

Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
MorganHoarau: The current implementation of `m_OnFinished` doesn't allow that. A public API is exposed and let user set/replace (not add) the `m_OnFinished` action and I would be worried to break the existing behaviour.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
MorganHoarau: Renamed `isEditorEventPassthroughActive `

Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
ekcoh: OK, thanks for explaining

ekcoh:
Failed code analyser is CI instability? Looked like it.

Packages/com.unity.inputsystem/InputSystem/Runtime/Events/InputEventTrace.cs
ekcoh: Agree, and this is all internal it seems so we can always revisit if needed later

ekcoh: Thanks for investigating

ekcoh (approved):
Reviewed up-to-and-including https://github.com/Unity-Technologies/InputSystem/pull/2407/commits/6d1bd88a92137451b3bd2553266f171b372e7bbf. IMO lets land this fix.

ekcoh:
@MorganHoarau noticed @Pauliusd01 have not checked this PR yet so would advise adding him for QA check.

MorganHoarau:
Code analyzer is failing because 2022.3 CI doesn't work anymore. Are we still supporting 2022.3?

Pauliusd01 (approved):

MorganHoarau:
I don't know how to properly refresh the code analyzer, but I ran it twice and it passed: https://unity-ci.cds.internal.unity3d.com/job/67910814

Merging now.

PR #2411, opened 2026-04-29
FIX: uum 138143 button press points for stickcontrol and vector2 add functionality

Darren-Kelly-Unity:
## Purpose of this PR

Purpose of this PR
This change fixes inconsistent press detection between action-level helpers (InputAction.IsPressed, WasPressedThisFrame, WasReleasedThisFrame) and bindings that use a Press interaction with its own pressPoint. Previously, those APIs could use a ButtonControl’s pressPoint while the interaction used a different threshold, so “pressed” could disagree with the interaction’s press/release behavior.

It also aligns Vector2 / stick bindings with the intended model: action-level press polling uses InputSettings.defaultButtonPressPoint unless a Press interaction sets an explicit threshold (no separate per-vector pressPoint on the composite control for this path).

Tracking: [UUM-138143](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-138143) (adjust link if your team uses Jira instead of Issue Tracker).

Docs touched: RespondingToActions.md, Interactions.md; implementation in InputAction, InputActionState, InputControlExtensions, ButtonControl, PressInteraction, StickControl (XML remarks), plus focused tests in ActuationPressPointTests and updates in CoreTests_Actions / CoreTests_Controls.

## Release Notes

Release Notes
Fixed: InputAction.IsPressed, WasPressedThisFrame, and WasReleasedThisFrame now respect a binding’s Press interaction pressPoint when set, so they stay consistent with the interaction’s press and release behavior instead of always following the bound ButtonControl’s pressPoint when both are set. [UUM-138143](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-138143)
Changed: For bindings to Vector2 / stick controls, action-level press APIs no longer use a per-control pressPoint on the vector; use a Press interaction for a custom threshold or rely on defaultButtonPressPoint. [UUM-138143](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-138143)

## Functional Testing status

- Automated: new **`ActuationPressPointTests`** and updates in **`CoreTests_Actions`** / **`CoreTests_Controls`** cover press-point actuation paths.
- Manual: repro project **IN-136917** / scene **Input Bug** from the ticket; verify console no longer shows **`IsPressed`** before processed magnitude reaches the configured press threshold for vector / stick cases.

## Performance Testing Status

I have added performance testing for the hot path methods that Hakan flagged and I don't see any major issues when comparing. There is a slight increase in cost of performance of about 0.1ms for 1000 iterations on the GetActuationPressThreshold() method that was asked to be checked.

No new per-frame allocations identified in scope.

## Overall Product Risks

Technical Risk: 2 — Touches core action state, actuation thresholds, and several control paths; behavior change for anyone relying on the old mismatch or on vector-level pressPoint for these APIs.

Halo Effect: 2 — Affects any project using Press + custom pressPoint with IsPressed / frame-edge helpers, and Vector2/stick bindings that assumed the old threshold source.

## Documentation Impact

**Changes analyzed:** `origin/develop...HEAD` (4 commits).

**User-facing:** New public **`IActuationPressPoint`**; **`InputAction`** press-helper remarks/behavior; **`ButtonControl`**, **`Vector2Control`**, **`StickControl`** docs around **`pressPoint`**; **`RespondingToActions.md`** updated for polling vs **`pressPoint`**.

### Documentation updates in this PR

- **`RespondingToActions.md`** — Clarifies **`IsPressed`** / frame helpers vs interactions and **`pressPoint`**; review wording and cross-refs after merge.

### Follow-up (optional)

- If QA/docs want a dedicated **`IActuationPressPoint`** callout in a control reference page, add a short subsection and link from **[Responding to Actions](xref:input-system-responding)**.

---

**Jira (fetched via MCP):** Bug, **In Progress**, **TBD** priority, assignee **Darren Kelly**, label **Input-Actions**. Ticket summary: *InputAction.IsPressed() and IsInProgress() triggers on wrong values when rewriting the default values of Vector2 interactions.*


Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
```diff
@@ -604,6 +604,7 @@ public unsafe InputControl activeControl
                     if (controlIndex != InputActionState.kInvalidIndex)
                         return state.controls[controlIndex];
                 }
+
```
ekcoh: Noticed there is a lot of white space formatting on this file, is it manual or by formatter?

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/ButtonControl.cs
```diff
@@ -67,6 +67,10 @@ public class ButtonControl : AxisControl
         /// </summary>
         public float pressPointOrDefault => pressPoint > 0 ? pressPoint : s_GlobalDefaultButtonPressPoint;
 
+        float IActuationPressPoint.pressPoint => pressPoint;
```
ekcoh: To me it look odd that presspoint (new and previous is modelled on the control instead of on the interaction), what happens when two interactions with different press points bind to the same button?

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/ButtonControl.cs
```diff
@@ -67,6 +67,10 @@ public class ButtonControl : AxisControl
...
+
+        float IActuationPressPoint.pressPointOrDefault => pressPointOrDefault;
```
ekcoh: Is this needed compared to implicitly implementing the interface? It just forwards it seems

ekcoh (commented):
@Darren-Kelly-Unity Would it be possible to get rid of all the whitespace diffs on this PR? It would be much easier to review if this was corrected.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
ekcoh: For visibility: I tried running formatter on this PR and this yielded nothing - and same result on develop so to me its unclear where these white space changes come from.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/ButtonControl.cs
Darren-Kelly-Unity: Confirmed this with AI but you can see in InputActionState.GetActuationPressThreshold() that:
 
 There is an order, it is control-first for device semantics, interaction when the control does not override, then the default settings value.
 
 AI Brief answer:
 
 "Press threshold on the control models the physical control’s default; each binding’s interaction can still use its own press point when the control does not fix one, and separate bindings to the same button keep separate interaction state and thresholds unless the control’s own pressPoint is set, in which case that value wins for unified resolution."
 
 There is also a longer text/ investigation done if you want to see that too?

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/ButtonControl.cs
Darren-Kelly-Unity: The first one is needed due to the value being able to be configured as a parameter in a layout.  [InputControl(parameters = "pressPoint=0.234")]

The second was redundant so removed that.

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Layout-configured press threshold, or a value less than or equal to zero when unset.
+        /// </summary>
+        float pressPoint { get; }
+
+        /// <summary>
+        /// Effective press threshold: <see cref="pressPoint"/> when set, otherwise the global default.
+        /// </summary>
+        float pressPointOrDefault { get; }
```
ekcoh: Why do we need two? Would it be better to have one, e.g. pressPoint that returns default if not having a layout-configured press threshold? Since both go through an interface anyway I wouldn't expect it to be an optimisation?

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// press-style checks such as <see cref="UnityEngine.InputSystem.InputAction.IsPressed"/>.
+    /// </summary>
+    /// <remarks>
+    /// Implemented by <see cref="ButtonControl"/> and <see cref="Vector2Control"/> (including
+    /// <see cref="StickControl"/>). A negative or zero <see cref="pressPoint"/> means "use default
+    /// resolution" via <see cref="pressPointOrDefault"/>; see each type for details.
+    /// </remarks>
+    /// <seealso cref="ButtonControl"/>
+    /// <seealso cref="Vector2Control"/>
+    public interface IActuationPressPoint
```
ekcoh: Not having a problem with this, but be mindful that this PR adds API surface which means committing to it - calls for careful testing and review.

ekcoh: Is the reason this is public to support the IActuationPressPoint on plugin control types? E.g. XR-specific controls or similar?

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
```diff
@@ -1744,7 +1763,7 @@ public unsafe float GetTimeoutCompletionPercentage()
 
                     if (interactionState.totalTimeoutCompletionTimeRemaining > 0)
                     {
-                        return (interactionState.totalTimeoutCompletionDone + timerCompletion * interactionState.timerDuration)  /
+                        return (interactionState.totalTimeoutCompletionDone + timerCompletion * interactionState.timerDuration) /
```
ekcoh, line 1774: Looks like pure formatting diff and could be reverted

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
```diff
@@ -1596,12 +1596,51 @@ private void ProcessControlStateChange(int mapIndex, int controlIndex, int bindi
...
         }
 
+        private BindingState* GetBindingStateForInteractionParameters(BindingState* bindingStatePtr)
+        {
+            if (bindingStatePtr->isPartOfComposite)
+                return &bindingStates[bindingStatePtr->compositeOrCompositeBindingIndex];
+            return bindingStatePtr;
+        }
+
+        private float? TryGetExplicitPressInteractionPressPoint(BindingState* bindingStateForInteractions)
```
ekcoh: Minor: But is `float?` better here than `bool TryGetExplicitPressInteractionPressPoint(BindingState* bindingStateForInteractions, ref float out)`?

ekcoh: Maybe its just a matter of taste...

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
```diff
@@ -1596,12 +1596,51 @@ private void ProcessControlStateChange(int mapIndex, int controlIndex, int bindi
...
+        {
+            var count = bindingStateForInteractions->interactionCount;
+            if (count == 0)
+                return null;
+            var start = bindingStateForInteractions->interactionStartIndex;
+            for (var i = 0; i < count; ++i)
```
ekcoh: Corner case: I am confused around this code, it looks like if there are multiple `PressInteraction`s it will return the pressPoint of the first one? While this is probably OK, it might be good documenting it on PressInteraction or similar?

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
```diff
@@ -1596,12 +1596,51 @@ private void ProcessControlStateChange(int mapIndex, int controlIndex, int bindi
...
+                    return press.pressPoint;
+            }
+
+            return null;
+        }
+
+        private float GetActuationPressThreshold(InputControl control, BindingState* bindingStatePtr)
+        {
+            var bindingForInteractions = GetBindingStateForInteractionParameters(bindingStatePtr);
+            var explicitPress = TryGetExplicitPressInteractionPressPoint(bindingForInteractions);
```
ekcoh: This cost is paid also when the first return is triggered. Did you compare any performance benchmark for this to rule out that these changes do not regress performance for various press configurations? Might be worthwhile to check since we have a combination here of virtual calls, fixed overhead evaluation, type checks etc.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/InputControlExtensions.cs
```diff
@@ -65,8 +68,8 @@ public static bool IsPressed(this InputControl control, float buttonPressPoint =
                 throw new ArgumentNullException(nameof(control));
             if (Mathf.Approximately(0, buttonPressPoint))
             {
-                if (control is ButtonControl button)
-                    buttonPressPoint = button.pressPointOrDefault;
+                if (control is IActuationPressPoint actuation)
```
ekcoh: Beware that this type check might be more expensive than the previous. Previously checking against a specific type, now checking for multiple types implementing interface. Its probably negligible but might we worth checking since hot-path.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/Vector2Control.cs
```diff
@@ -36,6 +36,27 @@ public class Vector2Control : InputControl<Vector2>
...
+        /// <seealso cref="pressPointOrDefault"/>
+        /// <seealso cref="IActuationPressPoint"/>
+        /// <seealso cref="UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint"/>
+        public float pressPoint = -1;
+        float IActuationPressPoint.pressPoint => pressPoint;
+
+        /// <summary>
+        /// Return <see cref="pressPoint"/> if set, otherwise return <see cref="UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint"/>.
+        /// </summary>
+        public float pressPointOrDefault => pressPoint > 0 ? pressPoint : ButtonControl.s_GlobalDefaultButtonPressPoint;
```
ekcoh:
```suggestion
        public float pressPointOrDefault => pressPoint > 0.0f ? pressPoint : ButtonControl.s_GlobalDefaultButtonPressPoint;
```

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/Vector2Control.cs
```diff
@@ -36,6 +36,27 @@ public class Vector2Control : InputControl<Vector2>
...
+        /// <remarks>
+        /// By default, this property is set to -1. If the value of the property is negative,
+        /// <see cref="UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint"/> is used unless a
+        /// <see cref="UnityEngine.InputSystem.Interactions.PressInteraction"/> on the binding sets an explicit
+        /// <c>pressPoint</c> (see <see cref="UnityEngine.InputSystem.InputAction.IsPressed"/> remarks).
+        /// </remarks>
+        /// <seealso cref="pressPointOrDefault"/>
+        /// <seealso cref="IActuationPressPoint"/>
+        /// <seealso cref="UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint"/>
+        public float pressPoint = -1;
```
ekcoh:
```suggestion
        public float pressPoint = -1.0f;
```

ekcoh (commented):
Reviewed, I had quite a lot of questions and some remaining from last review so leaving this as comment-only for now.

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
Darren-Kelly-Unity: You are right I will make this internal.

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
Darren-Kelly-Unity: This is just following the format that was already there for now, not sure about removing one as it was publicly available before.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
Darren-Kelly-Unity: I have left a few of the formatting changes in for now as they are correct. This one also looks ok I think?

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
Darren-Kelly-Unity: You are right, think it's better with bool.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
Darren-Kelly-Unity: Yes, It was already this way for buttonControl's previously. I have added documentation for this now.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/InputControlExtensions.cs
Darren-Kelly-Unity: I added performance tests for this vs the legacy system, we can maybe chat about it in a call if that makes sense but I don't see much change here.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/ButtonControl.cs
ekcoh: Thanks, in other words (shorter): 

Press threshold is priority in falling order:
1. Interaction press threshold (part of binding)
2. Control press threshold
3. Global default press threshold

?

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
ekcoh: If it was public we cannot remove it now (unless it fits a scheme to deprecate), but I am afraid two variants creates confusion around the interface contract that a user should implement.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
ekcoh: Agreed, marking this one resolved

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
ekcoh: Thanks, marking this one resolved

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/InputControlExtensions.cs
ekcoh: I think the most scalable way is adding any such observations (quantified) to the PR description instead?

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
ekcoh: But if the new interface only implemented the previous implicit function:
`pressPointOrDefault` wouldn't this still work since it takes the same value if available and otherwise fall back?

ekcoh: Since `if (control is IActuationPressPoint actuation)` is the new condition also covering `Button`?

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/ButtonControl.cs
Darren-Kelly-Unity: Yes exactly!

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/InputControlExtensions.cs
Darren-Kelly-Unity: Yep I added it to the PR description already!

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
```diff
@@ -84,9 +84,10 @@ Finally, there are three methods you can use to poll for button presses and rele
...
+The applicable press point is chosen from, in order: a positive [`pressPoint`](xref:UnityEngine.InputSystem.Controls.ButtonControl.pressPoint) on the driving control when it is a [ButtonControl](xref:UnityEngine.InputSystem.Controls.ButtonControl), or a positive [`pressPoint`](xref:UnityEngine.InputSystem.Controls.Vector2Control.pressPoint) when it is a [Vector2Control](xref:UnityEngine.InputSystem.Controls.Vector2Control) or [StickControl](xref:UnityEngine.InputSystem.Controls.StickControl); otherwise, if the binding lists one or more [Press](xref:UnityEngine.InputSystem.Interactions.PressInteraction) interactions, the `pressPoint` from the first in the binding's interaction order whose `pressPoint` is greater than zero (any interactions before it that are not `Press`, or `Press` with the default `pressPoint` of zero, are skipped for this step); otherwise [defaultButtonPressPoint](xref:UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint). On [composites](xref:input-system-action-bindings#composite-bindings), interaction parameters are taken from the composite binding.
```
MorganHoarau: Looks like the generated preview put that line in the array. 

Image

Make sure to add doc to review the doc changes

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
Darren-Kelly-Unity: I can try that and see if it works, you may be right yes!

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
```diff
@@ -1232,8 +1238,16 @@ private int ExpectedFrame()
         /// to a <see cref="StickControl"/>, the control will be considered "pressed" once the magnitude
         /// of the Vector2 of the control has crossed the press threshold.
         ///
-        /// Finally, note that custom button press points of controls (see <see cref="ButtonControl.pressPoint"/>)
-        /// are respected and will take precedence over <see cref="InputSettings.defaultButtonPressPoint"/>.
+        /// Press threshold (based on <see cref="InputControl.EvaluateMagnitude()"/> for the driving control): for a
```
MorganHoarau: 1st copy-paste from line 1157. This is a maintenance burden.
I'd recommend going for only defining for IsPressed() and then just a oneline:
/// The press threshold is resolved as described in <see cref="IsPressed"/> remarks.
This is what you did for WasPressedThisDynamicUpdate and WasReleasedThisDynamicUpdate.

Note that the block (starting by `Also note that because ...`) above is also duplicated and would benefit from the same treatment.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
```diff
@@ -1326,8 +1342,16 @@ public unsafe bool WasPressedThisDynamicUpdate()
...
```
MorganHoarau: 2nd copy-paste from line 1157. Same as above.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
```diff
@@ -1155,8 +1154,16 @@ public void Reset()
...
```
MorganHoarau: Very dense, even with the formatting in doc. I had to write it down myself and split it in a bullet point to understand. I don't know if it's possible to have a numbered list in xml doc? Would be a question for doc team.

I asked Claude, supposedly this is the way in xml doc:
```cs
/// The press threshold is resolved in priority order:
/// <list type="number">
/// <item>A positive <see cref="ButtonControl.pressPoint"/> or <see cref="Vector2Control.pressPoint"/>
/// on the driving control.</item>
/// <item>The <see cref="PressInteraction.pressPoint"/> from the first
/// <see cref="PressInteraction"/> on the binding with a value greater than zero.
/// For composites, interactions are read from the composite binding.</item>
/// <item><see cref="InputSettings.defaultButtonPressPoint"/>.</item>
/// </list>
```

However I am doubtful that our in house documentation parser is able to work with it because it only support a subsets of tag and rules.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/Vector2Control.cs
```diff
@@ -36,6 +36,27 @@ public class Vector2Control : InputControl<Vector2>
         [InputControl(offset = 4, displayName = "Y")]
         public AxisControl y { get; set; }
 
+        /// <summary>
+        /// Minimum vector magnitude before the control is considered pressed for purposes such as
+        /// <see cref="UnityEngine.InputSystem.InputAction.IsPressed"/> when this control drives the action.
+        /// </summary>
+        /// <remarks>
+        /// By default, this property is set to -1. If the value of the property is negative,
```
MorganHoarau: nit
```suggestion
        /// By default, this field is set to -1. If the value of the property is negative,
```

MorganHoarau:
The tickets mentions a deadzone issue: "Deadzone(0.3) WasPerformedThisFrame triggers incorrectly at Magnitude 0.000". 

Have you tried to repro that issue? I could not find tests for it.

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
suearkinunity:
```suggestion

The applicable press point is chosen from, in order: 

1. A positive [`pressPoint`](xref:UnityEngine.InputSystem.Controls.ButtonControl.pressPoint) on the driving control when it is a [ButtonControl](xref:UnityEngine.InputSystem.Controls.ButtonControl).
2. A positive [`pressPoint`](xref:UnityEngine.InputSystem.Controls.Vector2Control.pressPoint) when it is a [Vector2Control](xref:UnityEngine.InputSystem.Controls.Vector2Control) or [StickControl](xref:UnityEngine.InputSystem.Controls.StickControl).
3. If the binding lists one or more [Press](xref:UnityEngine.InputSystem.Interactions.PressInteraction) interactions, the `pressPoint` from the first interaction order whose `pressPoint` is greater than zero (any interactions before this that are either not `Press`, or `Press` with the default `pressPoint` of zero, are skipped for this step).
4. [defaultButtonPressPoint](xref:UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint). 

On [composites](xref:input-system-action-bindings#composite-bindings), interaction parameters are taken from the composite binding.
```

Packages/com.unity.inputsystem/Documentation~/RespondingToActions.md
```diff
@@ -84,9 +84,10 @@ Finally, there are three methods you can use to poll for button presses and rele
...
-|[`InputAction.WasPressedThisFrame()`](xref:UnityEngine.InputSystem.InputAction.WasPressedThisFrame)|True if the level of [actuation](xref:UnityEngine.InputSystem.InputControl.EvaluateMagnitude) on the action has, at any point during the current frame, reached or gone above the [press point](xref:UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint).|
-|[`InputAction.WasReleasedThisFrame()`](xref:UnityEngine.InputSystem.InputAction.WasReleasedThisFrame)|True if the level of [actuation](xref:UnityEngine.InputSystem.InputControl.EvaluateMagnitude) on the action has, at any point during the current frame, gone from being at or above the [press point](xref:UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint) to at or below the [release threshold](xref:UnityEngine.InputSystem.InputSettings.buttonReleaseThreshold).|
+|[`InputAction.IsPressed()`](xref:UnityEngine.InputSystem.InputAction.IsPressed)|True if the level of [actuation](xref:UnityEngine.InputSystem.InputControl.EvaluateMagnitude) on the action has crossed the applicable press point and did not yet fall to or below the [release threshold](xref:UnityEngine.InputSystem.InputSettings.buttonReleaseThreshold).|
```
suearkinunity:
```suggestion
|[`InputAction.IsPressed()`](xref:UnityEngine.InputSystem.InputAction.IsPressed)|True if the level of [actuation](xref:UnityEngine.InputSystem.InputControl.EvaluateMagnitude) on the action has crossed the applicable press point and hasn't fallen to or below the [release threshold](xref:UnityEngine.InputSystem.InputSettings.buttonReleaseThreshold).|
```

suearkinunity (approved):
Left suggestions, but I'm approving the PR so you're not blocked waiting for me to approve later.

Pauliusd01 (approved):
Seems to work fine, tested the things mentioned in the test plan as well as the user project that I had to modify. Just not sure if this counts as a fix or something new (FIX vs NEW) and whether that is the right way to fix this. Here's the tests I used if you want to double check: 

[PR_2411.zip](https://github.com/user-attachments/files/27955731/PR_2411.zip)
[InputTester.zip](https://github.com/user-attachments/files/27955740/InputTester.zip)

Packages/com.unity.inputsystem/InputSystem/Controls/IActuationPressPoint.cs
Darren-Kelly-Unity: I tried what you mentioned but we cannot have just pressPointOrDefault in the interface as we couldn't tell then if it's default being chosen or not.

The IActuationPressPoint does cover buttons now also yes!

MorganHoarau (approved):

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
```diff
@@ -1591,12 +1591,56 @@ private void ProcessControlStateChange(int mapIndex, int controlIndex, int bindi
...
+            if (bindingStatePtr->isPartOfComposite)
+                return &bindingStates[bindingStatePtr->compositeOrCompositeBindingIndex];
+            return bindingStatePtr;
+        }
+
+        // Resolves the explicit press threshold from PressInteraction on the binding. When multiple PressInteraction
+        // instances exist, uses the first in interaction list order with pressPoint > 0. GetActuationPressThreshold
+        // prefers this over a control pressPoint so IsPressed / WasPressedThisFrame / WasReleasedThisFrame stay aligned
+        // with PressInteraction.
+        private bool TryGetExplicitPressInteractionPressPoint(BindingState* bindingStateForInteractions, ref float explicitPressInteraction)
```
ekcoh: Minor: It feels like this method could easily fold into GetActuationPressThreshold? Seems like it's only used in one place? It looks like it might even read better.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputActionState.cs
```diff
@@ -1591,12 +1591,56 @@ private void ProcessControlStateChange(int mapIndex, int controlIndex, int bindi
...
+            }
+
+            return false;
+        }
+
+        internal float GetActuationPressThreshold(InputControl control, BindingState* bindingStatePtr)
+        {
+            var bindingForInteractions = GetBindingStateForInteractionParameters(bindingStatePtr);
+            float explicitPressThreshold = 0.0f;
+            var hasExplicitPressThreshold = TryGetExplicitPressInteractionPressPoint(bindingForInteractions, ref explicitPressThreshold);
```
ekcoh: Would it be fine to just return the explicit one directly even if control doesn't implement IActuationPressPoint?

ekcoh: (Also see comment around merging this function with the one above - I think it will actually increase readability) - however I am surprised we are not just returning the "explicitPressThreshold" directly if we have one? - maybe rename to "interactionPressThreshold" to make it clearer what it is as well?.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/Vector2Control.cs
```diff
@@ -36,6 +36,27 @@ public class Vector2Control : InputControl<Vector2>
...
+        /// By default, this field is set to -1. If the value of the property is negative,
+        /// <see cref="UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint"/> is used for action-level press
+        /// APIs unless a <see cref="UnityEngine.InputSystem.Interactions.PressInteraction"/> on the binding sets an
+        /// explicit <c>pressPoint</c>, in which case that interaction threshold takes priority (see
+        /// <see cref="UnityEngine.InputSystem.InputAction.IsPressed"/> remarks).
+        /// </remarks>
+        /// <seealso cref="pressPointOrDefault"/>
+        /// <seealso cref="ButtonControl"/>
+        /// <seealso cref="UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint"/>
+        public float pressPoint = -1.0f;
```
ekcoh: This looks obsolete? We only need pressPointOrDefault now right? (after recent changes)

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/Vector2Control.cs
```diff
@@ -36,6 +36,27 @@ public class Vector2Control : InputControl<Vector2>
...
+
+        /// <summary>
+        /// Return <see cref="pressPoint"/> if set, otherwise return <see cref="UnityEngine.InputSystem.InputSettings.defaultButtonPressPoint"/>.
+        /// </summary>
+        public float pressPointOrDefault => pressPoint > 0.0f ? pressPoint : ButtonControl.s_GlobalDefaultButtonPressPoint;
```
ekcoh: This should just be `=> ButtonControl.s_GlobalDefaultButtonPressPoint` right?

ekcoh (commented):
Reducing interface to a single method makes a lot of sense. Added some remaining suggestions/questions - once those are sorted I will approve this PR.

Packages/com.unity.inputsystem/InputSystem/Runtime/Controls/Vector2Control.cs
ekcoh: But if it just returns default anyway - that would also be the InputActionState fallback - so is there any point at all implementing the interface?

ekcoh (approved):
Discussed some simplifications to be done before. Will approve now to unblock.

Darren-Kelly-Unity:
> ## PR Reviewer Guide 🔍
> Here are some key observations to aid the review process:
> 
> ⏱️ **Estimated effort to review**: 3 🔵🔵🔵⚪⚪
> Medium effort: the PR has extensive test churn, but the core behavior change is concentrated in the action press-threshold resolution logic.
> 
> 🏅 **Score**: 84
> The change is well tested and targeted, but the new threshold-selection rule can still make action-level press polling disagree with supported stacked `Press` interactions.
> 
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Threshold mismatch**](https://github.com/Unity-Technologies/InputSystem/pull/2411/files#diff-b77be6054c214ebc44deeb7f6b4309194f2dab8ee301458f1d22869aeba538caR1608-R1616)
> Skipping earlier `PressInteraction` entries whose `pressPoint` is left at the default `0` can still desynchronize action-level helpers from interaction behavior. For example, with `interactions: "press,press(pressPoint=0.8)"`, the first interaction will press at the default threshold, but `IsPressed`/`WasPressedThisFrame` will stay false until `0.8` because this scan ignores the first `PressInteraction`. That reintroduces the inconsistency this change is trying to remove for valid stacked-press setups.
> 
> ```cs
> var count = bindingForInteractions->interactionCount;
> if (count > 0)
> {
>     var start = bindingForInteractions->interactionStartIndex;
>     for (var i = 0; i < count; ++i)
>     {
>         if (interactions[start + i] is PressInteraction press && press.pressPoint > 0)
>         {
>             return press.pressPoint;
> ```
> 
> * [ ]  Update review
> 
> 🤖 Helpful? Please react with 👍/👎 | Questions❓Please reach out in Slack [#ask-u-pr](https://unity.enterprise.slack.com/archives/C08MXQ4MDNW)

I have checked this & ran it through an AI check too. This is not completely true what's said and it's documented for this case also.

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputAction.cs
```diff
@@ -1232,8 +1242,21 @@ private int ExpectedFrame()
         /// to a <see cref="StickControl"/>, the control will be considered "pressed" once the magnitude
         /// of the Vector2 of the control has crossed the press threshold.
         ///
-        /// Finally, note that custom button press points of controls (see <see cref="ButtonControl.pressPoint"/>)
-        /// are respected and will take precedence over <see cref="InputSettings.defaultButtonPressPoint"/>.
+        /// Press threshold (based on <see cref="InputControl.EvaluateMagnitude()"/> for the driving control):
+        ///
+        /// 1. If the binding lists one or more <see cref="UnityEngine.InputSystem.Interactions.PressInteraction"/>
```
Nobuna-no: So you decided to keep the duplicated comments 3 times?

Darren-Kelly-Unity: Oops! Fixed.

Pauliusd01 (approved):
LGTM rechecked the test plan from here: https://github.com/Unity-Technologies/InputSystem/pull/2411#issuecomment-4474537003 as well as the user project with the latest commit. Attaching the test plan scripts if you want to double check them: 

[PR_2411.zip](https://github.com/user-attachments/files/28140318/PR_2411.zip)

PR #2412, opened 2026-04-29
FIX: ISXB-1097 Changed default event handled policy and deprecated event handling policy SuppressStateUpdates since broken by design.

ekcoh:
#### Summary

Fixes [ISXB-1097](https://jira.unity3d.com/browse/ISXB-1097): Setting `InputEventPtr.handled = true` does not prevent actions from triggering when switching devices.

##### Root Cause

The default event handled policy (`SuppressStateUpdates`) discards handled events entirely, including their state updates. This desynchronizes the Input System's internal device state from the actual source state. When the user switches to a different device, the non-handled event arrives while the previous device still has stale state, causing spurious action triggers.

##### Fix

- **Changed the default policy** from `SuppressStateUpdates` to `SuppressActionEventNotifications`. Under the new default, handled events still propagate state updates (keeping device state synchronized) but suppress action callback notifications (`Started`/`Performed`/`Canceled`). The pull-based action APIs (`WasPressedThisFrame`, `WasPerformedThisFrame`) are also gated by the suppression flag and return `false` for suppressed events.
- **Deprecated `SuppressStateUpdates`** with `[Obsolete]` (warning, not error) and an explanatory message. Existing code continues to compile but will receive warnings.
- **Fixed accidental `handled` flag** in `ResetDevice`: the synthetic reset event used `eventId = -1` as a sentinel, which set all bits including the handled flag (bit 31 of `eventId`). Changed to `InputEvent.InvalidEventId` so the handled bit is not accidentally set.
- **Fixed `RebindingOperation`** to only override the global event handled policy when the user makes an explicit choice via `WithActionEventNotificationsBeingSuppressed()`. Previously, the uninitialized field defaulted to `SuppressStateUpdates`, silently overriding whatever policy was configured globally.
- **Changed event suppression from map-wide to per-action tracking.** All `WasXxxThisFrame()` and `WasXxxThisDynamicUpdate()` polling APIs now check a per-action `Suppressed` flag on `TriggerState` (bit 7 of the existing flags byte) instead of a shared `m_Suppressed` bool on `InputActionState`. The flag is stamped when each action's phase changes, so it correctly reflects whether the specific event that triggered that action was handled. This fixes false positives/negatives when multiple events with mixed handled/unhandled states arrive in the same frame. Also fixes `WasReleasedThisFrame()`, `WasCompletedThisFrame()`, and their `DynamicUpdate` variants which previously did not check suppression at all.

#### Behavioral Changes

| Scenario | Before (SuppressStateUpdates) | After (SuppressActionEventNotifications) |
|----------|-------------------------------|------------------------------------------|
| Handled event device state | Not updated (discarded) | Updated (state propagates) |
| Handled event action callbacks | Not fired (event never processed) | Not fired (suppressed by `m_Suppressed`) |
| `control.isPressed` after handled press | `false` | `true` (reflects actual device state) |
| `action.WasPressedThisFrame()` after handled press | `false` (no state update) | `false` (gated by `IsSuppressed`) |
| `action.WasReleasedThisFrame()` after handled release | `false` | `false` (now gated by `IsSuppressed`) |
| `action.WasCompletedThisFrame()` after handled event | `false` | `false` (now gated by `IsSuppressed`) |
| Device switch after handled event | Spurious action triggers (stale state) | No spurious triggers (state synchronized) |

#### Migration for Existing Projects

 Existing projects are **not affected** by this change. The event handled policy is serialized as part of `InputManager` state, so projects that were using the old default (`SuppressStateUpdates`) will retain it upon deserialization. 

To opt in to the new behavior in an existing project, set the policy explicitly:

```csharp
InputSystem.manager.inputEventHandledPolicy = InputEventHandledPolicy.SuppressActionEventNotifications;
```

**UPDATE May 11th (OPEN ISSUE): The above is not correct. I overlooked that the event handling policy isn't a true setting and also that the above is internal API which hence is not possible for a user to change directly. If we want users to allow opting out of this we need to introduce a setting that is directly deprecated to allow switching back. That would at least allow opting out via InputSettings file OR we need to expose API to switch this in runtime (simpler) but would also be an API that is deprecated directly since we do not really want this to change. So we need to align on this and make a decision.**

Or equivalently:

```csharp
InputSystem.manager.inputEventHandledPolicy = InputEventHandledPolicy.Default;
```

Projects explicitly using `SuppressStateUpdates` will receive an `[Obsolete]` compiler warning guiding them to switch.

#### Open Design Question

When a device is reset (`InputSystem.ResetDevice`), pass-through actions currently emit both `Canceled` and `Performed(0f)` — because pass-through actions fire on *any* value change, including the reset-to-default. This is consistent with the pass-through contract but inconsistent with button/value actions (which only emit `Canceled`). If we decide to suppress the `Performed(0f)` on reset, the synthetic reset event should be explicitly marked as `handled` rather than relying on sentinel values. See inline design note in `CoreTests_Actions.cs`.

### Testing status & QA

Have run editor, playmode and player tests locally to verify functional behaviour essentially verifying:
- [X] Verify `Events_HandledEventsShouldNotTriggerActionsWhenSwitchingDevices` passes (direct ISXB-1097 regression test — device switching)
- [X] Verify `Events_HandledPressEdgeInMultiEventFrameShouldNotTriggerActions` passes (ISXB-1097 regression test — multi-event frame, e.g. DualSense 600 Hz)
- [X] Verify `Events_AllWasXxxThisFrameAPIsRespectEventSuppression` passes (ISXB-1097 regression test — all WasXxx polling APIs gated by per-action suppression)                                                                                    
- [X] Verify `Events_PerActionSuppressionWithMixedHandledEvents` passes (ISXB-1097 regression test — mixed handled/unhandled events in same frame, per-action correctness) 
- [X] Verify `EventHandledPolicy_ShouldReflectUserSetting` passes with updated defaults
- [X] Verify `Actions_InteractiveRebinding_CanSuppressEventsWhileListening` passes for both `SuppressStateUpdates` and `SuppressActionEventNotifications` policies
- [X] Verify `Actions_ResettingDevice_CancelsOngoingActionsThatAreDrivenByIt` passes (pass-through Performed(0f) behavior)
- [X] Run full action and event test suites to check for regressions

Did basic testing with Rebinding example. 

Tested repro case associated with ticket with DualSense **remains to be done**.

**Suggest that before merge we also cover**:
- [x] Simulated touch tooling since it relies on setting events as "handled" and check for regressions.
- [x] Thorough testing of Rebinding sample.
- [x] Testing OnScreen controls and VirtualMouse or other virtual or emulated devices.
- [ ] Test upgrade experience for a user already having the previous default event handling behaviour setting.
- [x] Test experience when creating a new project.
- [ ] Make sure XR, e.g. XRI have verified integration with these changes.

### Overall Product Risks

- Complexity: High - changes are relatively small but affect core behaviour of the system.
- Halo Effect: High - may potentially have side effects.

### Comments to reviewers

Apart from reviewing solution and test cases I also want us to decide on **Open Design Question**. Not emitting Performed makes much more sense and makes value types consistent in my opinion, but it would imply another behaviour change.

### Checklist

Before review:

- [X] Changelog entry added.
    - Extended existing tests and added new regression tests: `Events_HandledEventsShouldNotTriggerActionsWhenSwitchingDevices`, `Events_HandledPressEdgeInMultiEventFrameShouldNotTriggerActions`, `Events_AllWasXxxThisFrameAPIsRespectEventSuppression`, 
    `Events_PerActionSuppressionWithMixedHandledEvents` 
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
```diff
@@ -1300,7 +1300,7 @@ public unsafe bool WasPressedThisFrame()
         public unsafe bool WasPressedThisDynamicUpdate()
         {
             var state = GetOrCreateActionMap().m_State;
-            if (state != null)
+            if (state != null && !state.IsSuppressed)
```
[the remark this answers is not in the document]
ekcoh: @u-pr   Good catch — you're right that m_Suppressed is a single bool on InputActionState (shared across all actions in the map), not per-action. However, the issue is slightly different from what you         
  described.                                                                                                                                                                                              
                                                            
  m_Suppressed is not sticky across the frame — it's reassigned on every ProcessControlStateChange invocation (line 1564 in InputActionState.cs). So during event processing within the update, each event
   correctly gets its own suppression: event A (handled) sets m_Suppressed = true and its callbacks are suppressed, then event B (not handled) sets m_Suppressed = false and its callbacks fire normally. 
  The callback-based APIs (started/performed/canceled) are not affected by this issue.
                                                                                                                                                                                                          
  The problem is with the polling APIs (WasPressedThisFrame, WasReleasedThisFrame, etc.). These are called by user code after the update completes, at which point m_Suppressed reflects whatever the last
   ProcessControlStateChange in the frame set it to. This creates two incorrect scenarios:
                                                                                                                                                                                                          
  1. False negatives: If the last event in the frame was handled but an earlier unhandled event triggered an action, IsSuppressed is true and all WasXxx APIs return false — incorrectly suppressing      
  results from the unhandled event.
  2. False positives: If the last event was unhandled but an earlier handled event triggered an action, IsSuppressed is false and WasXxx APIs return true — incorrectly exposing results that should have 
  been suppressed.                                                                                                                                                                                        
   
  This is a pre-existing limitation in WasPressedThisFrame and WasPerformedThisFrame, which already had the IsSuppressed check before this PR. Our change only makes WasReleasedThisFrame,                
  WasCompletedThisFrame, and the DynamicUpdate variants consistent with the same pattern. Fixing this properly would require per-action (or per-actuation) suppression tracking rather than a map-wide
  flag, which is a larger scope change.                                                                                                                                                                   
                                                            
  In practice, the "mixed handled/unhandled events in the same frame for the same map" scenario is uncommon — typical usage is either suppressing all events (rebinding) or suppressing events from a     
  specific device (the ISXB-1097 scenario, where the onEvent listener handles all events). But it's a real gap worth tracking for a future improvement.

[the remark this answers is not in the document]
ekcoh: I decided this previous "limitation" have problems that could be seen as surprising side-effects and doesn't fully make sense so I decided to address it in https://github.com/Unity-Technologies/InputSystem/pull/2412/commits/a7038c042841eff8823dd7c90bd9a4f6cd989788.

ekcoh:
@u-pr Can you check if all previously raised concerns have been addressed in this PR?

Assets/Tests/InputSystem/CoreTests_Actions.cs
```diff
@@ -3748,6 +3748,13 @@ public void Actions_ResettingDevice_CancelsOngoingActionsThatAreDrivenByIt()
             // does not cause the action to start back up. For pass-through actions, that is different
             // as *any* value change performs the action. So here, we see *both* a cancellation and then
             // immediately a performing of the action.
+            //
```
ekcoh, line 3751: This should be rephrased into a decision when decision exists after review/discussion.

Pauliusd01:
FYI: likely will only have time to take a look at this on the latter half of next week

ekcoh:
@Pauliusd01 I can do some more testing myself to smoke this further and reach out to XR. We also have a decision to make which might call for additional changes to this.

MorganHoarau:
> When a device is reset (InputSystem.ResetDevice), pass-through actions currently emit both Canceled and Performed(0f) — because pass-through actions fire on any value change, including the reset-to-default. **This is consistent with the pass-through contract** [...]

I could not find info about that particular case in the doc. Even if it make sense, it's not obvious that "any value" also means system-synthetic event. Might be worth documenting this subtlety. 

> [...] If we decide to suppress the Performed(0f) on reset, the synthetic reset event should be explicitly marked as handled rather than relying on sentinel values.

I do not think we should change it as some users might be relying on the existing behaviour. It also gives a chance to sync any user-owned state linked to that passthrough action. Though, users could also reset such state through the `Canceled` callback, because we are now feeding them with synthetic data anyway.

Was that a long standing issue or requested change of behaviour? I might be missing some context here. I'm just having a hard time figuring out the user story behind it.

Even if getting rid of the `Performed(0f)` sounds appealing, I just do not see real value for the potential risk.

Assets/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -1436,6 +1452,222 @@ public void Events_ShouldRespectHandledPolicyUponUpdateAndSuppressedPressTransit
...
+
+        var southAction = new InputAction(name: "south", type: InputActionType.Button,
+            binding: "<Gamepad>/buttonSouth");
+        var northAction = new InputAction(name: "north", type: InputActionType.Button,
+            binding: "<Gamepad>/buttonNorth");
+        southAction.Enable();
+        northAction.Enable();
+
+        // Handle events that press buttonSouth, but let buttonNorth events through.
+        InputSystem.onEvent += (eventPtr, device) =>
```
MorganHoarau, line 1711: Not sure to understand why we need to subscribe an empty lambda here...

Assets/Tests/InputSystem/CoreTests_Events.cs
```diff
@@ -1436,6 +1452,222 @@ public void Events_ShouldRespectHandledPolicyUponUpdateAndSuppressedPressTransit
...
+
+        // Next frame: verify a genuine new press (unhanded) does trigger normally.
+        handleNextEvent = false;
+        InputSystem.QueueStateEvent(gamepad, new GamepadState()); // release
+        InputSystem.Update();
+        InputSystem.QueueStateEvent(gamepad,
+            new GamepadState().WithButton(GamepadButton.South)); // new press
+        InputSystem.Update();
+
+        Assert.That(performedCount, Is.EqualTo(1),
```
MorganHoarau, line 1647: Should not be needed because covered by other test in other scenarios, but I would have added the check for `Assert.That(action.WasPressedThisFrame(), Is.True);` to get the expected symmetry and specific scenario covered.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fixed `InputEventPtr.handled` not preventing actions from triggering when switching devices. The default event handled policy has been changed from `SuppressStateUpdates` (now deprecated) to `SuppressActionEventNotifications`, which keeps device state synchronized while suppressing action callbacks for handled events. [ISXB-1097](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1097)
+- Fixed all `InputAction.WasXxxThisFrame()` and `WasXxxThisDynamicUpdate()` polling APIs to use per-action suppression tracking instead of a map-wide flag. Previously, when multiple events arrived in the same frame with mixed handled/unhandled states, the last event's suppression state could incorrectly affect all actions in the map. [ISXB-1097](https://issuetracker.unity3d.com/product/unity/issues/guid/ISXB-1097)
```
MorganHoarau, line 13: Move to `Changed` section?

MorganHoarau (approved):
Just a few minor comments.

Good automated test coverage, every time I had ideas to improve a test, the next one was covering it. 

I like that approach because it moves the suppression ownership to the appropriate layer (`InputActionState.TriggerState`) and makes the system more resilient without adding complexity.

Packages/com.unity.inputsystem/InputSystem/Actions/InputAction.cs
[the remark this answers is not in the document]
ekcoh: This was confirmed to be fixed by u-pr so resolving this conversation

ekcoh:
> > When a device is reset (InputSystem.ResetDevice), pass-through actions currently emit both Canceled and Performed(0f) — because pass-through actions fire on any value change, including the reset-to-default. **This is consistent with the pass-through contract** [...]
> 
> I could not find info about that particular case in the doc. Even if it make sense, it's not obvious that "any value" also means system-synthetic event. Might be worth documenting this subtlety.
> 
> > [...] If we decide to suppress the Performed(0f) on reset, the synthetic reset event should be explicitly marked as handled rather than relying on sentinel values.
> 
> I do not think we should change it as some users might be relying on the existing behaviour. It also gives a chance to sync any user-owned state linked to that passthrough action. Though, users could also reset such state through the `Canceled` callback, because we are now feeding them with synthetic data anyway.
> 
> Was that a long standing issue or requested change of behaviour? I might be missing some context here. I'm just having a hard time figuring out the user story behind it.
> 
> Even if getting rid of the `Performed(0f)` sounds appealing, I just do not see real value for the potential risk.

Fair point about the fact that projects in the wild could be relying on the side-effect of an additional Performed callback (non-consistent) for Passthrough. That is why I didn't change that part now but ensured its called out in the test. Agree we could potentially also clarify this in the manual or scripting manual since it's undocumented - and honestly not something I was aware of before this PR. Since it is inconsistent I am not sure its intentional or a mistake from the past that it performs differently, personally I suspect the latter since it is called out as a "quirk" in the test case - which made want to raise this to discuss since its also not documented. I am not sure what a user would even though with the additional Performed callback carrying a false synthetic state.

There is no request for change of behavior - I am just calling it out since it makes no sense to me. I am completely fine keeping it as is for now. 

@MorganHoarau Did I understand it correctly that you where ok with the bigger behaviour change of this PR I am suggesting - changing the default event consumption mode?

ekcoh:
I notice now this PR looks kind of broken after it was merged with latest changes from develop :/

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Agreed on the first one, the second one seems to instead be a long lasting bug which gave different result deepening on what map and action belonged to.

Assets/Tests/InputSystem/CoreTests_Events.cs
ekcoh: Agreed, good catch, I'll get it added

Assets/Tests/InputSystem/CoreTests_Events.cs
ekcoh: Good catch, something interesting happened here, will revisit :)

ekcoh:
Fixed previous bad merge, this now ok to review and test again

AswinRajGopal (approved):
Per-action check for the suppression bit looks broader fix compared to the action map wide suppression state which fixes the mixed handled/unhandled events received in the same frame; Now only the affected action is suppressed. Looks good to me :) Thanks.

K-Tone:
This looks a good change to me, and I appreciate that it's a fairly contained and readable PR. As to the design questions inline, I didn't develop a preference yet to be honest.

K-Tone (approved):

Pauliusd01:
FYI I am mostly done with testing from: https://github.com/Unity-Technologies/InputSystem/pull/2412#issuecomment-4368448964 as well as some of my own tests which I'm attaching here in case you want to run them yourself: [InputPRTests.zip](https://github.com/user-attachments/files/27441167/InputPRTests.zip)

But I've yet to finish the ones you listed in your description so expecting to do that tomorrow

Pauliusd01 (changes requested):
I think upgrading for the first time might be broken, steps to repro: open the user project -> observe that device switching works with the public package -> add your local package to upgrade -> run the scene again and observe that mouse clicks, space or gamepad don't trigger any logs
(If you restart the editor the user project works correctly, it is only happening on the first upgrade)

Pauliusd01:
Note since the issue above is a one time upgrade issue that gets auto resolved with a restart I'd be okay with it staying if the fix can be complicated/risky. Just had to mentioned it

Pauliusd01:
I've reproduced the upgrade issue in develop, likely something recent is causing it so I'm gonna ignore it here

Pauliusd01 (approved):
Checked tests from here: https://github.com/Unity-Technologies/InputSystem/pull/2412#issuecomment-4388511927 as well as the tests mentioned in the PR description.

ekcoh:
@Pauliusd01 @LeoUnity @MorganHoarau @K-Tone I updated description with this:
**UPDATE May 11th (OPEN ISSUE):** The previous comment about existing projects not being affected is incorrect. I overlooked that the event handling policy isn't a true setting and also that the above is internal API which hence is not possible for a user to change directly. If we want users to allow opting out of this we need to introduce a setting that is directly deprecated to allow switching back. That would at least allow opting out via InputSettings file OR we need to expose API to switch this in runtime (simpler) but would also be an API that is deprecated directly since we do not really want this to change. So we need to align on this and make a decision.**

It's probably easiest we discuss this and make a decision based on shared perspective.

ekcoh:
Conclusion after exploring opt-in / out-out possibilities on: https://github.com/Unity-Technologies/InputSystem/tree/isxb-1097-event-suppression-fix-with-opt-out 

Discussed with @LeoUnity and we landed in the decision to this this PR as-is. A separate branch (linked above) was implemented with an opt-out solution, but since opt-in pattern for existing projects is not technically viable we leave this one only for future reference and delete it later if not needed. 

So we may land this PR if you agree @Pauliusd01?

PR #1424, opened 2021-10-15
Swapping InputManager state buffers and update step counts back to latest player update after editor update

jimon:
### Description

When reading actions or device states from FixedUpdate or early update, the values are incorrect in play mode because we're reading from last frames editor update. This was not a case <1.1.0-pre.6 because editor updates were ignored in play mode when game view had focus.

### Changes made

- Injected a callback in Unity PlayerLoop.Initialization system which runs before early update / FixedUpdate.
- Swap state buffers, update step counts and time offset if previous frame last update was editor update, and update before that was player update.

PR #2419, opened 2026-05-12
FIX: Re-inject player loop hook after user-driven loop reset (UUM-140343)

LeoUnity:
## Summary

Fixes [UUM-140343](https://jira.unity3d.com/browse/UUM-140343) — `PlayerLoop.SetPlayerLoop(PlayerLoop.GetDefaultPlayerLoop())` called from user code mid-play wipes the InputSystem's editor-only `InputSystemPlayerLoopRunnerInitializationSystem`, causing input state to freeze in `FixedUpdate` and other reads that occur between editor updates.

## Root cause

The InputSystem injects an editor-only subsystem into `PlayerLoop.Initialization` to drive `OnPlayerLoopInitialization`, which switches input state buffers back to player mode after an editor update. PR #1424 introduced this as a self-described hot-fix (see comment at `NativeInputRuntime.cs:147-148`). When the user resets the loop to Unity's default, that injection is dropped — the default loop only contains Unity's native subsystems — and the buffer-switch never fires again.

## Fix

Subscribe a watchdog to `EditorApplication.update` that re-injects the subsystem if it's missing. `EditorApplication.update` fires from `SceneTracker::Update()` (`Editor/Src/Selection/SceneInspector.cpp:742`), independently of the player loop, so a user `SetPlayerLoop` cannot wipe it. The watchdog is gated on `EditorApplication.isPlaying`:
- Edit mode: zero work (early return).
- Play mode: one `PlayerLoopSystem`-array scan per editor tick. Common case (injection present) returns immediately.

The original `OnPlayerLoopInitialization` mechanism is left untouched — the watchdog only re-establishes the injection when needed.

## Why not other approaches

Investigated and ruled out (Unity source-confirmed):
- **Add to default player loop**: no managed-side API. `s_defaultLoop` is built from C++ `PLAYER_LOOP_INJECT(...)` macros (`Runtime/Misc/PlayerLoop.cpp:270-292`).
- **Hook `SetPlayerLoop` event**: no callback fires (`PlayerLoop.cpp:368-389` — buffer swap, no events).
- **Move restore into `OnUpdate`**: tried (commit `12a5921f7`), preserved in branch history. Works but breaks the contract that manual `InputSystem.Update(Editor)` callers observe `s_LatestUpdateType == Editor` immediately after; required a `m_ManualUpdateDepth` counter to gate the restore. Reverted in favor of this approach.

## Test plan
- [x] New `[UnityTest]` `Integration_InputUpdatesContinue_AfterResettingPlayerLoopToDefault` reproduces the bug pre-fix and passes post-fix.
- [x] `Devices_CanHandleFocusChanges` (parametrized variants) still pass — no manual-update contract changes.
- [ ] Full `Unity.InputSystem.IntegrationTests` assembly via Yamato.
- [ ] Editor functional tests via Yamato (Windows + macOS + Linux).
- [ ] Manual walkthrough of the Jira repro scripts (LMB-in-FixedUpdate logger and RMB freeze script).

## Branch history

The branch keeps Option B (`OnUpdate`-based restore) committed and reverted (`12a5921f7` + `209331a5b`) so reviewers can compare both approaches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
@@ -148,32 +148,62 @@ namespace UnityEngine.InputSystem.LowLevel
                 // TODO move it to a proper native callback instead
                 if (value != null)
                 {
-                    // Inject ourselves directly to PlayerLoop.Initialization as first subsystem to run,
-                    // Use InputSystemPlayerLoopRunnerInitializationSystem as system type
-                    var playerLoop = UnityEngine.LowLevel.PlayerLoop.GetCurrentPlayerLoop();
-                    var initStepIndex = playerLoop.subSystemList.IndexOf(x => x.type == typeof(PlayerLoop.Initialization));
-                    if (initStepIndex >= 0)
-                    {
-                        var systems = playerLoop.subSystemList[initStepIndex].subSystemList;
+                    EnsurePlayerLoopInjection();
 
-                        // Check if we're not already injected
-                        if (!systems.Select(x => x.type)
-                            .Contains(typeof(InputSystemPlayerLoopRunnerInitializationSystem)))
-                        {
-                            ArrayHelpers.InsertAt(ref systems, 0, new UnityEngine.LowLevel.PlayerLoopSystem
-                            {
-                                type = typeof(InputSystemPlayerLoopRunnerInitializationSystem),
-                                updateDelegate = () => m_PlayerLoopInitialization?.Invoke()
-                            });
-
-                            playerLoop.subSystemList[initStepIndex].subSystemList = systems;
-                            UnityEngine.LowLevel.PlayerLoop.SetPlayerLoop(playerLoop);
-                        }
-                    }
+                    // Watchdog for UUM-140343: user code calling
+                    // PlayerLoop.SetPlayerLoop(GetDefaultPlayerLoop()) wipes our injection,
+                    // and Unity provides no callback for SetPlayerLoop, no engine API to add to
+                    // the default loop, and no per-frame hook inside the player loop that we
+                    // could anchor to (anything in the loop can be wiped the same way).
+                    // EditorApplication.update is invoked by SceneTracker, not by the player loop
+                    // (see Editor/Src/Selection/SceneInspector.cpp), so it survives a user reset
+                    // and lets us re-inject on the next editor tick. The cost is one cheap
+                    // PlayerLoopSystem-array scan per editor tick while playing -- effectively zero
+                    // when nothing's wrong. Outside play mode the early-return below skips it.
+                    UnityEditor.EditorApplication.update -= EnsurePlayerLoopInjectionIfPlaying;
+                    UnityEditor.EditorApplication.update += EnsurePlayerLoopInjectionIfPlaying;
+                }
+                else
+                {
+                    UnityEditor.EditorApplication.update -= EnsurePlayerLoopInjectionIfPlaying;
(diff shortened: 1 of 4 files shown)


Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
```diff
@@ -148,33 +148,63 @@ public Action onPlayerLoopInitialization
...
-                            });
-
-                            playerLoop.subSystemList[initStepIndex].subSystemList = systems;
-                            UnityEngine.LowLevel.PlayerLoop.SetPlayerLoop(playerLoop);
-                        }
-                    }
+                    EnsurePlayerLoopInjection();
+
+                    // Watchdog for UUM-140343: user code calling
+                    // PlayerLoop.SetPlayerLoop(GetDefaultPlayerLoop()) wipes our injection,
```
ekcoh, line 154: Is this only affecting the editor hook? We do a number of player loop hooks currently here: https://github.cds.internal.unity3d.com/unity/unity/blob/trunk/Modules/Input/Private/InputModuleRegistration.cpp#L26-L39

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
```diff
@@ -148,33 +148,63 @@ public Action onPlayerLoopInitialization
...
+            });
+
+            playerLoop.subSystemList[initStepIndex].subSystemList = systems;
+            UnityEngine.LowLevel.PlayerLoop.SetPlayerLoop(playerLoop);
+        }
+
+        // Watchdog tick. Fires from EditorApplication.update (independent of the player loop) and
+        // re-injects only when actually needed. In edit mode we don't care; OnPlayerLoopInitialization
+        // is gated on gameIsPlaying anyway, so a wiped injection while not playing is harmless.
+        private void EnsurePlayerLoopInjectionIfPlaying()
```
ekcoh, line 202: This is a lot of work to do on every tick and currently also triggering LINQ it seems - another idea - let me know what you think, do an active watchdog check instead, e.g. on each editor tick where the player hook is actually called we store the current timestamp `realTImeSinceStartup`, then this tick could just check e.g. `WasPlayerloopHookCalledLastXXX` or similar. That would imply that we never run anything heavier than a single if-statement for the happy case and the loop injection just triggers after XXX passes without an editor cycle executed?

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
LeoUnity: We discussed this offline, but adding here just for context.
GetDefaultPlayerLoop only contains systems registered in Native, the problem is that we have one(?) system today that is registered in managed, so if users SetPlayerLoop using GetDefaultPlayerLoop, those managed registered systems gets removed.
If we were to somehow move all (one? InputSystemPlayerLoopRunnerInitializationSystem) systems registrations to native it would solve the problem.

Packages/com.unity.inputsystem/InputSystem/Runtime/NativeInputRuntime.cs
LeoUnity: Completely agree, I don't think this PR should move forward, it was just to illustrate the issue and a (bad) fix. I don't think we should invest on a watchdog at all, unless there is something I am missing, we should just aim to make GetDefaultPlayerLoop to contain all the hooks we need always.

PR #2426, opened 2026-06-03
Clarify InputBinding.Matches mask semantics

AswinRajGopal:
### Description

 Clarifies the XML documentation for `InputBinding.Matches` to state that it performs mask/filter matching, not equality comparison.
The updated docs call out that a binding may not match itself when one of its mask fields, such as `groups`, contains a value that does not match as a filter. The docs now alsopoint users to `Equals(InputBinding)` / `==` for binding equivalence checks.

### Testing status & QA

Documentation-only change. No tests run.

### Overall Product Risks

- Complexity:
- Halo Effect:

### Comments to reviewers

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputBinding.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputBinding.cs
@@ -780,7 +780,7 @@ namespace UnityEngine.InputSystem
         /// Note that all comparisons are case-insensitive.
         ///
         /// This method doesn't test bindings for equality. A binding may not match itself if one of
-        /// its mask properties contains a value that does not match as a filter. For example, a binding
+        /// its mask properties contains a value that doesn't match as a filter. For example, a binding
         /// with <see cref="groups"/> set to an empty string has a group filter but no non-empty group
         /// to match against. To compare bindings for equivalence, use <see cref="Equals(InputBinding)"/>
         /// or <see cref="operator=="/>.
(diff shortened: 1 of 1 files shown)


Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputBinding.cs
```diff
@@ -779,6 +779,12 @@ internal bool TriggersAction(InputAction action)
         ///
         /// Note that all comparisons are case-insensitive.
         ///
+        /// This method does not test bindings for equality. A binding may not match itself if one of
```
suearkinunity:
```suggestion
        /// This method doesn't test bindings for equality. A binding may not match itself if one of
```

Packages/com.unity.inputsystem/InputSystem/Runtime/Actions/InputBinding.cs
```diff
@@ -779,6 +779,12 @@ internal bool TriggersAction(InputAction action)
...
+        /// its mask properties contains a value that does not match as a filter. For example, a binding
```
suearkinunity:
```suggestion
        /// its mask properties contains a value that doesn't match as a filter. For example, a binding
```

suearkinunity (approved):
Two small suggestions - we use "don't" instead of "does not".

ritamerkl (approved):

PR #2428, opened 2026-06-05
FIX: OnMouseUpAsButton dropped in Play mode on Game view focus change

LeoUnity:
PR #93985 — input/isx-2424/add-focus-event-to-module introduced a regression to the MonoBehaviour.OnMouseUp due to the UNITY_INPUTSYSTEM_SUPPORTS_MOUSE_SCRIPT_EVENTS not being defined in the test assembly. This PR addresses both the regression and ensuring the tests are now running.

## Fix: legacy mouse-event pipeline driven from editor updates

Focus events (engine >= 6000.5.0a8) cause an `InputUpdateType.Editor` update to run in
Play mode when the Game view focus changes. `FinalizeUpdate` drove the legacy
`SendMouseEvents` pipeline from that editor update, where `Pointer.current` reflects the
editor state buffer (position (0,0), not pressed). That produced a spurious mouse-up
that cleared the press target, so `OnMouseUpAsButton`/`OnMouseUp` were skipped on the
real release. Guard the legacy mouse send so it does not run for `InputUpdateType.Editor`
updates (consistent with existing guards in this file). Focus-event delivery is
unaffected. Covered by `CoreTests.MouseEvents_CanReceiveOnMouseUpAsButton`.

## Fix: MouseEvents tests were silently excluded from CI

`UNITY_INPUTSYSTEM_SUPPORTS_MOUSE_SCRIPT_EVENTS` was declared as a `versionDefine` only
on the runtime asmdef (`Unity.InputSystem.asmdef`). `versionDefines` are scoped to the
declaring asmdef and do not propagate to referencing assemblies, so the symbol was
always undefined in the test assembly. The entire `CoreTests_MouseEvents.cs` file is
gated by `#if UNITY_INPUTSYSTEM_SUPPORTS_MOUSE_SCRIPT_EVENTS` — which meant every
`MouseEvents_CanReceive*` test (including the regression test that catches the bug
above) silently compiled out and never ran in CI on any editor version. This PR adds
the same versionDefine to `Unity.InputSystem.Tests.asmdef`, mirroring how the sibling
`UNITY_INPUTSYSTEM_SUPPORTS_FOCUS_EVENTS` define is already replicated across both
asmdefs.

## Commits
1. `FIX: OnMouseUpAsButton dropped in Play mode on Game view focus change`
2. `FIX: Add UNITY_INPUTSYSTEM_SUPPORTS_MOUSE_SCRIPT_EVENTS to tests asmdef so MouseEvents tests compile`


VeraMommersteeg (approved):
looks like a simple fix to prevent it from reading from the wrong buffer. if the tests are all passing, then it lgtm. The only comment I would add is do we now have a test to cover this case so we catch this next time?

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - Fixed `InputAction.IsPressed`, `WasPressedThisFrame`, and `WasReleasedThisFrame` using a `ButtonControl`'s `pressPoint` when a binding also had an explicit `PressInteraction` with its own `pressPoint`, which could make those APIs disagree with the interaction's press and release behavior. Action-level press APIs now follow the interaction threshold when both are set explicitly.
 - Fixed `IndexOutOfRangeException` in `InputDeviceBuilder` when connecting an HID gamepad whose report descriptor declares a hat switch with Report Size 8 (e.g. ESP32-BLE-Gamepad). The HID layer now anchors the hat's directional sub-controls to the hat's own byte instead of letting the layout system auto-allocate a fresh byte for each [UUM-143659](https://jira.unity3d.com/browse/UUM-143659).
+- Fixed `OnMouseUpAsButton` and `OnMouseUp` being dropped in Play mode when the Game view's focus changes between a press and its release on Unity 6000.5.0a8 and newer. The legacy `SendMouseEvents` pipeline is no longer driven from `InputUpdateType.Editor` updates, which read the editor state buffer (position (0,0), not pressed) and produced a spurious mouse release that cleared the press target.
```
Pauliusd01, line 35: A changelog entry is not needed as this was introduced in development and fixed in development

Pauliusd01 (approved):
Seems simple, checked onMouse test scene as well as other mouse related ones in playmode and player and they work just fine

PR #2431, opened 2026-06-10
FIX: Touchscreen losing focus then input fails on UI afterwards

Darren-Kelly-Unity:
## Purpose of this PR

Fixes a bug where UI Toolkit elements become unresponsive to touchscreen input after an Alt+Tab cycle on Windows. The root cause was that `InputSystemProvider.OnFocusChanged()` did not reset pointer state on focus loss, leaving `PointerState.ClickCount` (and other fields) stale across the focus boundary. The fix assigns `= default` to all three pointer states in `OnFocusChanged(false)`, which zeroes every field including `ClickCount` — a full reset that `.Reset()` alone does not provide.

## Release Notes

- **[Input System]** Fixed touchscreen UI Toolkit elements (sliders, radio buttons, etc.) becoming unresponsive after Alt+Tabbing away from and back to a Windows player build.

## Functional Testing status

- Automated: new regression test `AfterAltTab_FirstTouchPress_GeneratesCleanButtonPressedEvent` in `InputForUIFocusRegressionTests` — fails on the unfixed code (clickCount=2), passes after the fix (clickCount=1).
- Manual: not possible to automate the full OS focus cycle; verified via the test that the fix exercises the correct code path (`m_Provider.OnFocusChanged` is called directly since `ScheduleFocusChangedEvent` drives only the InputManager `BackgroundBehavior` path, not the `IEventProviderImpl` path).
- Tested background behavior: `ResetAndDisableNonBackgroundDevices`.

## Performance Testing Status

No performance impact. The change adds a small number of field assignments inside `OnFocusChanged`, which is called at most twice per Alt+Tab cycle.

## Overall Product Risks

- **Technical Risk: 1** — Change is isolated to `OnFocusChanged`, a method that only fires on application focus transitions. No hot-path code affected.
- **Halo Effect: 1** — Resetting pointer state on focus loss could affect pen/mouse event handling across an Alt+Tab boundary, but this matches the documented intent and mirrors how `InputManagerProvider` already behaves (it polls fresh state every frame so stale state is never possible there).

## Class diagram

Analyzed range: `origin/main...HEAD`

<details>
<summary>InputSystemProvider — OnFocusChanged pointer-state reset</summary>

```mermaid
classDiagram
    class IEventProviderImpl {
        <<interface>>
    }

    class PointerState {
    }

    class InputSystemProvider {
        -PointerState m_MouseState
        -PointerState m_TouchState
        -PointerState m_PenState
        +OnFocusChanged(bool focus)
    }

    InputSystemProvider ..|> IEventProviderImpl
    InputSystemProvider "1" *-- "3" PointerState

    note for InputSystemProvider "OnFocusChanged now resets\nm_MouseState, m_TouchState,\nand m_PenState on focus loss"
```

</details>

### DIFF
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
@@ -313,6 +313,7 @@ namespace UnityEngine.InputSystem.Plugins.InputForUI
                 m_SeenTouchEvents = false;
                 m_SeenPenEvents = false;
                 m_ResetSeenEventsOnUpdate = false;
+                m_RepeatHelper.Reset();
                 m_Events.Clear();
             }
         }
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
@@ -284,13 +284,32 @@ namespace UnityEngine.InputSystem.Plugins.InputForUI
 
             if (!focus)
             {
-                // Replace state structs with default instances rather than calling Reset(),
-                // because Reset() preserves ClickCount (used for double-tap tracking within a
-                // session) but on focus loss we want a completely clean slate — stale ClickCount
-                // causes the first tap after Alt+Tab to be misidentified as a double/triple tap.
+                // Preserve the last known cursor/pen positions before resetting.
+                // Mouse and pen positions are needed so that a click-to-refocus
+                // (where the pointer hasn't moved and OnPointerPerformed won't fire)
+                // still dispatches to the correct screen location via OnClickPerformed.
+                // Touch positions are NOT preserved — every new touch always arrives
+                // with a fresh position from OnPointerPerformed.
+                var mouseLastValid   = m_MouseState.LastPositionValid;
+                var mouseLastPos     = m_MouseState.LastPosition;
+                var mouseLastDisplay = m_MouseState.LastDisplayIndex;
+
+                var penLastValid     = m_PenState.LastPositionValid;
+                var penLastPos       = m_PenState.LastPosition;
+                var penLastDisplay   = m_PenState.LastDisplayIndex;
+
+                // Use default assignment rather than Reset() because Reset() intentionally
+                // preserves ClickCount for within-session double-tap tracking.
+                // After a focus loss that ClickCount is stale and must be zeroed.
                 m_MouseState = default;
                 m_TouchState = default;
                 m_PenState = default;
+
+                if (mouseLastValid)
+                    m_MouseState.OnMove(m_CurrentTime, mouseLastPos, mouseLastDisplay);
+                if (penLastValid)
+                    m_PenState.OnMove(m_CurrentTime, penLastPos, penLastDisplay);
+
                 m_SeenTouchEvents = false;
                 m_SeenPenEvents = false;
                 m_ResetSeenEventsOnUpdate = false;
--- a/Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
+++ b/Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
@@ -312,7 +312,6 @@ namespace UnityEngine.InputSystem.Plugins.InputForUI
 
                 ResetSeenEvents();
                 m_ResetSeenEventsOnUpdate = false;
-                m_RepeatHelper.Reset();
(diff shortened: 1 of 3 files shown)


Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -281,6 +281,21 @@ static int SortEvents(Event a, Event b)
...
         {
             m_InputEventPartialProvider.OnFocusChanged(focus);
+
+            if (!focus)
+            {
+                // Replace state structs with default instances rather than calling Reset(),
+                // because Reset() preserves ClickCount (used for double-tap tracking within a
+                // session) but on focus loss we want a completely clean slate — stale ClickCount
+                // causes the first tap after Alt+Tab to be misidentified as a double/triple tap.
+                m_MouseState = default;
```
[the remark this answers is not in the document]
Darren-Kelly-Unity: Made changes to solve this.

MorganHoarau:
Your new unit test is failing on CI

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -281,6 +281,41 @@ static int SortEvents(Event a, Event b)
...
+                m_TouchState = default;
+                m_PenState = default;
+
+                if (mouseLastValid)
+                    m_MouseState.OnMove(m_CurrentTime, mouseLastPos, mouseLastDisplay);
+                if (penLastValid)
+                    m_PenState.OnMove(m_CurrentTime, penLastPos, penLastDisplay);
+
+                m_SeenTouchEvents = false;
```
MorganHoarau: Instead of resetting both `m_SeendTouchEvents` and `m_SeenPenEvents` here, you should use existing [ResetSeenEvents](https://github.com/Unity-Technologies/InputSystem/blob/431fe27e4b07bb9a33d6306c8348280a43932032/Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs#L176)

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
```diff
@@ -281,6 +281,41 @@ static int SortEvents(Event a, Event b)
...
+                m_SeenPenEvents = false;
+                m_ResetSeenEventsOnUpdate = false;
+                m_RepeatHelper.Reset();
```
MorganHoarau: Why do you reset navigation repeat state here? Looks unrelated to pointer clickCount bug and no mentioned in PR description. Is there a separate bug for it maybe?

MorganHoarau (approved):
PR description is different from actual diff after you applied bot suggestion. Please update accordingly.

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
Darren-Kelly-Unity: It was a suggestion that seemed to make sense from u-pr below!
https://github.com/Unity-Technologies/InputSystem/pull/2431/changes#r3388665182

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
Darren-Kelly-Unity: Done!

Packages/com.unity.inputsystem/InputSystem/Runtime/Plugins/InputForUI/InputSystemProvider.cs
MorganHoarau: If that was not a known issue, we should at least test it first before applying to know if there is an issue in the first place. It doesn't look like it could introduce any regression, but I'm just considering that any line of code not part of fixing existing a bug might be a potential liability for subtle regression. 

I haven't tested locally so I'll let QA confirm

Darren-Kelly-Unity: reverted it for now!

Pauliusd01:
Testing this one today

Pauliusd01 (approved):
Sorry for taking a while, I was battling my cheap touchscreen monitor and finally got it to work in 480p. I've checked the /test_plan tests and the original bug repro with a windows touch screen, mouse and pen device.

Test scene can be found here if you ever want to double check:
[PR_2431_V2.zip](https://github.com/user-attachments/files/29001009/PR_2431_V2.zip)

PR #2438, opened 2026-06-15
FIX: Input devices being lost up on upgrade to the new input version [ISX-2569]

K-Tone:
### Description

So the problem we're trying to fix here is that when you have a working project with pre-FEPM InputSystem and you then update the input package to the one that has the FEPM change, all input comes broken. The root problem appeared that all devices were lost during upgrade. 

The reason devices got lost during upgrade is somewhat tricky to explain, but it all comes down to the fact that native will only communicate the current devices once per Unity process. If you lose devices once, it's permanent - entering playmode doesn't list devices again. 

The way we lost devices was due to us renaming an in-memory scriptable object class that was used to keep state from InputSystemObject to InputSystemStateManager. Up on upgrade recompilation, Unity realised that the class for the InputSystemObject SO is no longer available so it shall be dropped, and all the devices died with it. 

Hereby, we bring back the old class with the old meta info so that the forward upgrade can work for now. 

### Testing status & QA

Tried locally. 

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fixed input devices being lost (no keyboard, mouse, gamepad, etc.) after upgrading the package while the Editor is open (related to the recent fast enter playmode change). [ISX-2569]
 - Fixed `InputSystemProvider` disabling project-wide actions on shutdown when UI Toolkit destroys its objects mid-play. The provider now scopes its lifecycle to the UI action map only and does not disable project-wide actions [UUM-134130](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-134130)
 - Fixed `InputActionRebindingExtensions.GetBindingDisplayString(InputAction, InputBinding, ...)` returning an empty string for composite bindings when the binding mask filters by group [UUM-141423](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-141423)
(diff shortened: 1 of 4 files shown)


ekcoh (approved):
Not sure I fully understand this fix but looks like fixing broken serialisation. The key takeaway here IMO is that this (not this PR; but how this work) is the wrong solution to the problem. I will link this issue/PR as a motivational case for why device ownership, enumeration, reconnects should live in the module (native) and not in the package. This would make this + any serialisation push/pop disappear from the package. Good fix!

K-Tone:
Thanks for the input Hakan! I would have totally loved if we could poll data from native at certain points instead of messing with ScriptableObjects. We do know keeping data is needed, just that if we keep copies in managed that appears to cost much more. I would have totally loved trying to minimise our managed state that needs to survive domain reload.

Pauliusd01:
Just leaving my status that I am currently busy with our module port, it is unlikely to be checked this week

K-Tone:
Simply to have it said somewhere, I'm on a PTO in July so if it's not this week then either someone else will have to address feedback on this PR or still myself but in August.

Pauliusd01:
I'll give it a go today

Pauliusd01 (approved):
LGTM, checked upgrading, reimporting and did some basic fast enter play mode rechecks.

PR #2460, opened 2026-07-29
CHANGE: docs tools update

ritamerkl:
### Description

Upgrading docs package tools. DocFX is no longer part of the package and needs to be installed separately. 
Also the .csproj file creation is not handled by the docs package anymore, so I included a command that creates the project files via the rider package. 

Additionally I added a small format fix in [‎Packages/com.unity.inputsystem/Documentation~/built-in-interactions.md‎](https://github.com/Unity-Technologies/InputSystem/pull/2460/changes#diff-b7dae7bdaf4c564b4323b0af6e3a9a25d4ab1b4b9bc5bd7dfe19e89865bebd05). 

### Testing status & QA

None.

### Overall Product Risks

- Complexity: Low
- Halo Effect: Low

### Checklist

Before review:

- [ ] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Assets/Tests/InputSystem/DocumentationBasedAPIVerficationTests.cs
+++ b/Assets/Tests/InputSystem/DocumentationBasedAPIVerficationTests.cs
@@ -48,9 +48,7 @@ class DocumentationBasedAPIVerficationTests
         // On CI (fresh clone, no IDE installed), SyncAll() is a no-op and no .sln/.csproj files
         // exist yet. Force solution generation via reflection since SyncVS is internal in Unity 6.x.
         var syncVsType = Type.GetType("UnityEditor.SyncVS, UnityEditor");
-        syncVsType?.GetMethod("SyncSolution",
-            BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static)
-            ?.Invoke(null, null);
+        syncVsType?.GetMethod("SyncSolution", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static)?.Invoke(null, null);
 
 #if HAVE_DOCTOOLS_INSTALLED
         (_documentationBuilderLogs, _docsFolder) = Documentation.Instance.GenerateEx(inputSystemPackageInfo, InputSystem.version.ToString(), docsPath);
--- a/Assets/Tests/InputSystem/DocumentationBasedAPIVerficationTests.cs
+++ b/Assets/Tests/InputSystem/DocumentationBasedAPIVerficationTests.cs
@@ -3,7 +3,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Text;
 using HtmlAgilityPack;
 using Mono.Cecil;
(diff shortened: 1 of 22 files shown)


Tools/CI/Settings/InputSystemSettings.cs
```diff
@@ -21,9 +21,37 @@ public class InputSystemSettings : AnnotatedSettingsBase
...
+    // So no extra .NET SDK install step is needed here.
+    //
+    // NuGet source reachability - `dotnet tool install` needs to resolve the docfx package from a feed.
+    // There's no nuget.config at the repo root (only Tools/CI/nuget.config, which NuGet won't discover
+    // from here since it only walks upward from the working directory), so we pin --add-source
+    // explicitly below to Unity's internal Artifactory NuGet proxy - the same source Tools/CI/nuget.config
+    // uses, and one we know CI agents can already reach since the recipe-regeneration job restores
+    // packages through it. Default sources (nuget.org) are likely unreachable from these locked-down
+    // build agents. Still worth confirming on the first real CI run that Artifactory actually mirrors
+    // the "docfx" package specifically (vs. only packages requested before).
```
jfreire-unity: Do we need all this comments? Some parts read as something to be confirmed

jfreire-unity (commented):
The change looks like it upgrades the PMDT but CI still fails, so we should circle back to the docs engineering team and let us know it's not fixing the problem.

Also, have you tried locally if this change generates the documentation?

Tools/CI/Settings/InputSystemSettings.cs
ritamerkl: Addressed in [shortened comments](https://github.com/Unity-Technologies/InputSystem/pull/2460/commits/e8abed85d5cc4181e5056201a4811e42c7cc70bc)

Assets/Tests/InputSystem/DocumentationBasedAPIVerficationTests.cs
```diff
@@ -44,6 +44,56 @@ public void OneTimeSetup()
...
+                csprojContent.AppendLine("  <ItemGroup>");
+                foreach (var refPath in asm.compiledAssemblyReferences)
+                    csprojContent.AppendLine(
+                        $"    <Reference Include=\"{Path.GetFileNameWithoutExtension(refPath)}\">" +
+                        $"<HintPath>{refPath}</HintPath></Reference>");
+                csprojContent.AppendLine("  </ItemGroup>");
+                csprojContent.AppendLine("</Project>");
+                File.WriteAllText(csprojPath, csprojContent.ToString(), Encoding.UTF8);
+            }
+        }
```
jfreire-unity: Is this strictly necessary? Is there no better way to avoid this or generate this XML without essentially concatenating strings of XML together? I'd report this to the docs engineering team if this is the only because it feels bad one if one has to do this to make docs work on CI.

ritamerkl: To my knowledge state it is necessary because the .csproj file is needed for the doc tooling in version 3x. It's generated by code for two main reasons:  
1. DefineConstants — includes Unity version defines (UNITY_6000_0_52),        
  platform defines (PLATFORM_STANDALONE_OSX), and project-specific feature      
  flags. These change per Unity version, OS, and project. DocFX needs accurate  
  defines to parse #if blocks correctly, so hardcoding them would cause wrong   
  API members to be included/excluded.
  2. Reference HintPaths (lines 87-90) — the compiled assembly references use
  absolute paths to the machine's Library/ScriptAssemblies/ folder. Inherently  
  machine-specific.

ritamerkl: solved in [Added .csproj creation into CI flow](https://github.com/Unity-Technologies/InputSystem/pull/2460/commits/362daf7907f384b117822f88a3ae1f9c50ed9535)

jfreire-unity (approved):
Looks good! Good to know CI gets fixed by updating the docs tooling.
Don't know if u-pr P2 concern makes sense, I'll leave it for you to decide!

PR #2465, opened 2026-08-05
FIX: Input Actions editor window throws a "Failed to load asset" exception on startup when its layout is restored in a different project [UUM-144318]

Pauliusd01:
### Description

> [!NOTE]
> This pull request was generated automatically. Please review carefully before merging.

`InputActionsEditorWindow.CreateGUI()` persists its target asset's GUID (`m_AssetGUID`) in the editor window layout, which is shared across projects. When the window's layout is restored in a project where that GUID resolves to no asset, the restore branch loads a null asset and throws, and the surrounding `catch` surfaces it via `Debug.LogException` on editor startup.

The `asset == null` branch now closes the window quietly instead of throwing. The close is scheduled through `EditorApplication.delayCall` rather than called inline: `CreateGUI()` runs inside the UIToolkit repaint, and closing the window synchronously there tears down its panel mid-repaint, which then produces a follow-up `NullReferenceException` in the render tree. Deferring the close lets `CreateGUI()` return cleanly and the window closes on the next editor tick.

### Testing status & QA

- No automated test added - the branch only runs on an editor domain-reload window-layout restore where the persisted asset GUID does not resolve in the current project, which `InputTestFixture` / edit-mode tests cannot reproduce.
- QA can re-run the manual repro from https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-144318 and confirm no exception is logged on startup.

### Overall Product Risks

- Complexity: low
- Halo Effect: low (internal-only edit confined to one branch of a private method; no public API surface touched)
- Risk rating: 3/5 - The changed asset==null branch runs only on a real domain-reload/cross-project restore, which the edit-mode tests never reproduce, so the specific behaviour is untested and a regression could land uncaught despite the delta being confined to suppressing one log.

### Comments to reviewers

The window shows an empty panel for a single editor tick before `delayCall` closes it; this is expected given the close is deferred out of the repaint. The sibling `catch` block a few lines below still calls `Close()` synchronously for genuinely unexpected restore exceptions and has the same latent teardown risk, but it was left unchanged to keep this fix scoped to the reported scenario.

Note: if fixing it this way is undesirable then maybe changing the type of the message from error to a warning would be enough?

### Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.

### DIFF
--- a/Packages/com.unity.inputsystem/CHANGELOG.md
+++ b/Packages/com.unity.inputsystem/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fixed the Input Actions editor window logging a "Failed to load asset" exception on editor startup when its saved window layout was restored in a project where the referenced asset GUID did not resolve; the window now closes quietly instead [UUM-144318](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-144318)
 - Empty foldouts are no longer shown for processors and interactions that have no settings (e.g. "Invert") [UUM-144325](https://issuetracker.unity3d.com/product/unity/issues/guid/UUM-144325)
 - Fixed the Input Actions editor toolbar buttons being clipped when the Action Properties panel grew tall enough to show a scrollbar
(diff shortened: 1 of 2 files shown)


MorganHoarau (approved):
Safe and simple. Comment is a bit verbose. Discussed on slack.

PR #2469, opened 2026-08-07
NEW: Add LocationSensor for Input System

MorganHoarau:
## Description

Addresses [ISX-2227](https://jira.unity3d.com/browse/ISX-2227) is part of the Input Manager parity epic [ISX-2108](https://jira.unity3d.com/browse/ISX-2108).

This PR is the managed layer only. Refer to the native bridge PR [here](https://github.cds.internal.unity3d.com/unity/unity/pull/118994).

Add a `LocationSensor` device that exposes the device's GPS and lifecycle through the standard Input System surface:

- Readings as event-driven controls: `latitude`, `longitude`, `altitude`, `horizontalAccuracy`, `verticalAccuracy`, `timestamp`.
- Lifecycle via `InputSystem.EnableDevice` / `DisableDevice` + `status` and `isEnabledByUser` as query properties.
- `Configure(desiredAccuracyInMeters, updateDistanceInMeters)` + `ResetConfiguration()`, with project defaults `InputSettings.locationAccuracy` / `locationDistanceThreshold` (both 10, matching legacy `Start()`), surfaced in Project Settings.

Backed by the shared native `LocationService`.

<details>
<summary>Class diagram: LocationSensor and command surface</summary>

```mermaid
classDiagram
    class Sensor
    class LocationSensor {
        +AxisControl latitude
        +AxisControl longitude
        +AxisControl altitude
        +AxisControl horizontalAccuracy
        +AxisControl verticalAccuracy
        +DoubleControl timestamp
        +LocationServiceStatus status
        +bool isEnabledByUser
        +static LocationSensor current
        +Configure(float, float)
        +ResetConfiguration()
    }
    class LocationState {
        <<internal, IInputStateTypeInfo>>
        +double timestamp
        +float latitude
        +float longitude
        +float altitude
        +float horizontalAccuracy
        +float verticalAccuracy
    }
    class InputSettings {
        +float locationAccuracy
        +float locationDistanceThreshold
    }
    class QueryLocationStatusCommand
    class QueryLocationEnabledByUserCommand
    class ConfigureLocationCommand

    Sensor <|-- LocationSensor
    LocationSensor ..> LocationState : stateType
    LocationSensor ..> QueryLocationStatusCommand : status (LSTA)
    LocationSensor ..> QueryLocationEnabledByUserCommand : isEnabledByUser (LUSR)
    LocationSensor ..> ConfigureLocationCommand : Configure (LCFG)
    LocationSensor ..> InputSettings : ResetConfiguration reads defaults
```
</details>

### Documentation Impact

- Public API added (`LocationSensor` + controls + `status` / `isEnabledByUser` / `Configure` / `ResetConfiguration`, `InputSettings.locationAccuracy` / `locationDistanceThreshold`, three `public` command structs). All carry XML docs.
- Migration table updated: `Documentation~/corresponding-old-new-api.md` now maps `Input.location` -> `LocationSensor` with a usage snippet.

## Testing status & QA

ℹ️ **PLEASE REFER TO**: https://github.cds.internal.unity3d.com/morgan-hoarau/internal-input-manual-test-projects/pull/2

* 6 unit tests added (`CoreTests_Devices.cs`): reading, `status`, `isEnabledByUser`, fallback path (no native impl), `Configure`, `ResetConfiguration`. 
* On-device (via the InputFeatureParity Validation project): 
  * Tested on Android, iOS and Web (Windows, Mac, Android and iOS) in both input backend "Both" and "Input System (new)".

## Overall Product Risks

* Complexity: 1
  * Additive, self-contained device; fallback to `Stopped` / `false` where no native impl exists (editor/desktop), so no regression to existing sensors. 
* Halo Effect: 1
  * Touches shared `InputSettings` (two new serialized fields + settings UI). 

## Comments to reviewers

## Checklist

Before review:

- [x] Changelog entry added.
    - Performance tests.
    - Integration tests.
    - Usage code examples added.

During merge:

    - `NEW: ___`.
    - `FIX: ___`.
    - `DOCS: ___`.
    - `CHANGE: ___`.
    - `RELEASE: 1.1.0-preview.3`.


Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Sensor.cs
```diff
@@ -60,6 +60,30 @@ internal struct LinearAccelerationState : IInputStateTypeInfo
...
     }
+
+    /// <summary>
+    /// Low-level input state for <see cref="LocationSensor"/>.
+    /// </summary>
+    internal struct LocationState : IInputStateTypeInfo
+    {
+        public static FourCC kFormat => new FourCC('L', 'O', 'C', ' ');
+
+        // Order matches native LocationInfo (timestamp first). Do not reorder.
```
MorganHoarau:
```suggestion
        // Order matches native LocationInfo. Do not reorder.
```

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
```diff
@@ -115,5 +115,5 @@ Note: [`UnityEngine.TouchScreenKeyboard`](https://docs.unity3d.com/ScriptReferen
...
-[`Input.location`](https://docs.unity3d.com/ScriptReference/Input-location.html)|No corresponding API yet.
+[`Input.location`](https://docs.unity3d.com/ScriptReference/Input-location.html)|[`LocationSensor`](xref:UnityEngine.InputSystem.LocationSensor).<br/>Start/Stop:<br/>`InputSystem.EnableDevice(LocationSensor.current);`<br/>`InputSystem.DisableDevice(LocationSensor.current);`<br/>Read the last fix from the controls, for example `LocationSensor.current.latitude.ReadValue()` (also `longitude`, `altitude`, `horizontalAccuracy`, `verticalAccuracy`, `timestamp`).<br/>Query lifecycle and permission with [`LocationSensor.status`](xref:UnityEngine.InputSystem.LocationSensor) and [`LocationSensor.isEnabledByUser`](xref:UnityEngine.InputSystem.LocationSensor).<br/>Set the requested accuracy and update-distance with `LocationSensor.current.Configure(accuracyInMeters, updateDistanceInMeters)`, or `ResetConfiguration()` to revert to the project defaults [`InputSettings.locationAccuracy`](xref:UnityEngine.InputSystem.InputSettings) and [`InputSettings.locationDistanceThreshold`](xref:UnityEngine.InputSystem.InputSettings).
```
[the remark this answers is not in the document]
MorganHoarau: fixed

ekcoh (commented):
Only glanced at this PR so far but my main reaction is that Command types are defined in package instead of being internal types within module. I would strongly suggest placing them in module instead. If there are issues with that which I have overlooked I am happy to discuss to find a good way forward.

Exposing them like this basically makes internal ABI part of the API which is something I think we would benefit to move away from

MorganHoarau:
Made the 3 location command structs internal, so the native ABI is no longer public API.

However, moving them into the module is blocked by dependency as they're built on `InputDeviceCommand` / `InputDeviceCommandInfo` / `FourCC`, which live in the package, and the module can't reference the package. 

It's the same for every command struct. So I'd let that module migration as a separate initiative.

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
MorganHoarau, line 1: FYI @suearkinunity

Pauliusd01 (approved):
I've only been asked to review the testing already done by Morgan, I've looked at his project before and the testing coverage seemed sufficient to me

todi1856:
Could you fix the link in the description "This PR is the managed layer only. Refer to the native bridge PR here." "here" doesn't point to anything ?

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Sensor.cs
```diff
@@ -694,4 +718,188 @@ protected override void FinishSetup()
             base.FinishSetup();
         }
     }
+
+    /// <summary>
+    /// Status of the location service backing a <see cref="LocationSensor"/>.
+    /// </summary>
+    public enum LocationServiceStatus
```
todi1856, line 725: Are we fine that this name will collide with https://docs.unity3d.com/6000.1/Documentation/ScriptReference/LocationServiceStatus.html ?

Because it's common that in scripts we have using UnityEngine;

Thus most people will hit this name collision.

It's minor, but curious about your opinion?

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSettings.cs
```diff
@@ -466,6 +466,48 @@ public float multiTapDelayTime
...
+        public float locationAccuracy
+        {
+            get => m_LocationAccuracy;
+            set
+            {
+                // ReSharper disable once CompareOfFloatsByEqualityOperator
+                if (m_LocationAccuracy == value)
+                    return;
+                m_LocationAccuracy = Mathf.Max(0f, value);
+                OnChange();
```
todi1856, line 486: Is this API callable at runtime? if so, who's reponsible for calling 
```
            var command = ConfigureLocationCommand.Create(desiredAccuracyInMeters, updateDistanceInMeters);
            ExecuteCommand(ref command);
```

to update this settings on the native side ?

the same for locationDistanceThreshold ?

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Sensor.cs
```diff
@@ -694,4 +718,188 @@ protected override void FinishSetup()
...
+        /// <param name="desiredAccuracyInMeters">Desired horizontal accuracy, in meters.</param>
+        /// <param name="updateDistanceInMeters">Minimum distance the device must move before a new reading is reported, in meters.</param>
+        /// <remarks>
+        /// By default, the sensor is configured with the values from <see cref="InputSettings.locationAccuracy"/>
+        /// and <see cref="InputSettings.locationDistanceThreshold"/>. <see cref="ResetConfiguration"/> returns to those defaults.
+        ///
+        /// If the sensor is already enabled, readings may briefly pause while they are applied.
+        /// If the sensor is disabled, values apply when the device is enabled.
+        /// </remarks>
+        public void Configure(float desiredAccuracyInMeters, float updateDistanceInMeters)
```
ekcoh, line 890: Related to a comment on the native PR: on iOS the ENBL and LCFG paths end up constructing a `CLLocationManager`, which Core Location requires to happen on a thread with an active run loop. Off the main thread the delegate callbacks never fire, and `isEnabledByUser` gets stuck at `false` permanently with nothing to indicate why.

InputSystem.cs:49 already says the API is main-thread-only with a few listed exceptions, so this isn't a new rule — but LocationSensor is stricter than the native contract it sits on (InputDeviceIOCTL explicitly permits parallel execution for the same callback), and right now nothing on either side records that.

Could `Configure`, `ResetConfiguration`, `status` and `isEnabledByUser` document the requirement on their `<remarks>`, and assert main-thread before issuing the command? A failure here is silent and permanent, which is the case where an assert earns its keep — the alternative is a user reporting "location never works" with no way to trace it back.

ekcoh (commented):
Reviewed alongside the native PR. The managed side is in good shape — introducing `InputSystem.LocationServiceStatus` to break the legacy module dependency was the right call, and the command structs and test coverage for the query/configure paths are solid. Thanks for making the commands internal in 2a473d8, that matches the sibling sensor states.

On my earlier point about interop types, the visibility half landed and thanks for that. On placement you're right that the command structs can't move: they embed `InputDeviceCommand`, implement `IInputDeviceCommandInfo` and expose `FourCC`, and module-to-package is the wrong direction. Same for `LocationState` with `[InputControl]` and `IInputStateTypeInfo`. I'll drop that part.

But I'd separate the envelope from the field types inside it. `LocationServiceStatus` (enum) has no package types in it at all, and the same now exists in Runtime/Input/LocationService.h, in UnityEngine.LocationServiceStatus, and here, held together only by comments. A package struct can take a module-owned field type, and ISX already consumes  several runtime files from module. So `QueryLocationStatusCommand` could stay put with its status field typed as a module-owned internal enum, and native could `static_assert` its half in the associated sensor device translation unit.

I realise that rubs against the "only NativeInputRuntime.cs should reference UnityEngineInternal.Input" convention, so it's worth a conversation rather than a drive-by change. The reason I keep pushing is that this enum is the one place where drift is silent. A wrong `FourCC` or struct size fails immediately and loudly, whereas a status enum that drifts just reports Running when the service actually failed.

A few smaller notes:
- `locationAccuracy` / `locationDistanceThreshold` are only pushed to native from `OnAdded`. Changing either afterwards has no effect until someone calls `ResetConfiguration()` by hand. Should `OnChange` re-push to the device?
- Nothing covers enable/disable mapping to `ENBL`/`DSBL`, which is the path that actually starts and stops the service. Worth a test given the native side keys its own enabled state off it.
- `status` returning `Stopped` when the command fails conflates "no native backend" with "service is stopped". It's called out in the comment, just noting it's observable from the public API.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Sensor.cs
MorganHoarau: Yeah I had to update the validation project for that so I get what you mean, but IMHO it is fine because this is usually only impacting a few files at best. 
At first I wanted to use the `UnityEngine.LocationServiceStatus` itself, but it is part of IM module... so we cant or we will need to migrate it later when deprecating IM.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSettings.cs
MorganHoarau: I've provided a new `LocationSensor.Configure` that is responsible for calling it. It will restart the service on the spot too. 

It replaces all the overrides of IM `LocationService.Start`. 

Does that sounds reasonable to you?

todi1856: Sorry, I still don't get it a bit, is there a scenario, where me as user could could call setters for InputSettings.locationAccuracy, or InputSettings.locationDistanceThreshold at runtime (on Android device) ? If so, what exactly would that to if I would call these ?

ekcoh:
A follow-up based on my previous feedback as well: currently it seems there is no type dependency from package to module. While this makes CI happy atm it also hides a real dependency. If e.g. the enum would come from module, the package also need a version based asmdef define to enable the code selectively here in this decoupled package. While a bit more work I think that is fine since it makes things more robust when type sits in the correct place. It could of course also be decided to handle it later as part of a bigger refactor for this anti-pattern at a later stage as long as its internal.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSettings.cs
MorganHoarau: No, these values are editor set only, they are used as default value when enabling the Location sensor for the first time (through `Configure` itself)

At this [line](https://github.com/Unity-Technologies/InputSystem/pull/2469/changes#diff-02d5e7f5d7220c6a0276c44cbcb7c20bbc36126206db4c5406f5ae709e147008R893).

todi1856: but this API is not Editor only ? That's why it's confusing... oh well

todi1856 (approved):

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
```diff
@@ -115,5 +115,5 @@ Note: [`UnityEngine.TouchScreenKeyboard`](https://docs.unity3d.com/ScriptReferen
...
...urrent);`<br/>Read the last fix from the controls, for example `LocationSensor.current.latitude.ReadValue()` (also `longitude`, `altitude`, `horizontalAccuracy`, `verticalAccuracy`, `timestamp`).<br/>Query lifecycle and permission with [`LocationSensor.current.status`](xref:UnityEngine.InputSystem.LocationSensor.status) and [`LocationSensor.current.isEnabledByUser`](xref:UnityEngine.InputSystem.LocationSensor.isEnabledByUser).<br/>Set the requested accuracy and update-distance with `LocationSensor.current.Configure(accuracyInMeters, updateDistanceInMeters)`, or `LocationSensor.current.ResetConfiguration()` to revert to the project defaults [`InputSettings.locationAccuracy`](xref:UnityEngine.InputSystem.InputSettings.locationAccuracy) and [`InputSettings.locationDistanceThreshold`](xref:UnityEngine.InputSystem.InputSettings.locationDistanceThreshold).<br/>Note: on Android and iOS the location permission and usage description are auto-added to the build only when your code references `LocationSensor` directly; if you use it solely through an `.inputactions` binding, reference the type in code (for example `LocationSensor.current`) or add the permission and usage description manually.
```
suearkinunity:
```suggestion
[`Input.location`](https://docs.unity3d.com/ScriptReference/Input-location.html)|[`LocationSensor`](xref:UnityEngine.InputSystem.LocationSensor).<br/>Start/Stop:<br/>`InputSystem.EnableDevice(LocationSensor.current);`<br/>`InputSystem.DisableDevice(LocationSensor.current);`<br/>Read the last fix from the controls, for example `LocationSensor.current.latitude.ReadValue()` (also `longitude`, `altitude`, `horizontalAccuracy`, `verticalAccuracy`, `timestamp`).<br/>Query lifecycle and permission with [`LocationSensor.current.status`](xref:UnityEngine.InputSystem.LocationSensor.status) and [`LocationSensor.current.isEnabledByUser`](xref:UnityEngine.InputSystem.LocationSensor.isEnabledByUser).<br/>Set the requested accuracy and update-distance with `LocationSensor.current.Configure(accuracyInMeters, updateDistanceInMeters)`, or `LocationSensor.current.ResetConfiguration()` to revert to the project defaults [`InputSettings.locationAccuracy`](xref:UnityEngine.InputSystem.InputSettings.locationAccuracy) and [`InputSettings.locationDistanceThreshold`](xref:UnityEngine.InputSystem.InputSettings.locationDistanceThreshold).<br/>Note: on Android and iOS the location permission and usage description are auto-added to the build only when your code references `LocationSensor` directly. If you use `LocationSensor` only through an `.inputactions` binding, you can either reference the type in code (for example `LocationSensor.current`) or add the permission and usage description manually.
```

suearkinunity (approved):
I have one small suggestions, just to break down a very long sentence.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSettings.cs
MorganHoarau: This particular API is in the InputSettings scriptable object, which make that data transient at runtime. Can only be edited and save during editor session.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Sensor.cs
MorganHoarau: Fair, I'll add a remark about threading concern for them.

MorganHoarau: Already have the assert for main thread in native.

ekcoh: OK, then just make sure failing that assert results in Console error or managed exception and not termination or undetermined state.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSettings.cs
```diff
@@ -466,6 +466,48 @@ public float multiTapDelayTime
...
+        /// <seealso cref="locationAccuracy"/>
+        public float locationDistanceThreshold
+        {
+            get => m_LocationDistanceThreshold;
+            set
+            {
+                // ReSharper disable once CompareOfFloatsByEqualityOperator
+                if (m_LocationDistanceThreshold == value)
+                    return;
+                m_LocationDistanceThreshold = Mathf.Max(0f, value);
```
ekcoh, line 506: Might have to guard against NaN and Infinity here or should. those be accepted? Not sure what the consequence is...

Packages/com.unity.inputsystem/InputSystem/Runtime/InputSettings.cs
```diff
@@ -466,6 +466,48 @@ public float multiTapDelayTime
...
+        /// <seealso cref="locationDistanceThreshold"/>
+        public float locationAccuracy
+        {
+            get => m_LocationAccuracy;
+            set
+            {
+                // ReSharper disable once CompareOfFloatsByEqualityOperator
+                if (m_LocationAccuracy == value)
+                    return;
+                m_LocationAccuracy = Mathf.Max(0f, value);
```
ekcoh, line 485: Might have to guard against NaN and Infinity here or should. those be accepted? Not sure what the consequence is...

ekcoh (approved):
I believe this PR generally looks good @MorganHoarau!

VeryHappyYoung (approved):

ekcoh (approved):
Approving this and let you judge whether any of the questions or small comments are worth action or not.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Sensor.cs
MorganHoarau: I ended up reverting the constraint on other platform than iOS (as it is the only platform affected). 
Opened a ticket for it (see [thread](https://github.cds.internal.unity3d.com/unity/unity/pull/118994/commits/3b522320d18731cd51b83f61888abd6e067fb44d#r1069621)). 

Also added a remark in managed to warn user about threading affinity.

PR #2475, opened 2026-08-17
DOCS: point backButtonLeavesApp migration at Application (ISX-2229)

m-dejean-unity:
`Input.backButtonLeavesApp` is replaced by `Application.backButtonLeavesApp` in Unity 6.7, so the
migration table no longer says "No corresponding API yet". The row also moves out of "Touch and Pen"
into a new "Application" section, for members of the old `Input` class that control application
behavior rather than reading input, and notes that the `Input` property is deprecated in 6.7 with the
API updater rewriting call sites.

Docs only — the property is engine-side and the package has nothing to add.

Engine PR: `unity/unity#120391` (internal)
Jira: ISX-2229


m-dejean-unity:
Superseded by unity/unity#120783, which makes the same migration-table change on `input/core-package-im-parity-staging`.

Per the delivery shape agreed on 2026-08-19, managed work for this epic lands in the core package inside `unity/unity` rather than here. Reopen if we decide to backport to the package repo — that isn't bound by feature freeze.

PR #2476, opened 2026-08-18
NEW: Pen.isSupported, Mouse.isSupported and Touchscreen.isPressureSupported [ISX-2046, ISX-2079]

ekcoh:
### Description

Legacy `UnityEngine.Input` conflates "is X supported on this platform" with "is an X present right now", and Input System has no equivalent for the first question. `Input.stylusTouchSupported` reports true on any iPad new enough to pair an Apple Pencil, paired or not. `Input.mousePresent` is a hardcoded `true` on Windows, macOS, Linux and WebGL, and genuine detection only on iOS, Android, UWP and the consoles. Windows looks like detection but is not: `TouchPhaseEmulation::GetMousePresent()` is `return true;`.

This adds the capability half.

- `Pen.isSupported`
- `Mouse.isSupported`
- `Touchscreen.isPressureSupported`

For [ISX-2046](https://jira.unity3d.com/browse/ISX-2046) that is two of its three requested mappings: `stylusTouchSupported` → `Pen.isSupported` and `touchPressureSupported` → `Touchscreen.isPressureSupported`. The third, `multiTouchEnabled` → `Touchscreen.multiTouchEnabled`, is **not** here. The ticket notes it is a setting rather than a hardware capability, and suppressing additional touches globally is behavioural work rather than a query, so it wants its own change. The migration doc points users at `primaryTouch` meanwhile, with the two behavioural differences spelled out. `Mouse.isSupported` is the short-term scope of [ISX-2079](https://jira.unity3d.com/browse/ISX-2079), whose remaining half is a real presence primitive rather than a capability one.

The queries are answered by a system endpoint in the engine: a reserved device id that is never registered, so it never appears in `InputSystem.devices` and no real device can answer a question about the platform. Internally, three internal command structs and an internal `InputCapabilitySupport` tristate, dispatched through a new `InputManager.ExecuteSystemCommand`, queried once and cached.

**Presence is a separate question, and it needs both halves.** ISX-2079 is explicit that `Mouse.current != null` is the wrong replacement for `mousePresent`: it tracks the device object's lifecycle, and several platforms register a `Mouse` or `Pen` unconditionally, so it is non-null with no hardware attached. The pattern is:

```csharp
if (Mouse.isSupported) { }                              // could a mouse ever work on this platform
if (Mouse.current != null && Mouse.current.enabled) { } // is one usable right now
```

The pair is asymmetric on purpose. `current != null` goes false when a platform reports removal; `enabled` never reflects a disconnect, it reflects whether the device is switched on. Neither alone is a presence check. Where a platform registers unconditionally and never reports removal this still over-reports, which is what the follow-up issues listed on the engine PR cover. A `device.IsUsable()` extension could hide the pair later without changing the semantics.

> [!IMPORTANT]
> **Depends on an engine change that has not landed yet** (PR #120240, same ticket). Until it does, `UNITY_INPUTSYSTEM_SUPPORTS_CAPABILITY_QUERIES` never fires and everything here compiles out, so **red or empty CI on this PR is expected and is not a defect**. Draft is up to open the design discussion, not to merge first.

### Testing status & QA

Verified on macOS M1 against a locally built `6000.7.0a6` Editor carrying the engine endpoint, which is the only configuration where this code compiles at all.

- **16 new tests, all passing**, three of them the cross-checks below. Each of the three states including `Unknown` for each property; the properties answering with no device added; a query never being delivered to a real device; repeated reads issuing exactly one command; nothing answering reporting `false` rather than throwing.
- **Three cross-check tests** assert the package's hand-written mirrors against the engine's own constants: enum wire values against `CapabilityState`, the FourCC codes against `NativeInputCapabilities`, and the payload size against the single byte the engine's validation accepts. Nothing generates these mirrors, so these are what catch the two repos drifting apart.
- **Regression check:** full `Devices_` category, since `InputManager`, `Pen`, `Mouse` and `Touchscreen` are all touched. 392 tests, 383 passing, 0 failures, 9 pre-existing not-run.

**For QA:** nothing to verify until the engine PR lands. After that, the interesting cases are the platforms where the answer is not obviously true. macOS should report `Pen.isSupported == true`, which contradicts legacy `stylusTouchSupported`, and `isPressureSupported == false`. Windows also reports `isPressureSupported == false`, even though Win32 can supply pressure: Unity's `ProcessPointerTouch` tests `TOUCH_MASK_PRESSURE` against `touchFlags` rather than `touchMask`, so the branch never runs and every touch reports a constant `1.0f`. That is a separately tracked engine bug, and the answer deliberately reports what a caller would actually receive rather than what the OS could provide.

### Overall Product Risks

- **Complexity:** Low. Three properties over one new command path. No existing behaviour changes, no state machine or event path is touched, and the whole surface compiles out on Editor versions that cannot answer, so on every currently shipping Editor this is a no-op.
- **Halo Effect:** Low. `InputManager` gains three cached fields and one method; `Pen`, `Mouse` and `Touchscreen` gain one static property each. The one shared-code change is replacing the unused `ExecuteGlobalCommand` with `ExecuteSystemCommand` (see below). Editor-branch risk is limited to the `versionDefines` expression being wrong, which would make the API silently absent rather than misbehave.

### Comments to reviewers

**`ExecuteGlobalCommand` is replaced by `ExecuteSystemCommand`.** The old helper had zero callers, orphaned when ISXB-927 deleted `UseWindowsGamingInputCommand`, and its premise was wrong: it addressed device id `0` on the comment that the engine routes such commands by FourCC alone, but `InputDeviceIOCTL` resolves the id against the device registry and `0` is the invalid-device sentinel, so nothing sent there could ever have been answered. The engine still declares the matching `'UWGI'` code that nothing handles; being cleaned up separately.

**The commands and `InputCapabilitySupport` are internal**, unlike everything else in `Devices/Commands/`. They target the system endpoint, so there is no custom device a user could answer them for, and it avoids freezing a capability model the low level input API is expected to revisit. **This is the decision I'd most like challenged**, since it is a visible deviation from the folder's convention. Note that ISX-2046 asks for the managed interop structs to live in `InputModule` rather than the package, and they do: `CapabilityState` and the payload are in `UnityEngineInternal.Input`. The package-side command structs here are what ISX needs in order to issue a command at all, and are a separate thing from those.

**Three command structs rather than one**, because `IInputDeviceCommandInfo` couples a struct to one static FourCC and a single struct taking the code as a parameter would make `typeStatic` a lie. Costs about ten duplicated lines each.

**The Device Simulator force-false is deliberately not replicated.** ISX-2079 asks for parity with legacy `mousePresent` including its `!simulateTouchEnabled` clause. That clause was a workaround for legacy having no device concept, where lying about the capability was the only lever available. ISX solves the same problem structurally: `DeviceSimulator/InputSystemPlugin.cs:61` disables the native `Mouse` and `Pen` while simulating and re-enables them afterwards, so `enabled` is the signal that moves. A platform capability should not know about an Editor simulation. Flagging it because it is a stated requirement on that ticket which this consciously does not meet.

**Touch pressure is platform-scoped, not per touchscreen.** That is the scope at which the answer exists: iOS asks the device model, Windows reads a WM_POINTER property, Android asks the platform. None enumerate digitizers, so a per-device channel would deliver system-scoped truth dressed as per-instance precision. If a platform ever gains touchscreens that genuinely differ, a per-device command can be added and preferred over this one, additively.

**The API is version-gated rather than always-present**, so users get a compile error on older Editors instead of a silent `false`. The expression is currently `6000.7.0a6`, the local development Editor, and **must be updated to whichever version the engine change actually ships in before this merges.**

Open questions: internal vs public commands; whether the properties should be instance rather than static (they are static because the answer is platform-scoped); whether `Touchscreen.multiTouchEnabled` should be folded in here or tracked separately; whether the manual section blocks merge.

### Checklist

Before review:

- [x] Changelog entry added.
    - `Added` section, with a before/after migration snippet.
    - JIRA ticket linked.
    - [ ] Jira port for the next release set as "Resolved" — not done, pending the engine change landing.
    - Functional tests, 16 of them, following the surrounding file's `Devices_Thing_Behaviour` naming.
    - Performance tests: not applicable, each query runs once and returns a cached bool.
    - Integration tests: not applicable, the cross-repo contract is covered by the three cross-check tests plus native tests on the engine side.
    - [x] Xmldoc cross references set.
    - [x] Explanation of how the API works, including why capability and presence are different questions.
    - [x] Usage code examples on all three properties.
    - [ ] **Manual not updated.** Wants a section on choosing between `isSupported` and `current != null`. Deliberately deferred: the presence half of that guidance misfires today on macOS and iOS, which register pens unconditionally, and those are separately tracked engine bugs. Worth deciding whether this blocks merge.

During merge:

- [x] Squash-merge commit message prefixed `NEW:`.


ekcoh:
Added @K-Tone and @MorganHoarau for visibility. Not sure branch/PR is done yet.

Packages/com.unity.inputsystem/InputSystem/Unity.InputSystem.asmdef
```diff
@@ -101,6 +101,11 @@
             "name": "Unity",
             "expression": "6000.5.0a8",
             "define": "UNITY_INPUTSYSTEM_SUPPORTS_FOCUS_EVENTS"
+        },
+        {
+            "name": "Unity",
+            "expression": "6000.7.0a6",
```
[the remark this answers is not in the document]
ekcoh: Correct, and it is a genuine blocker rather than a nit. Cannot be resolved yet, since the version this should point at does not exist until the engine side lands (unity PR #120240, same ticket).

What I have done instead of leaving it implicit: added a ////TODO at the top of InputSystem/Runtime/Devices/Commands/InputCapabilitySupport.cs spelling out that the expression is a local development Editor version and must be set to the shipping version before merge. It went there because asmdef files are JSON and cannot carry a comment of their own, which is part of why this is easy to miss.

The PR description also calls it out, and the dependency direction is stated at the top: this branch must not merge before the engine change.

Worth keeping this thread open until the version is pinned, rather than resolving it on the strength of a TODO.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
```diff
@@ -257,6 +257,46 @@ public class Mouse : Pointer, IInputStateCallbackReceiver
...
+        /// <remarks>
+        /// This answers "could a mouse work here", which is the question to ask when deciding
+        /// whether to offer mouse-specific functionality in a UI. To ask whether a mouse is
+        /// available to read from right now, compare <see cref="current"/> against <c>null</c>
+        /// instead.
+        ///
+        /// Legacy <c>UnityEngine.Input</c> conflated the two under <c>Input.mousePresent</c>, which
+        /// is genuine detection on Windows and the consoles but a hardcoded true on macOS and Linux.
+        ///
+        /// This is false when the platform has not been able to answer. The property exists only on
```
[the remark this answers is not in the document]
ekcoh: Fixed in [9577ea4](https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/9577ea477bb5e27412e40f97dd5044eb16b00b8f), on all three properties rather than just Mouse.

It was self-contradictory: "This is false when the platform has not been able to answer" followed immediately by "so false here means the platform said no". The docs now say false means the platform either does not support it or could not determine it, that the two are not distinguished, and that it never means the Editor was unable to ask.

Worth fixing rather than wording around, because the "could not determine" case is real: the engine side answers Unknown for touch pressure on at least one platform, where the OS supplies a pressure value whether or not the digitizer measures one. `Touchscreen.isPressureSupported` now says so explicitly.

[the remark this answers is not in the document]
ekcoh: @u-pr Verify this has been truly addressed.

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
```diff
@@ -67,6 +67,29 @@ Directly reading hardware controls bypasses the new Input System's action-based
...
+hardcoded constant meaning roughly "this platform has this kind of device", and on others they performed real
+hardware detection.
+
+The new Input System separates the two:
+
+- **Can this platform do it at all?** Use the capability properties: [`Mouse.isSupported`](xref:UnityEngine.InputSystem.Mouse),
+  [`Pen.isSupported`](xref:UnityEngine.InputSystem.Pen) and [`Touchscreen.isPressureSupported`](xref:UnityEngine.InputSystem.Touchscreen).
+  These never change while the application runs, so read them once and decide whether to offer device-specific
+  functionality.
+- **Is a device available to read from right now?** Use `Device.current != null && Device.current.enabled`.
```
MorganHoarau: I can see this causing a lot of trial and error for users, especially the `enabled` part. I would suggest adding a new `Device.isAvailable` to do that check, which would go well with `isSupported` from a semantic perspective.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Commands/InputCapabilitySupport.cs
```diff
@@ -0,0 +1,43 @@
+////TODO: the UNITY_INPUTSYSTEM_SUPPORTS_CAPABILITY_QUERIES version expression in
```
MorganHoarau: That is PR context, need to get rid of this comment.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
```diff
@@ -257,6 +257,66 @@ public class Mouse : Pointer, IInputStateCallbackReceiver
...
+        /// mouse is connected right now.
+        /// </summary>
+        /// <value>True if the platform supports mouse input.</value>
+        /// <remarks>
+        /// This answers "could a mouse work here", which is the question to ask when deciding
+        /// whether to offer mouse-specific functionality in a UI. To ask whether a mouse is
+        /// available to read from right now, check both <see cref="current"/> and
+        /// <see cref="InputDevice.enabled"/>, as described below.
+        ///
+        /// Legacy <c>UnityEngine.Input</c> conflated the two under <c>Input.mousePresent</c>, which is
```
MorganHoarau: This remarks is bloated with context that end user should not care about. Could you trim it down to only relevant information? 

Legacy input system context make sense in the `corresponding-old-new-api.md`, but has nothing to do here.

From a end-user perspective, I think the summary + first paragraph of remarks is enough.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Pen.cs
```diff
@@ -322,6 +322,79 @@ public class Pen : Pointer
...
         public new static Pen current { get; internal set; }
 
+#if UNITY_INPUTSYSTEM_SUPPORTS_CAPABILITY_QUERIES
+        /// <summary>
+        /// Whether the current platform can deliver pen input at all, regardless of whether a pen is
+        /// connected right now.
+        /// </summary>
+        /// <value>True if the platform supports pen input.</value>
+        /// <remarks>
+        /// This answers "could a pen work here", which is the question to ask when deciding whether
```
MorganHoarau: same as `Mouse.isSupported` remarks

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Touchscreen.cs
```diff
@@ -520,6 +520,60 @@ protected TouchControl[] touchControlArray
         /// <value>Current touch screen.</value>
         public new static Touchscreen current { get; internal set; }
 
+#if UNITY_INPUTSYSTEM_SUPPORTS_CAPABILITY_QUERIES
+        /// <summary>
+        /// Whether the current platform delivers a real pressure value with touch input.
+        /// </summary>
+        /// <value>True if the platform supports touch pressure.</value>
+        /// <remarks>
+        /// When this is false, <see cref="TouchControl.pressure"/> reports a constant 1 while a
```
MorganHoarau: same as `Mouse.isSupported` remarks

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
```diff
@@ -3125,14 +3125,79 @@ internal void ApplyActions()
...
         }
 
-        internal unsafe long ExecuteGlobalCommand<TCommand>(ref TCommand command)
+#if UNITY_INPUTSYSTEM_SUPPORTS_CAPABILITY_QUERIES
+        /// <summary>
+        /// Sends a command to the engine's system endpoint, which answers questions about the
+        /// platform rather than about any one device.
+        /// </summary>
+        /// <remarks>
+        /// This replaces an earlier ExecuteGlobalCommand, which addressed device id 0 on the premise
```
MorganHoarau: Leaking context PR in code. Not relevant in 2 years time.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
...upported` and `Touchscreen.isPressureSupported`, reporting what the current platform is capable of rather than what is connected right now. Use them to decide whether to offer device-specific functionality at all. These are not drop-in replacements for the legacy `UnityEngine.Input.stylusTouchSupported` and `Input.mousePresent`: those conflated capability with presence, and on platforms where legacy did real hardware detection the new properties report capability instead, so they can be `true` where legacy was `false`. To act on input, check `Device.current != null && Device.current.enabled` rather than `Device.current != null` alone; a non-null `current` means a device object is registered, which several platforms do unconditionally, and `enabled` is what tells you it is active. The properties require an Editor version that can answer the query and are not compiled in on older versions. `Pen.isSupported` and `Touchscreen.isPressureSupported` come from [ISX-2046](https://jira.unity3d.com/browse/ISX-2046); `Mouse.isSupported` is the short-term scope of [ISX-2079](https://jira.unity3d.com/browse/ISX-2079), whose remaining half is a real presence primitive rather than a capability one
```
MorganHoarau: Leaking IM parity context in changelog + explaining the whole feature. Better just reference the relevant doc.

Packages/com.unity.inputsystem/CHANGELOG.md
```diff
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+
+  ```csharp
```
MorganHoarau: Is code snippet supported by changelog? First time I see smth like this.

MorganHoarau (changes requested):

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
MorganHoarau: Someone shared this article about this common issue with claude: https://justindfuller.com/programming/accretive-editing

Maybe useful to understand the root cause of it. I'll update my local review skill to take that into account.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
```diff
@@ -3125,14 +3125,79 @@ internal void ApplyActions()
...
+        /// so it never appears in the device list. On an engine without the endpoint the id is simply
+        /// unknown and the command fails, which callers read as "we do not know".
+        /// </remarks>
+        internal unsafe long ExecuteSystemCommand<TCommand>(ref TCommand command)
             where TCommand : struct, IInputDeviceCommandInfo
         {
             var ptr = (InputDeviceCommand*)UnsafeUtility.AddressOf(ref command);
-            // device id is irrelevant as we route it based on fourcc internally
-            return InputRuntime.s_Instance.DeviceCommand(0, ptr);
+            return InputRuntime.s_Instance.DeviceCommand(NativeInputCapabilities.systemDeviceId, ptr);
```
MorganHoarau: Any reason for not using `m_Runtime` like in other call site in the file?

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
ekcoh: Agreed, and why I mentioned this in the PR description, there called `device.IsUsable()`, that is doable as an extension method which scales better than adding strongly coupled API - since it goes for all device types. However, its a simple addition and not load bearing but if you think I should fold that in and increase the API surface we commit to I am happy to, regardless, the parity it solved without it IMO.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Commands/InputCapabilitySupport.cs
ekcoh: Yes, its just there to make this unmergable before that version has been pinned.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
ekcoh: Totally agree - it should not be here. I will get it removed.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Pen.cs
ekcoh: Agree, comparing UnityEngine.Input to Input System should be in migration docs and nowhere else.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Touchscreen.cs
ekcoh: Agreed, I'll get it updated.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
ekcoh: Yes, agree, no need for a history lesson here.

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
ekcoh: Good catch, it should get the instance via static instance API so I get it updated - otherwise its not safe for mocked runtime

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Agreed, lets get rid of the bloat

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Lets just get rid of this

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
ekcoh: Might also be - as you say - root class is sufficient. Leaving this open until you have a chance to respond @MorganHoarau

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/CHANGELOG.md
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/InputSystem/Runtime/InputManager.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Touchscreen.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Pen.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Commands/InputCapabilitySupport.cs
ekcoh: Addressed in https://github.com/Unity-Technologies/InputSystem/pull/2476/commits/0a68b7625e66cc6e775ec504aaa6ae4961dda65c

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
MorganHoarau: Make sense, not a fan of "usable" as a term as I don't think I've ever seen any Unity API use it. I'm not sure to understand why an extension method scales better than adding a new property on Device though. In both case we are exposing a new API. But this could be move as a separate ticket to improve the clarity around the topic. 

However, I fear this would never be tackled on the side, so I would still add it here. It is a small addition and I don't see the concept going away anytime soon.

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
```diff
@@ -77,19 +101,19 @@ Directly reading hardware controls bypasses the new Input System's action-based
...
 [`Input.mousePosition`](https://docs.unity3d.com/ScriptReference/Input-mousePosition.html)|Use [`Mouse.current.position.ReadValue()`](xref:UnityEngine.InputSystem.Mouse)<br/>Example: `Vector2 position = Mouse.current.position.ReadValue();`<br/> **Note:** Mouse simulation from touch isn't implemented yet.
-[`Input.mousePresent`](https://docs.unity3d.com/ScriptReference/Input-mousePresent.html)|No corresponding API yet.
+[`Input.mousePresent`](https://docs.unity3d.com/ScriptReference/Input-mousePresent.html)|Use [`Mouse.isSupported`](xref:UnityEngine.InputSystem.Mouse) to check whether the platform supports mouse input at all.<br/>Example: `if (Mouse.isSupported) ShowMouseSettings();`<br/>**Note:** Not a drop-in replacement; see [Device capability and device availability](#device-capability-and-device-availability) above. Input System does not currently deliver mouse input on iOS, iPadOS or visionOS, so `Mouse.isSupported` is `false` there even though the platform itself supports indirect mice. Requires a recent Editor version.
```
MorganHoarau: > "Not a drop-in replacement;"

This doesn't sound good to me. Make sure to add @suearkinunity so she can review

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
```diff
@@ -257,6 +257,49 @@ public class Mouse : Pointer, IInputStateCallbackReceiver
...
+        /// This answers "could a mouse work here", which is the question to ask when deciding
+        /// whether to offer mouse-specific functionality in a UI. The answer cannot change while the
+        /// application runs.
+        ///
+        /// A false value means either that the platform does not support mouse input or that it could
+        /// not determine the answer. The two are deliberately not distinguished, because a caller
+        /// deciding whether to offer functionality wants the same behaviour in both cases. It never
+        /// means the Editor was unable to ask, since the property only exists where it can.
+        ///
+        /// Whether a mouse is available to read from right now is a separate question, and needs both
```
MorganHoarau: This does not read good for me as and enduser doc. As an enduser, I don't have time to read and just want access to the relevant information in a minimum effort. For instance, I would expect something like "To know if a mouse is available, use <see cref="current"/> and  <see cref="InputDevice.enabled"/>.". 

I would recommend running `enduser-docs` skill (located in the unity/unity repo's `.claude/skills` on this branch.

MorganHoarau (commented):
I still think enduser docs can be improve. I won't go over each for now, but I've provided more details for one + recommeded `enduser-docs` skill.

The doc might also mutate further if you end up adding the new isUsable/isAvailable API. 

I would personally add it, but better ask other reviewer their opinion too

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
MorganHoarau: Trying to be more specific about what is bothering me:

- The current formulation sounds like a meta commentary. It tells the reader that a distinction exists rather than just making the distinction.
- It buries the actionable content.
- There are casual filler (i.e. "read from right now").

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
ekcoh: I am not a fan of "usable" either, and not a fan of "available" either since I would assume its available as soon as its connected, but this also gates on device being enabled. 

I take the extension method thing back since I kind of ignored the fact that current design uses OO so having it on the base makes sense in this case instead of doing constrained generics.

I intererpret your reply as YES - add it? (Ignoring the open naming headache)

Assets/Tests/InputSystem/CoreTests_Devices.cs
```diff
@@ -5895,4 +5895,168 @@ public unsafe void Devices_DoesntErrorOutOnMaxTouchCount()
                 BeginTouch(i, new Vector2(i * 1.0f, i * 2.0f), time: 0);
         }, Throws.Nothing);
     }
+
+#if UNITY_INPUTSYSTEM_SUPPORTS_CAPABILITY_QUERIES
+    // Platform capability queries. These are addressed to the engine's system endpoint rather than
+    // to a device, because they answer "can this platform do X" rather than "is an X connected".
+    // What a given platform actually answers is asserted natively in the engine repository, from
+    // PlatformDependent, where the file location gates the test to that platform.
```
LeoUnity: "What a given platform actually answers is asserted natively in the engine repository, from PlatformDependent"
The way I read it it makes it sound like we have platfromdependent native test, but I don't think thats the case looking at the native PR.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Commands/QueryPenSupportedCommand.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <see cref="InputDevice.ExecuteCommand{TCommand}"/>.
+    ///
+    /// The FourCC must match <c>kInputFourCCIOCTLQueryPenSupported</c> in the engine's
+    /// <c>Modules/Input/InputFourCC.h</c>.
+    /// </remarks>
+    /// <seealso cref="Pen.isSupported"/>
+    [StructLayout(LayoutKind.Explicit, Size = kSize)]
+    internal struct QueryPenSupportedCommand : IInputDeviceCommandInfo
+    {
+        public static FourCC Type => new FourCC('Q', 'P', 'E', 'N');
```
LeoUnity, line 23: Nitpick: instead of redefining the FourCC here, can we use the `public FourCC(int code)` constructor with the NativeInputCapabilities.xxx int? this avoids things gettign out of sync and potentially makes the Devices_CapabilityQueryCodes_MatchTheEngineCodes obsolete.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
```diff
@@ -257,6 +257,49 @@ public class Mouse : Pointer, IInputStateCallbackReceiver
...
+        /// <see cref="current"/> and <see cref="InputDevice.enabled"/>. See the "Device capability and
+        /// device availability" section of the Input Manager migration documentation.
+        ///
+        /// Read this from the main thread: resolving the answer can require a platform API that is
```
LeoUnity: Is this necessary? this is not marked as ThreadSafe and by default all Unity APIs are main thread only

LeoUnity (approved):
Left a few nitpicks and suggestions.
I'm assuming this PR needs to be transferred to unity/unity at some point, consider splitting it into 2 PRs one with the code changes and another for Docs. 
The Docs deadline is much later, and this avoids doc specific reviews from blocking the delivery of the functionality

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
ekcoh: Agreed, removed from all three properties in 6026581ac.

There is a real constraint underneath it, in that the platform side resolves the answer once and caches it, so on iOS the first caller on any thread fixes the value process-wide. That is an implementation detail rather than something a caller can act on though, and by ISX convention main-thread-only is the default assumption, so documenting it was just noise.

I still wanted it to be thread-safe underneath for future proofing.

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Commands/QueryPenSupportedCommand.cs
ekcoh: Checked this and it is viable via `InternalsVisibleTo` so `NativeInputCapabilities` is reachable. I have left it as is for now for one reason worth flagging.

Those constants are` const int`, so C# inlines them at the call site at compile time. While the package ships independently of the engine, referencing them would bake the literal into the package rather than create a runtime dependency, which turns a mismatch into a silently stale value instead of a caught one. It would also reduce `Devices_CapabilityQueryCodes_MatchTheEngineCodes` to comparing a constant against itself, so we would lose the check rather than make it obsolete.

Once this is converted to "in-repo" and package and engine ship lockstep, both objections go away and your version is strictly better. Happy to do it as part of that move - which is next on the list after this review concludes.

There is also `NativeInputRuntime.cs`, which states it should be the only file referencing `UnityEngineInternal.Input`. That rule is already loose, `InputManager.cs` and `InputEvent.cs` reference it too, but it is another reason to make this change deliberately rather than as a nit.

Assets/Tests/InputSystem/CoreTests_Devices.cs
ekcoh: You read it right, and the comment was wrong. There are no per-platform tests in the engine's platform code at all. The native branch touches 13 platform files and not one of them is a test. The native tests live with the input module, and what they assert is that whichever platform they happen to run on returns a valid state, plus the endpoint contract itself: payload size validation, device codes rejected, and the enum values the managed mirror assumes.

So no test anywhere asserts the answer a named platform gives. A wrong per-platform answer is caught by cross-platform compilation and review, not by CI. Reworded the comment to say exactly that in e46584ae3.

Additional related cleanup in 3068ef5df.

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
ekcoh: I built it, as `InputDevice.isActive` returning `added && enabled`, and then took it back out. Writing the tests made it clear it does not earn its place, at least not for the pattern we are documenting.

Past a `current != null` guard, `isActive` is exactly `enabled`. Removal resets `current` to `null`, see `Mouse.OnRemoved`, so `added` is always true on that path. It is also the same number of characters, so `current != null && current.isActive` buys nothing over `current != null && current.enabled`.

Where it genuinely differs is a cached device reference. `RemoveDevice` only clears `m_DeviceIndex` and never touches the Disabled flags, so a device that has been removed still reports enabled as `true`. That is a real trap, and worth having a name for, but it is one that documentation fixes as well as new API does.

So instead of the property, I documented the trap in the migration doc in 40f5b7dce, with an example contrasting a cached reference against reading `current`. If we still think it pays for itself once that is written down, I am happy to revisit it on the core-package recreated PR, where the API review happens in the engine repo anyway.

Two shapes I ruled out on the way, in case they come up:
- `current?.enabled` does not compile as a condition, since `?.` on a bool member yields `bool?`. It needs `== true`, which is not shorter, and we use `?. `nowhere else in the docs.
An extension method that absorbs the null check is considered an anti-pattern generally so its dismissed since it looks like an instance call that should throw on `null` and silently does not.

You were right that the extension method does not scale better than a property when the design is already OO, so that part of my earlier answer was wrong regardless of where this lands.

OK to continue this discussion offline or in repurposed PR?

Packages/com.unity.inputsystem/InputSystem/Runtime/Devices/Mouse.cs
ekcoh: Agreed on all three points. Rewrote the remarks on all three properties in 6026581ac using the enduser-docs skill.

The skill also turned up two things I had not looked for. cs-engine-property.mdx says `<value>` tags are no longer recommended and to remove existing ones, so those are gone. release-notes.mdx says "legacy" should only be used for features with "legacy" in the name, so the CHANGELOG.md no longer says it.

One rule I did not follow: the same file says <seealso> is "not recommended" on properties. I kept them, because every neighbouring property in these files has them and stripping only these looked arbitrary. Happy to remove them if you would rather follow the guidance strictly.

Packages/com.unity.inputsystem/Documentation~/corresponding-old-new-api.md
ekcoh: Fair, it read as a warning label rather than guidance. Reworded in 40f5b7dce to say what actually differs.

Adding @suearkinunity as part of repurposed PR to decide on how to hand doc - improve it or do what @LeoUnity suggests and handle it outside.

Assets/Tests/InputSystem/CoreTests_Devices.cs
LeoUnity: Better, but I still think this comment is very verbose, do we need to say that:
"The engine's own tests assert that whichever platform they run on returns a
    // valid state and that the endpoint rejects malformed payloads. No test asserts the answer a
    // named platform gives, so a wrong per-platform answer is caught by review, not by CI."

Its a bit weird to comment what we are not doing and the reasoning is not that strong to justify it...

ekcoh: Completely dropped this slop in 0951ea02e - value was zero

ekcoh:
To reviewers @LeoUnity and @MorganHoarau: this PR will be closed as soon as pending comments are concluded, then it will be reopened as a internal PR #121126.

